PR #5, opened 2016-03-02
Add beta VIIRS L1B reader

davidh-ssec:
I've created a preliminary version of the VIIRS L1B format. A lot of the code was copied from the VIIRS SDR reader. I ended up having to modify a few important things to get stuff to work (see RawConfigParser to ConfigParser). The config file also got kind of long, but I think all data loading works as expected.


davidh-ssec:
Alright well it looks like my change for ConfigParser doesn't work in python 3. I guess I'll have to do this the "long" way.

PR #21, opened 2017-01-11
Mock pyresample.ewa

mitkin:
Mock pyresample.ewa to prevent sphinx from importing the module.

---
This change is [Reviewable](https://reviewable.io/reviews/pytroll/satpy/21)


davidh-ssec:
Does this stop readthedocs from completing?

mitkin:
yeah, it throws an error. Pyresample is already mocked, so it can't import ewa from it.

PR #25, opened 2017-01-26
Add another modis filepattern

ninahakansson:


davidh-ssec:
Sorry, I think all the tests passed. I tried to make travis skip the builds that we know fail but it looks like it considered that as a failure.

I'll figure this out, but thanks for the addition. Just curious, where is this file pattern from?

PR #27, opened 2017-01-30
WIP: Refactor Scene dependency tree

davidh-ssec:
Summary of changes made in this PR:
 - [x] Fix `Scene.available_composites` so it properly lists composites with modifiers
 - [x] Remove a majority of the dependency tree logic out of the Scene and in to another object (subclass of `Node` called `DependencyTree` that acts as root?)
 - [x] Create DatasetID static comparison method for names and wavelengths and `__eq__` for better handling of DatasetIDs in `list`.
 - [x] Allow modified composites to be loaded via `scn.load([DatasetID(..., modifiers=(...))])`
 - [x] Fix creation of dependency tree and loading of composites so that prereqs DatasetIDs are fully qualified as soon as possible.

Questions:
1. What is the correct DatasetID for a `true_color`?
    a. `DatasetID(name='true_color')`
    b. `DatasetID(name='true_color', resolution=742, calibration='reflectance', modifiers=('sunz_corrected', 'rayleigh_corrected'))`
2. Is a composite's DatasetID known before it creates a Dataset? Can a DatasetID be modified by a compositor during creation?
    - Examples may include a composite that changes the resolution of a prereq, or performs some sort of special calibration?
3. Does a `Dataset.info` have keys for every element in a `DatasetID` (which is held in `Dataset.info['id']`? What keeps them in sync? Should they be assumed to never change? `DatasetID` is a tuple so it could be a special case in `Dataset.info.__getitem__` but that requires yet another `dict` subclass.

### DIFF
--- a/satpy/__init__.py
+++ b/satpy/__init__.py
@@ -30,6 +30,6 @@
 """
 
 from satpy.version import __version__
-from satpy.scene import Scene
 from satpy.projectable import Dataset, Projectable
-from satpy.readers import DatasetID, DatasetDict
+from satpy.readers import DatasetID, DatasetDict, DATASET_KEYS
+from satpy.scene import Scene
--- a/satpy/__init__.py
+++ b/satpy/__init__.py
@@ -32,3 +32,4 @@
 from satpy.version import __version__
 from satpy.scene import Scene
 from satpy.projectable import Dataset, Projectable
+from satpy.readers import DatasetID, DatasetDict
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -33,7 +33,7 @@ import yaml
 from satpy.config import (CONFIG_PATH, config_search_paths,
                           recursive_dict_update)
 from satpy.projectable import InfoObject, Projectable, combine_info
-from satpy.readers import DatasetID, DatasetDict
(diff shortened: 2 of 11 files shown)


davidh-ssec:
Wow this was tough to get working, but I think it functionally works as expected. I've refactored composite config loading and the dependency tree used to read/load datasets. Important changes made so far:

 - Compositor objects are loaded and instantiated during `Scene.__init__`. This made dependency tree creation much easier to separate from the Scene. Previously these were loaded *iff* a dataset couldn't be found in the loaded readers
 - Multiple calls to `Scene.load` should work properly
 - Calls to `Scene.load` in a resampled Scene _should_ work if all dependencies needed from readers already exist, but it hasn't really been tested.
 - `Scene.available/all_composite_ids/names` methods only return non-modifier composites

Answers to original questions after talking with @mraspaud:

1. We would prefer to have the fully qualified DatasetIDs (b). Currently this PR only does the simple DatasetID (a). I don't think it is possible to do (b) reliably.
2. It would be nice to know the composite DatasetID before, but given the flexibility of a compositor I don't think this is possible. Currently this PR only handles the DatasetID on config read.
3. Yes, there are elements for each element in the DataseID.

TODO:
- [ ] Tests
- [ ] Documentation
- [ ] Try to definitively answer question 1.

PR #38, opened 2017-03-30
conda based install

bmu:
This adds an environment file for conda based insatllation. 
A small documentation is added.


bmu:
1. they are not required! they are created by `conda env export`. I can remove them.
2. it wasn't available in version 1.3.1 when I started! Now I it is I see! can add it to the conda part of installations

davidh-ssec:
1. If conda export produced then it's probably best to just leave it in.
2. We (I) have access to the conda-forge feedstock to update the pyresample package, but the other maintainer from conda-forge is pretty quick to update the package. You could leave the version off of pyresample and pykdtree if that's allowed in these environment scripts.

davidh-ssec:
I might also prefer the file be suffixed with the version of python, but I'm open to suggestions. My hope is that we could also make a python 3.6 version of the environment.

bmu:
I would switch to conda for pyresample! The only reason I didn't use it was, that it wasn't available!

bmu:
I have some time tomorrow and I think a python 3.6 version should be possible.

davidh-ssec:
Awesome thanks.

bmu:
cf-python is not included up to now as it introduces a lot of additional dependencies and packages to install! 
however it is quite well documented in their documentation how to install it using conda!

But maybe we should mention this or make it more clear in the documentation ("environment files to install satpy and all its dependencies" which is the actual text, is not really true!).

bmu:
Ok, I would rather stick with pyresample from pip as the conda-forge version will add a lot of additional dependencies!

bmu:
Seems like a 3.6 requirement file isn't as easy to create, as I thought. For some dependencies e.g. gdal, there is no py36 version in the default channel, so one has to switch to conda-forge. This introduces "UnsatisfiableError"`s than.

My proposal is to stick with 2.7 and to accept this request for now. 
We just need a decision on pyresample and I have to do some rewording in the documentation.

bmu:
I did the changes. should be ready now.

PR #53, opened 2017-06-05
Add CLAVR-x reader

davidh-ssec:
This PR adds the CLAVR-x reader. It is the first reader in SatPy that dynamically determines what variables are available for loading based on the contents of the provided/discovered files. This required only a few small changes to the yaml reader base class and only a small amount of extra work in the clavrx reader. It is not perfect or generic enough to work for all possible dynamic dataset readers, but it can at least be used as a good proof of concept.

CLAVR-x is "The Clouds from AVHRR Extended" processing package that utilizes the NOAA Algorithm Working Group (AWG) algorithms. Its main and most interesting set of products are cloud based, but it also includes a moon reflectance product based on VIIRS DNB imagery.


davidh-ssec:
This PR is currently failing due to a requirement by the netcdf4-python library on cython during installation. I can add it to the travis config, but am currently discussing the issue with the netcdf4-python maintainers which may result in a fix in the setup.py.

PR #55, opened 2017-06-22
Added yaml configs in enhancements to setup

ch-k:
Otherwise `satpy/etc/enhancements/generic.yaml` is not installed and something like `scene.show(chs)` might not work.


davidh-ssec:
This has been fixed in the `feature-clavrx-reader` branch. We were waiting on a newer version of the netCDF4 python library to fix a bug in the unittests running on travis. See #53. I'm trying to merge this in the next hour or so.

ch-k:
Nice, thanks a lot!

PR #56, opened 2017-07-07
Fix abil1b reader close files

ch-k:
This fix ensures that *.nc files are closed. It's not nice, but works for now (Some attributes should be read in advance to avoid extensive file opening and closing).

### DIFF
--- a/satpy/readers/abi_l1b.py
+++ b/satpy/readers/abi_l1b.py
@@ -45,12 +45,12 @@ class NC_ABI_L1B(BaseFileHandler):
     def __init__(self, filename, filename_info, filetype_info):
         super(NC_ABI_L1B, self).__init__(filename, filename_info,
                                          filetype_info)
-        self.nc = h5netcdf.File(filename, 'r')
-
+        self.filename = filename
         platform_shortname = filename_info['platform_shortname']
         self.platform_name = PLATFORM_NAMES.get(platform_shortname)
         self.sensor = 'abi'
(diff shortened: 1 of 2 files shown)


davidh-ssec:
We may want to switch the ABI reader to use something like my netcdf wrapper in `satpy/readers/netcdf_utils.py` (it uses netCDF4 instead of h5netcdf) but it would handle this a little cleaner. It wastes a little memory storing all the attribute in the file, but that isn't a big deal.

We could also make a new wrapper that uses h5netcdf if desired.

davidh-ssec:
I want to take this PR as an opportunity to discuss something about opening files. Hopefully the 3 of us (@ch-k, @mraspaud) can figure out something that works. Right now my utility classes for wrapping netcdf and hdf files closes the file as soon as it reads in all the metadata and reopens the file whenever a dataset is loaded. This was to prevent files from sitting open forever in a class that is many abstraction layers below what a normal user deals with (File Handler -> Reader -> Scene).

Would it be better to let the user have a keyword argument to switch from "always open" to "close immediately"? There are pros/cons for each. We could also try to come up with something in between where file handlers have a context manager like interface for the main Reader to use to say "keep all files open while I'm in this block of code (loading datasets) then close them".

davidh-ssec:
I'm closing this as we have now migrated to using xarray and dask. This means we use `xarray.open_dataset` and lazy load datasets from the file. I'm not sure what this means for opening/closing the file, but it does mean that this PR would need to be completely rewritten.

PR #65, opened 2017-09-11
Feature bwcompositor

pnuu:
This PR adds a single channel compositor so that black-and-white "composites" are configurable via the composite and enhancement yaml files.


davidh-ssec:
So what's with all the style/line width changes?

More importantly, what was the conversation that you and @mraspaud had that lead to this being needed?

pnuu:
The line changes came from autopep8 that defaults to 78(?) columns wide lines. I'll see if I can turn it off and restore the original formating.

This compositor is needed with trollflow(-sat) to create single channel images in the same way as the other (RGB) composites; by defining the name in the config. With this it's also possible to have separate modifiers (eg. stretch between A and B % reflectance, or invert the IR data and add some static stretching) for different wavelengths.

PR #69, opened 2017-09-13
Modis viewing geometry and atm correction

adybbroe:
Read the sun-satellite angles from the eos-hdf modis file and pass the data on so that they can be used by the compositers. E.g. required for the Rayleigh or IR atm correction with pyspectral.


satpy/readers/hdfeos_l1b.py
```diff
@@ -137,6 +137,26 @@ def get_area_def(self, *args, **kwargs):
         raise NotImplementedError
 
     def get_dataset(self, key, info, out=None, xslice=None, yslice=None):
+        if key.name in ['solar_zenith_angle', 'solar_azimuth_angle',
+                        'satellite_zenith_angle', 'satellite_azimuth_angle']:
+            sat_zen = self.sd.select('SensorZenith')
+            sun_zen = self.sd.select('SolarZenith')
+            sat_azi = self.sd.select('SensorAzimuth')
+            sun_azi = self.sd.select('SolarAzimuth')
```
davidh-ssec: In most of my readers I've handled stuff like this by providing a `file_key` in the config and accessing that from the `info` passed to this method. Otherwise, is there a reason you can't do:

```
if key.name == 'solar_zenith_angle':
    var = self.sd.select('SensorZenith')
```

Also it looks like you have some typos. `sat_zen` versus `sun_zen` below.

PR #72, opened 2017-09-14
Add DayNightCompositor

pnuu:
This PR adds a compositor that takes two composites; one is used on the day-side of the image and the other on night-side. The transition is faded using cosine of Sun zenith angle. Any 1 or 3 channel composite can be used on either side.

Example composite `natural_with_night_fog` is included in the built-in composites. Same with the corresponding enhancement.


pnuu:
There is a bug when using colorized single band composite, the mask shape doesn't match the colorized data. Fix is on the way.

pnuu:
I agree, that switching back and forth between `trollimage.Image` and `Dataset` is ugly. Couldn't think of a easier solution.

PR #104, opened 2017-11-22
Adding to more options to add_overlay. This to better control which l…

TAlonglong:
…evels of coast(GSHHS) and borders (WDB_II) are put on the plot

This must be combined PR on pycoast


TAlonglong:
Ah, bad idea. I pushed to develop again meaning this PR has two changes. Not the intention ...

TAlonglong:
Close this as PR 108 does the same

PR #111, opened 2017-11-30
Fixed area information to safe_sar_c reader to allow for resampling

eysteinn:
Datasets get area info as SwathDefinition


davidh-ssec:
I don't know much about this satellite/instrument, but usually if the data needs a `SwathDefinition` (over an `AreaDefinition`) the datasets are specified with `coordinates: [longitude_dataset_name, latitude_dataset_name]` (see other readers for examples). The base class will automatically generate the SwathDefinition for you and join the SwathDefinitions if needed. Is there a reason this can't be done with this one?

eysteinn:
Ok, I didnt know.  Im not very familiar with this either.  I will look further into it.

PR #118, opened 2017-12-06
Add file pattern for MODIS L1B from LAADS WEB

mitkin:
NASA's LAADS WEB file pattern is slightly different


davidh-ssec:
Wait, they use `.hdf` for HDF5 files?

PR #126, opened 2017-12-14
Improvemements to AHI True color imagery

### LINKED ISSUE #125: AHI True color imagary - poor stretch
For a code sample see the Jupyter notbook [here](https://github.com/pytroll/pytroll-examples/blob/master/satpy/ahi_true_color_pyspectral.ipynb)

#### Problem description
Observation by Andy Prata:

> Thanks for the response. I have installed the new version of satpy (0.7.5) and re-ran my script using 'true_color_reducedsize' for the rgb_composite string. Note: I also needed to comment out Dave's workaround, which is this line: local_scene = global_scene.resample(global_scene['B03'].info['area']) for the code to run without errors.
> 
> I've attached the output. It looks basically the same as before and quite different compared to the output of your jupyter noteboook example. I've also attached the new debug output.

#### Expected Output
Image shown in the above mentioned notebook example.

Log indicates that only a standard stretch and no CIRA stretch is performed.

Verified using Python 2.7.5, NumPy 1.13.1

adybbroe:

 - [x] Closes #125
 - [x] Tests passed

### DIFF
--- a/satpy/etc/composites/ahi.yaml
+++ b/satpy/etc/composites/ahi.yaml
@@ -20,6 +20,8 @@ modifiers:
 
   rayleigh_corrected:
     compositor: !!python/name:satpy.composites.PSPRayleighReflectance
+    atmosphere: us-standard
+    aerosol_type: marine_clean_aerosol
     prerequisites:
     - wavelength: 0.46
       modifiers: [reducer2, sunz_corrected]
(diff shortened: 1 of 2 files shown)


davidh-ssec:
Shouldn't we just make it semi-standard that any type of true color has a standard_name of `true_color`? I know it is isn't CF compliant, but neither is what we are doing now.

PR #128, opened 2017-12-15
Atm corrections

adybbroe:

 - [x] Tests passed (for all non-documentation changes)
 - [x] Passes ``git diff origin/develop **/*py | flake8 --diff`` (remove if you did not edit any Python files)

### DIFF
--- a/satpy/etc/composites/olci.yaml
+++ b/satpy/etc/composites/olci.yaml
@@ -5,9 +5,42 @@ composites:
     compositor: !!python/name:satpy.composites.RGBCompositor
     prerequisites:
     - name: 'Oa08'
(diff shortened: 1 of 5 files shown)


adybbroe:
I think it might be discussed at some point whether the user should be able to specify the atm-correction (atm and aerosol profiles) when doing corrections with pyspectral. Right now this PR specifies a few different recipes.

davidh-ssec:
Just brainstorming: What if we had something like `scn.configure_corrections(...)` or...maybe we need to have a way to customize individual composites given some of the conversations @mraspaud and I have been having.

adybbroe:
Yes, I think we eventually will need a way to customize individual composites, @davidh-ssec

PR #145, opened 2018-01-02
Add snow RGB, add r37-based and natural RGB recipes specific to

### LINKED ISSUE #144: 3.7 reflectance RGB recipes are wrong or missing, in particular for SEVIRI

```
from satpy.scene import Scene
debug_on()
from glob import glob
FILES = glob("/home/a000680/data/hrit/20150420/*201504201000*__")
glbscene = Scene(filenames=FILES, reader='hrit_msg')
glbscene.load(['day_microphysics'])
glbscene.show('day_microphysics')
```
#### Problem description

Loading the `day_microphysics` RGB for SEVIRI actually wrongly use the HRVIS channel, which then result in a KeyError running the above code. The reason is that the recipe is described only in the visir.yaml file.

Further there is no wintertime recipe of the above RGB, and there is no snow RGB either. Also the shortwave (VIS/NIR) channels are not sun-zenith corrected, which they should!

The above example result in a KeyError:
```
/home/a000680/usr/src/satpy/satpy/scene.pyc in show(self, dataset_id, overlay)
    593 
    594         from satpy.writers import get_enhanced_image
--> 595         get_enhanced_image(self[dataset_id], overlay=overlay).show()
    596 
    597     def images(self):

/home/a000680/usr/src/satpy/satpy/scene.pyc in __getitem__(self, key)
    305     def __getitem__(self, key):
    306         """Get a dataset."""
--> 307         return self.datasets[key]
    308 
    309     def __setitem__(self, key, value):

/home/a000680/usr/src/satpy/satpy/readers/__init__.pyc in __getitem__(self, item)
    181         key = self.get_key(item)
    182         if key is None:
--> 183             raise KeyError("No dataset matching '{}' found".format(str(item)))
    184         return super(DatasetDict, self).__getitem__(key)
    185 

KeyError: "No dataset matching 'day_microphysics' found"
```

SatPy version 0.7.7
Python 2.7.5

adybbroe:
SEVIRI, and fix sun-zenith correction

Fixing the RGBs using the 3.75 reflectance as derived with PySpectral.

 - [x] Closes #144 
 - [x] Tests passed 

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -370,8 +370,13 @@ class PSPRayleighReflectance(CompositeBase):
                              atmosphere=atmosphere,
                              aerosol_type=aerosol_type)
 
-        refl_cor_band = corrector.get_reflectance(
-            sunz, satz, ssadiff, vis.id.wavelength[1], blue)
+        try:
+            refl_cor_band = corrector.get_reflectance(sunz, satz, ssadiff, vis.id.name, blue)
+        except KeyError:
+            LOG.warning("Could not get the reflectance correction using band name: %s", vis.id.name)
+            LOG.warning("Will try use the wavelength, however, this may be ambiguous!")
+            refl_cor_band = corrector.get_reflectance(sunz, satz, ssadiff,
+                                                      vis.id.wavelength[1], blue)
 
         proj = Dataset(vis - refl_cor_band,
                        copy=False,
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -387,6 +387,16 @@ class NIRReflectance(CompositeBase):
         """Get the reflectance part of an NIR channel. Not supposed to be used
         for wavelength outside [3, 4] µm.
         """
+        nir, dummy = projectables
+        proj = Dataset(self._get_reflectance(projectables, optional_datasets) * 100, **nir.info)
+
+        proj.info['units'] = '%'
+        self.apply_modifier_info(nir, proj)
+
+        return proj
+
+    def _get_reflectance(self, projectables, optional_datasets):
+        """Calculate 3.x reflectance with pyspectral"""
         try:
             from pyspectral.near_infrared_reflectance import Calculator
         except ImportError:
@@ -413,11 +423,21 @@ class NIRReflectance(CompositeBase):
             lons, lats = nir.info["area"].get_lonlats()
             sun_zenith = sza(nir.info['start_time'], lons, lats)
 
-        refl39 = Calculator(nir.info['platform_name'], nir.info['sensor'], nir.id.name)
-        # nir.info['sensor'], nir.id.wavelength[1])
+        self._refl3x = Calculator(nir.info['platform_name'], nir.info['sensor'], nir.id.name)
+        return self._refl3x.reflectance_from_tbs(sun_zenith, nir, tb11, tb_ir_co2=tb13_4)
 
-        proj = Dataset(refl39.reflectance_from_tbs(sun_zenith, nir, tb11, tb_ir_co2=tb13_4) * 100, **nir.info)
-        proj.info['units'] = '%'
+
+class NIREmissivePartFromReflectance(NIRReflectance):
+
+    def __call__(self, projectables, optional_datasets=None, **info):
+        """Get the emissive part an NIR channel after having derived the reflectance. 
+        Not supposed to be used for wavelength outside [3, 4] µm.
+        """
+        nir, dummy = projectables
+        dummy = self._get_reflectance(projectables, optional_datasets)
+        proj = Dataset(self._refl3x.emissive_part_3x(), **nir.info)
+
+        proj.info['units'] = 'K'
         self.apply_modifier_info(nir, proj)
 
         return proj
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -416,7 +416,7 @@ class NIRReflectance(CompositeBase):
         refl39 = Calculator(nir.info['platform_name'], nir.info['sensor'], nir.id.name)
         # nir.info['sensor'], nir.id.wavelength[1])
 
-        proj = Dataset(refl39.reflectance_from_tbs(sun_zenith, nir, tb11, tb13_4) * 100, **nir.info)
+        proj = Dataset(refl39.reflectance_from_tbs(sun_zenith, nir, tb11, tb_ir_co2=tb13_4) * 100, **nir.info)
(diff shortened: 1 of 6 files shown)


satpy/composites/__init__.py
```diff
@@ -413,13 +423,21 @@ def __call__(self, projectables, optional_datasets=None, **info):
             lons, lats = nir.info["area"].get_lonlats()
             sun_zenith = sza(nir.info['start_time'], lons, lats)
 
-        refl39 = Calculator(nir.info['platform_name'],
-                            nir.info['sensor'], nir.id.wavelength[1])
+        self._refl3x = Calculator(nir.info['platform_name'], nir.info['sensor'], nir.id.name)
```
pnuu: Maybe this should be a separate stage prior to calling  `_get_reflectance()`? My suggestion is to add `_init_refl3x()` that does everything else but doesn't return the reflectance, and `_get_reflectance()` would have a check `if self._refl3x is None: self._refl3x()` before the return.

satpy/composites/__init__.py
```diff
@@ -413,13 +423,21 @@ def __call__(self, projectables, optional_datasets=None, **info):
...
-        proj.info['units'] = '%'
+
+class NIREmissivePartFromReflectance(NIRReflectance):
+
+    def __call__(self, projectables, optional_datasets=None, **info):
+        """Get the emissive part an NIR channel after having derived the reflectance. 
+        Not supposed to be used for wavelength outside [3, 4] µm.
+        """
+        nir, dummy = projectables
+        dummy = self._get_reflectance(projectables, optional_datasets)
```
pnuu: Use `_init_refl3x()` suggested above.

satpy/composites/__init__.py
```diff
@@ -413,13 +423,21 @@ def __call__(self, projectables, optional_datasets=None, **info):
...
-                       **nir.info)
-        proj.info['units'] = '%'
+
+class NIREmissivePartFromReflectance(NIRReflectance):
+
+    def __call__(self, projectables, optional_datasets=None, **info):
+        """Get the emissive part an NIR channel after having derived the reflectance. 
+        Not supposed to be used for wavelength outside [3, 4] µm.
+        """
+        nir, dummy = projectables
```
pnuu: `nir, _ = projectables`

pnuu:
Oh, forgot to mention that the title points to this PR (#145) not to the issue (#144).

adybbroe:
@pnuu  The title is long and continues below inside the PR, that is what the number 145 tells you. Nothing I did on purpose. The continuation is "SEVIRI, and fix sun-zenith correction"

satpy/composites/__init__.py
adybbroe: Yeah, I see what you mean, but isn't it a bit overkill here?
I think the amount of code would increase actually, and it is not likely this code will be reused elsewhere. And I think you then need even more new attributes besides _refl3x, to hold the nir, and tb11 for instance...

pnuu: The amount of code would increase a little, but the usage on line 437 wouldn't rely on a side-effect any more, and it would be much clearer what is happening.

Lines 407 to 424 would go with the `return` line to `_get_reflectance()`, and lines 400 to 406 + 426 would go to `_init_refl3x()`. (Line 406 would be in both.)

adybbroe:
@pnuu I have followed your suggestions. Hope I understood you right!?
Does it look better now?

satpy/composites/__init__.py
```diff
@@ -387,18 +387,32 @@ def __call__(self, projectables, optional_datasets=None, **info):
         """Get the reflectance part of an NIR channel. Not supposed to be used
         for wavelength outside [3, 4] µm.
         """
+        self._init_refl3x(projectables, optional_datasets)
+        proj = Dataset(self._get_reflectance(optional_datasets) * 100, **self._nir.info)
```
pnuu: If you pass also `projectables` to `_get_reflectance()` you don't need to set the `self._nir` etc. in `_init_refl3x()`

satpy/composites/__init__.py
```diff
@@ -410,17 +424,24 @@ def __call__(self, projectables, optional_datasets=None, **info):
...
-        self.apply_modifier_info(nir, proj)
+
+class NIREmissivePartFromReflectance(NIRReflectance):
+
+    def __call__(self, projectables, optional_datasets=None, **info):
+        """Get the emissive part an NIR channel after having derived the reflectance. 
+        Not supposed to be used for wavelength outside [3, 4] µm.
+        """
+        self._init_refl3x(projectables, optional_datasets)
+        _ = self._get_reflectance(optional_datasets)
```
pnuu: This isn't required anymore.

satpy/composites/__init__.py
adybbroe: But what about `self._nir.info`then?
Then I would need to do this again:
```
_nir, _ = projectables
```
Right?

pnuu: Yep (or `nir, _ = projectables`). Cleaner than setting `self._nir` anyways :-)

satpy/composites/__init__.py
adybbroe: How do you mean!?
The `__call__` method of the super class is not called, so I need this here as well, right?
I have done as you suggested, but the code increased by three lines.
I think the main problem with our concept of reusing code is that we have no `__init__`.

adybbroe: The following actually works, but is it really better!?
```
    def __call__(self, projectables, optional_datasets=None, **info):
        """Get the emissive part an NIR channel after having derived the reflectance. 
        Not supposed to be used for wavelength outside [3, 4] µm.
        """

        _ = super(NIREmissivePartFromReflectance, self).__call__(projectables, optional_datasets, **info)
        del(_)
        # self._init_refl3x(projectables)
        #_ = self._get_reflectance(projectables, optional_datasets)
        _nir, _ = projectables
        proj = Dataset(self._refl3x.emissive_part_3x(), **_nir.info)

        proj.info['units'] = 'K'
        self.apply_modifier_info(_nir, proj)

        return proj
```

pnuu:
I think there's an unnecessary call to `_get_reflectance()` on line 442, but otherwise everything looks like what I was thinking :-)

adybbroe:
@pnuu I understand you don't like the fact that reflectances need to be derived when you want to get the emissive part only. But this is indeed necessary as pyspectral is implemented currently.

pnuu:
Yeah, I missed that the side-effect is in pyspectral. Let's go with this and refactor pyspectral later :-)

PR #172, opened 2018-01-26
Bugfix viirs crefl composite - wasn't fully xarray/dask compliant

adybbroe:

The viirs crefl composite got broken after merging the xarray feature branch. 
This should now be fixed.

 - [x] Tests passed 

### DIFF
--- a/satpy/composites/crefl_utils.py
+++ b/satpy/composites/crefl_utils.py
@@ -1,11 +1,12 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2010, 2011, 2012, 2014, 2015.
+# Copyright (c) 2010-2018 PyTroll Developers
 
 # Author(s):
 
 #   Martin Raspaud <martin.raspaud@smhi.se>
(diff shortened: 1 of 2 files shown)


satpy/composites/crefl_utils.py
```diff
@@ -302,10 +305,9 @@ def get_atm_variables(mus, muv, phi, height, coeffs):
     sphalb0 = csalbr(tau_step)
 
     air_mass = 1.0 / mus + 1 / muv
-    ii, jj = np.where(np.greater(air_mass, MAXAIRMASS))
-    air_mass[ii, jj] = -1.0
+    air_mass = air_mass.where(air_mass > MAXAIRMASS, -1.0)
```
davidh-ssec: I think this is backwards. See https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.where.html compared to http://xarray.pydata.org/en/stable/generated/xarray.DataArray.where.html

adybbroe:
@davidh-ssec You were right!
Now we have a nice looking crefl true color, thanks!

PR #173, opened 2018-01-31
Update install.rst with more pip details and list packages which are not pulled in by requirements

howff:
- [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -15,6 +15,23 @@ To install the `satpy` package and all of its python dependencies from PyPI:
 
     $ pip install satpy
 
+To install with optional packages:
+
+.. code-block:: bash
+
+    $ pip install satpy pyorbital pyproj pycoast pyspectral pygac aggdraw h5netcdf python-hdf4 dask xarray toolz numexpr
+
+Example for Ubuntu using virtualenv:
+
+.. code-block:: bash
+
+    $ sudo apt-get install python-pip python-gdal
+    $ sudo pip install virtualenv
+    $ virtualenv /path/to/pytroll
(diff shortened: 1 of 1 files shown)


davidh-ssec:
It might be best to use `pip install satpy[extra]` which will look for `extra` in the `extras_require` section of the `setup.py`. See https://github.com/pytroll/satpy/blob/master/setup.py#L105

@mraspaud Maybe it is time we add an `all` extra that combines all the extras of the other sections.

doc/source/install.rst
```diff
@@ -15,6 +15,23 @@ To install the `satpy` package and all of its python dependencies from PyPI:
...
+.. code-block:: bash
+
+    $ pip install satpy pyorbital pyproj pycoast pyspectral pygac aggdraw h5netcdf python-hdf4 dask xarray toolz numexpr
+
+Example for Ubuntu using virtualenv:
+
+.. code-block:: bash
+
+    $ sudo apt-get install python-pip python-gdal
+    $ sudo pip install virtualenv
```
davidh-ssec: I'm not a huge fan of suggesting that users install stuff as `sudo`. I know it is necessary if you are using the system python, but I'm wondering if there is a better way. I would say conda, but I know not everyone uses conda.

davidh-ssec:
I think I'm going to use this PR as an opportunity to update the conda installation instructions as well since the environment file in the repository is pretty old. After we release 0.9 I think we should attempt to get satpy in to conda-forge for easier installation and simpler installation instructions.

Does anyone have suggestions on how we can have clean installation instructions when we have so many optional dependencies?

PR #180, opened 2018-02-02
Replace BW and RGBCompositor with a more generic one

pnuu:
This PR replaces to existing compositors with a more generic one, and it makes it possible to create also *LA* and *RGBA* composites.

 - [ ] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented


davidh-ssec:
I've been working on updating this for xarray and I've come up with a couple questions after comparing how @mraspaud updated the previous version of the `RGBCompositor` and how @pnuu did the `GenericCompositor`.

1. When should `wavelength`, `calibration`, `units`, and `modifiers` get removed from the metadata? Only for modes that aren't "L" and "LA" mode? For all?
2. @mraspaud had done some stuff with calculating the times/mid-time of the input datasets, but it looks like it modifies the metadata of the inputs which I'm not a fan of. @mraspaud what was the end goal of this functionality? Is the 'time' being modified a dimension/coordinate or a metadata attribute?

davidh-ssec:
Ok so the commit history for all this got real messed up, but this should merge fine in to `bugfixes-xarray`. It includes the bug fixes that were merged to develop which is what I originally based my changes on. So...after we answer my questions and resolve the TODOs I added I think we can merge this.

PR #182, opened 2018-02-05
Feature cloud compositor

pnuu:
This PR implements a compositor that can be used to create "cloud overlay" images from IR or VIS channels. As an example, composite and enhancement configs have been added for a composite `ir_cloud_day`. This kind of images can be used on top of static background maps to show cloud movements.

 - [x] Tests passed 


pnuu:
For some reason the `kwargs` are not propagated to the compositor, so currently only the default values given inside the compositor function are available.

PR #183, opened 2018-02-05
Add reader for KNMI's MetOp Ascat wind products

eysteinn:
Add reader for KNMI's MetOp Ascat wind products.


djhoese:
This has been replaced by #544.

PR #184, opened 2018-02-06
Update abi_l1b.yaml to allow C01 filename

howff:
C01 filename from Eumetcast has -{chid:6d}_0 suffix just like C02


davidh-ssec:
This should be merged to develop and then cherry picked to `maint/0.8` and then we should probably make a new 0.8.x release. The travis tests failing are likely not the fault of this PR given the setuptools issues we've been having.

PR #195, opened 2018-02-16
Add ABI-specific configs for Airmass composite

goodsonr:
Add ABI specific configs for Airmass composite

 - [ ] Tests added 
 - [x] Tests passed 

 - [ ] Fully documented

### DIFF
--- a/satpy/etc/enhancements/generic.yaml
+++ b/satpy/etc/enhancements/generic.yaml
@@ -70,7 +70,7 @@ enhancements:
         stretch: crude
         min_stretch: [-25, -40, 243]
         max_stretch: [0, 5, 208]
-  airmass_abit:
+  airmass_abi:
     standard_name: airmass
     sensor: abi
     operations:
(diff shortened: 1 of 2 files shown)


satpy/etc/enhancements/generic.yaml
```diff
@@ -70,6 +70,16 @@ enhancements:
         stretch: crude
         min_stretch: [-25, -40, 243]
         max_stretch: [0, 5, 208]
+  airmass_abi_default:
```
davidh-ssec: Could you remove the `_default` here? People are starting to think it means something when really I was just using `_default` to mean "not specific to any sensor or other specific identifying feature". @mraspaud not sure if you have an opinion on this.

goodsonr:
I'm happy either way you two want it.  I thought maybe you wanted the "default" here to say it is the default for abi.  Want me to change it now or wait for @mraspaud  to chime-in?

davidh-ssec:
Let's wait. We should probably define a rule. Even if it is an undocumented one for now, @mraspaud and I should agree on something.

davidh-ssec:
@goodsonr I just talked to Martin on slack. We think that the `_default` should be removed for anything that isn't generic. So any sensor specific enhancements should end in `_<sensor>` when possible. So your enhancement should be `airmass_abi`. In the future we might create separate yaml files per sensor. I think we allow that now, but it isn't heavily tested.

PR #216, opened 2018-03-08
Fix dataset writing so computations are shared between tasks

davidh-ssec:
This PR adds a `compute` keyword argument to all save_dataset, save_datasets, save_image methods in the Scene and Writers. It defaults to True in all of these methods so individual use cases have not changed. It is also handled by the higher level methods so that the user doesn't see a difference except in performance.

Handling writes in the way this PR implements will allow all dask objects to be computed at the same and is how it is implemented in the higher level methods. This means that if tasks share computations (input datasets) they will be shared and shouldn't have to be recalculated.

Additionally this will allow for more flexibility in the future. If a user calls `scn.save_datasets(..., compute=False)` they will get a dask `Delayed` object back which they can then use with other delayed objects.

If compute is True or a writer does not support returning a delayed object (and they don't complain about the compute keyword being passed) then functions will return whatever the result of the delayed function call was. Typically this is just None. This should also allow future flexibility if a writer should be returning something (maybe filenames of the files created?).

WIP: Docstrings and tests to make sure this functionality completely works.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -766,7 +766,7 @@ class Scene(MetadataObject):
         return writer
 
     def save_dataset(self, dataset_id, filename=None, writer=None,
-                     overlay=None, **kwargs):
+                     overlay=None, compute=True, **kwargs):
         """Save the *dataset_id* to file using *writer* (default: geotiff)."""
         if writer is None:
             if filename is None:
@@ -776,18 +776,19 @@ class Scene(MetadataObject):
                     os.path.splitext(filename)[1], **kwargs)
         else:
             writer = self.get_writer(writer, **kwargs)
-        writer.save_dataset(self[dataset_id],
-                            filename=filename,
-                            overlay=overlay, **kwargs)
+        return writer.save_dataset(self[dataset_id], filename=filename,
+                                   overlay=overlay, compute=compute,
+                                   **kwargs)
 
-    def save_datasets(self, writer="geotiff", datasets=None, **kwargs):
+    def save_datasets(self, writer="geotiff", datasets=None, compute=True,
+                      **kwargs):
         """Save all the datasets present in a scene to disk using *writer*."""
         if datasets is not None:
             datasets = [self[ds] for ds in datasets]
         else:
             datasets = self.datasets.values()
         writer = self.get_writer(writer, **kwargs)
-        writer.save_datasets(datasets, **kwargs)
+        return writer.save_datasets(datasets, compute=compute, **kwargs)
 
     def get_writer(self, writer="geotiff", **kwargs):
         """Get the writer instance."""
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -760,25 +760,26 @@ class Scene(MetadataObject):
             with open(conf_fn) as fd:
                 conf = recursive_dict_update(conf, yaml.load(fd))
         writer_class = conf['writer']['writer']
+        init_kwargs, kwargs = writer_class.separate_init_kwargs(kwargs)
         writer = writer_class(ppp_config_dir=self.ppp_config_dir,
                               config_files=config_files,
-                              **kwargs)
-        return writer
+                              **init_kwargs)
+        return writer, kwargs
 
     def save_dataset(self, dataset_id, filename=None, writer=None,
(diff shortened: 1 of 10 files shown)


satpy/writers/simple_image.py
```diff
@@ -23,6 +23,7 @@
 
 import logging
 
+import dask
```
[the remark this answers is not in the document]
davidh-ssec: @mraspaud Bug in stickler? This import is used.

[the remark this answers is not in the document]
mraspaud: I don't see it being used in this file ?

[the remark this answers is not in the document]
davidh-ssec: Ha I was looking at the wrong file. This was from a previous commit that stickler didn't comment on I think.

satpy/writers/__init__.py
```diff
@@ -325,17 +327,44 @@ def get_filename(self, **kwargs):
                 "No filename pattern or specific filename provided")
         return self.filename_parser.compose(kwargs)
 
-    def save_datasets(self, datasets, **kwargs):
+    def save_datasets(self, datasets, compute=True, **kwargs):
```
mraspaud: The docstring should be updated to exlain `compute`, and how that affects the return value.

satpy/writers/__init__.py
```diff
@@ -325,17 +327,44 @@ def get_filename(self, **kwargs):
...
         Subclasses can use this method to save all datasets to one single
         file or optimize the writing of individual datasets. By default
         this simply calls `save_dataset` for each dataset provided.
         """
+        sources = []
+        targets = []
         for ds in datasets:
-            self.save_dataset(ds, **kwargs)
+            res = self.save_dataset(ds, compute=False, **kwargs)
+            if isinstance(res, tuple):
```
mraspaud: I can't see `self.save_dataset` returning a tuple, even longer down the call chain, am I missing something ?

satpy/writers/__init__.py
```diff
@@ -325,17 +327,44 @@ def get_filename(self, **kwargs):
...
+
+        delayed = dask.delayed(sources)
+        if compute:
+            return delayed.compute()
+        else:
+            return delayed
 
-    def save_dataset(self, dataset, filename=None, fill_value=None, **kwargs):
+    def save_dataset(self, dataset, filename=None, fill_value=None,
+                     compute=True, **kwargs):
```
mraspaud: Docstring should mention `compute`

satpy/writers/__init__.py
```diff
@@ -360,15 +389,18 @@ def __init__(self,
         self.enhancer = Enhancer(ppp_config_dir=self.ppp_config_dir,
                                  enhancement_config_file=enhancement_config)
 
-    def save_dataset(self, dataset, filename=None, fill_value=None, overlay=None, decorate=None, **kwargs):
+    def save_dataset(self, dataset, filename=None, fill_value=None,
+                     overlay=None, decorate=None, compute=True, **kwargs):
```
mraspaud: Docstring should mention `compute` and the return value

satpy/writers/geotiff.py
```diff
@@ -229,5 +178,64 @@ def save_image(self, img, filename=None, floating_point=False, **kwargs):
 
         dst_ds.SetMetadata(tags, '')
 
-        # Close the dataset
-        dst_ds = None
+    def save_image(self, img, filename=None, floating_point=False,
+                   compute=True, **kwargs):
+        """Save the image to the given *filename* in geotiff_ format.
```
mraspaud: Docstring should mention `compute` and the return value

satpy/scene.py
```diff
@@ -766,7 +766,7 @@ def load_writer_config(self, config_files, **kwargs):
         return writer
 
     def save_dataset(self, dataset_id, filename=None, writer=None,
-                     overlay=None, **kwargs):
+                     overlay=None, compute=True, **kwargs):
```
mraspaud: Docstring should mention `compute` and what the function returns.

PR #237, opened 2018-03-21
Make parsing NetCDF date attributes Python 3 compatible

RutgerK:
This PR fixes the parsing of NetCDF date attributes in Python 3. It basically adds `.decode()` before using `strptime` to get a string instead of bytes.

Some readers didn't work on Python 3 at all (`nc_olci` and `nc_slstr` for example). Other readers handled it by using a try/except approach and using `.astype(str)` to convert. 

Strangely, in the master branch, the `abi_l1b` reader already uses `.decode()`. But apparently this is never committed to the develop branch.

 - [ ] Tests added
 - [ ] Tests passed

I tried running the tests on Windows and get some errors (`FAILED (failures=5, errors=3)`). But these dont seem to be related to changes in this PR.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "C:\\Miniconda3\\envs\\satpy\\python.exe"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.pythonPath": "C:\\Miniconda3\\envs\\satpy\\python.exe"
-}
--- a/satpy/readers/abi_l1b.py
+++ b/satpy/readers/abi_l1b.py
@@ -162,11 +162,11 @@ class NC_ABI_L1B(BaseFileHandler):
 
     @property
     def start_time(self):
-        return datetime.strptime(self.nc.attrs['time_coverage_start'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        return datetime.strptime(self.nc.attrs['time_coverage_start'].decode(), '%Y-%m-%dT%H:%M:%S.%fZ')
 
     @property
     def end_time(self):
-        return datetime.strptime(self.nc.attrs['time_coverage_end'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        return datetime.strptime(self.nc.attrs['time_coverage_end'].decode(), '%Y-%m-%dT%H:%M:%S.%fZ')
 
     def __del__(self):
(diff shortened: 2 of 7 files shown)


davidh-ssec:
@RutgerK Did you actually test the ABI code that you changed? Or test it without the change? I have been testing the `develop` branch with ABI data for the last couple weeks on python 3 and have not had any issues. I suppose it is possible that it has just been comparing the date strings instead of datetime objects.

Also do you know if these changes are needed even when the netcdf file is loaded with xarray's `open_dataset`?

RutgerK:
@davidh-ssec Maybe I have been fooling myself and something went wrong with the pip installation of the develop branch in my Conda environment. Let me check that again tomorrow. 

If it's not necessary as you say? Does that mean that conversions to string in the other readers have no use and can be removed, or is this different per reader?

I made this PR because the latest (stable) release doesn't work on Python 3. Checking the code I noticed readers handling the `strptime` in several different ways, or not at all. And the code hasn't changed in the develop branch since.

davidh-ssec:
It looks like none of the readers you've modified have been updated except for ABI. These all need to be updated to work with xarray (@mraspaud, didn't you update these readers already?). Here is a simple test with the way the ABI L1B reader works that seems to work fine on python 3:

```
$ ipython
Python 3.6.4 | packaged by conda-forge | (default, Dec 23 2017, 16:54:01) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.1.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import dask.array as da; import xarray as xr; import numpy as np

In [2]: nc = xr.open_dataset("/data/data/abi/20170901/OR_ABI-L1b-RadC-M3C01_G16_s20172451802150_e20172451804523_c20172451804570.nc")

In [3]: nc.attrs['time_coverage_start']
Out[3]: '2017-09-02T18:02:15.0Z'
```

RutgerK:
@davidh-ssec You are right. I must have ran the test against a version which didn't use xarray yet, sorry for the confusion.

Some other readers use xarray as well, like OLCI. But not all, SLSTR for example still uses h5netcdf in the develop branch. 

If the objective is to convert all readers to use xarray then this PR can be closed i guess, fixing something that is going to be removed anyway doesn't make much sense.

davidh-ssec:
@RutgerK Makes sense. Sorry for you having to waste time on this. If you feel like updating the PRs to work with xarray it would be very welcome ;). I understand that that is much more of an undertaking though. Thanks for your contributions regardless.

Issue #278, opened 2018-05-02
msg native reader fails on full disk image


sjoro:
#### Code Sample, a minimal, complete, and verifiable piece of code

```python
nat = Scene(f,reader='native_msg')
```
#### Problem description

MSG Native reader gets a ValueError when using a full disk image file

#### Expected Output
no error

#### Actual Result, Traceback if applicable
```
~/fork/satpy/satpy/readers/native_msg.py in _get_memmap(self)
     99             hdr_size = self.header.dtype.itemsize
    100             import ipdb; ipdb.set_trace()
--> 101             return np.memmap(fp_, dtype=dt, shape=(self.data_len,),
    102                              offset=hdr_size, mode="r")
    103 

~/dev/lib/python3.4/site-packages/numpy-1.14.2-py3.4-linux-x86_64.egg/numpy/core/memmap.py in __new__(subtype, filename, dtype, mode, offset, shape, order)
    262         bytes -= start
    263         array_offset = offset - start
--> 264         mm = mmap.mmap(fid.fileno(), bytes, access=acc, offset=start)
    265 
    266         self = ndarray.__new__(subtype, shape, dtype=descr, buffer=mm,

ValueError: mmap length is greater than file size

```
#### Versions of Python, package at hand and relevant dependencies
3.4.6
Thank you for reporting an issue !

sjoro:
Fixed and merged to `develop`

PR #282, opened 2018-05-04
Fix native_msg reader for ROI input and multi-part order file patterns

### LINKED ISSUE #277: msg_native reader fails when order number has a hyphen in it

# Your code here
from satpy import Scene
infile	=	['/indir/MSG2-SEVI-MSG15-0100-NA-20101101135742.336000000Z-20101101135754-1260206-1.nat']
global_scene = Scene(filenames=infile,sensor="seviri", reader="native_msg")

#### Problem description

The above code will fail with an error. If I edit the order number (1260206-1) to something that doesn't contain the hyphen (1260206) then satpy works OK. It appears that the code isn't set up to deal with orders that span multiple parts (-1, -2, etc).

#### Expected Output

Satpy loads MSG native data

  File "Make_MSG_new.py", line 3, in <module>
    global_scene = Scene(filenames=infile,sensor="seviri", reader="native_msg")
  File "/local/home/proud/VEnv/lib/python2.7/site-packages/satpy/scene.py", line 144, in __init__
    self.info.update(self._compute_metadata_from_readers())
  File "/local/home/proud/VEnv/lib/python2.7/site-packages/satpy/scene.py", line 162, in _compute_metadata_from_readers
    for x in self.readers.values())
  File "/local/home/proud/VEnv/lib/python2.7/site-packages/satpy/scene.py", line 162, in <genexpr>
    for x in self.readers.values())
  File "/local/home/proud/VEnv/lib/python2.7/site-packages/satpy/readers/yaml_reader.py", line 406, in start_time
    raise RuntimeError("Start time unknown until files are selected")
RuntimeError: Start time unknown until files are selected

2.7

sjoro:
Fixes issues #277 and #278. Fixes a bug in native_msg.py calibration, a wrong index was selected for the calibration coefficients. Fixes the area_extent-definition for native_msg. Adds two new areas to area.def. Performs code clean-up.

 - [x] Closes #277, #278
 - [x] Tests passed for native_msg


satpy/etc/readers/native_msg.yaml
```diff
@@ -8,9 +8,7 @@ reader:
 file_types:
   native_msg:
     file_reader: !!python/name:satpy.readers.native_msg.NativeMSGFileHandler ''
-    file_patterns: ['{satid:4s}-{instr:4s}-MSG{product_level:2d}-0100-NA-{processing_time1:%Y%m%d%H%M%S.%f}000Z-{processing_time2:%Y%m%d%H%M%S}-{order_id:d}.nat',
-                    '{satid:4s}-{instr:4s}-MSG{product_level:2d}-0100-NA-{processing_time1:%Y%m%d%H%M%S.%f}000Z-NA.nat',
-                    '{satid:4s}-{instr:4s}-MSG{product_level:2d}-0100-NA-{processing_time1:%Y%m%d%H%M%S.%f}000Z-{order_id:d}.nat']
+    file_patterns: ['{satid:4s}-{instr:4s}-MSG{product_level:2d}-0100-NA-{processing_time1:%Y%m%d%H%M%S.%f}000Z-{order_id:s}.nat']
```
djhoese: This doesn't include the first original pattern which had a `processing_time2`. If this file pattern doesn't actually exist (check with person who originally added it) then this is fine.

djhoese:
And by "not included before" I mean in old satpy or mpop.

satpy/etc/readers/native_msg.yaml
sjoro: ´processing_time2` is not used anywhere, the rest of the filename is now caught under `order_id`.

djhoese: Ok so you expect the processing time to just be "eaten up" by the order id? Ok, I'd rather have two file patterns, but I'm ok with this if @adybbroe and @mraspaud are.

sjoro: yes, exactly. those file parts are not needed.

djhoese:
@sjoro My question about ROI changes: This reader was part of mpop right? Or at least it existed in satpy before (version <0.9), right? Were people not using this reader with ROI datasets before this?

Basically, why were you the first one to find this bug? I figured it was some mistake you made in converting the reader to xarray/dask, but this PR looks like it is all new logic in the reader for figuring out of it is an ROI dataset.

sjoro:
ah, ok. yes, the reader was in mpop, but support for ROI-files was never implemented, only full disk files. we just got the ROI-area support fixed at the last workshop in Oslo. so, basically ROI-features were not tested before.

djhoese:
Ah ok, sounds good then.

PR #285, opened 2018-05-07
Fix native_msg calibration bug

sjoro:
Fix a bug in `get_dataset`-method in `native_msg.py`. Calibrated dataset values were accidentally set back to counts.

 - [ x] Passes ``git diff origin/develop **/*py | flake8 --diff``


sjoro:
Changes in scene.py removed.

PR #308, opened 2018-05-30
Update issue templates

djhoese:
Add separate bug and feature request issue templates


djhoese:
Looks like github didn't want to include both templates I made through their UI.

PR #322, opened 2018-06-12
Update MultiScene to work with xarray/dask

### LINKED ISSUE #157: Add animation example
#### Problem description

We should have an example in the documentation or https://github.com/pytroll/pytroll-examples of using satpy + ffmpeg to create an animated gif. Maybe on of two or more datasets being swapper for a single time or multiple times for a single channel (requires multiple Scene objects). I know it is fairly simple to generate a GIF using PIL so we could provide a utility method for this. Maybe that is the best way to go? Although ffmpeg could allow someone to make a movie (mp4, mov, etc).

### LINKED ISSUE #310: Create MultiScene from list of files

It is common in some user environments or GUI applications to have users open a series of files and ask for the software to just "go". Right now this isn't possible because a single Scene object expects all of the files provided to it to be for one particular time/orbit.

I suggest we add a classmethod to the MultiScene (which is not super well-tested/used) that would take a reader name and a series of files and automatically create the individual Scene's needed to load those files. This could add to the functionality provided in `readers/__init__.py` that uses the file patterns for readers to sort files in to groups based on start_time in the file. The classmethod would probably need to have keyword arguments for additionally grouping arguments including the resolution of the start_time.

I'm not sure how the MultiScene currently feels about SwathDefinitions and having sub-Scenes that are each different SwathDefinitions or different AreaDefinitions.

Initially, no. This would be a single new feature. In the future this may affect the individual Scene class or other utility functions we have like 'find_files_and_readers', but these conflicts aren't needed initially...I think.

No additional dependencies.

No changes to the build process.

I originally thought this would just be a separate utility function and while the core functionality will probably be in a utility function, the main interface advertised to users should probably be in the MultiScene since that it typically how these types of things would want to be used. Plus you can always get an iterator of Scenes from the MultiScene.

**Additional context**

While typing this up I also thought about the possibility that this could be helped by the idea of making the Scene an XArray a `xarray.Dataset` where the dimensions would pretty much be `(time, y, x)`. Just a thought.

Lastly, it would be a really cool example to give a bit list of files (and a reader name) to the MultiScene classmethod, run a multiscene method to save the scenes as a set of animations, and then show the animation. This could all _probably_ happen in a jupyter notebook (.gif).

djhoese:
This PR udpates the MultiScene to work with xarray and dask. I added a few useful properties and would like to add at least one helper method to it for creating GIFs or movies using the `imageio` library (see #157).

If I have time I'd also like to add the functionality discussed in #310. I marked that issue as being for v0.10, but if it is straight forward I wouldn't mind it showing up in v0.9 as an experiment feature.

I need to add documentation and maybe an example notebook for various multiscene operations. I might add it to the regular sphinx docs with the title ending in "EXPERIMENTAL" until it has been further evaluated after SciPy 2018.

 - [x] Closes #157 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
         - MAIN_CMD='python setup.py'
         # 'krb5' is added as a workaround for an issue with gdal
         # see: https://github.com/conda-forge/gdal-feedstock/issues/205
-        - CONDA_DEPENDENCIES='xarray dask toolz Cython sphinx cartopy pillow matplotlib scipy pyyaml pyproj pyresample coveralls coverage codecov behave netcdf4 h5py h5netcdf gdal rasterio pyhdf mock krb5'
+        - CONDA_DEPENDENCIES='xarray dask toolz Cython sphinx cartopy pillow matplotlib scipy pyyaml pyproj pyresample coveralls coverage codecov behave netcdf4 h5py h5netcdf gdal rasterio imageio pyhdf mock krb5'
         - PIP_DEPENDENCIES='trollsift trollimage pyspectral pyorbital'
         - SETUP_XVFB=False
         - EVENT_TYPE='push pull_request'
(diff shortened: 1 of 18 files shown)


djhoese:
I did not have time so I am not implementing the feature discussed in #310. I'll do that in a future PR.

PR #352, opened 2018-06-29
Remove loose print statement and fix future warning

### LINKED ISSUE #345: Future warning - xarray
**Describe the bug**
I get a Future warning from xarray when creating a scene object

**To Reproduce**

```python
# Your code here
from satpy import Scene
from datetime import datetime

DATA_DIR = "/home/a000680/laptop/Nordisk/EARS_NWC/data/case_20180314/ears"

myfiles = find_files_and_readers(base_dir=DATA_DIR,
                                 start_time=datetime(2018, 3, 14, 10, 5),
                                 end_time=datetime(2018, 3, 14, 10, 11),
                                 reader='nc_nwcsaf_pps')
scene = Scene(filenames=myfiles)
scene.load(['ct'])
```

**Expected behavior**
No future warning

**Actual results**

```
(0, ['/home/a000680/usr/src/satpy/satpy/etc/readers/nc_nwcsaf_pps.yaml'], ['nc_nwcsaf_pps'])

/usr/lib64/python2.7/_abcoll.py:391: FutureWarning: iteration over an xarray.Dataset will change in xarray v0.11 to only include data variables, not coordinates. Iterate over the Dataset.variables property instead to preserve existing behavior in a forwards compatible manner.
  return list(self)
/usr/lib64/python2.7/_abcoll.py:391: FutureWarning: calling len() on an xarray.Dataset will change in xarray v0.11 to only include data variables, not coordinates. Call len() on the Dataset.variables property instead, like ``len(ds.variables)``, to preserve existing behavior in a forwards compatible manner.
  return list(self)
/home/a000680/.local/lib/python2.7/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters

```

Also, notice the first line above! There seem to be a loose print somewhere!?

**Environment Info:**
 - OS: Linux
 - SatPy Version: 0.9.0b1.dev0

adybbroe:

Fixing issue 345. A loose print statement is removed. And a future warning from xarray is dealt with in the ears-nwc reader

 - [x] Closes #345
 - [ ] Tests passed 

### DIFF
--- a/satpy/readers/__init__.py
+++ b/satpy/readers/__init__.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2015-2017.
+# Copyright (c) 2015-2018.
 
 # Author(s):
 
@@ -574,7 +574,6 @@ def load_readers(filenames=None, reader=None, reader_kwargs=None,
 
     for idx, reader_configs in enumerate(configs_for_reader(reader, ppp_config_dir)):
         if isinstance(filenames, dict):
-            print(idx, reader_configs, reader)
             readers_files = set(filenames[reader[idx]])
         else:
             readers_files = remaining_filenames


adybbroe:
I have removed the print statement that affected all readers. But not yet looked at the warning from xarrat

djhoese:
@adybbroe Do you have an example of the future warning? Maybe someone else could look at the warning and fix it so that we could merge this for 0.9.1.

djhoese:
FYI I found at least one small future warning in the viirs l1b reader where I was doing `if 'dim_0' in data:` and now it needs to be `if 'dim_0' in data.dims:` because in the future it will search the values of the data array which is not what we want. Maybe your future warning is something similar, but I haven't actually looked. I will make a pull request for my warning's fix in the next couple days hopefully.

djhoese:
@adybbroe If you have a chance to test this that would be great. The xarray related warnings should be fixed now. If you don't get to it before the tests pass I'll just merge it anyway and if you still have the warning we'll open the issue up again.

PR #363, opened 2018-07-11
Add utility function to save delayed writer results in one go

pnuu:
This PR adds an slightly modified utility function from Polar2Grid that computes/saves all given delayed results from (multiple) calls to `res = scn.save_datasets(..., compute=False)` in one go.

 - [x] Tests added
 - [x] Tests passed
 - [x] Fully documented

### DIFF
--- a/doc/source/writers.rst
+++ b/doc/source/writers.rst
@@ -101,4 +101,18 @@ The above example can be used in enhancements YAML config like this:
 
 .. _trollimage: http://trollimage.readthedocs.io/en/latest/
 
+Saving multiple Scenes in one go
+================================
 
+As mentioned earlier, it is possible to save ``Scene`` datasets directly
+using :meth:`~satpy.scene.Scene.save_datasets` method:.  However,
+sometimes it is beneficial to collect more ``Scene``s together and process
+and save them all at once.
+
+    >>> from satpy.writers import compute_writer_results
+    >>> res1 = scn.save_datasets(file_pattern="/tmp/{name}.png",
+    ...                          writer='simple_image')
+    >>> res2 = scn.save_datasets(file_pattern="/tmp/{name}.tif",
+    ...                          writer='geotiff')
+    >>> results = [res1, res2]
+    >>> compute_writer_results(results)
--- a/doc/source/writers.rst
+++ b/doc/source/writers.rst
@@ -105,9 +105,11 @@ Saving multiple Scenes in one go
 ================================
 
 As mentioned earlier, it is possible to save ``Scene`` datasets directly
-using :meth:`~satpy.scene.Scene.save_datasets` method:.  However,
+using :meth:`~satpy.scene.Scene.save_datasets` method.  However,
 sometimes it is beneficial to collect more ``Scene``s together and process
 and save them all at once.
+
+::
 
     >>> from satpy.writers import compute_writer_results
     >>> res1 = scn.save_datasets(file_pattern="/tmp/{name}.png",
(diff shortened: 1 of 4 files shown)


satpy/writers/__init__.py
```diff
@@ -381,6 +381,46 @@ def to_image(dataset, copy=False, **kwargs):
...
+            # source, target to be passed to da.store
+            # Check for empty results
+            if len(res) == 0:
+                continue
+            num = len(res[0])
+            if num == 0:
+                continue
+            for i in range(num):
+                sources.append(res[0][i])
+                targets.append(res[1][i])
```
djhoese: I'm not sure it matters performance-wise but couldn't this be replaced with `extend` and `zip`? Then you could probably also remove the check for zero length.

doc/source/writers.rst
```diff
@@ -101,4 +101,20 @@ The above example can be used in enhancements YAML config like this:
 
 .. _trollimage: http://trollimage.readthedocs.io/en/latest/
 
-
+Saving multiple Scenes in one go
+================================
+
+As mentioned earlier, it is possible to save ``Scene`` datasets directly
+using :meth:`~satpy.scene.Scene.save_datasets` method:.  However,
```
djhoese: Extra colon after method?

pnuu:
Added tests for using `compute_writer_results()` for `simple_image` and `geotiff` writer output. Also simplified the source/target collection as suggested. For `mitiff` and `cf` writers there are commented-out placeholders, as I didn't figure out what metadata they need.

satpy/tests/test_writers.py
```diff
@@ -253,13 +256,15 @@ def test_enhance_with_sensor_entry2(self):
 
 
 class TestYAMLFiles(unittest.TestCase):
+
```
djhoese: I'm 99% sure this is an undesired non-PEP8 change (blank line between class declaration and class docstring).

pnuu: Hmm. Must be something `autopep8` did, as I didn't touch the existing tests. Removed the blank lines.

pnuu:
Hmm. Maybe. Yes. Then `save_datasets()` could call `res = self.save_dataset(composite, ..., compute=False)`, collect the results and then call `compute_writer_results(results)`.

My use case for this utility function is in operations (trollflow and trollflow-sat) for saving several composites -- possibly from several different resampled scenes -- that all might have different filename patterns and writers.

pnuu:
I adjusted `satpy.writers.Writer.save_datasets()` so that it also uses this utility function. I needed to adjust one unit test for `simple_image` writer as now `res = scn.save_datasets(..., compute=False)` always returns a `list` of one or more objects of either `dask.delayed` or `tuple` of two lists (sources and targets), so a direct `res.compute()` won't work.

I didn't find anything in the writer documentation mentioning using `res.compute()`, so I think this won't affect anyone. Maybe need to check the PyTroll examples, too...

pnuu:
The change should be safe also for `pytroll-examples`. And the documentation addition I already did uses the `compute_writer_results()` utility function, so it should be the most obvious for the users. I hope.

satpy/tests/test_writers.py
```diff
@@ -30,7 +30,7 @@
 import numpy as np
 import xarray as xr
 
-from satpy.writers import show, to_image
+from satpy.writers import show, to_image, compute_writer_results
```
djhoese: Could you move the satpy imports to be inside the tests so that import errors/failures on satpy's side of things don't cause all tests to not run?

PR #373, opened 2018-07-19
Fix path separator issue on Windows

### LINKED ISSUE #372: "find_files_and_readers" doesn't work on Windows
As already briefly discussed at issue #366, parsing the reader yaml files causes the pattern to change the path separator depending on the platform. This results in the `satpy.find_files_and_readers` function to never return any results.

The path separator in the yaml files are, by convention, specified as `"/"`, for example the SLSTR yaml:

`'{mission_id:3s}_SL_{processing_level:1s}_{datatype_id:_<6s}_{start_time:%Y%m%dT%H%M%S}_{end_time:%Y%m%dT%H%M%S}_{creation_time:%Y%m%dT%H%M%S}_{duration:4d}_{cycle:3d}_{relative_orbit:3d}_{frame:4d}_{centre:3s}_{mode:1s}_{timeliness:2s}_{collection:3s}.SEN3/{dataset_name}_radiance_an.nc'`

The `AbstractYAMLReader` splits this pattern on that same separator, but then uses `os.path.join` to join in back to a path. In doing this, the `os.path.join` will use the platform depending separator. So on Windows, the pattern shown above transforms to:

`'{mission_id:3s}_SL_{processing_level:1s}_{datatype_id:_<6s}_{start_time:%Y%m%dT%H%M%S}_{end_time:%Y%m%dT%H%M%S}_{creation_time:%Y%m%dT%H%M%S}_{duration:4d}_{cycle:3d}_{relative_orbit:3d}_{frame:4d}_{centre:3s}_{mode:1s}_{timeliness:2s}_{collection:3s}.SEN3\\{dataset_name}_radiance_an.nc'`

The code doing this is:
```            
file_patterns = [os.path.join(*pattern.split('/'))
                             for pattern in filetype_info['file_patterns']]
``` 
https://github.com/pytroll/satpy/blob/master/satpy/readers/yaml_reader.py#L99

When searching for files, the [`get_filebase` function](https://github.com/pytroll/satpy/blob/master/satpy/readers/yaml_reader.py#L66) splits on `"/"`, but this separator won't occur because of the `os.path.join` mentioned above. Hence the length of the split will always return 1, causing the function to only return the filename, whereas it should return the parent folder as well. 

The resulting [`fnmatch` attempt](https://github.com/pytroll/satpy/blob/master/satpy/readers/yaml_reader.py#L78) will always fail because its comparing:

`fnmatch('S1_radiance_an.nc', '???_SL_?_??????_????????T??????_????????T??????_????????T??????_????_???_???_????_???_?_??_???.SEN3\\*_radiance_an.nc')`

Personally I think that using `os.path.sep` in `get_filebase` should work, since the pattern is created/parsed by using `os.path.join`. On Windows it does work, I tested it for the OLCI L1 and SLSTR L1 readers. But @djhoese expressed some concerns about this:
https://github.com/pytroll/satpy/issues/366#issuecomment-404569222

I'm not familiar enough with SatPy's internals to understand the implications of such a change.

RutgerK:
This PR fixes the path separator issue #372 which failed to find files on Windows.

Details are in the issue linked above. In short, because the yaml reader splits the path's and joins them again using `os.path.join` the `get_filebase` function should use that same separator. This can be ensured by using `os.path.sep`. 

Closes #372

I have only been able to test this on Windows 10 (Python 3.5) for the OLCI and SLSTR readers. Since this relates to platform specific behavior, it should definitely also be tested on the other platforms. 

I also added `.vscode` to the gitignore since i keep pushing my IDE settings to Github. But its unrelated to the issue of course.

### DIFF
--- a/.eggs/README.txt
+++ b/.eggs/README.txt
@@ -1,6 +1,0 @@
-This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
-
-This directory caches those eggs to prevent repeated downloads.
-
-However, it is safe to delete this directory.
-
--- a/.eggs/README.txt
+++ b/.eggs/README.txt
@@ -1,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
+
+This directory caches those eggs to prevent repeated downloads.
+
+However, it is safe to delete this directory.
+
--- a/.gitignore
+++ b/.gitignore
@@ -58,5 +58,8 @@ nosetests.xml
 # PyCharm Settings
 .idea
 
+# VSCode Settings
+.vscode
+
 # vi / vim swp files
 *.swp
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,9 @@
+{
+    "python.linting.enabled": true,
+    "python.linting.pylintEnabled": true,
+    "python.linting.flake8Enabled": true,
+    "python.linting.mypyEnabled": false,
+    "editor.formatOnSave": true,
+    "editor.formatOnType": true,
+    "python.formatting.provider": "flake8"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,0 @@
-{
-    "python.linting.enabled": true,
-    "python.linting.pylintEnabled": true,
-    "python.linting.flake8Enabled": true,
-    "python.linting.mypyEnabled": false,
-    "editor.formatOnSave": true,
-    "editor.formatOnType": true,
-    "python.formatting.provider": "flake8"
-}
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -66,7 +66,7 @@ def listify_string(something):
 def get_filebase(path, pattern):
     """Get the end of *path* of same length as *pattern*."""
     # A pattern can include directories
-    tail_len = len(pattern.split('/'))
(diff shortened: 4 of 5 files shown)


djhoese:
@RutgerK could you look at the tests on appveyor and trying running the unit tests yourself. If I had to guess I would say that our tests were incorrect (which is why we didn't catch this before) and now that you've fixed the code the incorrect tests are failing.

djhoese:
@RutgerK Any update on the appveyor tests?

.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
```
djhoese: Please remove this file and directory (`.eggs/README.txt`).

RutgerK:
Good point, and you're right. The test fail because of the same issue. 

Both tests use a hard-coded separator, similar as it is in the yaml config files. I think that's fine, since it is the starting point. But the yaml reader class parses these patterns before calling the functions that now fail. They fail because the tests don't perform this parsing step. 

Parsing the pattern happens at:
https://github.com/pytroll/satpy/blob/master/satpy/readers/yaml_reader.py#L99

Unfortunately the parsing is not a separate function, so tests can't use the same code without having a instantiation of the yaml reader class. 

I've added the parsing step to the tests just to see what AppVeyor makes of it. But it's of course bad practice to put the logic in the test itself. But what would be the best way to resolve that?  I can move that line to a separate function which both the tests and the yaml reader can use.

PR #374, opened 2018-07-19
Use rasterio for reading generic image formats

pnuu:
This PR makes it possible to read generic image formats in a lazy way using `rasterio`.

If the image has georeferencing that `rasterio` and GDAL can read, like in GeoTIFF, the information is parsed and stored in the `Scene` so that the scene can be resampled.

 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/generic_image.py
+++ b/satpy/readers/generic_image.py
@@ -71,13 +71,11 @@ class GenericImageFileHandler(BaseFileHandler):
         data['bands'] = BANDS[data.bands.size]
 
         # Mask data if alpha channel is present
-        if data.bands.size in (2, 4):
-            mask = data.data[-1, :, :] == np.iinfo(data.dtype).min
-            data = data.astype(np.float64)
-            masked_data = da.stack([da.where(mask, np.nan, data.data[i, :, :])
-                                    for i in range(data.shape[0])])
-            data.data = masked_data
-            data = data.sel(bands=BANDS[data.bands.size - 1])
+        try:
+            data = mask_image_data(data)
+        except ValueError as err:
+            logger.warning(err)
+
         data.attrs = attrs
         self.file_content['image'] = data
 
@@ -121,3 +119,17 @@ def get_geotiff_area_def(filename, crs):
                               proj4_str_to_dict(crs),
                               x_size, y_size, area_extent)
     return area_def
+
+
+def mask_image_data(data):
+    """Mask image data if alpha channel is present."""
+    if data.bands.size in (2, 4):
+        if not np.issubdtype(data.dtype, np.integer):
+            raise ValueError("Only integer datatypes can be used as a mask.")
+        mask = data.data[-1, :, :] == np.iinfo(data.dtype).min
+        data = data.astype(np.float64)
+        masked_data = da.stack([da.where(mask, np.nan, data.data[i, :, :])
+                                for i in range(data.shape[0])])
+        data.data = masked_data
+        data = data.sel(bands=BANDS[data.bands.size - 1])
+    return data
--- a/satpy/readers/generic_image.py
+++ b/satpy/readers/generic_image.py
@@ -103,7 +103,7 @@ class GenericImageFileHandler(BaseFileHandler):
 
 
 def get_geotiff_area_def(filename, crs):
-    """"""
+    """Read area definition from a geotiff."""
     from osgeo import gdal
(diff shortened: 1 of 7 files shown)


pnuu:
Added first unittests, but have some failure in _saving_ the test data. Reading works locally from other GeoTIFFs, so there's some kind of a bug when creating the test data.

pnuu:
The saving issue was just a copy&paste error. Works better if the area definitions are actually assigned to the correct datasets.

satpy/tests/reader_tests/test_generic_image.py
```diff
@@ -0,0 +1,172 @@
...
+                                             '20180101_0000_test_rgb.tif'),
+                         writer='geotiff')
+        scn.save_dataset('rgba', os.path.join(self.base_dir,
+                                              'test_rgba.tif'),
+                         writer='geotiff')
+
+    def tearDown(self):
+        """Remove the temporary directory created for a test"""
+        try:
+            import shutil
```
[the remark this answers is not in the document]
pnuu: This is fixed, but for some reason isn't cleared.

djhoese:
@pnuu Looks like this includes changes from your `save_datasets` fixes. This could have been added to 0.9.1 but let's move it to 0.10 because of the commit confusion.

satpy/tests/reader_tests/test_generic_image.py
```diff
@@ -0,0 +1,223 @@
...
+
+import xarray as xr
+import dask.array as da
+import numpy as np
+
+from satpy.scene import Scene
+from satpy.readers.generic_image import GenericImageFileHandler
+from satpy.readers.generic_image import get_geotiff_area_def
+from satpy.readers.generic_image import mask_image_data
+from satpy import CHUNK_SIZE
```
djhoese: Could you move all satpy-related imports to inside the tests so that import errors don't cause all tests to fail? @mraspaud was the first person to point this out to me.

satpy/readers/generic_image.py
```diff
@@ -55,14 +94,42 @@ def start_time(self):
...
+        logger.debug("Reading %s.", key)
+        return self.file_content[key.name]
+
+
+def get_geotiff_area_def(filename, crs):
+    """Read area definition from a geotiff."""
+    from osgeo import gdal
+    from pyresample.geometry import AreaDefinition
+    from pyresample.utils import proj4_str_to_dict
+    fid = gdal.Open(filename)
```
djhoese: Don't `rasterio`'s CRS objects have this functionality (pulling the geotransform/etc out of the file)? Or the rasterio objects themselves? I thought there was a `to_rasterio` method but I don't see it. There is an example of getting the Affine object which is the same as the geotransform is purpose: http://xarray.pydata.org/en/stable/generated/xarray.open_rasterio.html

pnuu: The extents could be derived from the rasterio attributes it seems, but there is the PCS ID that needs to be read from the geotiff metadata not available via rasterio. One option would be to use `my_area` instead, but then the definition in the files would not be identical.

pnuu: The `rasterio`'s `.crs`attribute is just a string, e.g. `'+init=epsg:3035'`.

pnuu:
Oh, yeah, I think I created the new branch from the previous branch I was working on.

djhoese:
@pnuu when you get a chance do you want to try rebasing/merging this with master? I merged the other PR you had with the writer save utility function.

pnuu:
I'll take a look at rebasing in the morning.

pnuu:
For some reason the tests segfault in Travis, always at the same stage when [saving](https://github.com/pytroll/satpy/blob/feature-rasterio-generic_image/satpy/tests/reader_tests/test_generic_image.py#L101) the mock-images . Any idea what could be happening?

djhoese:
Hm either a bug in the specific version of pillow/libjpeg/libpng/etc that the tests are using and are currently available on conda OR there is something fishy (race condition?) with the temporary files you are reading/writing. You could try making them not using the named temporary file, maybe? Not sure it is worth the time right now.

djhoese:
Wow, what happened to the commit history on this? Github seems to be having trouble with the merge.

pnuu:
Probably the rebase messed up something. I'll see tolorrow if there's something to be done with resetting and doing the rebase again.

PR #381, opened 2018-08-06
Fix keyword arguments not being properly passed to writers

djhoese:
With changes that happened recently with how XRImage objects were being created and passed to writers something caused keyword arguments to stop being passed to the right locations. This fixes that.

 - [ ] Tests added 
 - [x] Tests passed 


djhoese:
Not sure how coverage decreased when all I did was remove lines of code...

PR #384, opened 2018-08-08
Natural color rgb

adybbroe:
Rename the natural_color RGB to false_color RGB (it was the MODIS naming of this type of RGB) and adjust the natural_color RGB to fit the EUMETSAT recipe.

 - [x] Tests passed 

### DIFF
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -113,7 +113,7 @@ composites:
       modifiers: [sunz_corrected, rayleigh_corrected]
     standard_name: true_color
 
-  natural:
+  natural_color:
     compositor: !!python/name:satpy.composites.SelfSharpenedRGB
     prerequisites:
     - wavelength: 1.63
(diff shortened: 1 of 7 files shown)


djhoese:
Awesome, thanks for fixing this. I thought the recipes I was using in Polar2Grid were the same as the EUMETSAT natural recipe. For historical purposes I should probably explain that the MODIS "false_color" uses the wavelength that it uses due to some bad detectors iirc. The VIIRS natural should probably use those same channels for proper comparison between the two instruments, but having this recipe called "false_color" for now I think is good enough.

@adybbroe could you double check that I didn't screw up the ABI and AHI recipes too?

adybbroe:
@djhoese Yes, I checked, it is fine for AHI and ABI! Also checked SEVIRI, MSI and VIIRS

adybbroe:
I think this one can be merged, right?

djhoese:
So a bit of a consistency issue. Are we calling it `natural_color` or `natural` or something else? We already do `true_color` so I guess `natural_color` would be better. If we change to `natural_color` then all of "natural" composites should be fixed and we may have to update some of the examples.

adybbroe:
@djhoese Now it should be consistent throughout. natural_color, natural_color_sun, etc...

djhoese:
Do you need to update the enhancements too?

adybbroe:
Yes, you were right, thx for the reminder @djhoese

PR #395, opened 2018-08-17
Check that the areas have same shapes

pnuu:
This PR fixes an oversight in DayNightCompositor that results in zero-sized composite if the areas of the two unresampled composites do not match.

 - [ ] Tests passed 


pnuu:
Oh. One learns something new every day. I'll update later today.

PR #400, opened 2018-08-21
Convert IASI L2 reader for xarray/dask

pnuu:
This PR converts the IASI L2 reader to be compatible with xarray/dask version of SatPy.

This is one part of #399 .

 - [x] Tests added 
 - [x] Tests passed 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
         - NUMPY_VERSION=stable
         - MAIN_CMD='python setup.py'
         - CONDA_DEPENDENCIES='xarray dask toolz Cython sphinx cartopy pillow matplotlib scipy pyyaml pyproj pyresample coveralls coverage codecov behave netcdf4 h5py h5netcdf gdal rasterio imageio pyhdf mock libtiff'
-        - PIP_DEPENDENCIES='trollsift trollimage pyspectral pyorbital'
+        - PIP_DEPENDENCIES='trollsift trollimage pyspectral pyorbital libtiff'
         - SETUP_XVFB=False
         - EVENT_TYPE='push pull_request'
         - SETUP_CMD='test'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ environment:
     MINICONDA_VERSION: "latest"
     CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\ci-helpers\\appveyor\\windows_sdk.cmd"
     CONDA_DEPENDENCIES: "xarray dask toolz Cython sphinx cartopy pillow matplotlib scipy pyyaml pyproj pyresample coverage netcdf4 h5py h5netcdf gdal rasterio imageio pyhdf mock libtiff"
-    PIP_DEPENDENCIES: "trollsift trollimage pyspectral pyorbital"
+    PIP_DEPENDENCIES: "trollsift trollimage pyspectral pyorbital libtiff"
     CONDA_CHANNELS: "conda-forge"
 
   matrix:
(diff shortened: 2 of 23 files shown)


satpy/readers/iasi_l2.py
```diff
@@ -105,62 +113,69 @@ def get_dataset(self, key, info):
...
     dsid = DSET_NAMES[key.name]
-    data = fid["/PWLR/" + dsid].value
+    dset = fid["/PWLR/" + dsid]
+    if dset.ndim == 3:
+        dims = ['y', 'x', 'level']
+    else:
+        dims = ['y', 'x']
+    data = xr.DataArray(da.from_array(dset.value, chunks=CHUNK_SIZE),
+                        name=key.name, dims=dims).astype(np.float32)
+    data = xr.where(data > 1e30, np.nan, data)
```
djhoese: I don't know if there is a performance difference, but FYI there is also a `data.where(data <= 1e30)` which will fill anything with NaN where the condition is False. But what you have is probably equivalent under the hood.

satpy/readers/iasi_l2.py
```diff
@@ -105,62 +113,69 @@ def get_dataset(self, key, info):
...
-    each of IASI scans."""
+    """Calculate seconds since EPOCH from days and milliseconds for each of IASI scan."""
 
     all_datetimes = []
     for i in range(days.size):
         day = int(days[i])
         msec = msecs[i]
         scanline_datetimes = []
-        for j in range(VALUES_PER_SCAN_LINE / 4):
+        for j in range(int(VALUES_PER_SCAN_LINE / 4)):
```
djhoese: Is this ever going to be not a whole number? I suppose the `int` ends up flooring it anyway.

satpy/readers/iasi_l2.py
```diff
@@ -105,62 +113,69 @@ def get_dataset(self, key, info):
...
             usec = 1000 * (j * VIEW_TIME_ADJUSTMENT + msec)
             for k in range(4):
                 delta = (dt.timedelta(days=day, microseconds=usec))
-                scanline_datetimes.append(EPOCH + delta)
+                scanline_datetimes.append(delta.total_seconds())
```
djhoese: This results in a different result than what this function was doing previously, right? Is this going to confuse/affect what users see?

djhoese:
How hard would it be to add tests for this reader?

Coverage...coverage...coverage!

satpy/readers/iasi_l2.py
pnuu: Good to know, although I prefer to have the chosen data shown explicitly. At least for this is more understandable.

satpy/readers/iasi_l2.py
pnuu: I added the call to `int()` as `range()` doesn't allow the argument to be a float, which it is for every (single-`/`) division in Python 3. For Python 2 the call isn't needed.

satpy/readers/iasi_l2.py
pnuu: Pyresample doesn't want to have the datetime objects in arrays, so this was more or less a must. Maybe this was handled differently in the non-xarray SatPy. And I'm quite certain I've been the only user of this so far, so we shouldn't get any complaints 8-)

pnuu:
Tests should be quite easy, just need to create an dummy file and use that. The format is simple. I'll add something, maybe even today.

pnuu:
I added tests, but just realized I'm using `Scene()` to test the reader. Should I use another day to convert and test directly the reader class?

djhoese:
It is probably more unit-y if you used the Reader directly, but it is not easy to use unless you look at some of the other reader tests. I'm fine either way.

pnuu:
Coverage, coverage, coverage added: 70+ % total.

PR #401, opened 2018-08-22
Wrap solar and satellite angles in xarray

pnuu:
In the initial xarray/dask conversion the solar and satellite angles were missed and were not properly converted to xarrays. This PR adds the wrapping.

 - [x] Tests passed 


djhoese:
Please change the title to be more clear about what reader this is in.

PR #408, opened 2018-09-07
Add support for CLAVR-x geostationary products

rayg-ssec:
This is a small fix addressing invalid values showing up in AWIPS as unmasked, exhibited when generating SCMI data from CLAVR-x AHI products. It applies scale_factor and add_offset to valid_min and valid_max, then masks off data outside those limits. It also passes that range downstream for use in the scmi writer. Tested primarily on AHI CLAVR-x output but may also apply to polar output. 

 - [x] Tests passed 


djhoese:
I'm not sure what problem @stickler-ci is having. The stickler configuration hasn't changed and their docs don't seem to say they've dropped flake8 support. I'm just going to ignore it for now.

PR #423, opened 2018-09-24
Fixed that save_dataset does not propagate fill_value

TAlonglong:
Fixed that save_dataset does not propagate fill_value into get_enhanced_image. Further does not get_enhanced_image propagate fill_value into add_overlay and add_decorate. Those in turn does not propagate fill_value into pil_image.

### DIFF
--- a/satpy/writers/__init__.py
+++ b/satpy/writers/__init__.py
@@ -179,7 +179,7 @@ def _determine_mode(dataset):
 
 
 def add_overlay(orig, area, coast_dir, color=(0, 0, 0), width=0.5, resolution=None,
-                level_coast=1, level_borders=1):
+                level_coast=1, level_borders=1, fill_value=None):
     """Add coastline and political borders to image, using *color* (tuple
     of integers between 0 and 255).
     Warning: Loses the masks !
@@ -224,7 +224,7 @@ def add_overlay(orig, area, coast_dir, color=(0, 0, 0), width=0.5, resolution=No
 
         LOG.debug("Automagically choose resolution %s", resolution)
 
-    img = orig.pil_image()
+    img = orig.pil_image(fill_value=fill_value)
     cw_ = ContourWriterAGG(coast_dir)
     cw_.add_coastlines(img, area, outline=color,
                        resolution=resolution, width=width, level=level_coast)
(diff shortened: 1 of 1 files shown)


djhoese:
So I've looked at this more and it makes sense that `fill_value` needs to be passed to the overlay/decorate functions so that `pil_image` can "finalize" the image properly (adding or not adding an alpha channel as desired). From what I understand the series of events for the bad performing use case is:

1. Provide L or RGB DataArray to `get_enhanced_image`
2. Add decoration or overlay (using associated kwarg to `get_enhanced_image`.
3. These call `pil_image` on the XRImage object which by default will make an 8-bit LA or RGBA DataArray. Which ends up getting returned from `get_enhanced_image`.
4. Writer tries to save LA/RGBA data to a JPEG/PNG using the PillowWriter (trollimage.xrimage.XRImage.pil_save). This calls `XRImage.finalize` which sees that fill_value was provided so it fills any invalid (null) pixels in the DataArray with the fill value *but* it does not remove the Alpha channel. XRImage assumes that since the user provided an Alpha channel that they want that to still exist. Which I'm guessing fails when saved to JPG and produces undesired output with PNG, right?

So the above situation makes sense for why this is happening and producing undesired results. I have no issue with this solution. I do have an issue with the performance of the functions as implemented. IMO the multiple calls to `pil_image` and `/ 255.0` to renormalize the data are wasteful. This of course only happens if an image is decorated and "overlayed" in the same call. I don't want this PR to include a change for this but just wanted to mention it as an issue.

satpy/writers/__init__.py
```diff
@@ -350,10 +351,10 @@ def get_enhanced_image(dataset,
         enhancer.apply(img, **dataset.attrs)
 
     if overlay is not None:
-        add_overlay(img, dataset.attrs['area'], **overlay)
+        add_overlay(img, dataset.attrs['area'], fill_value=fill_value, **overlay)
 
     if decorate is not None:
-        add_decorate(img, **decorate)
+        add_decorate(img, fille_value=fill_value, **decorate)
```
djhoese: Typo `fill_value` versus `fille_value`.

TAlonglong:
Thanks

Regarding the pil_image; yes it is a bad solution if both are invoked. But at the time this was implemented I could not see another solution.

I'm happy to discuss this at some point in time.

djhoese:
Best I could think of is pass the PIL image to these functions. It makes them harder to use outside of the `get_enhanced_image` function, but not sure that is important. Regardless, it is fine the way it is.

PR #428, opened 2018-09-26
Start and End Time are now returned as datetime objects not Lists

ColinDuff:
Start Time and End Time should be returned as an object not as a list. This causes some issues when using SIFT


djhoese:
Looks good to me. Maybe @sjoro or @adybbroe can comment on if there was a specific reason this was done originally.

Note to self: I need to cherry pick this to the 0.9 maintenance branch and make a 0.9 release.

sjoro:
i can't remember any particular reason, but i think somewhere along the line of the development for the reader this might have been necessary? dates back to `mpop` perhaps?

PR #431, opened 2018-09-27
Handle file type specific resolution in YAML files

sfinkens:
In some cases (e.g. GOES Imager) a channel's spatial resolution changes from one platform to another while the wavelength and file format remain the same. In order to avoid multiple YAML reader definitions for the same file format, it would be nice if satpy supported one resolution per ``file_type``. For example:

```yaml
file_types:
    goes_12to13_nc_10_7:
        file_reader: !!python/name:satpy.readers.goes_nc.GOESNCFileHandler
        file_patterns: ['goes12.{start_time:%Y.%j.%H%M%S}.BAND_04.nc',
                        'goes13.{start_time:%Y.%j.%H%M%S}.BAND_04.nc']
    goes_14to15_nc_10_7:
        file_reader: !!python/name:satpy.readers.goes_nc.GOESNCFileHandler
        file_patterns: ['goes14.{start_time:%Y.%j.%H%M%S}.BAND_04.nc',
                        'goes15.{start_time:%Y.%j.%H%M%S}.BAND_04.nc']

datasets:
  '10_7':
    name: '10_7'
    wavelength: [10.2, 10.7, 11.2]
    calibration:
      counts:
        standard_name: counts
        units: 1
      radiance:
        units: mW m-2 cm-1 sr-1
      brightness_temperature:
        standard_name: brightness_temperature
        units: K
    coordinates:
        - longitude_10_7
        - latitude_10_7
    file_type:
        goes_14to15_nc_10_7:
          resolution: 4000
        goes_12to13_nc_10_7:
          resolution: 8000

  longitude_10_7:
      name: longitude_10_7
      file_type:
          goes_14to15_nc_10_7:
            resolution: 4000
          goes_12to13_nc_10_7:
            resolution: 8000
      standard_name: longitude
      units: degrees_east

  latitude_10_7:
    name: latitude_10_7
    file_type:
        goes_14to15_nc_10_7:
          resolution: 4000
        goes_12to13_nc_10_7:
          resolution: 8000
    standard_name: latitude
    units: degrees_north
```

This PR tries to accomplish that by:

  - making ``FileYAMLReader._preferred_filetype()`` accept a dictionary
  - updating the dataset's resolution by the time we know the file types: ``FileYAMLReader.update_ds_ids_from_file_handlers()``

I'm neither sure whether this is the best way nor what the consequences might be :) Feel free to suggest changes.

 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -205,6 +205,7 @@ class AbstractYAMLReader(six.with_metaclass(ABCMeta, object)):
             if 'coordinates' in dataset and \
                     isinstance(dataset['coordinates'], list):
                 dataset['coordinates'] = tuple(dataset['coordinates'])
+
             # Build each permutation/product of the dataset
             id_kwargs = []
             for key in DATASET_KEYS:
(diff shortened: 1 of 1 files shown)


sfinkens:
@djhoese Brilliant! It was right in front of me, but I didn't notice it. This solution is much simpler and helps keeping both the reader and the YAML definition clean.

What do you think of adding a ``resolution`` property to ``BaseFileHandler`` to make contributors aware of that possibility?

```python
@property
def resolution(self):
    return None
```

djhoese:
Closing as it seems we decided the resolution attribute is an easier solution for now. Note, I'm not against this being added but I think there may be better ways to do the combination of these things, see #434.

PR #447, opened 2018-10-08
Fix masking in DayNightCompositor when composites have partial missing data

pnuu:
If one of the two composites passed to `DayNightCompositor` has partially missing data the resulting merged composite had holes at those locations. This discrepancy between the composites can be e.g. due to reflectances below zero on the night side  being masked out by the reader. This behavior can be correct for other use cases, but here it creates distracting holes in the images.

This PR replaces the masked ares with zeros if they are valid in the other composite.

 - [ ] Tests added 
 - [ ] Tests passed 

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -867,6 +867,10 @@ class DayNightCompositor(GenericCompositor):
         day_data = add_bands(day_data, night_data['bands'])
         night_data = add_bands(night_data, day_data['bands'])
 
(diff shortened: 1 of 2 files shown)


pnuu:
Indeed, that `data1.where(~nans, 0)` kept the attributes intact. Updated accordingly. Thanks @djhoese!

pnuu:
Don't have time for adding unit tests during this week, and next week I'm out of office.

PR #451, opened 2018-10-09
Allow readers to accept pathlib.Path instances as filenames.

### LINKED ISSUE #450: Allow readers to accept pathlib.Path instances
I would like to create a scene from a list of filenames, like this:

```python
scene = Scene(filenames=files)
```
where `files` is a list of `pathlib.Path` instances, instead of strings.

IMHO, the solution should be a one-liner somewhere in the code. It should not have any impact elsewere.

honnorat:
Allow readers to accept pathlib.Path instances as filenames. 

 - [x] Closes #450
 - [x] Tests added 
 - [x] Tests passed
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -67,7 +67,7 @@ def get_filebase(path, pattern):
     """Get the end of *path* of same length as *pattern*."""
     # A pattern can include directories
     tail_len = len(pattern.split(os.path.sep))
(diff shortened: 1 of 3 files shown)


djhoese:
Wow is that really all that is needed? Nice.

Would you mind adding a simple unit test? I'm guessing you'll have to add something to check if pathlib is available and skip the test if not (python 2).

honnorat:
I hope this is enough, I don't know where to test elsewhere.

PR #452, opened 2018-10-09
Improve handling of missing requirements

### LINKED ISSUE #147: SEVIRI HRIT reading: More userfriendly warning when no EPI/PRO files are present

```
from satpy.scene import Scene
debug_on()
seviri = Scene(filenames=..., reader='hrit_msg')

```
#### Problem description

It often happens that a user forgets to copy the EPI and PRO files to the data directory. When instantiating the Scene object you will get a RuntimeError exception as depicted below.
Running with debug info on you will also get some warnings like:

```
[WARNING: 2018-01-05 10:50:09 : satpy.readers.yaml_reader] Missing requirements for /data/lang/satellit/geo/hrit/2015/201511/16/12/H-000-MSG1__-MSG1________-IR_120___-000005___-201511161200-__
```

But it is not obvious it comes from the missing prologue and epilogue files.

#### Expected Output

A warning message saying that this might most probably be due to the missing header/trailer files:

`WARNING: No EPI/PRO files found!`

```
    102                                                     reader_kwargs=reader_kwargs,
    103                                                     metadata=metadata)
--> 104         self.info.update(self._compute_metadata_from_readers())
    105         self.datasets = DatasetDict()
    106         self.cpl = CompositorLoader(self.ppp_config_dir)

/home/a000680/usr/src/satpy/satpy/scene.pyc in _compute_metadata_from_readers(self)
    119         if self.readers:
    120             mda['start_time'] = min(x.start_time
--> 121                                     for x in self.readers.values())
    122             mda['end_time'] = max(x.end_time
    123                                   for x in self.readers.values())

/home/a000680/usr/src/satpy/satpy/scene.pyc in <genexpr>((x,))
    119         if self.readers:
    120             mda['start_time'] = min(x.start_time
--> 121                                     for x in self.readers.values())
    122             mda['end_time'] = max(x.end_time
    123                                   for x in self.readers.values())

/home/a000680/usr/src/satpy/satpy/readers/yaml_reader.pyc in start_time(self)
    410     def start_time(self):
    411         if not self.file_handlers:
--> 412             raise RuntimeError("Start time unknown until files are selected")
    413         return min(x[0].start_time for x in self.file_handlers.values())
    414 

RuntimeError: Start time unknown until files are selected
```

Python 2.7.5
Satpy 0.7.7

sfinkens:
Improve the warning about a missing requirement (e.g. Epilogue, Prologue) and abort if there are insufficient requirements to read any data.

 - Make it a user warning (instead of a logger warning)
 - Include the name of the requirement into the message
 - Raise a ``RuntimeError`` if there are insufficient requirements to read any data -> No more ``Start time unknown until files are selected`` errors. In particular it ensures a consistent exception if
    - all requirements are missing
    - at least one requirement is present and at least one is missing but alltogether that is not enough to read any images. Example: Epilog present, Prolog missing. Currently this case passes silently until the user tries to access the data.

The modifications still allow to read as much data as possible. E.g. missing requirements for one out of 10 images will trigger warnings, but no exceptions.

Example output:

```
/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/readers/yaml_reader.py:406: UserWarning: Missing requirement 'HRIT_PRO' for /cmsaf/cmsaf-ops6/precip/data/geo/east_atlantic/2018/01/18/H-000-MSG3__-MSG3________-IR_097___-000002___-201801181800-__
/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/readers/yaml_reader.py:406: UserWarning: Missing requirement 'HRIT_PRO' for /cmsaf/cmsaf-ops6/precip/data/geo/east_atlantic/2018/01/18/H-000-MSG3__-MSG3________-IR_097___-000003___-201801181800-__
  File "test.py", line 16, in <module>
    scene = Scene(filenames=filenames, reader='hrit_msg')
  File "/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/scene.py", line 152, in __init__
    reader_kwargs=reader_kwargs)
  File "/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/scene.py", line 197, in create_reader_instances
    ppp_config_dir=self.ppp_config_dir)
  File "/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/readers/__init__.py", line 602, in load_readers
    reader_instance.create_filehandlers(loadables)
  File "/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/readers/yaml_reader.py", line 528, in create_filehandlers
    raise RuntimeError('Insufficient requirements (e.g. Prolog, '
RuntimeError: Insufficient requirements (e.g. Prolog, Epilog) to read any data
```
On-the-fly decompression is addressed by #436.

 - [x] Closes #147 
 - [ ] Tests added
 - [ ] Tests passed
 - [x] Passes ``git diff origin/master **/*py | flake8 --diff`


satpy/readers/yaml_reader.py
```diff
@@ -386,11 +398,13 @@ def new_filehandler_instances(self, filetype_info, filename_items):
...
-                continue
-            except KeyError:
-                logger.warning("Missing requirements for %s", filename)
+            except (KeyError, RuntimeError) as req:
+                # Monkeypatch warnings module to omit the source line - just
+                # for this warning.
+                warnings.formatwarning = formatwarning_no_source
+                warnings.warn("Missing requirement {} for {}".format(
+                    req, filename))
+                warnings.formatwarning = formatwarning_orig
```
djhoese: Why remove the source line from the warning?

satpy/readers/yaml_reader.py
```diff
@@ -322,22 +331,25 @@ def check_file_covers_area(file_handler, check_area):
...
-                    # FIXME: Isn't this super wasteful? filename_info.items()
-                    # every iteration?
-                    if (all(item in filename_info.items()
-                            for item in fhd.filename_info.items())):
+                    if set(fhd.filename_info.items()).issubset(filename_info):
                         req_fh.append(fhd)
                         break
                 else:
-                    raise RuntimeError('No matching file in ' + requirement)
+                    raise RuntimeError(requirement)
```
djhoese: Even if this gets caught and a warning issued it would be nice to have a proper error message. Maybe `"No matching requirement file of type: {}".format(requirement)`?

satpy/readers/yaml_reader.py
```diff
@@ -479,6 +493,14 @@ def new_filehandlers_for_filetype(self, filetype_info, filenames):
         filtered_iter = self.filter_fh_by_metadata(filehandler_iter)
         return list(filtered_iter)
 
+    def _is_requirement(self, filetype):
+        """Determine whether the given filetype is required by others"""
+        for _, filetype_info in self.config['file_types'].items():
```
djhoese: Could be replaced with `.values()`.

satpy/readers/yaml_reader.py
```diff
@@ -498,6 +520,14 @@ def create_filehandlers(self, filenames):
                     filehandlers,
                     key=lambda fhd: (fhd.start_time, fhd.filename))
 
+        # Abort if there are insufficient requirements to read any data
+        # (e.g. no file handlers at all or only requirements).
+        only_requirements = all([self._is_requirement(filetype)
+                                 for filetype in self.file_handlers])
+        if not self.file_handlers or only_requirements:
+            raise RuntimeError('Insufficient requirements (e.g. Prolog, '
+                               'Epilog) to read any data')
```
djhoese: This assumes that required files don't have any other useful data in them. While that might be the case for the Prolog/Epilog-based readers that may not be true for all future readers. In the Scene we *could* do a `self.available_dataset_names()` check and if filenames were provided and no datasets are available then raise an exception. @mraspaud thoughts?

If this *has* to be done here then I would prefer that a reverse requirements lookup dictionary be created when the Reader is created so that the `_is_requirement` function is just a dictionary `filetype in self.filetype_required_by` or something. This loop is `O(N^2)` (if I'm correct) so any way to reduce how many times the file types are iterated over would be great.

Lastly, I'd be ok with this type of error being ignored here. Technically a reader could have files added to it over time and not all in one call (I think) so this check may be premature. The `not self.file_handlers` is handled by the loader already so it wouldn't need to be here too if the requirements check was removed. I'm open to this if it can be done quickly and not slow down other readers (that don't have required file types), just convince me :wink: .

satpy/readers/yaml_reader.py
sfinkens: Only the last line of the statement is included which is not very informative in this case: ``req, filename))``

satpy/readers/yaml_reader.py
sfinkens: Haha, true!

satpy/readers/yaml_reader.py
sfinkens: Good point. I use the exception to pass the name of the missing requirement. If I catch ``KeyError`` and ``RuntimeError`` separately, I can use the ``RuntimeError`` message as a warning.

satpy/readers/yaml_reader.py
sfinkens: Having said that, I could of course store the message in a variable so that the warning statement fits in one line ^^ But even then it does not provide much additional information. On the other hand, the current approach is less readable/maintainable.

satpy/readers/yaml_reader.py
sfinkens: Oh, many good points! I have to think about that. Not sure my understanding of the whole process is good enough to come up with a general solution.

sfinkens: Agree, the check in the reader is premature. Your idea to check ``self.available_dataset_names()`` (or ``_ids()`` which would be a little bit faster) in ``Scene.__init__`` is very good. It is much simpler and leads to the same behaviour.

satpy/readers/yaml_reader.py
sfinkens: "readability counts" :)

satpy/scene.py
```diff
@@ -150,6 +150,12 @@ def __init__(self, filenames=None, reader=None, filter_parameters=None, reader_k
         self.readers = self.create_reader_instances(filenames=filenames,
                                                     reader=reader,
                                                     reader_kwargs=reader_kwargs)
+
+        if filenames and not self.available_dataset_ids(reader_name=reader):
```
sfinkens: The only problem I could not resolve is: If no file matches the ``filter_parameters`` the exception is triggered mistakenly. But I would expect ``create_reader_instances`` to throw an exception in this case? Previously you would run into the ``Start time unknown until files are selected`` error.

satpy/readers/yaml_reader.py
djhoese: Our maximum line width is now 120 characters so it should be fine to put this all on one line and remove the special `no_source` function.

satpy/readers/yaml_reader.py
```diff
@@ -386,11 +391,12 @@ def new_filehandler_instances(self, filetype_info, filename_items):
             try:
                 req_fh = self.find_required_filehandlers(requirements,
                                                          filename_info)
-            except RuntimeError:
-                logger.warning("Can't find requirements for %s", filename)
+            except KeyError as req:
+                msg = "Missing requirement {} for {}".format(req, filename)
```
djhoese: Doesn't this put a string version of the exception `req` inside the error message? Should this be the specific requirement that is missing?

satpy/scene.py
djhoese: Sorry I'm a little confused. So if none of the files match (due to filter_parameters or other reasons) then `create_reader_instances` isn't raising an exception or it is?

Also, you should probably remove the `reader_name=reader` part since filenames can be a dictionary mapping `reader: [filenames]`. Just to check if there are any available datasets for any possible readers.

sfinkens: ``create_reader_instances`` does _not_ raise an exception in this case. Instead the subsequent call to ``_compute_metadata_from_readers``  causes the following exception:

```
  File "test.py", line 21, in <module>
    end_time=datetime.datetime(2010, 1, 2)
  File "/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/scene.py", line 159, in __init__
    self.attrs.update(self._compute_metadata_from_readers())
  File "/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/scene.py", line 177, in _compute_metadata_from_readers
    for x in self.readers.values())
  File "/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/scene.py", line 177, in <genexpr>
    for x in self.readers.values())
  File "/cmsaf/nfshome/sfinkens/software/devel/satpy/satpy/readers/yaml_reader.py", line 296, in start_time
    raise RuntimeError("Start time unknown until files are selected")
RuntimeError: Start time unknown until files are selected
```

djhoese: So maybe `load_readers` (in `satpy/readers/__init__.py` which is called by `create_reader_instances`) should change:

```python
    if not reader_instances:
        raise ValueError("No supported files found")
```

to something like:

```python
if not reader_instances or all(reader.available_dataset_ids for reader in reader_instances):
    raise ValueError("No supported files found")
```

Then I think you could remove the `scene.py` change. If the error messages would be clearer I suppose the above if statement could be split in to `if not reader_instances:` and `elif all(...):` so that it is more clear "you gave valid files, but not enough, or not useable ones" or something like what you had.

satpy/readers/yaml_reader.py
sfinkens: The string version of the ``KeyError`` already contains the missing requirement. Using the string version is recommended (as far as I know). The actual problem is using ``RuntimeError.message`` because this attribute was removed in Python 3. I tested this with Python 2 :see_no_evil:

djhoese: That's fine by me I guess. If you have an example of what the final error message looks like that would be appreciated.

sfinkens: The warnings would be :
1) ``UserWarning: No handler for reading requirement 'HRIT_PRO' for H-000-MSG4__-MSG4________-IR_108___-000006___-201809050900-__`` if none of the provided scans is complete
2) ``UserWarning: No matching requirement file of type HRIT_PRO for H-000-MSG3__-MSG3________-IR_108___-000008___-201801181200-__`` if at least one of the provided scans is complete and at least one is incomplete

satpy/scene.py
sfinkens: Good idea! I moved the check & exception to ``load_readers``

satpy/readers/yaml_reader.py
```diff
@@ -322,22 +323,26 @@ def check_file_covers_area(file_handler, check_area):
...
 
     def find_required_filehandlers(self, requirements, filename_info):
-        """Find the necessary fhs for the current filehandler.
+        """Find the necessary file handlers for the given requirements.
 
         We assume here requirements are available.
+
+        Raises:
+            KeyError, if no handler for the given requirements is available.
+            RumtimeError, if there is a handler for the given requirements,
```
djhoese: Small typo `Runtime` versus `Rumtime`. I can fix this for you if I don't see anything else that needs changing.

satpy/scene.py
djhoese: You said you moved this, but it still exists and doesn't seem to exist in `load_readers`. Forgot to push something?

sfinkens: Strange, my local and remote branch are up-to-date. Also, the ``Files Changed`` tab shows the diffs I commited, especially no changes to ``scene.py``. Maybe if I fix the typo and push again, it will be updated...

satpy/readers/__init__.py
```diff
@@ -611,4 +611,9 @@ def load_readers(filenames=None, reader=None, reader_kwargs=None,
                 remaining_filenames)))
     if not reader_instances:
         raise ValueError("No supported files found")
+    elif not all([list(r.available_dataset_ids)
+                  for r in reader_instances.values()]):
```
djhoese: Is there any reason or use case that this should maybe be `not any(...)`? That is, we could allow one or more readers to not have any datasets available but not *all* of them. Thinking if a user has multiple readers available and wants to load some dataset from at least one of them.

Granted the multiple reader functionality isn't really used by anyone right now.

sfinkens: :+1: Good idea!

PR #454, opened 2018-10-09
Update github pull request template with better git diff command

djhoese:
The git diff command (seen below) in the pull request template is not compatible with versions of bash that don't support `globstar` (`**` expansion). The one this PR will replace it with should work for all modern shells including older versions of bash like those on OSX.

See #429 for a discussion.


djhoese:
I guess I could have just pushed this directly to master.

PR #455, opened 2018-10-09
Ensures BaseFileHandler.filename is a str and is used by its subclasses.

### LINKED ISSUE #453: Review subclasses of BaseFileHander
This is to carry on with #450 (and PR #451), where @djhoese asked:

> One question, since you had to change the fake hdf5 handler to make the test work, does this mean that the base file handler (satpy/readers/file_handlers.py) might also need to be updated? What readers have you tried this with?

**Describe the bug**
Using `pathlib.Path` instances as filenames in subclasses of `BaseFileHander` may not always work. Most of the time, BaseFileHander.filename is used with open() or compatible calls (h5py.File(), xr.open_rasterio(), etc.) which is fine. But some classes may need adjustments:

- GRIBFileHandler
- HDF4BandReader and HDF4FileHandler 
- NativeMSGFileHandler
- NcNWCSAFMSG which uses h5netcdf (side note: why not h5py like most other readers ?)
- classes from `satpy.readers.nc_slstr`

honnorat:
All subclasses of BaseFileHandler have been reviewed to be sure that their constructor `Subclass.__init__` uses `BaseFileHandler.filename` instead of the constructor argument.

BaseFileHandler.filename is forced to be a string. The result is that `pathlib.Path` or strings can be used equally for building an instance of any subclass.

 - [x] Closes #453
 - [ ] Tests added 
 - [x] Tests passed 
 - [x] Passes ``git diff origin/master '**/*py' | flake8 --diff``

### DIFF
--- a/satpy/readers/abi_l1b.py
+++ b/satpy/readers/abi_l1b.py
@@ -51,7 +51,7 @@ class NC_ABI_L1B(BaseFileHandler):
         super(NC_ABI_L1B, self).__init__(filename, filename_info,
                                          filetype_info)
         # xarray's default netcdf4 engine
-        self.nc = xr.open_dataset(filename,
+        self.nc = xr.open_dataset(self.filename,
                                   decode_cf=True,
                                   mask_and_scale=False,
                                   chunks={'x': CHUNK_SIZE, 'y': CHUNK_SIZE})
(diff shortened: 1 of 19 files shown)


djhoese:
Why didn't you force `str(filename)` when the base reader class creates the file handlers?

honnorat:
What do you mean by "the base reader class" ?

djhoese:
Sorry I was trying to catch a bus and had to type fast 😜 

I meant `FileYAMLReader` the "reader" class that creates the individual "file handler" classes. If you changed this line I think it would do the equivalent: https://github.com/pytroll/satpy/blob/master/satpy/readers/yaml_reader.py#L396

honnorat:
I see your point. The advantage of this PR is that the changes it proposes are independent of the complex machinery in `satpy.readers.yaml_readers`. It ensures that when you use the FileHandlers, you are sure that MyFileHandler.filename is a string, whatever the way it was created.

All changes in the subclasses only make this use uniform across the entire code base.

Your solution is good enough for today, though.

PR #474, opened 2018-10-24
Improve handling of dependency loading when reader has multiple matches

djhoese:
This is to address issues brought up related to #458 where `scn.available_composite_names()` would result in an AssertionError instead of the proper result. It now shows me only `'overview'`, does that sound right @mraspaud?

This PR adds a subclass of KeyError to the DatasetDict key checks called `TooManyResults` to distinguish itself from a regular KeyError where the key doesn't exist. Since it is a subclass of KeyError it shouldn't cause any issues with other code that was specifically catching `KeyError`.

Note: This is technically a workaround because SatPy isn't smart enough right now to handle datasets with different names but the same exact wavelength.

This PR also sits on top of #458.

 - [x] Tests added 
 - [x] Tests passed 

### DIFF
--- a/satpy/etc/composites/slstr.yaml
+++ b/satpy/etc/composites/slstr.yaml
@@ -4,9 +4,9 @@ composites:
   overview:
     compositor: !!python/name:satpy.composites.GenericCompositor
     prerequisites:
-    - 0.65
-    - 0.86
-    - 10.8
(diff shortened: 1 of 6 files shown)


djhoese:
> This PR also sits on top of #458.

Yes.

PR #478, opened 2018-10-29
Allow writers to create output directories if they don't exist

djhoese:
If a user specifies a directory in their filename pattern it is not currently created which will causes errors in most readers. Example:

```python
    scn.save_datasets(filename='{start_time:%Y%m%d}/{name}_{start_time:%Y%m%d_%H%M%S}.tif')
```

Which should create a `20180101` directory and put a `C01_20180101_120000.tif` file in it.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented


djhoese:
@mraspaud I asked for you to review this PR but wanted to merge it as soon as possible because I wanted to see if my small change to setup.py fixes the readthedocs build. Feel free to still comment on anything you disagree with and I'll fix it tomorrow.

djhoese:
It *did* fix readthedocs!

PR #479, opened 2018-10-30
Add ll_bbox and xy_bbox to any cropped DataArray attributes

djhoese:
This adds metadata to any DataArray that is cropped using the `Scene.crop` method. Adds `ll_bbox` if provided to crop, adds `xy_bbox` if provided, otherwise nothing is added. This can be useful in later processing or in filename patterns for differentiating cropped data from the original data. I also added some utility methods for converting bboxes to strings. These may exist elsewhere but I couldn't find them.

Another option is to add a suffix to the area_id of the new AreaDefinition like `-cropped` or something.

Honestly, I'm not super happy with this solution but this is important information to provide to the user. Time to discuss.

 - [ ] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -186,10 +186,7 @@ class Scene(MetadataObject):
         else:
             return set(self.attrs['sensor'])
 
-    def create_reader_instances(self,
-                                filenames=None,
-                                reader=None,
-                                reader_kwargs=None):
+    def create_reader_instances(self, filenames=None, reader=None, reader_kwargs=None):
         """Find readers and return their instances."""
         return load_readers(filenames=filenames,
                             reader=reader,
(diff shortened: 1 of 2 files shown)


djhoese:
After discussing this on slack and thinking more about it I'm going to close this without merging. The main use case was for my geo2grid project where having a bbox attribute for the user to use in the filename pattern would make it easier to include the bbox values in the output filename. @mraspaud pointed out that this information could/should probably come from the new AreaDefinition. Also if the user is specifying a latlon bbox on the command line, they can customize the filename themselves to include these values as well. Lastly, Martin and I talked about maybe putting `_cropped` on the end of the AreaDefinition name and/or in the description of the `AreaDefinition`. I've decided against this as it doesn't really solve any problems and may make users annoyed in the future when their area definition has a suffix they can't easily remove from the command line. I'm not necessarily against it, but don't think it is worth the time to add it.

PR #483, opened 2018-11-01
Add netcdf file version to end of ABI file patterns

djhoese:
Sometimes people decide to put `.nc4` as the extensions for their file patterns. This is a preventative measure when SIFT users will inevitably try to load ABI L1B data with these extensions.

 - [ ] Tests added 
 - [x] Tests passed 


djhoese:
I only changed a YAML file so coveralls is wrong.

PR #494, opened 2018-11-09
Fix end time calculation in nc_goes

sfinkens:
Currently, if the sector is unknown, a KeyError is thrown. The PR fixes this bug.

 - [x] Tests added 
 - [x] Tests passed 


sfinkens:
@djhoese Thanks for reviewing & merging!

PR #498, opened 2018-11-12
Make it possible to configure in-line composites

pnuu:
With this PR it will be possible to define composites in-line inside of another composites. This will deprecate several composites currently using `Dust` and `Airmass` compositors.

 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -434,8 +434,6 @@ class PSPRayleighReflectance(CompositeBase):
             suna = suna.data
             sunz = sunz.data
 
-        LOG.info('Removing Rayleigh scattering and aerosol absorption')
-
         # First make sure the two azimuth angles are in the range 0-360:
         sata = sata % 360.
         suna = suna % 360.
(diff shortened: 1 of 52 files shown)


satpy/composites/__init__.py
```diff
@@ -154,10 +153,21 @@ def _process_composite_config(self, composite_name, conf,
...
-                    # 'modifiers' being None otherwise it will be an empty
-                    # tuple
-                    item.setdefault('modifiers', None)
+                    # Handle in-line composites
+                    if 'compositor' in item:
+                        # Create an unique temporary name for the composite
+                        import uuid
+                        temp_comp_name = 'temp_' + str(uuid.uuid4())
+                        # Minimal composite config
+                        conf_tmp = {composite_type: {temp_comp_name: item}}
```
djhoese: Would it be possible to hash the composite arguments or something so it isn't unique? This way composites could be reused...eh maybe that's a bad idea. Actually that's the whole point of defining the dependencies not inline.

djhoese:
To be clear and document this could you show me what one of these inline compositors looks like?

pnuu:
I think I'm now finished with this, unless there are good suggestions. I didn't update the composite documentation, as it is basically missing and will require a full issue of its own.

djhoese:
My version of this PR shows that there is a conflict with compositor_tests? Probably from @mraspaud recent MODIS thin PR.

pnuu:
Thanks for pointing that out, rebased to current master.

djhoese:
@pnuu Is this ready to go? I know you asked me to do some performance testing, but other than that is this done?

djhoese:
Tests for these situations should probably be added to the `test_scene.py` module or somewhere similar, but these tests are admittedly out of control and complex.

pnuu:
I added a test that verifies the instrument specific composite overrides the one from `visir.yaml` also for the in-lined composites (the sub-composite prerequisites are using channel names instead of wavelengths).

PR #504, opened 2018-11-16
Change unnecessary warning messages to debug

djhoese:
There were a lot of warnings that always show up when doing certain types of processing. It was important when features were first developed, but as things have stabilized these types of messages are more for satpy developers rather than for users. If a user is curious why their product wasn't generated they will still enable debug logging to get the most information as possible.

Additionally there are a couple small changes in this PR:
 - Remove conda-forge workaround in travis for conda-forge/defaults incompatibility. Uses channel priority instead
 - Remove unnecessary 'abi' sensor in abi_l1b.yaml. The sensor is already added in the code.
 - Other small code style changes

 - [ ] Tests added 
 - [x] Tests passed 

### DIFF
--- a/satpy/enhancements/__init__.py
+++ b/satpy/enhancements/__init__.py
@@ -124,9 +124,6 @@ def crefl_scaling(img, **kwargs):
                                  attrs=band_data.attrs)
         return band_data
 
-    # mask reflectances that are below 0
-    bad_mask = (img.data < 0).any(axis=0)
-    img.data = img.data.where(~bad_mask)
     return apply_enhancement(img.data, func, separate=True)
 
 
--- a/satpy/enhancements/__init__.py
+++ b/satpy/enhancements/__init__.py
@@ -124,6 +124,9 @@ def crefl_scaling(img, **kwargs):
                                  attrs=band_data.attrs)
         return band_data
 
+    # mask reflectances that are below 0
+    bad_mask = (img.data < 0).any(axis=0)
+    img.data = img.data.where(~bad_mask)
     return apply_enhancement(img.data, func, separate=True)
 
 
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -892,7 +892,8 @@ class Scene(MetadataObject):
             missing = self.missing_datasets.copy()
             self._remove_failed_datasets(keepables)
             missing_str = ", ".join(str(x) for x in missing)
-            LOG.debug("The following datasets were not created: {}".format(missing_str))
+            LOG.warning("The following datasets were not created and may require "
+                        "resampling to be generated: {}".format(missing_str))
         if unload:
             self.unload(keepables=keepables)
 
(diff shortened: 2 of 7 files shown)


satpy/enhancements/__init__.py
```diff
@@ -124,6 +124,9 @@ def func(band_data, index=None):
                                  attrs=band_data.attrs)
         return band_data
 
+    # mask reflectances that are below 0
+    bad_mask = (img.data < 0).any(axis=0)
```
mraspaud: Won't that create holes in composite in very dark places ?

satpy/scene.py
```diff
@@ -900,8 +892,7 @@ def load(self, wishlist, calibration=None, resolution=None,
             missing = self.missing_datasets.copy()
             self._remove_failed_datasets(keepables)
             missing_str = ", ".join(str(x) for x in missing)
-            LOG.warning(
-                "The following datasets were not created: {}".format(missing_str))
+            LOG.debug("The following datasets were not created: {}".format(missing_str))
```
mraspaud: I think this is pretty important info, I'd like to keep this as a warning. What typically happens is that at the next step the user tries to show or save the image and it crashes. Having the reason visible above is a good thing.

mraspaud:
I'm ok with PRs fixing multiple small things, but please describe them in the description. The description on this PR makes it sound like there is only one change.

satpy/enhancements/__init__.py
djhoese: You are absolutely right. This turned in to a WIP after thinking more about it and testing things. That's why I asked the question about this on slack. The cira stretch and the crefl stretch behave differently regarding negative reflectances (cira does log on them so they become NaN). I am starting to think that the clipping should just be handled like in the above line. That is, if any reflectance is negative, clip all of them to 0 (black). Otherwise, this clipping could be made part of the rayleigh correction or the ratio sharpening code. If the bands don't share "special" pixels together then there are some weird effects on the terminator limb (like bright cyan pixels).

Bottom line: At night the dark parts of the earth are showing up as negative reflectances, should they be black or transparent. A negative reflectance isn't really a real reflectance, right?

satpy/scene.py
djhoese: Hhhhmm how about INFO level? Warning seems negative and this isn't really a bad thing necessarily. A little iffy. Either way I can add information like "The following datasets were not created, but will be created after resampling: {}". If this was the message does that change your mind about what the log level should be?

djhoese:
I completely agree. I honestly don't remember clicking submit on this. I wasn't done with it. I also didn't plan for this to include so many small things but I ran `git status` and ended up having a lot of little things hanging around in my local repository.

djhoese:
With the changes currently in this PR (masking out negative reflectances):

image

Previous results by clipping to 0 and 1 in crefl enhancement per band (note the bright cyan on the terminator):

image

Changing it to be as described and setting any pixels with negative reflectances to black (looks like there are still blue pixels, just less of them):

image

Final conclusion: After looking closely at all images the bright cyan pixels are present in all of them. They are just harder to see in some of them or are in different quantities.

PR #507, opened 2018-11-19
Change default enhancement for reflectance data to square root (gamma 1.5)

djhoese:
Having the default enhancement require the min/max of the entire array
kills performance. IMO this is a suitable default. In the future I'd like to do the same for brightness temperature data or at least *most* BT data, but that one will be harder to agree on. I'm not sure where to document this other than release notes.

 - [ ] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/satpy/etc/enhancements/generic.yaml
+++ b/satpy/etc/enhancements/generic.yaml
@@ -6,6 +6,15 @@ enhancements:
     - name: stretch
       method: &stretchfun !!python/name:satpy.enhancements.stretch ''
       kwargs: {stretch: linear}
+  reflectance_default:
+    standard_name: toa_bidirectional_reflectance
+    operations:
+    - name: linear_stretch
+      method: !!python/name:satpy.enhancements.stretch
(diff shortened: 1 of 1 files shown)


djhoese:
ABI Full Disk OK with you?

mraspaud:
sure

djhoese:
Channel 3 - Original:

image

Channel 3 - Square root (new):

image

Channel 2 - Original:

image

Channel 2 - Square root (new):

image

Forgot to profile, let me do that now. Interestingly the original scaling produces smaller geotiffs because it is able to compress the entire image instead of only compressing chunks.

djhoese:
Profiling results - old (3m16s total, dask computation 2m19s):

image

Profile results - new (2m2s total, dask computation 1m23s):

image

So more than twice as much memory, 1 minute extra time, and a "better" image. Also note that I was looking at the AHI space mask images and watching doctor who while generating these so...give or take a few seconds on the overall time.

PR #515, opened 2018-11-24
Fix all flake8 errors in satpy package code

### LINKED ISSUE #467: Flake8-ify all of satpy
Our pull request templates currently ask the user to only check their diff for flake8 issues. This command is overly complicated because a lot of satpy doesn't conform to PEP8 standards. We should try to get `flake8 satpy` to run without any messages.

djhoese:
I ran `flake8 satpy` from the root of the repository and got a lot of errors. I walked through these errors and fixed all but one of them. All of them were style changes except one.

The one I didn't fix is the `li_l2` reader which hasn't been migrated to xarray/dask yet. Instead of removing the failing import I added the reader to the ignored paths in setup.cfg.

The only non-style change was in the mitiff writer (@TAlonglong) where there were bare except statements (`except:`). I changed these to `except (KeyError, ValueError, RuntimeError):` to make flake8 happy and hopefully not change any functionality of the writer. @TAlonglong please verify that these exceptions are what you expected. Catching *all* errors by doing `except:` or `except Exception:` is considered bad practice.

Lastly, some of these changes were just personal preference changes, but still changes that I think go along with the conventions that @pytroll/core have talked about in the past on slack. Some were also to reduce number of broken lines now that we've agreed on 120 max line length.

 - [x] Closes #467 

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,3 +1,4 @@
 linters:
   flake8:
-    max-line-length: 120
+    python: 3
+    config: setup.cfg
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -27,6 +27,7 @@
 import logging
 import os
 import time
+import warnings
 from weakref import WeakValueDictionary
 
 import numpy as np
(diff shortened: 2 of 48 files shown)


djhoese:
Wow I really thought I would have gotten more than 0.06% coverage increase.

PR #517, opened 2018-11-25
Travis CI: Add flake8 tests for syntax errors, undefined names

cclauss:
__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,12 @@ matrix:
 install:
     - git clone --depth 1 git://github.com/astropy/ci-helpers.git
     - source ci-helpers/travis/setup_conda.sh
+    - pip install flake8
+before_script:
+    # stop the build if there are Python syntax errors or undefined names
+    - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+    # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+    - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 script:
 - coverage run --source=satpy setup.py test
 - coverage run -a --source=satpy -m behave satpy/tests/features --tags=-download


djhoese:
Interesting idea. We currently use stickler for this type of functionality to notify contributor's inside their PR. I guess I wouldn't be against this, but I know @mraspaud had some objections to mixing style checks and tests in the past. Also see #515 for fixing all flake8 issues.

cclauss:
These are _not_ style checks.  These are syntax errors and missing variable definitions.  As the commit message clearly states these errors have the potential to halt the runtime.

cclauss:
Output: https://travis-ci.org/pytroll/satpy/jobs/459492483#L1121

djhoese:
I understand that the way you have called flake8 here will catch situations that would stop code from operating. I did read your commit messages and comments in your changes. No need to get upset over a disagreement.

That said, flake8 is, at the end of the day, a style enforcement tool. Any syntax errors should be caught by unit tests, however, we are lacking some tests at this point. Some of the fixes you made in #516 were for defunct or rarely used modules which is the best excuse I have for their issues. 😉 

All of that said, I am not against adding this, but I am not the only satpy developer and the @pytroll/core devs will discuss this later. Thanks for the contributions.

djhoese:
So why close this one?

PR #519, opened 2018-11-26
Add xarray/dask bilinear resampling

### LINKED ISSUE #518: Make bilinear interpolation dask/xarray friendly
Bilinear interpolation doesn't work with the dask/xarray version of SatPy, so it should be converted.

It might be that pyresample needs some updates, too. If this is the case, I'll add a relevant issue there.

pnuu:
This PR adds bilinear interpolation by interfacing the corresponding functionality from pyresample. Note also https://github.com/pytroll/pyresample/issues/148 which is actually the bottleneck killing the performance.

 - [x] Closes #518 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/satpy/resample.py
+++ b/satpy/resample.py
@@ -606,10 +606,8 @@ class BilinearResampler(BaseResampler):
 
         del kwargs
 
-        source_geo_def = mask_source_lonlats(self.source_geo_def, mask)
-
         if self.resampler is None:
-            kwargs = dict(source_geo_def=source_geo_def,
+            kwargs = dict(source_geo_def=self.source_geo_def,
                           target_geo_def=self.target_geo_def,
                           radius_of_influence=radius_of_influence,
                           neighbours=32,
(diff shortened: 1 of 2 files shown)


satpy/resample.py
```diff
@@ -890,3 +900,28 @@ def resample_dataset(dataset, destination_area, **kwargs):
     new_data.attrs.update(area=destination_area)
 
     return new_data
+
+
+def mask_source_lonlats(source_def, mask):
```
djhoese: This function isn't supposed to be needed anymore now that pykdtree's `query` method accepts a `mask` keyword argument.

pnuu: Seemes to work without it, so great spot.

djhoese:
If you can move the caching to the base resampler but also provide kwargs somewhere (resamplers only accept two positional arguments in `__init__` right now) for turning certain caching on/off that would be good.

Should cache_dir be an init argument? Or a precompute argument only?

I think disk-caching should be off by default, memory caching on by default.

pnuu:
I agree that disk-caching should be off by default, like it is now. And memory-caching can be on by default, but I'd like a way to disable it: it will be necessary for large source/target areas and different resolutions and/or satellites.
For the rest, let's have another issue/PR for them, "refactor resamplers" or something.

PR #539, opened 2018-12-09
Hdf5 msg reader

BENR0:
Adds reader for MSG data in hdf5 format

Tests are not included yet


djhoese:
@BENR0 What's the status on this pull request?

BENR0:
Still didn't have the time to add tests. Apart from that from my point of view in order to be able to merge this the reader name needed to be changed to the official naming pattern and the HRV area extent is currently not handled correctly. In general there should be some refactoring possible to share area extent calculations with the other Seviri readers.
Since in our working group we still have some Seviri data as hdf files we are currently using this in all our Seviri processing chains.

PR #545, opened 2018-12-12
Add configurable parameters to solar zenith correctors

djhoese:
## Main Purpose

This PR adds YAML-configurable parameters to the two solar-zenith correctors/modifiers. The three parameters that are now available in both are:

* **correction_limit** (float): Maximum solar zenith angle to apply the correction in degrees. Pixels beyond this limit have a constant correction applied. Default 88.
* **min_sza** (float): Minimum solar zenith angle in degrees that is considered valid and correctable. Default 0.0.
* **max_sza** (float): Maximum solar zenith angle in degrees that is considered valid and correctable. Default 90.0.

## Behavior Change

Currently, this PR has changed the default values for these parameters based on some tests I did with ABI natural_color and true_color RGBs. Previously, the equivalent value for `max_sza` was `None` due to some undesired (in my opinion) usage of the fill values in the correction code. The way the corrections were previously set up is that for data past 88 degrees a constant correction is applied to the data regardless of SZA. So for SZA 0 to 88 the correction was applied normally, for 88-90+ a constant correction was being applied, even if SZA was NaN. In this PR I've changed it to be corrected from 0 to 88, constant from 88-90 (`max_sza`), and NaN after that. This allows satpy and the user to say "anything past this SZA isn't useful data". The main reason for this change was due to noticing that ABI natural color RGBs were having night time pixels "corrected", increasing the "useless" data values.

Note that the user can get the same behavior as before by specifying `max_sza: !!null` in the YAML for the `sunz_corrected`.

I am ok not including this behavior change (I'd still like it) because I can have it customized in my own projects (Polar2Grid/Geo2Grid).

## Examples

NOTE: The default True Color enhancement sets all negative reflectances to NaN/transparent.

### GOES-16 ABI True Color - Old

image

### GOES-16 ABI Natural Color - Old

image

### GOES-16 ABI True Color - 0 to 88 to 90

image

### GOES-16 ABI Natural Color - 0 to 88 to 90

image

### GOES-16 ABI True Color - 0 to 88 to 92

image

### GOES-16 ABI Natural Color - 0 to 88 to 92

image

### GOES-16 ABI True Color - 0 to 88 to 95

Same as 0 to 88 to 92 image.

### GOES-16 ABI Natural Color - 0 to 88 to 95

image

## Miscellaneous

I originally started working on this because the natural color image shows bright green and red pixels in the night time data. I was able to prove that negative reflectances were part of the problem and were being made worse by the SZA corrector and by the sharpening code. However, it didn't get rid of all of it and cutting off the SZA correction seemed to make more sense.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -313,14 +313,14 @@ class SunZenithCorrectorBase(CompositeBase):
 
     coszen = WeakValueDictionary()
 
-    def __init__(self, min_sza=0.0, max_sza=90.0, **kwargs):
+    def __init__(self, min_sza=0.0, max_sza=95.0, **kwargs):
         """Collect custom configuration values.
 
         Args:
             min_sza (float): Minimum solar zenith angle in degrees that is
                 considered valid and correctable. Default 0.0.
             max_sza (float): Maximum solar zenith angle in degrees that is
-                considered valid and correctable. Default 90.0.
+                considered valid and correctable. Default 95.0.
 
         """
         self.min_sza = np.cos(np.deg2rad(min_sza)) if min_sza is not None else None
@@ -370,7 +370,13 @@ class SunZenithCorrectorBase(CompositeBase):
 
 
 class SunZenithCorrector(SunZenithCorrectorBase):
-    """Standard sun zenith correction, 1/cos(sunz)."""
+    """Standard sun zenith correction using ``1 / cos(sunz)``.
+
+    In addition to adjusting the provided reflectances by the cosine of the
+    solar zenith angle, this modifier also forces all reflectances beyond a
+    solar zenith angle of `max_sza` to 0.
+
+    """
 
     def __init__(self, correction_limit=88., **kwargs):
         """Collect custom configuration values.
@@ -382,7 +388,7 @@ class SunZenithCorrector(SunZenithCorrectorBase):
             min_sza (float): Minimum solar zenith angle in degrees that is
                 considered valid and correctable. Default 0.0.
             max_sza (float): Maximum solar zenith angle in degrees that is
-                considered valid and correctable. Default 90.0.
+                considered valid and correctable. Default 95.0.
 
         """
         self.correction_limit = correction_limit
@@ -398,6 +404,10 @@ class EffectiveSolarPathLengthCorrector(SunZenithCorrectorBase):
 
     (2006): https://doi.org/10.1175/JAS3682.1
 
+    In addition to adjusting the provided reflectances by the cosine of the
+    solar zenith angle, this modifier also forces all reflectances beyond a
+    solar zenith angle of `max_sza` to 0 to reduce noise in the final data.
+
     """
 
     def __init__(self, correction_limit=88., **kwargs):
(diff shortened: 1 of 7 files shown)


djhoese:
I made some images with the DayNightCompositor and an ABI true_color and I think this may need to be changed to 92 to get a small transition between the two images.

djhoese:
Just so it is documented I'm reiterating what was discussed on slack here. @mraspaud and @pnuu thought it would be better for the "night" data to be set to 0 rather than NaNs. This serves my project just as well so I am ok with it. Martin also said that 95 degrees might be the best cutoff.

However, the resulting natural color image for this looks...ugly. I am going to try testing gradual dropoffs on the correction so that the data goes to 0 on the edge of the image. I'll have examples later.

djhoese:
The following natural color images were created with gradual SZA corrections between 88 and 95 degrees. I tried a linear one but it didn't do a good enough job of cutting off the noise in the night data so I'm not including it here. The "Natural Log (before inversion)" code is what is currently committed and what I think looks the best. Due to the angle limits these different methods have little to no effect on the true color images.

## Square root 

image

## Square root (before inversion)

image

## Natural Log

image

## Natural Log (before inversion)

image

mraspaud:
My favourite is natural log before inversion

satpy/etc/enhancements/generic.yaml
```diff
@@ -75,6 +75,7 @@ enhancements:
     operations:
     - name: stretch
       method: *stretchfun
+      #      kwargs: {stretch: crude, min_stretch: 0.0, max_stretch: 1.0}
```
mraspaud: This can be removed

satpy/utils.py
```diff
@@ -208,31 +209,42 @@ def proj_units_to_meters(proj_str):
...
+    angles, the correction is the same as at the ``limit`` if ``max_sza``
+    is `None`. The default behavior is to gradually reduce the correction
+    past ``limit`` degrees up to ``max_sza`` where the correction becomes
+    0. Both ``data`` and ``cos_zen`` should be 2D arrays of the same shape.
 
     """
 
     # Convert the zenith angle limit to cosine of zenith angle
-    limit = xu.cos(xu.deg2rad(limit))
+    limit_rad = np.deg2rad(limit)
```
mraspaud: I assume `xu` and `np` are equivalent here, since which version of xarray and numpy ? Is it something we need to indicate in setup.py ?

satpy/composites/__init__.py
```diff
@@ -313,17 +313,30 @@ class SunZenithCorrectorBase(CompositeBase):
 
     coszen = WeakValueDictionary()
 
+    def __init__(self, min_sza=0.0, max_sza=95.0, **kwargs):
```
mraspaud: I think `min_sza` should be remove, it isn't something that is used in practice.

satpy/utils.py
djhoese: http://xarray.pydata.org/en/stable/api.html#universal-functions

Mainly:

```
With recent versions of numpy, dask and xarray, NumPy ufuncs are now supported directly on all xarray and dask objects. This obliviates the need for the xarray.ufuncs module, which should not be used for new code unless compatibility with versions of NumPy prior to v1.13 is required.
```

mraspaud: So we need to fix setup.py to read `numpy >= 1.13.0` I suppose

djhoese: Yes, I suppose you are right.

djhoese:
I'm not simply a terrible person. Martin and I did discuss this on slack so even though he hasn't officially approved this PR he said it was ok for me to merge this.

PR #553, opened 2018-12-16
Fix DifferenceCompositor when areas are incompatible

### LINKED ISSUE #552: GOES Composites failling
**Describe the bug**
If I try to create a composite based upon a GOES-16 image then I get an error message saying that there is no matching dataset

**To Reproduce**

```python
# Your code here
from		satpy			 import Scene
from		glob				 import glob
import	os

BASE_DIR		=	{my_in_dir}
OUT_DIR		=	{my_out_dir}

count			=	0
fnames	=	glob(os.path.join(BASE_DIR, 'OR*-Rad*-*C01*.nc'))
fnames.sort()

incount			=	1
f		=	fnames[0]

infs	=	glob(f.replace('C01', 'C*')[:len(BASE_DIR) + 50] + '*.nc')
scn	=	Scene(reader='abi_l1b', filenames=infs)

print('	-	Processing',incount,'of',len(fnames))
scn.load(['convection'])
new	=	scn.resample(scn.min_area(), resampler='native')
outf	=	OUT_DIR+'GOES_'+str(incount).zfill(4)+'.png'

new.save_dataset('convection', outf)

**Expected behavior**
A PNG file to be created in OUT_DIR that shows the generated convection composite for a given timeslot.

**Actual results**
  File "/network/aopp/apres/users/proud/Progs/conda/lib/python3.6/site-packages/satpy/readers/__init__.py", line 301, in __getitem__
KeyError: 'convection'

  File "/network/aopp/apres/users/proud/Progs/conda/lib/python3.6/site-packages/satpy/scene.py", line 1043, in save_dataset
  File "/network/aopp/apres/users/proud/Progs/conda/lib/python3.6/site-packages/satpy/scene.py", line 616, in __getitem__
    return self.datasets[key]
  File "/network/aopp/apres/users/proud/Progs/conda/lib/python3.6/site-packages/satpy/readers/__init__.py", line 303, in __getitem__
    key = self.get_key(item)
  File "/network/aopp/apres/users/proud/Progs/conda/lib/python3.6/site-packages/satpy/readers/__init__.py", line 292, in get_key
    best=best, **dfilter)
  File "/network/aopp/apres/users/proud/Progs/conda/lib/python3.6/site-packages/satpy/readers/__init__.py", line 243, in get_key
KeyError: "No dataset matching 'DatasetID(name='convection', wavelength=None, resolution=None, polarization=None, calibration=None, level=None, modifiers=None)' found"



**Environment Info:**
Ubuntu 14.04 LTS + Ubuntu 18.04
Python 3.6, SatPy 0.10.0

**Additional context**
All 16 GOES NetCDF files are available in the BASE_DIR, missing channels is not the problem.

djhoese:
Fix #552

The `DifferenceCompositor` composite class does not do `check_areas` which causes issues in certain situations if the resolutions are different. The numpy operations would fail because the arrays were different shapes which would be seen by satpy as "this composite can't be made" rather than "we should try to make this composite after resampling".

 - [x] Closes #552 
 - [x] Tests added 
 - [x] Tests passed 

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -677,8 +677,8 @@ class DifferenceCompositor(CompositeBase):
 
     def __call__(self, projectables, nonprojectables=None, **info):
         if len(projectables) != 2:
-            raise ValueError("Expected 2 datasets, got %d" %
-                             (len(projectables), ))
+            raise ValueError("Expected 2 datasets, got %d" % (len(projectables),))
+        projectables = self.check_areas(projectables)
         info = combine_metadata(*projectables)
         info['name'] = self.attrs['name']
 
--- a/satpy/tests/compositor_tests/__init__.py
+++ b/satpy/tests/compositor_tests/__init__.py
@@ -275,6 +275,58 @@ class TestSunZenithCorrector(unittest.TestCase):
         np.testing.assert_allclose(res.values, np.array([[66.853262, 68.168939], [66.30742, 67.601493]]))
 
 
+class TestDifferenceCompositor(unittest.TestCase):
+
+    def setUp(self):
+        """Create test data."""
+        from pyresample.geometry import AreaDefinition
+        area = AreaDefinition('test', 'test', 'test',
+                              {'proj': 'merc'}, 2, 2,
+                              (-2000, -2000, 2000, 2000))
+        attrs = {'area': area,
+                 'start_time': datetime(2018, 1, 1, 18),
+                 'modifiers': tuple(),
+                 'resolution': 1000,
+                 'name': 'test_vis'}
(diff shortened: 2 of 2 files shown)


djhoese:
@pnuu I think most compositors are ok because they pass things to the GenericCompositor which does the check. I've added the checks to the two new compositors you made (sandwich and luminance) which looked like they were using `combine_metadata` to decide if the areas were the same somehow.

djhoese:
Also @pnuu I added a FIXME to the SandwichCompositor because there is a for loop over the bands that looks like it can be removed in favor of a simpler `luminance * rgb_image`.

PR #554, opened 2018-12-16
Fix AreaDefinition usage to work with newer versions of pyresample

djhoese:
AreaDefinitions argument names have changed. A lot of my usages of `AreaDefinition` specified `x_size` and `y_size` but these have been changed to `width` and `height` in https://github.com/pytroll/pyresample/pull/138 to match the `areas.yaml` format. This PR fixes all uses of those arguments by name.

This PR should probably only be merged after that pyresample PR is merged and released to be sure I fixed everything.

 - [ ] Tests added 
 - [x] Tests passed 


djhoese:
I tested this with the current pyresample master in a separate branch and all tests pass. That means that current tests pass with both the old version of pyresample and the soon-to-be-released version of pyresample.

PR #560, opened 2018-12-24
Fix available_composite_ids including inline comp dependencies

djhoese:
This PR fixes inline composite dependencies being listed in `available_composite_ids/names` and the `all_composite_ids/names`. The point of inline compositors is to not have to put as much work in to "officially" defining new composites and they shouldn't be needed by 99% of users. It can be confusing to see these dependencies (ex. `airmass_dep_0`) in the list coming from `Scene.available_composite_ids`.

This PR prefixes these inline dependencies with a `_` and then ignores composites with that name when listing them in the Scene's methods.

 - [ ] Tests added 
 - [x] Tests passed 

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -148,7 +148,7 @@ class CompositorLoader(object):
                     # Handle in-line composites
                     if 'compositor' in item:
                         # Create an unique temporary name for the composite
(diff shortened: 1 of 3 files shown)


djhoese:
The loss of test coverage is in a file I did not edit so not much I can do there. I'm going to merge this to rush a 0.11.1 release since the other @pytroll/core devs are on holiday.

PR #564, opened 2019-01-01
Add new ABI composites

jhbravo:
- [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


jhbravo:
Hi @djhoese 
renewing the combinations,
About the documentation, what should I do?
I use the same combination style that is in the yaml files. I guess the intention is not to use a function for each combination that requires a band difference
Also I made a branch for this, do not use the master

satpy/etc/composites/abi.yaml
```diff
@@ -147,3 +147,333 @@ composites:
...
+        - name: C05
+        - name: C02
+    standard_name: convection
+
+## Cloud Phase Distinction or Day Cloud Phase (name from NASA SPORT) or Cloud Type RGB(name from EUMETSAT)
+  cloud_phase_distinction_raw:
+    compositor: !!python/name:satpy.composites.SelfSharpenedRGB
+    prerequisites:
+    - name: C13
+    - name: C02
```
djhoese: To properly use `SelfSharpenedRGB` I think you need to include `high_resolution_band: green` as a keyword argument (same indentation level as `standard_name` below). Otherwise it will try to do the sharpening based on the "red" band by default which is not the high resolution band.

There is at least this use of `SelfSharpenedRGB` and the composite below. Maybe others.

jhbravo:
1.- No, none of `visir.yaml`

2.- I use recipes of `composites/ahi.yaml`
from this same file is where you copy `_eum`, some
for example airmass_eum I put it that way because it uses 10.8 and therefore I use C14 instead of C13 that is used in ABI
Do you think it is redundant or unnecessary? I can erase it
that of `WIS *` is described below in that same yaml
`WIS (WMO Information System)`
This is in the file from which I obtained the recipes
It occurs to me that what I can do is write the link to the file, and that anyone can see it, since they are Internet files
I can make the modifications that you and the core-developers believe necessary

3.- I wrote only in `enhancements/generic.yaml`, since I base myself on recipes of `composites/ahi.yaml` and they are written in this file,

4.- I remain aware that how this description can be generated

djhoese:
For points 2 and 4, I think it would be best to add a `description: ` field in the recipes and provide a string description of the recipe. It could include things like:

* whose recipe it is
* maybe why it is different from the default EUMETSAT recipe (if one exists)
* what groups use it (WIS) and the full expanded version of those acronyms if possible.

I don't think the URL for the recipe needs to be included in the `description` of the recipe but putting a comment right above the recipe `# http://...` would be greatly appreciated. As for the `description` string, if it needs to span multiple lines you could try using some of the methods described [here](https://stackoverflow.com/a/21699210/433202).

I also wouldn't mind input from @pytroll/core devs on this. Is there a better name than `description`?

> 3.- I wrote only in enhancements/generic.yaml, since I base myself on recipes of composites/ahi.yaml and they are written in this file,

If the recipe does not differ from the existing recipes then it would be best to reduce duplication and confusion. For example your enhancement for `dust_abi` seems to match the existing `dust_default` enhancement. I would prefer anything that matches existing enhancement configurations be removed and any enhancements that are specific to ABI get moved to `enhancements/abi.yaml`. If there are enhancements that are specific to AHI composites then I wouldn't be against moving those to `enhancements/ahi.yaml` even if that means they are replicated between `ahi.yaml` and `abi.yaml`.

As @pnuu mentioned in #429, I think it would be best to remove the `_abi` from the `ash/dust/fog` recipe names (so name them `ash`, `dust`, `fog`). If they are meant to produce the same or similar image as the default recipe then I think most users would be fine with the small change in channels/wavelengths used.  I understand that someone may want to compare the differences between the recipes, but I don't think that is the common case (@pytroll/core). As long as it is mentioned in the recipe `description` and/or comments around the recipe.

Wow, this got long, sorry.

satpy/etc/composites/abi.yaml
```diff
@@ -147,3 +147,345 @@ composites:
...
+    - name: C15
+    standard_name: ir_overview
+## Tropical Airmass RGB (name from EUMETSAT)
+  airmass_tropical:
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+    - compositor: !!python/name:satpy.composites.DifferenceCompositor
+      prerequisites:
+        - name: C10
+        - name: C08
```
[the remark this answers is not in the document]
jhbravo: hi @djhoese 

What I did was to use a series of recipes that exist and I just have the link to the file.

http://www.virtuallab.bom.gov.au/files/3114/7884/4308/NewRGBProductsNov2016RFGmeeting.pdf

that's where I base myself on most RGB compositions that I adapt to combine the bands using satpy

when I generate some other combination I will also leave the link to the file where I get the recipes

any comment I'm watching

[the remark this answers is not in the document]
djhoese: Based on the recipe on slide 6 in that PDF the Red and Green channels should be flipped (C10 - C08) instead of (C08 - C10).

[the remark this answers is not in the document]
jhbravo: I tried to follow as the recipes say

satpy/etc/composites/abi.yaml
```diff
@@ -147,3 +147,345 @@ composites:
...
+
+###
+## Airmass using Channel14 RGB (own)
+  airmass_eum:
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+    - compositor: !!python/name:satpy.composites.DifferenceCompositor
+      prerequisites:
+      - name: C08
+      - name: C10
```
djhoese: It was pointed out to me that my other comment was wrong and that your tropical airmass RGB recipe is correct. I was actually talking about this `airmass_eum` (which for some reason I thought was the tropical one). @jhbravo any idea where this `airmass_eum` came from?

satpy/etc/composites/abi.yaml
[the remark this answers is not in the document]
djhoese: As mentioned in my other comment, this tropical airmass RGB recipe is correct. I'm confused about the other one.

satpy/etc/composites/abi.yaml
jhbravo: that of `_eum`  I copied it from a couple of combinations that are in `enhancements/ahi.yaml`
and my intention was that you could make the **original combination of eumetsat** with the band 10.8 and also with the **advanced sensors ABI/AHI** with the band 10.3
I was really just curious to compare

mraspaud: It could be in the case of the eumetsat recipe that the enhancement are doing the inversion.

jhbravo:
hi @djhoese 
could you please provide me with an example of how I can add the description and / or document of the rgb combinations
where would it be? in `satpy/etc/composites/abi.yaml`
or in a separate file?

djhoese:
Yes, in the `abi.yaml` file. I mentioned this in my previous comment a while ago: https://github.com/pytroll/satpy/pull/564#issuecomment-450977071

jhbravo:
@djhoese 
Description and / or documentation of the combinations rgb for ABI.
what do you think?

satpy/etc/composites/abi.yaml
```diff
@@ -147,3 +147,261 @@ composites:
...
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+    - name: C13
+    - name: C08
+    - name: C10
+    standard_name: water_vapors1
+
+  water_vapors2:
+    ## Differential Water Vapor (name from NASA SPORT)
+    ## RGB Differential Water Vapor recipe source: http://rammb.cira.colostate.edu/training/visit/quick_guides/Simple_Water_Vapor_RGB.pdf
```
djhoese: The PDF should be http://rammb.cira.colostate.edu/training/visit/quick_guides/QuickGuide_GOESR_DifferentialWaterVaporRGB_final.pdf right?

satpy/etc/composites/abi.yaml
```diff
@@ -147,3 +147,261 @@ composites:
...
+    prerequisites:
+    - name: C03
+      modifiers: [sunz_corrected]
+    - name: C05
+      modifiers: [sunz_corrected]
+    - name: C07
+      modifiers: [nir_reflectance]
+    standard_name: snow
+
+  day_microphysics_abi:
```
djhoese: Could you move this one above `snow` so it is just below the other `day_microphysics_eum`?

satpy/etc/composites/abi.yaml
```diff
@@ -147,3 +147,261 @@ composites:
...
+    ## source: https://www.eumetsat.int/website/home/Images/ImageLibrary/DAT_2861499.html
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+    - name: C05
+      modifiers: [sunz_corrected]
+    - name: C06
+      modifiers: [sunz_corrected]
+    - name: C02
+      modifiers: [sunz_corrected, rayleigh_corrected]
+    standard_name: natural
```
djhoese: The `natural` composite is not consistently called `natural_color` in Satpy. Some sensors had `natural` some had `natural_color`, this was fixed a couple versions ago.

satpy/etc/composites/abi.yaml
```diff
@@ -147,3 +147,261 @@ composites:
...
+
+  cloud_phase_raw:
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+    - name: C05
+    - name: C06
+    - name: C02
+    standard_name: natural
```
djhoese: Same here, `natural_color`.

satpy/etc/enhancements/generic.yaml
```diff
@@ -783,3 +783,97 @@ enhancements:
           stretch: crude
           min_stretch: [0, 0, 0]
           max_stretch: [1, 1, 1]
+
+  ash_abi:
```
djhoese: I'd like all ABI specific enhancements to go in `enhancements/abi.yaml`.

mraspaud:
Looks good. Shouldn't references by a sub item of description?

djhoese:
My thought was that `description` is a string and references is a dictionary. That way, when documenting composites in the sphinx docs we could have a string description for each composite and a table or list with the references (the link text is the key of the references dictionary and the URL is the value of the references dictionary).

satpy/etc/composites/abi.yaml
jhbravo: yes I already corrected it

satpy/etc/composites/abi.yaml
jhbravo: It's ready

satpy/etc/composites/abi.yaml
jhbravo: It's ready

satpy/etc/composites/abi.yaml
jhbravo: It's ready

satpy/etc/enhancements/generic.yaml
jhbravo: It's ready

jhbravo:
using the same logic, I could place in `etc/readers/abi_l1b.yaml`
```
  C01:
    description: >
      ABI Band 1 (0.47 µm)
    references:
      CIRA Quick Guide: http://cimss.ssec.wisc.edu/goes/OCLOFactSheetPDFs/ABIQuickGuide_Band01.pdf
    name: C01
    wavelength: [0.450, 0.470, 0.490]
        ...
```

djhoese:
> using the same logic, I could place in etc/readers/abi_l1b.yaml

That is risky, but I like the idea. It definitely shouldn't go in this PR. I'd rather discuss this being done in the reader configs in a separate issue. It could easily start conflicts "why are you advertising CIRA's training guides instead of organization X".

jhbravo:
> > using the same logic, I could place in etc/readers/abi_l1b.yaml
> 
> That is risky, but I like the idea. It definitely shouldn't go in this PR. I'd rather discuss this being done in the reader configs in a separate issue. It could easily start conflicts "why are you advertising CIRA's training guides instead of organization X".

Ok, it's fine

jhbravo:
I found other guides, which are on the [GOES-R page](https://www.goes-r.gov/education/ABI-bands-quick-info.html) has a description of the bands
maybe that could have no conflict to be a description of GOES-R

```
  C01:
    description: >
      GOES-R ABI Band 1 (“Blue” visible)
    references:
      GOES-R Quick Guide: https://www.goes-r.gov/education/docs/ABI-bands-FS/ABI%20Band%201%20Fact%20Sheet_Revised%202.24.15.pdf
    name: C01
    wavelength: [0.450, 0.470, 0.490]
        ...
```
and as an additional resource
https://www.goes-r.gov/mission/ABI-bands-quick-info.html
They are the same as I put before, apparently they are not necessarily from CIRA, the quick guide does not have its logo, only NOAA and NASA and GOES-R

adybbroe:
@djhoese I like (prefer) your proposal on description attribute and a reference dictionary

mraspaud:
@jhbravo travis says there is a yaml error in the files, can you check it out?

mraspaud:
here is the error:
```
E                   yaml.scanner.ScannerError: while scanning a simple key
E                     in "/home/travis/build/pytroll/satpy/satpy/etc/composites/abi.yaml", line 617, column 7
E                   could not find expected ':'
E                     in "/home/travis/build/pytroll/satpy/satpy/etc/composites/abi.yaml", line 618, column 5
```

jhbravo:
> @jhbravo travis says there is a yaml error in the files, can you check it out?

Thnks @mraspaud, I was see this error in mi local repo and corrected it, but I forgot update it here, sorry

mraspaud:
@jhbravo Thanks for the PR and sorry for the delay in getting it reviewed and merged!

PR #574, opened 2019-01-12
Add code of conduct

djhoese:
See https://github.com/pytroll/pytroll.github.io/pull/2 for details.

Same options apply to this PR as that one. Mainly, if any of you (@mraspaud @adybbroe @pnuu) do not want to be contacted about violations let me know.


djhoese:
I thought the README links would work, but doesn't look like it. I'll fix it after the fact if they don't work when this is merged.

PR #582, opened 2019-01-22
Add 'reduce_data' keyword argument to disable cropping before resampling

pnuu:
This PR adds a way to disable data reduction before resampling. This is a work-around for #383 that might be useful even after fixing pyresample.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented 

Usage when calling resampling:
```python
lcl = glbl.resample(..., reduce_data=False)
```
The default is to apply data reduction.

Not sure where to document this, as resampling documentation is coming direct from resample.py docstrings, and this fix happens in scene.py. Suggestions?

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -900,7 +900,8 @@ class Scene(MetadataObject):
         if unload:
             self.unload(keepables=keepables)
 
-    def _resampled_scene(self, new_scn, destination_area, **resample_kwargs):
+    def _resampled_scene(self, new_scn, destination_area, reduce_data=True,
+                         **resample_kwargs):
         """Resample `datasets` to the `destination` area."""
         new_datasets = {}
         datasets = list(new_scn.datasets.values())
@@ -933,7 +934,7 @@ class Scene(MetadataObject):
             LOG.debug("Resampling %s", ds_id)
             source_area = dataset.attrs['area']
             try:
-                if resample_kwargs.get('reduce_data', True):
+                if reduce_data:
                     slice_x, slice_y = source_area.get_area_slices(
                         destination_area)
                     source_area = source_area[slice_y, slice_x]
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -968,7 +968,8 @@ class Scene(MetadataObject):
                 replace_anc(res, pres)
 
     def resample(self, destination=None, datasets=None, generate=True,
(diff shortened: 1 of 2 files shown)


satpy/scene.py
```diff
@@ -933,13 +933,16 @@ def _resampled_scene(self, new_scn, destination_area, **resample_kwargs):
...
+                if resample_kwargs.get('reduce_data', True):
+                    slice_x, slice_y = source_area.get_area_slices(
+                        destination_area)
+                    source_area = source_area[slice_y, slice_x]
+                    dataset = dataset.isel(x=slice_x, y=slice_y)
+                    assert ('x', source_area.x_size) in dataset.sizes.items()
+                    assert ('y', source_area.y_size) in dataset.sizes.items()
+                    dataset.attrs['area'] = source_area
+                else:
+                    LOG.info("Data reduction disabled by the user")
```
djhoese: I don't think a log message is needed here since the user explicitly asked for it. If it is needed, maybe DEBUG instead?

satpy/scene.py
```diff
@@ -933,13 +933,16 @@ def _resampled_scene(self, new_scn, destination_area, **resample_kwargs):
...
             source_area = dataset.attrs['area']
             try:
-                slice_x, slice_y = source_area.get_area_slices(
-                    destination_area)
-                source_area = source_area[slice_y, slice_x]
-                dataset = dataset.isel(x=slice_x, y=slice_y)
-                assert ('x', source_area.x_size) in dataset.sizes.items()
-                assert ('y', source_area.y_size) in dataset.sizes.items()
-                dataset.attrs['area'] = source_area
+                if resample_kwargs.get('reduce_data', True):
```
djhoese: Could you add this kwarg as an explicit kwarg in the method declaration? It should at least be popped from resample_kwargs so it doesn't get passed to the resamplers.

pnuu:
The test failures in `generic_image` reader seem to be related to version changes of either `gdal`, `pyproj` or `rasterio`. Locally the tests passed with `gdal=2.3.2`, `pyproj=1.9.5.1` and `rasterio=1.0.9` and fail after upgrading to latest packages.

djhoese:
I merged #595 which *should* fix the errors you were receiving for now. I restarted the tests and hopefully it picks up on the changes in master. Otherwise I'll merge master in to your branch for this.

PR #588, opened 2019-01-23
Fix start_time type in seviri_l1b_nc reader

BENR0:
Trying to write out geotiff files from data loaded with seviri_l1b_nc
reader gives Error because the reader sets start_time and end_time as
string instead of datetime. This commit fixes this error.


djhoese:
@sjoro @ColinDuff I assume one of you worked on this reader? Are these changes ok with you?

sjoro:
looks good to me!

ColinDuff:
fine by me

PR #589, opened 2019-01-23
Fix dask not importing sharedict automatically in dask 1.1+

djhoese:
Sharedict will be deprecated in dask in the future. As part of those changes the `sharedict` module is no longer automatically included as part of `import dask` with use `dask.sharedict`. See https://github.com/dask/dask/issues/4417 for migration changes needed in the future.

PR #591, opened 2019-01-25
Fix dask deprecation warning for sharedict in multiscene

djhoese:
See #589 for more details. The dask `sharedict` module/object is now deprecated in favor of `HighLevelGraph`. The module and object can be used in similar ways so this PR does an try/except to try to use the `HighLevelGraph` and avoids the deprecation warning.

Closes https://github.com/dask/dask/issues/4417

### DIFF
--- a/satpy/multiscene.py
+++ b/satpy/multiscene.py
@@ -26,12 +26,17 @@
 import logging
 import numpy as np
 import dask
-import dask.sharedict
 import dask.array as da
 import xarray as xr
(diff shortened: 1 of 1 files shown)


djhoese:
I can't do much about the coverage reduction, it is a try/except on ImportError so the environments are only going to have one solution and I'm not going to add a test just for this when the exception will eventually get removed with newer versions of dask.

PR #598, opened 2019-01-28
Deprecate old reader names so that they are no longer recognized

djhoese:
This is meant for version 0.15 of satpy. See #597 for details.


djhoese:
@mraspaud I was thinking maybe I should merge this as-is because maybe we want to keep the "old reader name" -> "new reader name" mapping for any future reader name changes. For this 0.15 release I would remove all of the current mappings as fully deprecated, but leave the dictionary there for future ones. Thoughts?

PR #605, opened 2019-02-04
Fix enhancements using dask delayed on internal functions

djhoese:
Internal functions are not dask serializable and fail on multi-process schedulers.

See https://github.com/pytroll/trollimage/pull/41 and https://github.com/pytroll/trollimage/pull/42 for more info. Bottom line: dask can't serialize delayed functions that are defined internally/inline another function. This PR moves 2 enhancements internal functions outside of them so they should be serializable.

@pnuu Could you test these for me with a normal threaded (default) scheduler? I don't think I have data that can make these composites.

 - [ ] Tests added 
 - [x] Tests passed 

### DIFF
--- a/satpy/enhancements/__init__.py
+++ b/satpy/enhancements/__init__.py
@@ -146,6 +146,11 @@ def cira_stretch(img, **kwargs):
     return apply_enhancement(img.data, func)
 
 
+def _lookup_delayed(luts, band_data):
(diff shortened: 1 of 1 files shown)


djhoese:
I'm assuming that since there has been no discussion on this that no one has an issue with it. Merging...

PR #614, opened 2019-02-12
Support for reduced resolution OLCI data

simonrp84:
The current L1B OLCI reader can only read the full resolution level 1b files (EFR___) due to the "_{frame:4d}_" portion of the filename pattern.
This PR allows reduced resolution (ERR___) files to be read as well. These files do not contain the frame ID, it is replaced by '____'


simonrp84:
I did consider just changing :4d to :4s, but think that this is more compatible in the long term...just in case someone needs to use the frame number in the future.

PR #615, opened 2019-02-13
Fix confusing error in abi_l1b reader when file fails to open

djhoese:
Sometimes there are cases where the ABI L1B reader can't open a file and it never creates `self.nc`. In these cases the `__del__` method was trying to access `self.nc.close()` and failing with an `AttributeError`. This AttributeError was getting through to the user and making the overall issue with their use of the reader confusing. This PR catches the AttributeError and stops it from being shown to the user. The rest of the reader should still properly report what went wrong with opening the file (not found, etc).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/abi_l1b.py
+++ b/satpy/readers/abi_l1b.py
@@ -45,8 +45,7 @@ PLATFORM_NAMES = {
 class NC_ABI_L1B(BaseFileHandler):
 
     def __init__(self, filename, filename_info, filetype_info):
-        super(NC_ABI_L1B, self).__init__(filename, filename_info,
-                                         filetype_info)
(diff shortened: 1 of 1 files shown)


djhoese:
For reference the message was something like:

```
  File "/home/graemem/workspace/g2g_1_0_0/geo2grid_v_1_0_0/lib/python3.6/site-packages/satpy/readers/abi_l1b.py", line 220, in __del__
    self.nc.close()
AttributeError: 'NC_ABI_L1B' object has no attribute 'nc'
```

PR #621, opened 2019-02-17
Add FAQ items on number of workers and threads

### LINKED ISSUE #620: Add FAQ about controlling number of threads for pykdtree and blas
@pnuu has done a lot of work learning how to control the number of threads used by dask workers. I also asked a related question on stackoverflow [here](https://stackoverflow.com/questions/53192729/how-many-threads-does-a-dask-worker-use-in-a-threaded-scheduler). It would be nice if a question could be added to the FAQ page of the documentation describing what users might see when looking at `top` with high CPU usage (more than `num_dask_workers * 100%`) and what environment variables can be used to control that more.

pnuu:
This PR adds some explanations to the FAQ on how the number of workers and number of threads affect the CPU and memory usage.

 - [x] Closes #620

### DIFF
--- a/doc/source/faq.rst
+++ b/doc/source/faq.rst
@@ -32,7 +32,12 @@ workers by doing the following at the **top** of your python code:
     # all other SatPy imports and code
 
 This will limit dask to using 8 workers. Typically numbers between 4 and 8
-are good starting points.
+are good starting points. This can be done by setting environment variable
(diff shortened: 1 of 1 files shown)


pnuu:
Thanks for checking the text, I'm bad at spotting bad wordings from my own writing :-)

PR #623, opened 2019-02-17
Fix mitiff writer handling negative values

TAlonglong:
… set to 0. Else these negative values will be showed as no data in mitiff makeing holes in the images

 - [ ] Tests passed 

### DIFF
--- a/satpy/writers/mitiff.py
+++ b/satpy/writers/mitiff.py
@@ -608,11 +608,8 @@ class MITIFFWriter(ImageWriter):
             img = get_enhanced_image(datasets.squeeze(), enhance=self.enhancer)
             for i, band in enumerate(img.data['bands']):
                 chn = img.data.sel(bands=band)
-                if (chn < 0).any():
-                    LOG.warning("Band has negative values. Set these to 0.")
-                    chn = chn.where(chn > 0, 0)
+                data = data.clip(0, 1)
                 data = chn.values * 254. + 1
-                data = data.clip(0, 255)
                 tif.write_image(data.astype(np.uint8), compression='deflate')
 
         del tif
--- a/satpy/writers/mitiff.py
+++ b/satpy/writers/mitiff.py
@@ -608,6 +608,9 @@ class MITIFFWriter(ImageWriter):
             img = get_enhanced_image(datasets.squeeze(), enhance=self.enhancer)
             for i, band in enumerate(img.data['bands']):
                 chn = img.data.sel(bands=band)
+                if (chn < 0).any():
+                    LOG.warning("Band has negative values. Set these to 0.")
+                    chn = chn.where(chn > 0, 0)
                 data = chn.values * 254. + 1
                 data = data.clip(0, 255)
                 tif.write_image(data.astype(np.uint8), compression='deflate')


TAlonglong:
Will do

Yes you are right. I was thinking to complicated.

TAlonglong:
No. Something is wrong.

I need to figure out.

TAlonglong:
This does not work as expected.

Sorry, need to come back later.

djhoese:
Something wrong with what this code is doing or are the negatives not the problem for the issue you were seeing?

TAlonglong:
What I believed was negative values might be Nans. So my first PR nor your review could not handle this correctly. 

This PR set both the data (what I believed to be negative)  inside the product and actual no data around the image to 1.

So this can not be fixed in a simple way.

djhoese:
Are you saying you need a way to distinguish the outside-swath pixels from the invalid pixels in the data?

TAlonglong:
Yes sort of.

I'm looking at a true_color product from VIIRS data. And some of the data values become negative ( and a clip will fix that), but also some data become NaNs. And then its impossible to distinguish between outside-swath pixels and these NaNs.

Then you would need some sort of mask telling you where your original data is.

TAlonglong:
Here is a dump of band 1 data values of true_color VIIRS coming from `get_enhanced_image`
```
   3.64499960e-01  2.36074584e-01  3.97148192e-01  4.66766842e-01
   3.69751845e-01  3.40644053e-01  2.79614665e-01  3.15402483e-01
   3.61698539e-01  3.55747276e-01  2.88489568e-01  1.77055436e-01
  -1.38144797e-01             nan -1.11956747e-01  1.90558498e-01
   3.34240002e-01  2.57897489e-01  1.21825900e-01 -4.68858150e-02
   1.30674019e-01 -2.84616778e-02  1.13713820e-02  7.20118916e-02
   3.58097057e-01  4.44507475e-01  4.30890742e-01  4.22765721e-01
```

djhoese:
@TAlonglong We (me and @mraspaud and @pnuu I think) discussed this in the past. To make sure we're talking about the same thing, what true color composite are you using? The default for VIIRS? What about the enhancement?

Right now I think there is an inconsistency between the "cira_stretch" and other stretch methods where cira_stretch makes negative reflectances NaNs. This kind of works for geostationary data where you get a transparent night side of the data, but does not work when there are negative values inside the day data.

TAlonglong:
Its the default composite:
```
  true_color:
    compositor: !!python/name:satpy.composites.RatioSharpenedRGB
    prerequisites:
    - name: M05
    - name: M04
    - name: M03
    optional_prerequisites:
    - name: I01
    standard_name: true_color
    high_resolution_band: red
```

with the default enhancement:

```
  true_color_default:
    standard_name: true_color
    operations:
    - name: cira_stretch
      method: !!python/name:satpy.enhancements.cira_stretch
```

So yes cira_stretch

djhoese:
Hhhmm maybe cira_stretch needs to be updated for satpy 0.13. @mraspaud complaints?

TAlonglong:
Looking at the `cira_stretch` function it might be that the` log10` gets negative `band_data` which produce nans

TAlonglong:
Is it possible the add before the line:
https://github.com/pytroll/satpy/blob/master/satpy/enhancements/__init__.py#L141

`band_data = band_data.clip_lower(0)`?

Or will this copy be to expesive?

I have not tested this, but it look like it might be a solution.

djhoese:
Yes exactly. I avoided changing it in the past until a later version. It seems now is the time. We may want to go as far as `band_data.clip(0, 1.2)`. I don't know if the data objects in the enhancement have a `clip_lower` method so `clip(0)` may be needed.

TAlonglong:
I have done some small tests with this and this works for my test data:

```
def cira_stretch(img, **kwargs):
    """Logarithmic stretch adapted to human vision.                                                                                                                      
                                                                                                                                                                         
    Applicable only for visible channels.                                                                                                                                
    """
    LOG.debug("Applying the cira-stretch")

    def func(band_data):
        log_root = np.log10(0.0223)
	denom = (1.0 - log_root) * 0.75
	band_data *= 0.01
        band_data = band_data.clip(np.finfo(float).eps)
        band_data = xu.log10(band_data)
        band_data -= log_root
        band_data /= denom
        return band_data

    return apply_enhancement(img.data, func)
```

only `clip(0)` will give `xu.log10` `-inf` which also give problems. So I propose to clip the `band_data` to the smallest possible float greater than 0.

PR #627, opened 2019-02-21
Add DNB satellite angles (DNB_SENZ, DNB_SENA) to VIIRS SDR reader

simonrp84:
Currently, the VIIRS SDR reader does not load the satellite zenith and azimuth angles from the geolocation file. This PR allows those datasets to be read if required (as is already done for I and M band data).


simonrp84:
A new push by me (for NWCSAF v2018 support) appears to also have been included in here. Advice about how to separate the VIIRS and NWCSAF commits would be appreciated!

simonrp84:
Problem with the inclusion of two commits in one should be resolved, this should only include the VIIRS changes.

PR #634, opened 2019-02-28
[Obsolete] Cache bilinear resampling LUTs in-memory

pnuu:
This PR adds in-memory caching of bilinear resampling parameters in the same way as nearest neighbour resampling is doing. Before this the parameters were always re-read from disk.

 - [x] Tests added 
 - [ ] Tests passed 


pnuu:
Closing as obsolete. Might revisit at a later time.

PR #643, opened 2019-03-11
In cira_strech clip values less or equal to 0 to avoid nans and -inf.

TAlonglong:
In the cira_stretch method add clipping if values less or equal to 0. This to avoid nans for nagtive values and -inf of 0 when doing the log10.

These nans and -inf causes ares with actual data to be handled as fill values causing "holes" in the data where there should be data. Small, yes but not fill values.

The clipping value is set to the smallest possible float larger than 0


TAlonglong:
I dont understamd why the checks fail

TAlonglong:
I think the tests went ok now.

djhoese:
@TAlonglong could you merge this with master to get my "fixes" for appveyor?

TAlonglong:
Not sure I understand. Do you want me to trigger the tests again?

The only way I know how to do that is to close the PR and reopen it.

PR #657, opened 2019-03-14
In LI reader, decode NC attribute time as ASCII

gerritholl:
In the Lightning Imager (LI) L2 reader, decode the NetCDF attributes containing the starting and ending times as ASCII, so that it can be read in using Python 3.

There are no tests yet for the LI reader, I will add relevant tests to PR #656.


gerritholl:
Redundant with #2271

PR #659, opened 2019-03-14
DOC: Refer to PyTroll coding guidelines

gerritholl:
In the documentation Developer Guide, refer to the general PyTroll coding guidelines.  This replaces the much less complete paragraph for SatPy coding guidelines.

### DIFF
--- a/doc/source/dev_guide/index.rst
+++ b/doc/source/dev_guide/index.rst
@@ -18,13 +18,10 @@ at the pages listed below.
 Coding guidelines
 =================
 
-SatPy tries to follow
-`PEP8 <https://www.python.org/dev/peps/pep-0008/>`_ style guidelines for
-all of its python code. We also try to limit lines of code to 80 characters
-whenever possible and when it doesn't hurt readability. SatPy follows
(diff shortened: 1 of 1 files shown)


djhoese:
I am ok merging this, but *only* after the [linked to guidelines](http://pytroll.github.io/guidelines.html) are updated to reflect current practices:

1. 120 max line length (I think we've completely gotten rid of 80 as a limit)
2. The "python3" compatibility could probably just be removed at this point since most/all projects will drop python 2 support by the end of 2019.
3. I think @mraspaud's maintainability guidelines are too strict for the average developer. IMO they end up making the code needlessly complex/long when they're supposed to do the opposite. Perhaps emphasizing limiting the number of the things mentioned (lines of code, parameters, etc) would be less scary? Or maybe start the section out with nicer language. I understand guidelines are guidelines, but still.
4. Mention "flake8" as a tool for checking style in style checking. Could also mention stickler which is used in most repositories PRs.
5. Remove link to releasing instructions and/or mention "RELEASING.md" instructions that should be in most if not all repositories.

PR #663, opened 2019-03-15
Feature add nwcsaf precipclouds composite

adybbroe:
Add support for the generation of the RGB of the NWCSAF probability based precipitation clouds product 

 - [ ] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented


satpy/composites/__init__.py
```diff
@@ -1014,6 +1014,46 @@ def zero_missing_data(data1, data2):
     return data1.where(~nans, 0)
 
 
+class PrecipCloudsRGB(GenericCompositor):
```
mraspaud: Could you move this class to the cloudproducts.py file instead ? I think it makes sense to have it there.

satpy/composites/__init__.py
```diff
@@ -1014,6 +1014,46 @@ def zero_missing_data(data1, data2):
...
+
+    def __call__(self, projectables, *args, **kwargs):
+        """Make an RGB image out of the three probability categories of the NWCSAF precip product"""
```
mraspaud:
```suggestion
        """Make an RGB image out of the three probability categories of the NWCSAF precip product."""
```

Don't forget the `.`

satpy/composites/__init__.py
djhoese: Can we also rename the module `cloud_products.py`?

mraspaud: to what ?

djhoese: Ah misunderstanding. You typed `cloudproducts.py` so I was suggesting `cloud_products.py`. Nevermind then I see it is already called `cloud_products.py`.

mraspaud: ah, sorry for the confusion

adybbroe: Ok, thanks guys. Done!
I haven't re-tested it yet however. Will do soon.
@mraspaud Should we look at solving the treatment of nodata as well (the white and transparent in the pictures)? Probably we should, just have no clue yet.

satpy/composites/__init__.py
adybbroe: Ok, done!

satpy/composites/__init__.py
mraspaud: yes we should. The question is if the reader needs to be fixed or the composite does.

adybbroe: I have tested it now. It works again after my last push this morning.
However, the white and nodata thingy I haven't touched. It did not appear in the testcase I had now. I think the issue is that when there is no precip in the AVHRR-only part it becomes nodata and then transparent. Something like this. I will have a short look and see if I can understand the issue. Otherwise I would like to move for another PR and go on. It doesn't really have an affect on the usage. It just doesn't look as nice as it should

adybbroe: ppsv2014_metopb_precipitation_probability_201903282049_euron1
ppsv2014_metopb_precipitation_probability_201903281908_euron1
ppsv2014_metopb_precipitation_probability_201903281730_euron1

I can't seem to find examples of the kind we saw earlier. Could we merge this now, and work on this later if it appears again?

PR #677, opened 2019-03-26
Fix MultiScene save_animation when distributed isn't installed

djhoese:
If distributed is not installed and the user doesn't explicitly ask for `client=False` then it still tried to use distributed and would fail. This fixes it so the client is not used if distributed is not installed.

 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/satpy/multiscene.py
+++ b/satpy/multiscene.py
@@ -45,6 +45,11 @@ try:
 except ImportError:
     imageio = None
 
+try:
+    from dask.distributed import get_client
+except ImportError:
+    get_client = None
+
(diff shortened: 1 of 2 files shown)


djhoese:
It was also brought up by @ColinDuff that imageio>2.5.0 may not work on all platforms or in all situations. This will need more investigation but downgrading for now is a good enough solution. There are no explicit tests in satpy that actually call in to imageio (we mock it).

@ColinDuff would you mind filing an issue with the error messages that you were getting (that is, the errors after passing `client=False`)?

PR #682, opened 2019-03-29
Make masking of Himawari space pixels optional

simonrp84:
Currently, when processing Himawari data SatPy will mask all pixels outside the Earth disk. This is not always the desired behaviour (for example, studying noctilucent clouds in AHI data requires the 'space' pixels to be preserved).
This PR makes the space pixel masking optional by setting up a new kwarg - `mask_space` - that can be set to either True or False. The default value is True, which mirrors the current behaviour of SatPy and means that users must explicitly disable masking.
This can be done by passing `reader_kwargs={'mask_space': False}` to a new Scene object.

An example of usage is included in the docstring for AHIHSDFileHandler and is copied below:
```
        import satpy
        import glob

        filenames = glob.glob('*FLDK*.dat')
                            reader='ahi_hsd',
                            reader_kwargs={'mask_space': False})
        scene.load([0.6])
```

 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/ahi_hsd.py
+++ b/satpy/readers/ahi_hsd.py
@@ -243,7 +243,7 @@ class AHIHSDFileHandler(BaseFileHandler):
 
     """
 
-    def __init__(self, filename, filename_info, filetype_info,mask_space=True):
+    def __init__(self, filename, filename_info, filetype_info, mask_space=True):
         """Initialize the reader."""
         super(AHIHSDFileHandler, self).__init__(filename, filename_info,
                                                 filetype_info)
--- a/satpy/readers/ahi_hsd.py
+++ b/satpy/readers/ahi_hsd.py
@@ -219,8 +219,29 @@ _SPARE_TYPE = np.dtype([
     ("spare", "S256")
 ])
 
+
 class AHIHSDFileHandler(BaseFileHandler):
-    """AHI standard format reader."""
+    """AHI standard format reader.
+    The AHI sensor produces data for some pixels outside the Earth disk (i,e: 
+    atmospheric limb or deep space pixels).
+    By default, these pixels are masked out as they contain data of limited
+    or no value, but some applications do require these pixels.
+    It is therefore possible to override the default behaviour and perform no
+    masking of non-Earth pixels.
+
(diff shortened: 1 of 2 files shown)


satpy/readers/ahi_hsd.py
```diff
@@ -444,7 +465,8 @@ def read_band(self, key, info):
         res = xr.DataArray(res, attrs=new_info, dims=['y', 'x'])
 
         # Mask space pixels
-        res = self._mask_space(res)
+        if (self.mask_space):
```
sfinkens: There is no need for extra parentheses here

simonrp84: I've removed them now.

satpy/readers/ahi_hsd.py
```diff
@@ -221,9 +221,29 @@
 
 
 class AHIHSDFileHandler(BaseFileHandler):
-    """AHI standard format reader."""
+    """AHI standard format reader.
+    The AHI sensor produces data for some pixels outside the Earth disk (i,e:
```
djhoese: Could you add a blank line between the first line and the first paragraph? I think PEP257 suggests this.

simonrp84: Done

simonrp84:
> Nice work! Can you please add a test making sure that `_mask_space` is only called if `mask_space==True`?

Does my most recent commit (3cd49eb) fulfill this?

satpy/tests/reader_tests/test_ahi_hsd.py
```diff
@@ -236,6 +236,12 @@ def test_read_band(self, calibrate, *mocks):
             ref_mask = np.logical_not(get_geostationary_mask(self.fh.area).compute())
             self.assertTrue(np.all(mask == ref_mask))
 
+        # Test if masking space pixels disables with appropriate flag
```
sfinkens: I think you have to indent this to make it work correctly

sfinkens: Because you need the ``ahi_hsd.open`` method to be patched as well

sfinkens:
LGTM

PR #696, opened 2019-04-04
Allow variations of image filename extensions in 'generic_image' reader

gerritholl:
In a generic image filename extension, apart from `.png`, `.jpg`, and
`.tif`, also permit the long versions `.jpeg` and `.tiff`, as well as
uppercase versions.  Upper case filenames occur in the wild and a user
may be confused by a file `DEM.TIF` is rejected.  This commit does not
allow mixed case extensions as those are very unusual and it would make
a very long yaml file to allow them all.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


gerritholl:
This is a WIP because I need to figure out if/how/where to add tests for this, and I don't know if it needs specific documentation and/or a changelog entry.

djhoese:
You could add tests for this, but I personally wouldn't add them if I were you. If the generic reader has tests it should be fairly obvious where a different filename could be put in.

PR #701, opened 2019-04-08
Clarify release instructions

pnuu:
This PR clarifies the release instructions. The previous version didn't explain which `username` should be given, and neither did it tell to add the contents of `CHANGELOG.temp` to `CHANGELOG.md` before commit. And the version numbers in steps 4. and 5. were 14 minor and 45 patch releases apart.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -12,8 +12,9 @@ loghub pytroll/satpy -u <github username> -st v<previous version> -plg bug "Bugs
 ```
 
 This command will create a CHANGELOG.temp file which need to be added
-to the top of the CHANGLOG.md file.  Remember to update also the
-version number to the same given in step 5. Don't forget to commit
+to the top of the CHANGLOG.md file.  The same content is also printed
+to terminal, so that can be copy-pasted, too.  Remember to update also
+the version number to the same given in step 5. Don't forget to commit
 CHANGELOG.md!
 
 5. Create a tag with the new version number, starting with a 'v', eg:
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -3,22 +3,26 @@
 1. checkout master
 2. pull from repo
 3. run the unittests
-4. run `loghub` and update the `CHANGELOG.md` file:
+4. run `loghub`.  Replace <github username> and <previous version> with proper
+   values.  To get the previous version run `git --tag` and select the most
+   recent with highest version number.
 
 ```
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -3,23 +3,27 @@
 1. checkout master
 2. pull from repo
 3. run the unittests
-4. run `loghub` and update the `CHANGELOG.md` file:
+4. run `loghub`.  Replace <github username> and <previous version> with proper
+   values.  To get the previous version run `git --tag` and select the most
```
djhoese: The command is `git tag`, `git --tag` fails.

RELEASING.md
```diff
@@ -3,23 +3,27 @@
...
 ```
-loghub pytroll/satpy -u <username> -st v0.8.0 -plg bug "Bugs fixed" -plg enhancement "Features added" -plg documentation "Documentation changes" -plg backwards-incompatibility "Backwards incompatible changes"
+loghub pytroll/satpy -u <github username> -st v<previous version> -plg bug "Bugs fixed" -plg enhancement "Features added" -plg documentation "Documentation changes" -plg backwards-incompatibility "Backwards incompatible changes"
 ```
 
-Don't forget to commit!
+This command will create a CHANGELOG.temp file which need to be added
+to the top of the CHANGLOG.md file.  Remember to update also the
+version number to the same given in step 5. Don't forget to commit
+CHANGELOG.md!
```
djhoese: Does `CHANGELOG.temp` differ from the stdout of the command? I've never actually used/opened the temp file. I've just copied the output from the terminal.

RELEASING.md
```diff
@@ -3,23 +3,27 @@
...
 
 5. Create a tag with the new version number, starting with a 'v', eg:
 
 ```
-git tag -a v0.22.45 -m "Version 0.22.45"
+git tag -a v<new version> -m "Version <new version>"
```
djhoese: Should we include an example of what this looks like to make it even more clear?

RELEASING.md
pnuu: Meh. First used `--tag` on command line and when writing the command to instructions had already forgot the correct form.

RELEASING.md
pnuu: Didn't do comparison, but I guess they were identical. I'll mention this.

RELEASING.md
pnuu: Thought about that, and I think it'll be better to have both. Will add examplels with actual calls.

djhoese:
I removed the `enhancement` label since that should probably be limited to code enhancements/features. I'll update the label description now.

PR #717, opened 2019-04-11
Fix 'viirs_compact' and 'viirs_l1b' readers to return WMO/Oscar platform name

### LINKED ISSUE #716: Reading the EUMETSAT compact viirs format returns wrong platform name (J01 instead of NOAA-20)
**Describe the bug**

When reading the VIIRS compact format data the `platform_name` returned from the reader is `J01` while it should be `NOAA-20` 

**To Reproduce**

```
from satpy.scene import Scene
from datetime import datetime
debug_on()

DATADIR = "/home/a000680/data/ears-viirs/noaa_new"

my_files = find_files_and_readers(base_dir=DATADIR,
                                  reader='viirs_compact',
                                  start_time=datetime(2019, 4, 9, 13, 1, 0),
                                  end_time=datetime(2019, 4, 9, 13, 18, 0))
scn = Scene(filenames=my_files)
scn.load(['true_color'])
scn.save_dataset('true_color', './true_color_NOAA20.png')

```

**Expected behavior**
No Warnings or errors from Pyspectral
and a nice image in the end

**Actual results**

Pyspectral looks for a file that doesn't exist, and then the true color either fails (pyspectral >=v0.8.5) or it just doesn't use the RSR data to derive the atm corrected VIIRS bands.

**Screenshots**

**Environment Info:**
 - OS: Linux
 - Satpy Version: [0.14.0]
 - Pyspectral Version: 0.8.4 and 0.8.7

**Additional context**

adybbroe:
Fix VIIRS readers (compact and l1b) to return the WMP standard `platform_name`

 - [x] Closes #716 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


satpy/readers/viirs_compact.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-# Copyright (c) 2014, 2015 Martin Raspaud
+# Copyright (c) 2014 - 2019 PyTroll
```
djhoese: I think we usually do `2014-2019` (no spaces) and it should be `Satpy developers`.

satpy/readers/viirs_l1b.py
```diff
@@ -59,13 +60,14 @@ def platform_name(self):
             res = self.get('/attr/platform',
                            self.filename_info['platform_shortname'])
         except KeyError:
-            res = 'Suomi-NPP'
+            res = 'Unknown'
```
djhoese: I *think* this was to handle older versions of VIIRS L1B that didn't have a platform name. Hopefully those files shouldn't be used anymore. I'm ok with this change though.

Actually...we would only get a KeyError if the filename didn't have a `platform_shortname`. Should this try/except be removed and the `self.filename_info` be changed to `.get`?

djhoese: Eh leave this the way it is.

PR #721, opened 2019-04-12
Consistent platform id attribute across NAT + HRIT SEVIRI readers

simonrp84:
**Bugfix**
Currently, the SEVIRI L1B Native reader will crash, as it is unable to calibrate IR channels:
`AttributeError: 'NativeMSGFileHandler' object has no attribute 'platform_id'`
This is because the Native reader stores `platform_id` inside `self.mda` whereas the calibration routine expects it as `self.platform_id`. The HRIT reader stores this attribute as `self.platform_id`.

This PR updates the Native reader to store the attribute as `self.platform_id`, thus making both SEVIRI readers consistent and preventing the `AttributeError` from occurring.

[edit] The method described in issue #722 is a better solution to this.

 - [x] Tests passed 


sjoro:
looks good to me.

PR #728, opened 2019-04-18
Add VIIRS Fire Temperature rgb

adybbroe:
Adding the Fire Temperature RGB for VIIRS

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -2,7 +2,6 @@
 Composites
 ==========
 
-
 Built-in Compositors
 ====================
 
@@ -23,78 +22,6 @@ that are specific to an instrument can be placed in YAML config files named
 accordingly (e.g., ``seviri.yaml`` or ``viirs.yaml``). See the
 `satpy repository <https://github.com/pytroll/satpy/tree/master/satpy/etc/composites>`_
 for more examples.
-
-Before we go on with the various compositors available first some words on
-creating/adding a default RGB composite with associated enhancement.
-
-
-RGB composites and associated enhancements
-------------------------------------------
-
-The default RGB composites available are all configured in YAML configuration
-files, as mentioned above. However, the composite YAML configuration files
-doesn't specify how the final image is enhanced (stretched). This is specified
-in the YAML configuration files for the enhancements (e.g. `generic.yaml`).  It
-is possible to have several recipes for the same combination of channels. This
-is for instace the case for the fire temperature RGB, where different agencies
-use slightly different enhancements for the same combination VIIRS bands.
-
-Here are two different enhancements used at EUMETSAT and NOAA::
-
-  fire_temperature_awips:
-    standard_name: fire_temperature
-    name: fire_temperature_awips
-    operations:
-    - name: stretch
-      method: *stretchfun
-      kwargs:
-        stretch: crude
-        min_stretch: [273.0, 0, 0]
-        max_stretch: [333.0, 100., 75.]
-    - name: gamma
-      method: *gammafun
-      kwargs: {gamma: [0.4, 1.0, 1.0]}
-  fire_temperature_eumetsat:
-    standard_name: fire_temperature
-    name: fire_temperature_eumetsat
-    operations:
-    - name: stretch
-      method: *stretchfun
-      kwargs:
-        stretch: crude
-        min_stretch: [273.0, 0, 0]
-        max_stretch: [350.0, 60., 60.]
-    - name: gamma
-      method: *gammafun
-      kwargs: {gamma: [1.0, 1.0, 1.0]}
-
-
-In the composite YAML file it os then necessary to map the right enhancement to
-the right composite using the `name` attribute::
-
-  fire_temperature_awips:
-    # CIRA: EUMETSAT
-    compositor: !!python/name:satpy.composites.GenericCompositor
-    prerequisites:
-    - name: M12
-    - name: M11
-    - name: M10
-    standard_name: fire_temperature
-    name: fire_temperature_awips
-    
-  fire_temperature_eumetsat:
-    # CIRA: AWIPS
-    compositor: !!python/name:satpy.composites.GenericCompositor
-    prerequisites:
-    - name: M12
-    - name: M11
-    - name: M10
-    standard_name: fire_temperature
-    name: fire_temperature_eumetsat
-
-
-
-
 
 GenericCompositor
 -----------------
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -2,6 +2,7 @@
 Composites
 ==========
 
+
 Built-in Compositors
 ====================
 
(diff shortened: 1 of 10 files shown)


adybbroe:
viirs_201807171156_mid_sweden_fire_temperature_crop
fire_temp_histograms

satpy/etc/enhancements/generic.yaml
```diff
@@ -79,6 +79,22 @@ enhancements:
     - name: gamma
       method: *gammafun
       kwargs: {gamma: 1.8}
+  fire_temperature_default:
```
djhoese: Is this enhancement specific to VIIRS? Maybe we should put it in `enhancements/viirs.yaml`?

adybbroe: I guess the same enhancement can be applied to any sensor with a 1.6, 2.2 and 3.7 micron channel....

adybbroe: I will continue on Tuesday, possibly tuning it further. But now Easter holiday. So, @djhoese 
don't merge yet. Just wanted to add it before closing down for today.

adybbroe: capture_viirs_201507042209_alaska_fire_temperature_298368_20100_1595_crop
And here the one from http://rammb.cira.colostate.edu/projects/alaska/blog/index.php/tag/fire-temperature-rgb/
fire_temperature_20150704_2209Z

adybbroe: After some further tuning...

adybbroe: capture_viirs_201507042209_alaska_fire_temperature_298368_20130_15130_gamma_11_12_12_crop

...after a bit of further tuning with @mraspaud using a bit of gamma we are even closer. This is my final bet!

adybbroe:
Below I show the results of the Alaska July 2015 fires for the four CIRA recipes and the one I tuned to look like the one shown on the CIRA web site.
First the original VIIRS:
viirs_201507042209_alaska_central_fire_temperature

Here the one where all bands use the reflectance - using the 3.9 reflective part, removing the emissive part:
viirs_201507042209_alaska_central_fire_temperature_39refl

Here the AWIPS version:
viirs_201507042209_alaska_central_fire_temperature_awips

Here the EUMETSAT one:
viirs_201507042209_alaska_central_fire_temperature_eumetsat

And here the test/smhi version:
viirs_201507042209_alaska_central_fire_temperature_smhi

adybbroe:
@djhoese and @mraspaud I added some documentation, please could you take what you feel is necessary and reorganize if needed?
I think as you both indicated it is preferable to have to specify only one composite recipe (when they are the same) and then couple the enhancement in another way. So, on default the recipe use one enhancement, if the user wants the "same" RGB with another enhancement it can be specified in the call. Like:
`scene.load([`fire_temperature`], enhancement_recipe="awips")`
?

doc/source/overview.rst
```diff
@@ -80,7 +80,24 @@ out certain features of the data. This includes using one dataset to enhance
...
+    >>> scene = Scene(filenames=myfiles)
+    >>> print(scene.available_composite_names())
+
+These default RGB composites have recipes that are specified in the
+`visir.yaml` and/or in the config file specific for the given sensor
+(e.g. `viirs.yaml` or `seviri.yaml`). The composites may use different
+enhancements, see below. For more details on how to add your own composite
+recipe to the configuration see :doc:`composites` and :doc:`enhancements`.
+
+Satpy also makes it possible to create your own custom composites
```
djhoese: I'm not sure this is the right place for this. This is the overview section and I don't think it should have real code in it or talk about the details of YAML files. You also mention creating your own composite, but the rest of the existing paragraph talks again about making your own custom composites. The `Scene` is missing a `reader` parameter.

doc/source/composites.rst
```diff
@@ -23,6 +24,78 @@ accordingly (e.g., ``seviri.yaml`` or ``viirs.yaml``). See the
...
+        stretch: crude
+        min_stretch: [273.0, 0, 0]
+        max_stretch: [350.0, 60., 60.]
+    - name: gamma
+      method: *gammafun
+      kwargs: {gamma: [1.0, 1.0, 1.0]}
+
+
+In the composite YAML file it os then necessary to map the right enhancement to
+the right composite using the `name` attribute::
```
djhoese: I'm worried about this too because it over simplifies the logic of enhancers and mentions matching based on `name` but doesn't say anything about `standard_name` even though it is also specified.

This is also being talked about in detail in the "Built-in composites" when maybe it should be put in a custom compositor section and this section can add a link to it.

djhoese:
@adybbroe Could you give me a summary of what I'm supposed to be seeing in your images? First, what is "original VIIRS" (your first image)? I know from being on the email chain that you were given two recipes so I understand what those are. Is there any reason to keep the SMHI version now that we have the "official" recipe?

adybbroe:
To answer your question @djhoese I repeat what I wrote on Slack (for completeness):
The “original viirs” is from Curtis Seaman’s slides. So it is the one he thinks works best. But awips doesn’t have access to radiances he said, therefore the awips one. Then he listed a Eumetsat one, thought it was appropriate to add that as well. I think time will have to tell what works best when and where.

They are all supposed to show fires obviously, but background temperatures might have influence on how each one of them perform.

mraspaud:
I'm pasting here the documentation @adybbroe wrote, so that they can be taken up in another PR while this one can be merged.

In composites.rst:
```rst
RGB composites and associated enhancements
------------------------------------------

The default RGB composites available are all configured in YAML configuration
files, as mentioned above. However, the composite YAML configuration files
doesn't specify how the final image is enhanced (stretched). This is specified
in the YAML configuration files for the enhancements (e.g. `generic.yaml`).  It
is possible to have several recipes for the same combination of channels. This
is for instace the case for the fire temperature RGB, where different agencies
use slightly different enhancements for the same combination VIIRS bands.

Here are two different enhancements used at EUMETSAT and NOAA::

  fire_temperature_awips:
    standard_name: fire_temperature
    name: fire_temperature_awips
    operations:
    - name: stretch
      method: *stretchfun
      kwargs:
        stretch: crude
        min_stretch: [273.0, 0, 0]
        max_stretch: [333.0, 100., 75.]
    - name: gamma
      method: *gammafun
      kwargs: {gamma: [0.4, 1.0, 1.0]}
  fire_temperature_eumetsat:
    standard_name: fire_temperature
    name: fire_temperature_eumetsat
    operations:
    - name: stretch
      method: *stretchfun
      kwargs:
        stretch: crude
        min_stretch: [273.0, 0, 0]
        max_stretch: [350.0, 60., 60.]
    - name: gamma
      method: *gammafun
      kwargs: {gamma: [1.0, 1.0, 1.0]}


In the composite YAML file it os then necessary to map the right enhancement to
the right composite using the `name` attribute::

  fire_temperature_awips:
    # CIRA: EUMETSAT
    prerequisites:
    - name: M12
    - name: M11
    - name: M10
    standard_name: fire_temperature
    name: fire_temperature_awips
    
  fire_temperature_eumetsat:
    # CIRA: AWIPS
    prerequisites:
    - name: M12
    - name: M11
    - name: M10
    standard_name: fire_temperature
    name: fire_temperature_eumetsat

```

In overview.rst:
```rst
Compositing
===========

Many users of satellite imagery combine multiple sensor channels to bring
out certain features of the data. This includes using one dataset to enhance
another, combining 3 or more datasets in to an RGB image, or any other
combination of datasets. Satpy comes with a lot of common composite
combinations built-in and allows the user to request them like any other
dataset.

The list of default RGB composites available depends obviously on the
sensor. For instance the AVHRR sensor does not allow to generate a true color
RGB, whereas that is possible with VIIRS and MODIS. It is straightforward to
see what default possibilities are available:

    >>> from satpy.scene import Scene
    >>> scene = Scene(filenames=myfiles)
    >>> print(scene.available_composite_names())

These default RGB composites have recipes that are specified in the
`visir.yaml` and/or in the config file specific for the given sensor
(e.g. `viirs.yaml` or `seviri.yaml`). The composites may use different
enhancements, see below. For more details on how to add your own composite
recipe to the configuration see :doc:`composites` and :doc:`enhancements`.

Satpy also makes it possible to create your own custom composites
and have Satpy treat them like any other dataset. See :doc:`composites`
for more information.
```

ameraner:
Hi @adybbroe , for the sake of clarity/tracking of recipes, do you remember where you got the "eumetsat" recipe? The Eumetrain recipe (now) uses the same parameters as the "awips" one (see Page 32 https://resources.eumetrain.org/RGBguide/recipes/RGB_recipes.pdf)

PR #732, opened 2019-04-23
Fix coord2area_def.py utillity script for python 3

adybbroe:
Fixing the utility script used to help users define a new area. Now it should be Python 3 compatible and it outputs the "new" area definition format for the areas.yaml file.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/utils/coord2area_def.py
+++ b/utils/coord2area_def.py
@@ -105,19 +105,7 @@ if __name__ == '__main__':
                    ",lon_0=" + str(lon_0) + ",ellps=WGS84").split(","))
 
     print(proj4_string)
-
-    print("REGION:", name, "{")
-    print("\tNAME:\t", name)
-    print("\tPCS_ID:\t", proj + "_" + str(lon_0) + "_" + str(lat_0))
-    print("\tPCS_DEF:\tproj=" + proj +
-          ",lat_0=" + str(lat_0) +
-          ",lon_0=" + str(lon_0) +
-          ",ellps=WGS84")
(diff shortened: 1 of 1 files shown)


utils/coord2area_def.py
```diff
@@ -104,19 +104,33 @@
...
+    print("\tNAME:\t", name)
+    print("\tPCS_ID:\t", proj + "_" + str(lon_0) + "_" + str(lat_0))
+    print("\tPCS_DEF:\tproj=" + proj +
+          ",lat_0=" + str(lat_0) +
+          ",lon_0=" + str(lon_0) +
+          ",ellps=WGS84")
+    print("\tXSIZE:\t", xsize)
+    print("\tYSIZE:\t", ysize)
+    print("\tAREA_EXTENT:\t", area_extent)
+    print("};")
```
djhoese: I personally think we should just get rid of the old format.

adybbroe: @djhoese Agreed!

adybbroe:
Are we okay with this as it is now? @djhoese and @mraspaud ?

djhoese:
@adybbroe Have you tested that the YAML produced can be read by pyresample?

adybbroe:
> @adybbroe Have you tested that the YAML produced can be read by pyresample?

@djhoese Yes! I took the output and pasted in the yaml file and it works. Enough?

PR #733, opened 2019-04-26
Update VIIRS EDR Active Fires

### LINKED ISSUE #725: Update 'viirs_edr_active_fires' reader to read newest algorithm output
This was first discussed on slack and brought up by @ColinDuff. The reader originally written by @rdaruwala was based on developer/beta versions of the products created using the CSPP Active Fires software. These files started with `AFEDR`, but the software has now been updated to have an `AFIMG` and `AFMOD` prefix for imagery (I band) and moderate (M band) resolution output. Some examples can be seen from the SSEC direct broadcast FTP location:

ftp://ftp.ssec.wisc.edu/pub/eosdb/npp/viirs/2019_04_14_104_2042/edr/

From what I've heard from @kathys, the `AFIMG` output is different than the `AFMOD` files so some more complex changes may be needed to fix this.

The file patterns for the current reader should be *replaced* with the new filename scheme. I'd also like to see if the files that can be ordered from NOAA CLASS can be read with this same reader, see https://data.nodc.noaa.gov/cgi-bin/iso?id=gov.noaa.ncdc:C00971. If it isn't just a simple file pattern addition then I'd consider support for CLASS files as out of scope for this issue. I've put some files [here](ftp://ftp.ssec.wisc.edu/pub/ssec/davidh) which will only last for 7 days.

I've asked @rdaruwala to work on this this week.

rdaruwala:
Updates the VIIRS EDR Active Fires reader to read `AFEDR`, `AFIMG`, and `AFMOD` NetCDF and CSV files.

For `AFEDR` and `AFMOD`, the following parameters can be accessed:
- confidence_pct
- power
- T13
- latitude & longitude

For `AFIMG`, the following parameters can be accessed:
- confidence_cat
- power
- T4
- latitude & longitude

 - [x] Closes #725
 - [x] Tests added
 - [x] Tests passed

### DIFF
--- a/satpy/etc/readers/viirs_edr_active_fires.yaml
+++ b/satpy/etc/readers/viirs_edr_active_fires.yaml
@@ -15,7 +15,6 @@ file_types:
         variable_prefix: "Fire Pixels/"
         file_patterns:
         - 'AFEDR_npp_d{start_time:%Y%m%d_t%H%M%S%f}_e{end_time:%H%M%S%f}_b{orbit:5d}_c{creation_time}_{source}.nc'
-        - 'AF_v1r1_npp_s{start_time:%Y%m%d%H%M%S%f}_e{end_time:%Y%m%d%H%M%S%f}_c{creation_time:%Y%m%d%H%M%S%f}.nc'
         - 'AFMOD_npp_d{start_time:%Y%m%d_t%H%M%S%f}_e{end_time:%H%M%S%f}_b{orbit:5d}_c{creation_time}_{source}.nc'
     fires_text:
         file_reader: !!python/name:satpy.readers.viirs_edr_active_fires.VIIRSActiveFiresTextFileHandler
--- a/satpy/etc/readers/viirs_edr_active_fires.yaml
+++ b/satpy/etc/readers/viirs_edr_active_fires.yaml
@@ -68,4 +68,4 @@ datasets:
         file_type: [fires_text, fires_netcdf_img]
         file_key: "{variable_prefix}T4"
         coordinates: [longitude, latitude]
-        units: 'K'1
+        units: 'K'
--- a/satpy/etc/readers/viirs_edr_active_fires.yaml
+++ b/satpy/etc/readers/viirs_edr_active_fires.yaml
@@ -59,7 +59,13 @@ datasets:
         units: 'MW'
     T13:
         name: T13
-        file_type: [fires_netcdf, fires_text, fires_netcdf_img]
+        file_type: [fires_netcdf, fires_text]
         file_key: "{variable_prefix}FP_T13"
         coordinates: [longitude, latitude]
         units: 'K'
+    T4:
+        name: T4
+        file_type: [fires_text, fires_netcdf_img]
+        file_key: "{variable_prefix}T4"
+        coordinates: [longitude, latitude]
+        units: 'K'1
--- a/satpy/etc/readers/viirs_edr_active_fires.yaml
+++ b/satpy/etc/readers/viirs_edr_active_fires.yaml
@@ -21,8 +21,6 @@ file_types:
         file_reader: !!python/name:satpy.readers.viirs_edr_active_fires.VIIRSActiveFiresTextFileHandler
         file_patterns:
             - '{data_id}_npp_d{start_time:%Y%m%d_t%H%M%S%f}_e{end_time:%H%M%S%f}_b{orbit:5d}_c{creation_time}_{source}.txt'
-#            - 'AFIMG_npp_d{start_time:%Y%m%d_t%H%M%S%f}_e{end_time:%H%M%S%f}_b{orbit:5d}_c{creation_time}_{source}.txt'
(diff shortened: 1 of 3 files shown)


satpy/readers/viirs_edr_active_fires.py
```diff
@@ -43,11 +49,13 @@ def get_dataset(self, dsid, dsinfo):
...
-        data = self[dsinfo.get('file_key', dsid.name)]
+
+        key = dsinfo.get('file_key', dsid.name).format(variable_prefix=self.prefix)
+        data = self[key]
         data.attrs.update(dsinfo)
 
-        data.attrs["platform_name"] = self['/attr/satellite_name']
-        data.attrs["sensor"] = self['/attr/instrument_name']
+        data.attrs["platform_name"] = self.get('/attr/satellite_name')
+        data.attrs["sensor"] = self.get('/attr/instrument_name')
```
djhoese: Does this ever not exist? These need to always be set to something not `None`.

satpy/etc/readers/viirs_edr_active_fires.yaml
```diff
@@ -5,44 +5,64 @@ reader:
...
-            - 'AFEDR_npp_d{start_time:%Y%m%d_t%H%M%S%f}_e{end_time:%H%M%S%f}_b{orbit:5d}_c{creation_time}_{source}.txt'
+            - '{satellite_name}_npp_d{start_time:%Y%m%d_t%H%M%S%f}_e{end_time:%H%M%S%f}_b{orbit:5d}_c{creation_time}_{source}.txt'
 
 datasets:
-    detection_confidence:
-        name: detection_confidence
-        standard_name: fire_confidence
+    confidence_cat:
+        name: confidence_cat
+        standard_name: confidence_cat
```
djhoese: I think it would be best to remove the standard name for this.

satpy/etc/readers/viirs_edr_active_fires.yaml
```diff
@@ -5,44 +5,64 @@ reader:
...
+        file_type: [fires_text, fires_netcdf_img]
```
djhoese: How can both fire confidences be in `fires_text`? Additionally the NetCDF file type should be listed before the text file type.

satpy/etc/readers/viirs_edr_active_fires.yaml
```diff
@@ -5,44 +5,64 @@ reader:
...
-        file_key: "Fire Pixels/FP_confidence"
+        file_key: "{variable_prefix}FP_confidence"
         coordinates: [longitude, latitude]
         units: '%'
     longitude:
         name: longitude
         standard_name: longitude
-        file_type: [fires_netcdf, fires_text]
-        file_key: "Fire Pixels/FP_longitude"
+        file_type: [fires_netcdf, fires_text, fires_netcdf_img]
```
djhoese: Same for these, netcdf for IMG should be first, then netcdf for MOD, then text for img, then text for mod.

satpy/etc/readers/viirs_edr_active_fires.yaml
rdaruwala: I use one parser for both MOD and IMG filetypes. Depending on the satellite_name attribute, I read the CSV into a dataset with the relevant headers

Do you think it's better to separate them into two different readers?

djhoese: If someone gives you a MOD and IMG text file, then Satpy will see them as two files that *both* need to be read to get the confidence data. This would result in a concatenation of the data (most likely) which is not what we want. Additionally, isn't the confidence in one of the text files percent and the other is category?

Also, the order needs to be swapped regardless, netcdf should come before text.

Unless you can point out why I'm wrong, I would separate the file types (the python code can still be one class).

satpy/readers/viirs_edr_active_fires.py
rdaruwala: There was a case, should be resolved now

satpy/readers/viirs_edr_active_fires.py
```diff
@@ -43,11 +49,22 @@ def get_dataset(self, dsid, dsinfo):
...
         """
-        data = self[dsinfo.get('file_key', dsid.name)]
+
+        key = dsinfo.get('file_key', dsid.name).format(variable_prefix=self.prefix)
+        data = self[key]
         data.attrs.update(dsinfo)
 
-        data.attrs["platform_name"] = self['/attr/satellite_name']
-        data.attrs["sensor"] = self['/attr/instrument_name']
+        if self.filename_info['satellite_name'].upper() == "NPP":
```
djhoese: Use a dictionary instead please and no print statements (that's what KeyError exceptions are for).

djhoese:
I was going to ask why the CLASS active fires files weren't included but then remembered that the files I gave you didn't have any fires in them. I went to CLASS to try to find some and can't even find the product to be able to download new files...so I guess we'll support them when they are more widely used/available.

PR #740, opened 2019-04-28
Add FY-3D MERSI-2 L1B Reader (mersi2_l1b)

djhoese:
This PR adds a reader for the MERSI-2 L1B data format which are HDF5 files. This PR also includes some clean up of the VIRR reader (removing the "level" from the DatasetID).

## TODO

 - [x] Closes #706 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``git diff origin/master -- "*py" | flake8 --diff`` 
 - [x] Fully documented

PR #745, opened 2019-04-30
Virr Improve Comments and Logic

wroberts4:
I added a comment above radiance_data = rad2temp(...) to better improve readability. I also changed the use of logging to be more standard and understandable.
 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
I've already handled the `LOG` stuff in one of my commits in #740. I had at first done the try/except but then decided later that the LOG message for the KeyError wasn't providing much more information than the normal KeyError traceback. In fact, the `yaml_reader.py` will log a warning reporting what *dataset* couldn't be loaded (instead of the specific key) and include the original KeyError traceback which I think is the most useful and least redundant.

I'll copy your comment to my PR (and keep you as the author) or I'll cherry pick the commit. Thanks for this.

PR #760, opened 2019-05-10
Avoid leaking file objects in NetCDF4FileHandler

gerritholl:
Even though `chunks` is specified in the kwargs, xarray uses dask.arrays only for data variables that have at least one dimension; for zero-dimensional data variables (scalar), it uses its own lazy
loading for scalars.  When those are accessed after file closure, xarray reopens the file without closing it again.  This will leave potentially many open file objects (which may in turn trigger a Segmentation Fault: https://github.com/pydata/xarray/issues/2954#issuecomment-491221266

This problem is avoided if the scalar data variables are loaded into memory before the context manager is exited.  This PR implements such a mitigation.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 

### DIFF
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -135,6 +135,16 @@ class NetCDF4FileHandler(BaseFileHandler):
             with xr.open_dataset(self.filename, group=group,
                                  **self._xarray_kwargs) as nc:
                 val = nc[key]
+                # Even though `chunks` is specified in the kwargs, xarray
+                # uses dask.arrays only for data variables that have at least
+                # one dimension; for zero-dimensional data variables (scalar),
+                # it uses its own lazy loading for scalars.  When those are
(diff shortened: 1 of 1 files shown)


djhoese:
This makes sense to me, but let's see what the xarray devs have to say in the related https://github.com/pydata/xarray/issues/2954 issue to make sure this can't be "auto-fixed" in a future xarray version.

gerritholl:
Does this need a unit test?  I'm not sure how to test this change in a unit test.

djhoese:
Probably not. Even if you did find a nice way to check open files or something, it would probably break in a newer version of xarray.

djhoese:
Summary of our talk in person:

1. HDF group says don't leave the file open
2. @gerritholl will let the xarray people know that they are leaving files open and we'll see how this issue progresses in the future.
3. I'll merge this for now to simplify @gerritholl's work flow and because it may help others this week.

PR #793, opened 2019-05-28
Fix composites with 3.x

adybbroe:
The default band wavelength for the 3.7-3.9 band that is used in many RGBs is set to 3.9.
This is okay for SEVIRI, but NOT for most other sensors that also have bands around 3.7 which is less affected by CO2 absorption.

3.75 should be the default I would say!

 - [ ] Closes #xxxx 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 


djhoese:
What is the wavelength in the official EUMETSAT recipes for these composites (assuming they came from EUMETSAT)? I think `visir.yaml` (generic VIS/IR) composites should use whatever the official recipe uses and any customization or "tweaking" to make it work for more sensors should be limited. I would prefer if separate sensor-specific composite configs were created instead.

adybbroe:
@djhoese I treat it as a bug. :-)
The 3.9 is not ideal as it is affected by CO2 absorption and a correction for that is applied in the original EUMETSAT recipes. Now, ABI, AHI and SEVIR only have one band in this area, and it is around 3.9. FCI will have this band around 3.8 (and less affected by CO2 absorption). All the polar orbiters have a 3.7-3.8 band which is the better for clouds. BUT they also have bands in the higher end around 4. See plots:
rsr_band_0350_0450_abi_ahi_seviri
rsr_band_0350_0430

I think the default should be to get the band for closest to the 3.7-3.8. We could perhaps put at 3.8 and then it should fit all the Geos as the bands are fairly wide?

adybbroe:
My suggestion would be to treat SEVIRI as a special case. And the seviri.yaml already have all the relevant RGB recipes formulated with the exception of `night_microphysics` (which is the same as `cloudtop`). The "official" name seems to be `night_microphysics`- `cloudtop` I assume comes from SMHI "cloud top" in Swedish is "Moln topp". 
Then we could use `3.8` in the `visir.yaml` and it should be the most correct for all other sensors. If a sensor has many bands in this region the one we want can be specified in the sensor specific YAML files.
Opinons?

djhoese:
@adybbroe Sounds good and I agree that the cloudtop/microphysics naming should be cleaned up if possible. Ohhhh maybe we need to make a DeprecatedCompositor that is given the real class it is supposed to use underneath, but otherwise produces a deprecation warning.

adybbroe:
Good, I see your point @djhoese . Good idea!

satpy/etc/composites/seviri.yaml
```diff
@@ -23,7 +23,7 @@ composites:
     standard_name: cloudtop
 
   cloudtop_daytime:
-    compositor: !!python/name:satpy.composites.GenericCompositor
+    compositor: !!python/name:satpy.composites.NotRecommendedCompositor
     prerequisites:
     - name: IR_039
       modifiers: [nir_emissive]
```
djhoese: `cloudtop_daytime` uses the `nir_emissive` modifier, doesn't this mean it shouldn't be deprecated without an equivalent being available?

adybbroe: I guess you are right. It depends on how we choose to to use this "deprecation/not-recommended" flagging. I had thought we could warn users that they should stop using certain RGB's. This particular one is more of experimental character, and thus hardly used. And, we have really no experience with it. But still I wanted to have it included showing what you can do with Pyspectral. Maybe it is sufficient with a comment in the YAML file?

djhoese: I would say in the long run we should probably avoid "experimental" composites being in the main satpy package...maybe. Maybe the compositor needs to be `DeprecatedCompositor` (to emphasize that things are going away more than not being recommended) that includes a `description` parameter in the YAML that says what composite to use instead. Otherwise, maybe the description says why the composite is being deprecated (available in another package, experimental, etc). Thoughts?

mraspaud:
Where are we here ? @adybbroe any plans forwards ?

mraspaud:
@adybbroe any progress on this ?

djhoese:
@adybbroe Where are we at with this?

adybbroe:
@djhoese and @mraspaud I can try find some time next week and get back into this again. I need dig out some test cases again I suppose. Would that be okay for you, or is it more of a hurry?

djhoese:
I think that's fine. I just wanted to rename the `night_microphysics_abi` composite to `night_microphysics` but realized that this PR also deals with composites so was going to wait.

gerritholl:
@adybbroe  I like the `NotRecommendedCompositor`.  Any plans to integrate this?

djhoese:
@adybbroe Any chance this PR is still useful and you have time to work on it?

PR #799, opened 2019-05-29
Add keyError feedback to virr_l1b

wroberts4:
Improved the error feedback when a key is not found in the file.

 - [ ] Closes #xxxx 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 


djhoese:
I think the proper fix here is in P2G, see my PR. P2G shouldn't be sending extra kwargs and I fix that in my PR.

PR #801, opened 2019-06-03
Limb cooling correction

adybbroe:
- [ ] Closes #xxxx 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 


djhoese:
Can we take this opportunity to move some of these composites somewhere other than `composites/__init__.py`? Something like atmospheric corrections for rayleigh and others going to `atmosphere.py`. Classes would still need to be imported in `composites/__init__.py` for now, possibly with a wrapper class for deprecation message so people know to point to the new class location in their YAML. I'm not sure where limb correction goes in this case. Even if we just moved the modifiers to `modifiers.py` that would feel cleaner. I think @mraspaud had ideas in the past for how to name these.

PR #804, opened 2019-06-04
Add compositor for adding an image as a background

### LINKED ISSUE #460: Creating day/night composites
Hi there,

I'm not sure if this is possible, but is there a way to produce day and night composites via GOES-16 ABI to create an image similar to this?

https://cdn.star.nesdis.noaa.gov/GOES16/ABI/FD/GEOCOLOR/20182831245_GOES16-ABI-FD-GEOCOLOR-1808x1808.jpg

The description of "GeoColor" is:

> GeoColor is a multispectral product composed of True Color (using a simulated green component) during the daytime, and an Infrared product that uses bands 7 and 13 at night. During the day, the imagery looks approximately as it would appear when viewed with human eyes from space. At night, the blue colors represent liquid water clouds such as fog and stratus, while gray to white indicate higher ice clouds, and the city lights come from a static database that was derived from the VIIRS Day Night Band.

Any help would be appreciated! Thanks for maintaining such a robust and useful project!

### LINKED ISSUE #734: Add a compositor that can use static images
## Feature Request

In https://github.com/pytroll/satpy/issues/460 it has been discussed that a compositor that can use a static image as a background would be very useful.

I've been thinking about this "`BackgroundCompositor`" for a while even before the issue was opened, and already have some ideas. In fact this might end up being two compositors:

1. `StaticImageCompositor` - read an image and use it as is (`L(A)` or `RGB(A)`), or as a channel in a composite (`L(A)`)
2. `BackgroundCompositor` - use a static image (either directly or using `StaticImageCompositor`) and overlay another composite on top of it (blend using `A` channel of the top-image as weight)

With these and already existing `DayNightCompositor` and `CloudCompositor` it would be possible to create images having:

- Earth at Night as a background and overlay of clouds
- mix of Earth at Night and Blue Marble as background and overlay of clouds
- a static map as a background and overlay of clouds

pnuu:
This PR adds two compositors that makes it possible to use pre-made images as a part of other composites:

- `StaticImageCompositor` - load image and use it as a part of other composites
- `BackgroundCompositor` - use a composite as a background for other composites

Add compositor for loading pre-made images as part of a composite.

 - [x] Closes #734
 - [x] Closes #460 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ dist
 build
 doc/build
 eggs
+*.eggs
 parts
 bin
 var
@@ -27,6 +28,7 @@ pip-log.txt
 .coverage
 .tox
 nosetests.xml
+htmlcov
 
 #Translations
 *.mo
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -392,7 +392,7 @@ To ensure that the images aren't auto-stretched and possibly altered,
 the following should be added to enhancement config (assuming 8-bit
 image) for both of the static images::
 
-      static_day:
+    static_day:
       standard_name: static_day
       operations:
       - name: stretch
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -189,6 +189,33 @@ reflectance.
     >>> colorized_ir_clouds = local_scene['colorized_ir_clouds']
     >>> composite = compositor([vis_data, colorized_ir_clouds])
 
+StaticImageCompositor
+---------------------
+
+    :class:`StaticImageCompositor` can be used to read an image from disk
+    and used just like satellite data, including resampling and using as a
+    part of other composites.
+
+    >>> from satpy.composites import StaticImageCompositor
+    >>> compositor = StaticImageCompositor("static_image", filename="image.tif")
+    >>> composite = compositor()
+
+BackgroundCompositor
+--------------------
+
+    :class:`BackgroundCompositor` can be used to stack two composites
+    together.  If the composites don't have `alpha` channels, the
+    `background` is used where `foreground` has no data.  If `foreground`
+    has alpha channel, the `alpha` values are used to weight when blending
+    the two composites.
+
+    >>> from satpy import Scene
+    >>> from satpy.composites import BackgroundCompositor
+    >>> compositor = BackgroundCompositor()
+    >>> clouds = local_scene['ir_cloud_day']
+    >>> background = local_scene['overview']
+    >>> composite = compositor([clouds, background])
+
 Creating composite configuration files
 ======================================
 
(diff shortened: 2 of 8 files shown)


pnuu:
I need some help to figure out how to make Satpy to recognize the composites based on `StaticImageCompositor`. Here's an example composite:

```yaml
  background:
    standard_name: background
    fname: /tmp/overview.tif
    # Optional for non-GeoTIFF image file
    # area: euro4
```
And here's a test script (use whatever data and area available):

```python
import glob
from satpy import Scene

fnames = glob.glob('/home/lahtinep/data/satellite/new/H*201409040500*')
glbl = Scene(reader='seviri_l1b_hrit', filenames=fnames)
glbl.load(['background'])
lcl = glbl.resample('euro4')
lcl.save_dataset('background', filename='/tmp/test.tif')
```

I get this stack trace:

```python
  File "/home/lahtinep/Software/pytroll/packages/satpy/satpy/readers/__init__.py", line 335, in __getitem__
KeyError: 'background'

  File "/tmp/test.py", line 14, in <module>
    lcl.save_dataset('background', filename='/tmp/test.png')
  File "/home/lahtinep/Software/pytroll/packages/satpy/satpy/scene.py", line 1257, in save_dataset
  File "/home/lahtinep/Software/pytroll/packages/satpy/satpy/scene.py", line 680, in __getitem__
    return self.datasets[key]
  File "/home/lahtinep/Software/pytroll/packages/satpy/satpy/readers/__init__.py", line 337, in __getitem__
    key = self.get_key(item)
  File "/home/lahtinep/Software/pytroll/packages/satpy/satpy/readers/__init__.py", line 326, in get_key
    best=best, **dfilter)
  File "/home/lahtinep/Software/pytroll/packages/satpy/satpy/readers/__init__.py", line 277, in get_key
KeyError: "No dataset matching 'DatasetID(name='background', wavelength=None, resolution=None, polarization=None, calibration=None, level=None, modifiers=None)' found"
```

If I instead use the compositor directly and place the composite to the scene resampling and saving works:

```python
import glob
from satpy import Scene

from satpy.composites import StaticImageCompositor
comp = StaticImageCompositor('static', fname='/tmp/overview.tif')
foo = comp()

fnames = glob.glob('/home/lahtinep/data/satellite/new/H*201409040500*')
glbl = Scene(reader='seviri_l1b_hrit', filenames=fnames)

glbl['foo'] = foo
lcl = glbl.resample('euro4')
lcl.save_dataset('foo', filename='/tmp/test.tif')
```

djhoese:
So I've tracked down the issue for why this doesn't behave the way you expect. The main issue is that the dependency tree, once constructed, is used in two ways:

1. "Give me all the leaves of the tree": Nodes/datasets that do not depend on anything else and try to load them from the existing readers.
2. "Give me all the trunk nodes of the tree": Nodes/datasets that depend on "leaf" nodes where we get prerequisites and then call the compositor object.

We have three easy options and one "No please don't make me do it" option for a fix.

1. Add a special "None"-like node that says "This doesn't depend on anything else, but is still considered a composite". This requires special handling in the dependency tree to not return the None node when providing "leaves" and to not try to load a None node. The None node can be assigned any time there are no `prerequisites`.
2. Modify the generic_image reader to be easier to use for cases like this so that you can name your image (via keyword argument passed to the reader)? Still may require manual loading from one Scene to another Scene that is doing the `.load`.
3. Change the Background compositor to load the `fname` if provided...crap this won't work when resampling is needed.
4. "No please": refactor the dependency tree to know the difference between something that comes from a reader and something that doesn't in a way independent of where it ends up on the tree.

I vote for 1 which I can try to implement if you and @mraspaud agree.

Also, could you rename `fname` to `filename`. Not a big fan of abbreviations in interfaces (except for maybe `cmap` but even then I don't know).

djhoese:
I've pushed fixes for option 1 above to your branch. Tests that need to be added and I think using your compositor is the best way to test them:

1. Test that loading an image updates the `name` attribute of the DataArray returned. Previously, before my commit, it still had the `image` name provided by the `generic_image` reader.
2. Test that the static image composite shows in the available_composite_names and in all_composite_names, but *not* in `all_dataset_names`.
3. Test that it can load from the Scene.

pnuu:
Couldn't yet figure out what and how to mock to get real results from the `scn.available_*()` methods. On ipython prompt it's easy to test with real data, but gets rather complicated without any files :-) All the tests @djhoese listed work in real world, but mocking things is hard. Will have another go tomorrow.

pnuu:
Here's an example (global version of the one given as example in composite documentation) image using `DayNightCompositor`, `StaticImageCompositor` and `BackgroundCompositor`.
background_compositor_example

pnuu:
If the image file has no `start_time` derived from the filename, `dt.datetime.utcnow()` is used. This works great for NRT operational production, but the kwarg you suggest would be a great adfition. Actually, there might be some cases where it could have use for all the compositors (mainly `DayNightCompositor`, and testing modifiers for time slots where one doesn't have ctual data), so maybe it could be added to `GenericCompositor` in another PR?

djhoese:
@pnuu I'm not sure I understand how it would be used by other compositors, but waiting for a later PR seems fine. Actually, it may even be possible to just add `start_time` to the compositor parameters in a nice hacky way like: https://stackoverflow.com/a/4206943/433202

PR #817, opened 2019-06-12
Add image 'mode' as identifying attribute to enhancements

djhoese:
Similar to how we have `name` and `standard_name` as identifying factors that someone can use in the enhancement configs, this PR adds `mode`. This is used in the new RGB default enhancement which does a crude linear stretch between 0 and 1. This is a nice thing to have when creating new RGBs and experimenting with them because we don't have to configure a new enhancement every time. This is also useful for DayNightCompositor composites which are already enhanced and on a 0 to 1 scale. This way you could make any number of DayNight composites and wouldn't need to make a new 0 to 1 enhancement for each one.

I should really add documentation for this and it needs tests but I wanted to have people look at this as soon as possible since I'd like this for my scipy tutorial.

The other thing that is *not* implemented here but has been talked about is the shortcut of allowing certain attributes to "shortcut" the enhancement process. So if an RGB is detected (via the enhancement mode) then a special RGB-friendly enhancement would be used which would default to color limits of 0 and 1 for each band but could also use a `color_limits` attribute or `vmin` and `vmax` attributes if they existed to override the 0 and 1. I'm not sure this is the right time to implement this functionality since it seems like a work around for the general problem of "creating enhancements is more annoying than it should be for simple composites".

 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented

### DIFF
--- a/satpy/enhancements/__init__.py
+++ b/satpy/enhancements/__init__.py
@@ -30,6 +30,15 @@ def noop(img, **kwargs):
     return img
 
 
+def unconfigured_noop(img, **kwargs):
+    """Do nothing and warn that no enhancement was configured."""
+    import warnings
+    name = img.data.attrs.get('name', '<unknown>')
+    warnings.warn("No enhancement configured for '{}'. Will not stretch and "
+                  "assuming already 0-1 normalized.".format(name))
+    return noop(img, **kwargs)
+
+
 def stretch(img, **kwargs):
     """Perform stretch."""
     return img.stretch(**kwargs)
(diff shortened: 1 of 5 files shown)


gerritholl:
@djhoese What is the status of this PR?  Is this still current?  Seems to have some merge conflicts.

djhoese:
@gerritholl I still like the idea of this PR but I think @pnuu and @mraspaud didn't like it so much. I can fix the conflicts, but am curious what their current opinion is.

PR #751, opened 2019-05-06
Add a reader for FY-4A AGRI level 1 data

zxdawn:
This PR adds a reader for the FY-4A L1 data format which are HDF5 files.

 - [x] Closes #603
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``git diff origin/master -- "*py" | flake8 --diff`` 
 - [x] Fully documented

PR #831, opened 2019-06-26
Fix EWA resampling tests not properly testing caching

djhoese:
This was first noticed by @zxdawn in #751. Starting with dask 2.0+ these tests failed. The main reason is that this test was not testing what it was supposed to be testing after changes a long time ago to `resample.py`. The other reason is that in dask 2.0+, any `map_blocks` function is called with fake/empty input data to check what array type object is going to be returned. This meant multiple calls to `ll2cr` when previously there were only calls for actual data calculations.

So this PR fixes:

1. The EWA tests so that they actually make sure things were cached.
2. Resampler caching so that hashing of `resample_kwargs` results in proper equality checks, meaning resampler's that should be re-used **are** reused.
3. Fix geoviews conversion methods only issuing a warning if geoviews was not imported (FYI @BENR0) even though nothing could be done if it failed to import.

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/resample.py
+++ b/satpy/resample.py
@@ -621,6 +621,11 @@ class EWAResampler(BaseResampler):
 
     def precompute(self, cache_dir=None, swath_usage=0, **kwargs):
         """Generate row and column arrays and store it for later use."""
+        if self.cache:
+            # this resampler should be used for one SwathDefinition
+            # no need to recompute ll2cr output again
+            return None
+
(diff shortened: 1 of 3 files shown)


djhoese:
FYI See https://github.com/dask/dask/issues/5007 for my questions/concerns about this dask metadata discovery.

PR #837, opened 2019-06-28
Fix Satpy tests to work with new versions of pyresample

djhoese:
See https://github.com/pytroll/pyresample/pull/196

In the above pyresample PR I make changes to pyresample so that pyproj's CRS object is used for intermediate storage of the CRS information used by `AreaDefinition` when pyproj 2.0+ is installed. However, pyproj 2.0+ may add, remove, or change parameter names depending on what is most accurate. This means that a lot of Satpy's tests that do `my_area.proj_dict == expected_proj_dict` won't work because new parameters will be provided. I fixed this by only checking the expected parameters exact values instead of the `dict` as a whole.

One other "gotcha" with this PR is discussed in https://github.com/pyproj4/pyproj/issues/357. Basically pyproj 2.0+ seems to be dropping `+wktext` from the PROJ parameters which from my experience was needed for some projections to be accepted by rasterio's `CRS` object. Since pyproj is removing it, these projections are no longer valid (`+proj=ob_tran` and `+proj=nsper`). I've skipped the checks for these projections in the Satpy tests.

If @TAlonglong has time I'd like his opinion on my changes to the mitiff writer which basically say "Try to get an EPSG code if at all possible before restoring to `area_def.proj_str`". This is to avoid pyproj/pyresample expanding EPSG codes when mitiff wants the original EPSG code to do further conversions. It'd be nice, if mitiff files could support it, if the writer could use pyproj 2.0+ to do the EPSG to PROJ conversion, but it seems the results are different which I think is the visualization clients fault.

Note: Satpy should still work fine with the new pyresample and old pyresample, but the *tests* in this PR are needed to pass with newer pyresample (once my pyresample PR is merged and released).

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/tests/reader_tests/test_ahi_hsd.py
+++ b/satpy/tests/reader_tests/test_ahi_hsd.py
@@ -71,10 +71,13 @@ class TestAHIHSDNavigation(unittest.TestCase):
                             'spare': ''}
 
             area_def = fh.get_area_def(None)
-            self.assertEqual(area_def.proj_dict, {'a': 6378137.0, 'b': 6356752.3,
-                                                  'h': 35785863.0, 'lon_0': 140.7,
-                                                  'proj': 'geos', 'units': 'm'})
-
+            proj_dict = area_def.proj_dict
+            self.assertEqual(proj_dict['a'], 6378137.0)
+            self.assertEqual(proj_dict['b'], 6356752.3)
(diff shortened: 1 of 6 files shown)


sfinkens:
Oh, that's deprecated. Nevermind!

PR #854, opened 2019-07-13
Add GOES-R GLM L2 Gridded product reader and small ABI L1b changes

deeplycloudy:
Add GOES Geostationary Lightning Mapper support to SatPy. The first type supported is the gridded GLM imagery. This is not the lower-level, L2 point data produced by GOES ground system, but is instead the files produced by `glmtools`.

 - [x] Finish adding all GLM imagery types to the yaml. 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented 

Update: This pull request also includes small changes to the ABI L1b reader. Mainly it drops the legacy satellite_altitude/longitude/latitude attributes from being added to loaded data. Other code was refactored to be shared between the GLM, ABI L1b, and ABI L2 readers.


djhoese:
Is this based on the L2 data? Would that then technically make this L3 data? I guess not. Wondering if a more specific reader name is needed like `glm_l2_imagery` or something. @deeplycloudy @mraspaud thoughts? Main concern is if we add a reader for the point data then it can't also be called `glm_l2`, but is probably a more "accurate" name for that reader verssus this reader.

deeplycloudy:
NWS is calling it a L2 product, which I agree with because it’s a fairly direct accumulation of the original L2 information. We restore the pixel shapes and make some choices about how to weight information when doing the accumulation, but there is no additional physical retrieval / synthesis / etc. 

I agree it would be helpful to carefully distinguish the different L2 data. How about glm_l2_lcfa and glm_l2_imagery, which correspond to LCFA and GLM{F|C|M} in the filenames? If the satpy philosophy is to make those separate readers, I want to do so.

I already have a mature L2 LCFA reader that we can repurpose when doing the point reader. Group centroids+imagery is a useful visualization/overlay for big stratiform discharges.

satpy/etc/readers/glm_l2.yaml
```diff
@@ -0,0 +1,45 @@
...
+  
+# Typical filenames from Unidata THREDDS server:
+# http://thredds.unidata.ucar.edu/thredds/catalog/satellite/goes/east/
+#     products/GeostationaryLightningMapper/CONUS/current/catalog.html 
+# OR_GLM-L2-GLMC-M3_G16_s20191920000000_e20191920001000_c20191920001380.nc
+
+file_types:
+  glm_l2_imagery:
+    file_reader: !!python/name:satpy.readers.glm_l2.NC_GLM_L2_IMAGERY
+    file_patterns: ['{system_environment:2s}_{mission_id:3s}-L2-{scene_abbr:s}-{scan_mode:2s}_{platform_shortname:3s}_s{start_time:%Y%j%H%M%S%f}_e{end_time:%Y%j%H%M%S%f}_c{creation_time:%Y%j%H%M%S%f}.nc']
```
djhoese: Could `scene_abbr` maybe be changed to say `GLM{area_code:s}` instead?

deeplycloudy:
I'm certainly supportive of reusing the ABI infrastructure since my goal has been to have the GLM imagery be just another ABI channel. If you want to point to the essential bits from which I could inherit, I have no objections to making that change.

I wrote my reader from scratch because the ABI reader has a lot of machinery for tasks specific to the radiance data (and surely other matters that I don't understand), whereas the GLM data can use the default xarray reader (since it comes from the xarray writer in the first place). So it was easier to do that as a first effort than to try to understand everything the ABI reader was doing.

djhoese:
Completely understood and I agree, but @yufeizhu600 did some work to get the ABI L2 reader working which included separating out the "shareable" functionality to one `abi_base.py` module. I'd still wait on #905 before starting the work on anything. It may be simple enough that I can just do it and have you review it, but I'm not sure my time availability coming up.

That said, I'm trying to get some last minute proof-of-concept things done for my talk at the Joint Satellite Conference in Boston in a week and a GLM/ABI composite from Satpy would be pretty sweet. But I also need to actually have slides to show so I should probably work on that first.

satpy/etc/readers/glm_l2.yaml
deeplycloudy: Some of the ABI L1b reader machinery/tests still use scene_abbr. What's the difference?

djhoese: The name (`scene_abbr` versus `area_code`) isn't necessarily the important part, but the current L2 reader uses `area_code`. The important part is that `GLM` is a static part of the filename that can be used to distinguish files from matching this reader versus the ABI L2 reader.

djhoese:
@deeplycloudy I made some important changes that look big but really aren't big to how the data is loaded.

1. I changed the file handler to be based on the ABI L1b/L2 base file handler. This let's the GLM file handler benefit from some work I've done on getting better more consistent precision with the GOES AreaDefinition.
2. I kept the `scene_abbr` in the file name pattern to be consistent with ABI L1b and changed the `area_code` in the ABI L2 (FYI @yufeizhu600). This way all three readers use the same names for these fields.
3. Fixed a couple trailing spaces that our linters didn't like and added docstrings.
4. I added extra metadata to the reader YAML based on some work I did relatively recently to make readers more usable for GUI applications and other tools.
5. I added an `available_datasets` method to dynamically add resolution information to the datasets when the files are loaded. This method also checks if the variable *actually* exists in the file. I did this after noticing that some of the L2 files on Unidata's THREDDS server that you pointed me to didn't have any data variables (I assume this is because there was no lightning data for the time I picked).

djhoese:
@mraspaud This is ready for re-review.

PR #874, opened 2019-08-12
Hardcoding of mersi2 l1b reader valid_range for channel 24 and 25 as these are wrong in the HDF data

TAlonglong:
Add harcoding of mersi2 l1b reader valid_range for channel 24(10.8) and 25(12.0) as these have wrong values in the HDF data


TAlonglong:
I can't see any version information. What the PR do is that it check the current known invalid value and corrects this.

When the format is updated the valid_range value is not altered.

TAlonglong:
flake 8 did not complain at least. But I removed the parentheses and fixed the space.

djhoese:
@mraspaud @TAlonglong What do we think on this? Turns out I need this and it doesn't look like a fix is coming soon (or at least there wasn't any kind of indication) as I had hoped.

FYI merged with master this should pass tests again.

PR #877, opened 2019-08-14
Fix standard_name and units for T4/T13 in viirs_edr_active_fires reader

djhoese:
Noticed this when working with fire data more. The units in the files are `"kelvins"` which, although accepted by udunits, is not the most common way of describing a temperature variable's units.

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
         - EVENT_TYPE='push pull_request'
         - SETUP_CMD='test'
         - CONDA_CHANNELS='conda-forge'
-        - CONDA_CHANNEL_PRIORITY='True'
+        - CONDA_CHANNEL_PRIORITY='strict'
 matrix:
   include:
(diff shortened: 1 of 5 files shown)


djhoese:
Ugh I should have done these CI tests on #875 because now Travis is failing because of those. Otherwise, this looks good with hacky/dev version of ci-helpers. Still working on other ci-helpers improvements to get appveyor to work.

djhoese:
So both of these are relatively passing for the changes that were made. The failures are caused by the issues that #875 addresses. I'm going to merge this since it is a relatively small and simple fix, merge master with #875, then merge #875. After that I will merge my ABI geolocation PR and other people can start merging master in to their stuff and we can finally start passing tests.

Otherwise, I need to wait on ci-helper PRs to be merged and we can switch back from using my hacky solution.

PR #879, opened 2019-08-15
Replace Numpy files with zarr for resampling LUT caching

pnuu:
This PR is an attempt to replace Numpy .npz files with Zarr for caching resampling LUT files.

 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented

### DIFF
--- a/satpy/resample.py
+++ b/satpy/resample.py
@@ -498,7 +498,7 @@ class KDTreeResampler(BaseResampler):
         mask_name = getattr(mask, 'name', None)
         filename = self._create_cache_filename(cache_dir,
                                                mask=mask_name, **kwargs)
-        if kwargs.get('mask') in self._index_caches:
+        if kwargs.get('mask') in self._index_caches and cache_dir is not False:
             self._apply_cached_indexes(self._index_caches[kwargs.get('mask')])
         elif cache_dir:
             cache = np.load(filename, mmap_mode='r', allow_pickle=True)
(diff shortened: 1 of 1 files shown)


pnuu:
As it now is, this version is slower than the Numpy-based version. So if someone finds any ways to improve performance, contributions, pointers and ideas are wellcome.

djhoese:
As mentioned on slack I think there are two things that could be done to improve performance:

1. Save all three index arrays to one zarr file. Right now I think it is computing them separately but they all use the same base calculations.
2. Remove, maybe, the `.persist()` logic. I'm not sure that is used by anything and I'm not sure why I had it in there initially. Perhaps it was meant to speed things up if you used the resampler multiple times so that the indexes would be in memory already. However, in the past we were using dask arrays based on numpy memmaps so this persist logic was never used. I think the logic as-is is now holding all of the indexes in memory which is a waste...probably.

pnuu:
Did some "quick" tests with option 2. Reading SEVIRI HRIT data, loading three composites "simple" (`night_fog`, `airmass`, `night_microphysics`) and resampling  to `euro1` area with nearest-neighbour. The first run calculates and saves the resampling LUTs, the other two use those cached values.

**npz (current master)**
first run: 78.95
cache run:29.56
third run: 29.66

**zarr**
first run: 78.93
cache run: 29.00
third run: 29.55

**zarr, no persist**
first run: 114.94
cache run: 34.53
third run: 34.30

So it seems that the persistence only affects the first run, that is when the resampling data is calculated anyway. Makes sense. I'll try to find some time tomorrow to check what's the performance when using a single `zarr` file.

djhoese:
Do you have the persist/no persist backwards? Not persisting should be *faster*, right (from reduced memory usage, etc)?

pnuu:
No, the order is correct. My guess is that without `persist` the computations are triggered several times for the separate datasets. Not sure, but sure feels like it.

pnuu:
Aargh, I made this PR from the wrong branch :man_facepalming:

PR #891, opened 2019-08-30
Fix HDF4 reading utility using dtype classes instead of instances

djhoese:
I think we were getting lucky in previous versions of numpy and xarray where a `dtype` class could be used relatively interchangeably with a dytpe instance. Now in numpy 1.17 this isn't magic anymore and we have to fix some of our usage. For example, `np.float32` is a class (a `type`) so things like `np.float32.itemsize` now return a property of a class *not* an instance.

This PR makes it so the HDF4 utility class uses instances of dtypes to avoid some errors I was getting.

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/readers/hdf4_utils.py
+++ b/satpy/readers/hdf4_utils.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2017 Satpy developers
+# Copyright (c) 2017-2019 Satpy developers
 #
 # This file is part of satpy.
 #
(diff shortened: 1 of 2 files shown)


djhoese:
I think the main issue here is with HDF4 (pyhdf's) object that we are adding a dtype to. Numpy is still fine with `np.array([1, 2, 3], dtype=np.float32)` and doing `my_arr.dtype.itemsize` works. I can't seem to get a combination of dask and numpy to not work with `np.float32` style usage.

djhoese:
So I did a little more investigating. The `var` in `from_sds` is a pyhdf `SDS` object. We add a dtype and a shape and then dask seems to be able to treat it like a numpy array. One odd thing is that dask normally looks for an `ndim` attribute on objects, but since our `shape` is a list it doesn't need it. If I make `shape` a tuple then it complains about not having `ndim`. I...don't know. Going to leave it since it works and when it breaks we can look in to it more.

PR #905, opened 2019-09-18
Fix ABI L2 reader not scaling and masking data

djhoese:
This PR does the following:

* Fix L2 reader so it scales and masks data (it wasn't doing this before)
* Fix ABI L1b and L2 readers so they preserve integer data types of possible (including preserving `_FillValue` attribute for use later in satpy)
* Add `calibration` to C01-C16 CMI products
* Move resolution metadata to the python `available_datasets` method to get it from the file. CSPP Geo's products are supposed to all be 2km but the operational products from NOAA vary.
* Add day and night versions of the COD and CSP products. CSPP Geo does not merge these algorithm's output together so you get two separate files that could both exist for the same time step.
* Fix the `area_code` part of the file name patterns so they catch M1 and M2 for mesoscale data.
* As part of the two above points, separate the plain COD and CSP file patterns so they won't conflict with the CODD/CODN/CSPD/CSPN names.
* Add CSPP Geo's NAV product file for Longitude and Latitude. This is a "nice to have" product that CSPP Geo passes on from the AIT software instead of hiding it from the user. It is an unofficial product.
* Adds the FSC (fractional snow cover) file and dataset to the reader

One more question: Should I remove the `ancillary_variables` attributes from the file so we avoid the warning about not being able to load the `DQF` variable?

@yufeizhu600 Please review/test if possible.

 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy``

### DIFF
--- a/satpy/etc/enhancements/abi.yaml
+++ b/satpy/etc/enhancements/abi.yaml
@@ -11,3 +11,12 @@ enhancements:
         kwargs: {gamma: 2.0}
       - name: contrast
         method: !!python/name:satpy.enhancements.abi.cimss_true_color_contrast
+  cmi_reflectance_default:
+    standard_name: toa_lambertian_equivalent_albedo_multiplied_by_cosine_solar_zenith_angle
+    operations:
+      - name: linear_stretch
+        method: !!python/name:satpy.enhancements.stretch
+        kwargs: {stretch: 'crude', min_stretch: 0.0, max_stretch: 100.}
+      - name: gamma
+        method: !!python/name:satpy.enhancements.gamma
+        kwargs: {gamma: 1.5}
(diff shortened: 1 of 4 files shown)


djhoese:
Tests are failing because of https://github.com/pydata/xarray/issues/3317

djhoese:
@yufeizhu600 Based on your comments on slack and the fixes I did, do you think this is ready for merge?

djhoese:
@yufeizhu600 Now that I have your most recent changes, do you think this is ready for merge?

@mraspaud Review if you'd like otherwise I'll probably merge while traveling on Sunday.

yufeizhu600:
Yes, I think it is good to go ahead.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #937, opened 2019-10-14
Add GLM + ABI highlight composite

djhoese:
Depends on #854. This is a proof of concept that highlights certain pixels in the ABI C14 channel based on the GLM flash_extent_density. It does this by increasing the R and G channels of an RGBA version of C14 and reducing the B channel. This makes the lightning show up as yellow pixels.

## TODO

 - [x] Add keyword arguments for the various parameters of the highlight compositor
 - [x] Decide the best way to handle the BT enhancement (CC @mraspaud). To make this composite not look bad I set the ABI BT enhancement to make the clouds white. This contradicts the default BT enhancement. If this is kept as-is then C08-C10 should be customized to reflect the different temperature ranges they are capable of detecting.

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented 

CC @deeplycloudy


djhoese:
Github doesn't support MP4s so here is a GIF animation of the result of this composite turned into a animation:

djhoese:
Code:

```
from satpy import Scene, MultiScene
from glob import glob
from satpy.readers import group_files

glm_files = sorted(glob('/data/satellite/glm/20191013/OR_GLM-L2-GLMC-M3_G16_s201928621*.nc'))
abi_files = sorted(glob('/data/satellite/glm/20191013/OR_ABI-L1b-RadC-M6C14_G16*'))

glm_groups = group_files(glm_files[9:], reader='glm_l2', time_threshold=60 * 5)
joined_groups = [{'abi_l1b': [abi_file], 'glm_l2': [glm_file]} for abi_file, glm_group in zip(abi_files, glm_groups) for glm_file in glm_group['glm_l2']]

mscn = MultiScene(Scene(filenames=jg) for jg in joined_groups)
mscn.load(['C14_yellow_lightning'])
mscn.save_animation("/data/satellite/glm/20191013/C14_yellow_lightning.mp4", fps=2)
```

djhoese:
@mraspaud This is ready for review now. My recent fixes in #1476 make this work as expected. I created a separate single band composite for the C14 channel so the custom "clouds are white" enhancement doesn't effect the default brightness temperature enhancements anymore.

mraspaud:
This is missing tests, right?

djhoese:
Oh those old things...didn't I already right these...no, no I didn't. :sob:

djhoese:
@mraspaud Tests added. Now it is ready. Sorry for wasting your time.

mraspaud:
No time wasted :) Looking at this now.

satpy/composites/glm.py
```diff
@@ -0,0 +1,121 @@
...
+                the linearly scaled highlight data and then added or
+                subtracted from the highlight channels. See class docstring
+                for more information.
+            highlight_channel (tuple): Series of booleans or None for every
+                channel in the RGBA image (4). True means apply the highlight
+                effect by adding, False means apply the highlight effect by
+                subtracting, and None means don't apply the highlight. By
+                default this will add to the Red and Green channels, subtract
+                from the Blue channel, and not effect the Alpha channel. This
+                results in yellow highlights in the resulting image.
```
mraspaud, line 66: I think I understand how this works, but I don't like the way `highlight_channel` is used: `True` for additive, `False` for substractive is kind of strange.
How about combining `max_factor` and `highlight_channel` in one argument? That would also allow more control over the resulting color. Like `highlight_channel = (.8, .6, -.8, 0)` for a more orange look.

satpy/composites/glm.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def _apply_highlight_effect(self, background_data, factor):
+        new_channels = []
+        for highlight_effect, band_name in zip(self.highlight_channel, "RGBA"):
+            if highlight_effect:
+                channel_factor = factor
+            elif highlight_effect is None:
+                channel_factor = None
+            else:
+                channel_factor = -factor
```
mraspaud, line 88: This should be a separate method.

satpy/composites/glm.py
```diff
@@ -0,0 +1,121 @@
...
+
+            new_channel = background_data.sel(bands=[band_name])
+            if channel_factor is not None:
+                new_channel = new_channel + channel_factor
+            new_channels.append(new_channel)
+        return new_channels
+
+    def __call__(self, projectables, optional_datasets=None, **attrs):
+        """Create B/W image with highlighted pixels."""
```
mraspaud, line 97: If this works only with B/W, should it raise an exception if it's colour?

satpy/composites/glm.py
```diff
@@ -0,0 +1,121 @@
...
+        highlight_product, background_layer = self.match_data_arrays(projectables)
+
+        # Enhance the background as normal (assume B/W image)
```
mraspaud, line 100: Unnecessary comment.

satpy/composites/glm.py
```diff
@@ -0,0 +1,121 @@
...
+        attrs = background_layer.attrs
+        img = get_enhanced_image(background_layer)
+        # Clip image data to interval [0.0, 1.0]
```
mraspaud, line 103: Unnecessary comment.

satpy/composites/glm.py
```diff
@@ -0,0 +1,121 @@
...
+        img.data = img.data.clip(0.0, 1.0)
+        # Convert to RGBA so we can adjust the colors later
+        img = img.convert('RGBA')
+        background_data = img.data
+
+        # Adjust the colors of background by highlight layer
+        factor = self._get_highlight_factor(highlight_product)
+        new_channels = self._apply_highlight_effect(background_data, factor)
+        new_data = xr.concat(new_channels, dim='bands')
+        new_data.attrs = attrs
```
mraspaud, line 113: This could be a separate method

mraspaud (changes requested):
This works, but I'd like to improve on the `highlight_channel` argument.

satpy/composites/glm.py
djhoese: Hhhhmmm, it converts whatever is provided to a RGBA image. Technically it would work fine if provided a background image that is already RGB but I'm not sure the use case. The B/W part of this docstring is just that it assumes that the enhanced version of the background image will be B/W. I think I'll update the docstring rather than limit the functionality of the compositor.

satpy/composites/glm.py
djhoese: Interesting idea. I'm ok with that. The `highlight_channel` argument came up as a last minute "this would be more flexible" thing so I'm happy to change `max_factor` to be a tuple/list as that should make more sense.

djhoese:
@mraspaud Ready for re-review. I think I've addressed everything.

mraspaud:
Indeed you have addressed everything.

However, I did more about this and I find there is an imbalance somewhere. 
The `min_highlight` and `max_highlight` are basically the min and max of the crude stretch function we have in trollimage. So why not enhance the front layer? This is already done for the background layer, so it would make sense to apply it to the front layer too. And that would ensure maximum generality :)

djhoese:
🤔 Good point. I'd be OK calling crude stretch directly but depending on the enhancement config would require another single band compositor and it feels like misdirection. I don't see a use case where anything but crude stretch makes sense. Hm this is a tough one.

mraspaud:
The way I see it, the highlight compositor should take in two composites and enhance them before putting them together. I think that the most logical and consistent way to do it.

djhoese:
I agreed with you until I thought about it more and realized that deferring the normalization to the enhancements is technically saying "we are adjusting one *image* with another *image*", but that isn't the purpose of this compositor. It is adjusting an image with the data values of a different product. The other clarification is that the resulting image for the "highlight layer" is never explicitly put "on top" of the other layer, it is only used to adjust the background layer. So it isn't two images on top of one another, it is one image being modified.

mraspaud:
Ok, I understand the point you're making. So in this case, would it be possible to have another compositor that does the normalisation? I really don't think it belongs in the highlight compositor itself.

djhoese:
I guess I'd rather go the enhancement route (with an additional single band compositor) rather than a new compositor just for normalizing data. This feels weird to me. I get what you are saying and I don't completely disagree, but it feels like a slippery slope. To me this normalization is part of the operation of the compositor, but when looking at other compositors there are only a few that do calculations that could be separated out to another compositor. The others require doing enhancements, but those are also dealing with images. I'll give it 10 minutes of thinking and then I'll just do it.

mraspaud:
If it hurts to do it in another way, fine, keep it as it is now :) I'll let you make the final decision now that I have voiced my concerns

djhoese:
Ok, it hurts. I even started doing the enhancement config and then it came to documenting it and I just couldn't make it feel right. If someone is upset about it then make a pull request and separate it out and document it. I'm not a fan. Leaving it as is.

PR #939, opened 2019-10-15
Fix MERSI-2 natural_color composite using the wrong band for sharpening

djhoese:
The ratio sharpening compositor uses the optional (4th) prerequisite as the high resolution band. For MERSI-2's natural_color composite I had the high/low res green bands swapped. Band 4 is the high res and should be optional, band 15 is the low res and should be the required. This should have minimal different on the red and blue bands but is should replace the green band in the final output with the higher resolution band 4 image.

 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy``

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,13 @@ env:
         - CONDA_CHANNEL_PRIORITY='strict'
 matrix:
   include:
-  - env: PYTHON_VERSION=2.7
+  - env:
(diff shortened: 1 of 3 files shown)


djhoese:
Examples created by @kathys with Polar2Grid:

## Before

Screen Shot 2019-10-15 at 3 03 22 PM

## After

Screen Shot 2019-10-15 at 3 29 54 PM

PR #953, opened 2019-10-25
Encode header attributes in CF writer

sfinkens:
This PR introduces encoding of header attributes in the CF writer. The same methods are employed as for dataset attributes (i.e. the standard json encoder plus some extra rules). This also fixes the problem of zero-valued header attributes not being included in the global nc attributes.

Furthermore, I harmonized the encoding of `bool` and `np.bool_` attributes. Currently we have `"True"` for `np.bool_` and `"true"` for boolean. I chose `"true"` and `"false"` because that is the json default.

 - [x] Tests passed 
 - [X] Passes ``flake8 satpy``


sfinkens:
Eh, it's probably better to call `encode_nc` on the header attributes, too

PR #967, opened 2019-11-13
Keep history attribute  when header_attrs are provided

ninahakansson:
Set the history attribute in cf_writer also when header_attrs are provided.

I did not add a test, but updated the one for cf_writer header_attrs.

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [ ] Fully documented


sfinkens:
Nice catch @ninahakansson!

PR #974, opened 2019-11-22
Fix available_composite_names including night_background static images

djhoese:
The builtin static image composites `night_background` and `night_background_hires` were being included in the `scn.available_composite_names()` results. This is misleading and can make users think these composites are coming from the reader data, which they are not.

This PR puts a `_` prefix on these composites which tells the Scene to not list them.

 - [x] Closes https://github.com/ssec/polar2grid/issues/239
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/etc/composites/seviri.yaml
+++ b/satpy/etc/composites/seviri.yaml
@@ -362,11 +362,11 @@ composites:
     standard_name: night_ir_with_background
     prerequisites:
       - night_ir_alpha
-      - night_background
+      - _night_background
 
   night_ir_with_background_hires:
     compositor: !!python/name:satpy.composites.BackgroundCompositor
     standard_name: night_ir_with_background_hires
     prerequisites:
       - night_ir_alpha
-      - night_background_hires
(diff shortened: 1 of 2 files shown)


djhoese:
@sfinkens For the first point, no it does not need to be updated. The name of the sections in the enhancement config should be purely for labeling/human-reading.

The second point, I would say a "better" solution might be to not include any subclasses of the StaticImageCompositor but that's a pretty broad rule/restriction. The only other way we currently have of "hiding" a composite dependency is to make it an inline composite dependency. In that case though we internally add the `_` prefix to hide it in the available_composite_names list :smile: 

I think a better solution would require an explanation for why it is important for the background image to be listed. Yes, it is good for debugging to make it more obvious, but if you're already debugging then you'll probably realize the name of the dependency starts with an underscore and to ask for it directly (ex. `scn.load(['_night_background'])`).

djhoese:
If no one else has an opinion, I'll merge this a little later.

PR #976, opened 2019-11-25
CF Writer Improvements

sfinkens:
* Fix crash with unlimited dimensions
* Align units of time coordinates and bounds. Currently we are saving absolute times but relative time bounds (seconds since start time)
* Sort datasets by name

 - [x] Tests passed 
 - [X] Passes ``flake8 satpy``

### DIFF
--- a/satpy/writers/cf_writer.py
+++ b/satpy/writers/cf_writer.py
@@ -539,7 +539,7 @@ class CFWriter(Writer):
         datas = {}
         start_times = []
         end_times = []
-        for ds in ds_collection.values():
+        for ds_name, ds in sorted(ds_collection.items()):
             if ds.dtype not in CF_DTYPES:
                 warnings.warn('Dtype {} not compatible with {}.'.format(str(ds.dtype), CF_VERSION))
             try:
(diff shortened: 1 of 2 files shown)


satpy/writers/cf_writer.py
```diff
@@ -259,22 +260,14 @@ def area2cf(dataarray, strict=False):
...
     end_time = min(end_time for end_time in end_times
                    if end_time is not None)
-    try:
-        dtnp64 = dataarray['time'].data[0]
-    except IndexError:
-        dtnp64 = dataarray['time'].data
-    time_bnds = [(np.datetime64(start_time) - dtnp64),
-                 (np.datetime64(end_time) - dtnp64)]
-    data = xr.DataArray(np.array(time_bnds)[None, :] / np.timedelta64(1, 's'),
+    data = xr.DataArray([[np.datetime64(start_time), np.datetime64(end_time)]],
```
sfinkens: I'm not entirely sure, why the current method uses the timestamp. But I guess it's just for encoding reasons (bounds relative to timestamp). By changing the bounds' datatype to `datetime64` xarray will handle the encoding.

PR #983, opened 2019-11-27
Add group method to MultiScene

sfinkens:
Add `group` method to `MultiScene`. By default, `MultiScene` only operates on dataset IDs shared by all scenes. Using the `group` method you can specify groups of datasets that shall be treated equally by `MultiScene`. Even if their dataset IDs differ (for example because the names or
 wavelengths are slightly different).

Groups must be specified as a dictionary `{group_id: dataset_names}` where the keys
must be of type `DatasetID`. 

One use case for this feature is creating a `MultiScene` from scenes recorded by different geostationary satellites (with slightly different wavelengths) at roughly the same time.

Examples:

a) Using the `MultiScene` constructor
```python
groups = {satpy.DatasetID('my_group', wavelength=(10, 11, 12)): ['B13', 'C13', 'IR_108']}
multi_scene = satpy.MultiScene([h8_scene, g16_scene, met10_scene])
multi_scene.group(groups)
```

b) Using `MultiScene.from_files`
```python
groups = {satpy.DatasetID('my_group', wavelength=(10, 11, 12)): ['C13', 'C14']}
# C13 from GOES-16 and C14 from GOES-17
multi_scene = satpy.MultiScene.from_files(
    ['OR_ABI-L1b-RadF-M6C13_G16_s20200011200218_e20200011209537_c20200011210022.nc',
     'OR_ABI-L1b-RadF-M6C14_G17_s20200011200358_e20200011209425_c20200011209490.nc'],
    reader='abi_l1b')
multi_scene.load(['C13', 'C14'])
multi_scene.group(groups)
```

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented


satpy/multiscene.py
```diff
@@ -109,15 +109,43 @@ def __getitem__(self, ds_id):
...
+def add_aliases(scenes, aliases):
+    """Add dataset aliases to the given scenes."""
+    for scene in scenes:
+        scene = scene.copy()
+        for alias, target_names in aliases.items():
+            target_ids = [DatasetID.from_dict(scene[name].attrs)
+                          for name in target_names if name in scene]
+            if len(target_ids) == 1:
+                target_id = target_ids[0]
+                new_ds = scene[target_id].copy()
```
sfinkens: I have to create another copy here. Otherwise the wavelength can't be changed.

djhoese:
Nice job! A couple questions/concerns after skimming the code but not really walking through each part:

1. Does this add the datasets with the new name or does it **replace** the old name with the new name?
2. If something isn't included in the groups dictionary but still exists in the input Scenes, is it discarded?
3. When I use the MultiScene I typically use `MultiScene.from_files` then do my load and resample calls. The way this is coded I'd have to provide my groups/aliases to `from_files` but I wouldn't have datasets loaded yet so this would fail, right?
4. Could this be made a method that can be called on an existing MultiScene to control the behavior in cases like my use case in point 3?

sfinkens:
1. Existing datasets are not replaced.
2. Datasets not contained in the group dictionary remain unchanged.
3. Good point. Yes, that would fail.
4. I think it should be possible. As long as that method is called before `blend()`.

doc/source/multiscene.rst
```diff
@@ -40,6 +40,46 @@ iteratively overlays the remaining datasets on top.
     >>> blended_scene = new_mscn.blend()
     >>> blended_scene.save_datasets()
 
+Grouping Datasets
```
djhoese, line 43: I'm worried about this being confused with `group_files` which is used from `MultiScene.from_files` and is mentioned in the last section of this document. Would "Grouping Similar Datasets" be too awkward? Or should there maybe be a `from_files` specific section?

sfinkens: No, "Grouping Similar Datasets" is much better. We could add two examples of how to create a multi scene (constructor, from_files) to the beginning of the chapter?

djhoese: Oh that's a good idea. Creation examples.

sfinkens: Added a chapter on multi scene creation

doc/source/multiscene.rst
```diff
@@ -13,6 +13,51 @@ examples will walk through some basic use cases of the MultiScene.
...
+    ...    Scene(reader='viirs_sdr', filenames=glob('/data/viirs/day_1/*t180*.h5')),
+    ...    Scene(reader='viirs_sdr', filenames=glob('/data/viirs/day_2/*t180*.h5'))
+    ... ]
+    >>> mscn = MultiScene(scenes)
+    >>> mscn.load(['I04'])
+
+or by using the :meth:`MultiScene.from_files <satpy.multiscene.MultiScene.from_files>`
+class method to create a ``MultiScene`` from a series of files. This uses the
+:func:`~satpy.readers.group_files` utility function to group files by start
+time.
```
djhoese, line 33: Could you add "by start time or other filenames parameters" since that can be controlled by the reader configuration (YAML) or the user by keyword argument? Just want to emphasize that it isn't just start time.

sfinkens: Sure, added!

PR #987, opened 2019-11-29
scene.aggregate will now handle a SwathDefinition

loerum:
scene.aggregrate can now handle a SwathDefinition, when computing the resolution.


loerum:
Hi Dave, yes, it will finally call pyresample's SwathDefinition.aggregate

PR #993, opened 2019-12-03
Fix VIIRS EDR Flood file patterns not working for AOI files

djhoese:
The file patterns for the VIIRS EDR Flood reader were not specific enough and were causing the Area Of Interest files to be matched against the full swath file pattern. It matched because "001_1157" in the filename was being matched to a single field that was only supposed to match a single number. This PR makes the file pattern more specific.

Closes https://github.com/ssec/polar2grid/issues/243

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``


djhoese:
Failures are unrelated and fixed in master. Merging now...

PR #997, opened 2019-12-04
Add check if prerequisites is used

TAlonglong:
If prerequisites is used in the mitiff writer be sure you use the right one.

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented


TAlonglong:
Hm I will have look if an additional test can test this.

PR #999, opened 2019-12-04
Add FCI L1C reader short and long name metadata

djhoese:
Adds short_name, long_name, and a longer description for the FCI reader. This is so GUI applications like my SIFT tool can display a "prettier" name for this reader.


djhoese:
@gerritholl FYI ^

gerritholl:
As far as I'm concerned we don't need the word "generic".  I think it's not meaningful.  A generic FCI L1C reader would read both FDHSI and HRFI.

djhoese:
Ok i reworded it based on the comments here and on slack.

PR #1003, opened 2019-12-06
Update copyright header in readers. Add and fix docstrings.

sjoro:
This PR fixes the copyright block in readers and adds and fixes some docstrings to pass flake8 check.

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/readers/geocat.py
+++ b/satpy/readers/geocat.py
@@ -1,25 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2017 Satpy developers
-#
-# This file is part of satpy.
-#
-# satpy is free software: you can redistribute it and/or modify it under the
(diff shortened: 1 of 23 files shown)


satpy/readers/geocat.py
```diff
@@ -283,7 +276,12 @@ def get_metadata(self, dataset_id, ds_info):
 
         return info
 
-    def get_dataset(self, dataset_id, ds_info, xslice=slice(None), yslice=slice(None)):
+    def get_dataset(self, dataset_id, ds_info, xslice=None, yslice=None):
+        """Get dataset."""
+        if xslice is None:
+            xslice = slice(None)
+        if yslice is None:
+            yslice = slice(None)
```
djhoese: Oh this is from the older versions of Satpy. We can probably remove the use of these slices entirely.

PR #1005, opened 2019-12-06
Remove unused reader xslice/yslice keyword arguments

djhoese:
Before we used dask these keyword arguments allowed us to limit how much data was loaded from disk. With dask this isn't needed since future slicing of the data should limit what dask chunks are loaded.

 - [ ] Tests passed 
 - [x] Passes ``flake8 satpy``


djhoese:
@mraspaud Approved on slack. Merging now. @sjoro if you have issues post-merge let me know.

Issue #909, opened 2019-09-19
computation between datasets loses geocoding information


ghislainp:
**Describe the bug**

I'm computing a ratio of two bands from the same image (OLCI sensor).
scene['new_band'] = scene['Oa05']  / scene['Oa01']
and then resample and save to geotif the result. The resulting image is correct, but not its geolocation.

Inspecting scene['new_band']  shows that the area attrs is not there. In fact, even the resolution is not copied, it is None. The Oa05 and Oa01 have both an area and it is the same (same reference)

**To Reproduce**

```python
# Your code here
scene.load(['Oa05', 'Oa01'])
scene['new_band'] = scene['Oa05'] / scene['Oa01']

scene['new_band'].area
```

**Expected behavior**
I'm expecting scene['new_band'] to have the combined metadata of the two bands used to compute the ratio, including area and resolution.

**Actual results**

raise an AttributeError because scene['new_band'] has no area attribute.

**Environment Info:**
 - OS: linux
 - Satpy Version: 0.16.1 (had the same probleme with 0.15)

djhoese:
This is by design in xarray and is not controlled by us in Satpy. Xarray, by default, will drop attributes when doing any operation on a DataArray. If you print `scene['new_band'].attrs` it should be empty. You will have to manually copy the attributes that you want to the new DataArray object.

There is a `satpy.dataset.combine_metadata` function to assist with this. You should be able to do:

```python
from satpy.dataset import combine_metadata
scene['new_band'] = scene['0a05'] / scene['0a01']
scene['new_band'].attrs = combine_metadata(scene['0a05'], scene['0a01'])
scene['new_band'].attrs['some_other_key'] = 'whatever_value_you_want'
```

Another thing is that you may want to create a temporary variable, assign the attributes, and *then* assign that variable to the Scene object to avoid the Scene adding its own attributes (thinking it knows best) before you have set your own:

```python
new_band = scene['0a05'] / scene['0a01']
new_band.attrs = ...
scene['new_band'] = new_band
```

Note: You should access DataArray metadata through `.attrs[key]` instead of `.key`.

ghislainp:
Thank you for your very clear response.

I found this PR: https://github.com/pydata/xarray/pull/2482  that may be relevant.
I tried: xr.set_options(keep_attrs=True) and it works in my case, without any manual copy of the attrs. Of course, you could not use this global option in satpy, but the PR proposes other options that seems more local.

[...]

PR #1008, opened 2019-12-10
Add documentation for combine_metadata function

BENR0:
Adds documentation to combine and add metadata when doing calculations,
which drops attributes from DataArrays.
This is based on answers given in #909.


BENR0:
You are right. After checking again the following attributes get added:
- resolution:    None
- calibration:   None
- polarization:  None
- level:         None
- modifiers:     ()

Name gets added too but is the actual name specified during assignment which makes sense.
Therefore I will remove the note.

PR #1017, opened 2019-12-11
Add pass through of xr compute, persist and chunk to Scene

### LINKED ISSUE #1015: Add compute method to Scene
## Feature Request

Sometimes I am in the situation that I need to compute all DataAarrays in the Scene. For example if I want to use a resampled Scene in two different processing streams and don't want to calculate the resampling twice.

The Scene could have a "compute()" method which just calls "compute()" on each DataArray:

```Python
def compute():
    for k in scn_msg.datasets.keys():
        scn[k] = scn[k].compute()
```
Maybe this can even be more generic like an "apply" method. This could then be used for all DataArray methods like "chunks()" etc.

This would not change the existing user workflows.

BENR0:
Adds the xarray interfaces `compute`, `persist` and `chunk` to Scene.

If for example `scn.compute()` is called it is iterated over all Datasets in the Scene and `compute` is called on every Dataset (which is a xarray.DataArray).

 - [x] Closes #1015
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented 

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -1143,6 +1143,36 @@ class Scene:
                                           **kwargs)
         return writer.save_datasets(dataarrays, compute=compute, **save_kwargs)
 
+    def compute(self, **kwargs):
+        """Call `compute` on all Scene datasets.
+
+        See :meth:`xarray.DataArray.compute` for more details.
+        """
+        new_scn = self.copy()
+        for k in new_scn.datasets.keys():
+            new_scn[k] = new_scn[k].compute(**kwargs)
+        return new_scn
+
+    def persist(self, **kwargs):
+        """Call `persist` on all Scene datasets.
+
+        See :meth:`xarray.DataArray.persist` for more details.
+        """
+        new_scn = self.copy()
+        for k in new_scn.datasets.keys():
+            new_scn[k] = new_scn[k].persist(**kwargs)
+        return new_scn
+
+    def chunk(self, **kwargs):
+        """Call `chunk` on all Scene datasets.
+
+        See :meth:`xarray.DataArray.chunk` for more details.
+        """
+        new_scn = self.copy()
+        for k in new_scn.datasets.keys():
+            new_scn[k] = new_scn[k].chunk(**kwargs)
+        return new_scn
+
     @staticmethod
     def _get_writer_by_ext(extension):
         """Find the writer matching the ``extension``.
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -1144,7 +1144,7 @@ class Scene:
         return writer.save_datasets(dataarrays, compute=compute, **save_kwargs)
 
     def compute(self, **kwargs):
-        """Call `compute` on all Scene datasets.
+        """Call `compute` on all Scene data arrays.
 
         See :meth:`xarray.DataArray.compute` for more details.
         """
@@ -1158,7 +1158,7 @@ class Scene:
         return new_scn
 
     def persist(self, **kwargs):
-        """Call `persist` on all Scene datasets.
+        """Call `persist` on all Scene data arrays.
 
         See :meth:`xarray.DataArray.persist` for more details.
         """
@@ -1172,7 +1172,7 @@ class Scene:
         return new_scn
 
     def chunk(self, **kwargs):
-        """Call `chunk` on all Scene datasets.
+        """Call `chunk` on all Scene  data arrays.
 
         See :meth:`xarray.DataArray.chunk` for more details.
         """
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -1149,7 +1149,7 @@ class Scene:
         See :meth:`xarray.DataArray.compute` for more details.
         """
         new_scn = self.copy()
-        for k in new_scn.datasets.keys():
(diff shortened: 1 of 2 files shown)


djhoese:
Nice start. I just thought of something, in order to match the xarray/dask interfaces these should probably return a *new* Scene object. Thoughts?

BENR0:
Haven't thought about that it's a good point. Yes I think if xarray and dask return new objects satpy should too. Maybe we can make that a default but add a parameter like "inplace"?

djhoese:
Looks like stickler isn't so happy with your indentation. I'm not sure I like the inplace kwarg, does it really provide you anything? Pandas, xarray, and dask are all not inplace anymore. Although I understand the `_apply` decorator, I'm worried it causes a little too much indirection. I'm willing to be convinced, but just worried at first glance.

BENR0:
Actually after adding the scn.copy() part I was not so convinced myself since in any way a Scene is returned so from my side we can remove the "inplace" parameter. 
It's kind of funny I wasn't sure of the decorator because the saving of repetitive code is minimal. So I guess you are thinking of moving it into the scene and just call it in each return?

djhoese:
> So I guess you are thinking of moving it into the scene and just call it in each return?

What is "it" in that question? I was thinking doing a `new_scn = scn.copy()` followed by the for loop with the `compute/persist/chunk` calls in each method was good enough.

BENR0:
I meant the apply function with "it". That would at least save some code. But sure we can also copy the loop to every method.

djhoese:
I rendered the sphinx locally for your changes and noticed there needs to be a blank line between the "subject" and the rest of the docstring otherwise they get rendered as one line in the HTML. I also tweaked the verb tense to make flake8-docstring happy. This then lead to flake8-docstring complaining that we were using the name of the method inside the method's docstring. I told flake8 to ignore that for now. @mraspaud thoughts?

I'll see if I can cleanly ignore the check just for those methods instead of globally.

sfinkens (approved):
Nice work, very useful!

satpy/scene.py
```diff
@@ -1143,6 +1143,36 @@ def save_datasets(self, writer=None, filename=None, datasets=None, compute=True,
...
         return writer.save_datasets(dataarrays, compute=compute, **save_kwargs)
 
+    def compute(self, **kwargs):
+        """Call `compute` on all Scene datasets.
+
+        See :meth:`xarray.DataArray.compute` for more details.
+        """
+        new_scn = self.copy()
+        for k in new_scn._datasets.keys():
+            new_scn[k] = new_scn[k].compute(**kwargs)
```
djhoese, line 1153: It would be nice if these methods could compute (same for the persist) all the DataArrays at the same time. As is this will likely recompute share dependencies of the DataArrays.

djhoese:
I'm not sure what codefactors problem is. If you merged with `main` then this isn't supposed to be an issue (the asserts in the test directory).

satpy/scene.py
BENR0: I think I don't exactly understand what you mean with "at the same time".

djhoese: Normally for dask arrays you want to call `res1, res2, res3 = dask.array.compute(array1, array2, array3)` so that all dependency calculations for generating those three arrays are only computed once. I'm not sure how that can be done with xarray. You could try passing the DataArrays to `da.compute` and see if that works.

BENR0: Ah I see. Yes indeed that would be useful. I will check if that is possible with xarray or as you suggested with `da.compute`.

satpy/scene.py
```diff
@@ -1143,6 +1143,45 @@ def save_datasets(self, writer=None, filename=None, datasets=None, compute=True,
                                           **kwargs)
         return writer.save_datasets(dataarrays, compute=compute, **save_kwargs)
 
+    def compute(self, **kwargs):
+        """Call `compute` on all Scene datasets.
```
djhoese, line 1147: Should we say "data arrays" here instead of datasets to avoid the future confusion when Scene is more dependent on xarray Dataset objects?

satpy/scene.py
```diff
@@ -1143,6 +1143,45 @@ def save_datasets(self, writer=None, filename=None, datasets=None, compute=True,
...
+
+        See :meth:`xarray.DataArray.compute` for more details.
+        """
```
djhoese, line 1150:
```suggestion
        See :meth:`xarray.DataArray.compute` for more details.
        Note that this will convert the contents of the DataArray to numpy arrays which
        may not work with all parts of Satpy which may expect dask arrays.
        """
```

djhoese (changes requested):
Two small suggestions, but otherwise looks good.

satpy/scene.py
BENR0: Yes definitely. I think this is better since this otherwise adds to confusion.

djhoese:
Looks like the jobs got hung up, I've restarted them.

djhoese (approved):
LGTM. @mraspaud or others, have any comments?

mraspaud (approved):
LGTM

PR #1027, opened 2019-12-30
Update SCMI writer and VIIRS EDR Flood reader to work for pre-tiled data

djhoese:
This is functionality needed by some folks at the SSEC as a proof of concept to get VIIRS EDR Flood data in to AWIPS. The files generated by their software is a composite of ABI and VIIRS data in to individual latlong tiles (15 degrees by 15 degrees each). These files and their naming scheme was not previously supported by the `viirs_edr_flood` reader. That is one of the features this PR adds.

For the SCMI writer, while working with this new data I noticed that the SCMI writer was needlessly creating "extra" tiles. This was mostly floating point error where the writer was determining that we needed to "shift" the current grid of tiles over by ~1e-15 pixels. This ends up resulting in a 1 pixel shift that isn't needed. This was fixed in this PR.

The remaining TODO for this PR (at the time of writing) is to make it so generated tile files can be "updated" by additional executions. This is needed for the way the ABI/VIIRS composites are generated with 15x15 degree tiles. We need to process one tile, generate the AWIPS tiles, process the next input tile, and update any of the "missing" space in the previous generated tiles.

 - [ ] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ env:
     global:
         # Set defaults to avoid repeating in most cases
         - PYTHON_VERSION=$TRAVIS_PYTHON_VERSION
-        - NUMPY_VERSION=stable
+        - NUMPY_VERSION=1.17
         - MAIN_CMD='python setup.py'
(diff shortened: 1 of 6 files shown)


djhoese:
@mraspaud This is ready for review, but as mentioned in #1028 one of the mitiff tests is failing and seems unrelated to my changes. I am unable to reproduce the failure locally it seems.

djhoese:
I talked with @mraspaud on slack. Ok with this being merged given the important CI bug fixes and that the SCMI and VIIRS EDR Flood changes shouldn't effect other parts of Satpy.

PR #1051, opened 2020-01-28
Return counts from satpy/avhrr_l1b_gaclac reader

### LINKED ISSUE #1050: Return counts from avhrr_l1b_gaclac reader
Implement and return digital counts from satpy avhrr_l1b_gaclac reader

helgaweb:
Implemented get_counts from pygac in satpy/avhrr_l1b_gaclac reader

 - [x] Closes #1050  
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented 


mraspaud:
@helgaweb you need to fix the unittests, they where relying on the function you renamed

mraspaud:
@helgaweb you need to add your name to the AUTHORS.md file, and maybe also add a small test for your new functionality ?

helgaweb:
> @helgaweb you need to add your name to the AUTHORS.md file, and maybe also add a small test for your new functionality ?

Alright! Yes the unit test are so far missing. Will do that. Since I'm away next week, it will be the week after. Sorry for the delay.

mraspaud:
No rush, thanks for contributing !

mraspaud:
@helgaweb did you have time to look at this ?

helgaweb:
Hi @mraspaud  Very sorry for the delay. Its planned for this Friday together with the MetOp reader...

mraspaud:
Sounds great !

satpy/etc/readers/avhrr_l1b_gaclac.yaml
```diff
@@ -15,6 +15,9 @@ datasets:
             units: '%'
           radiance:
             standard_name: toa_outgoing_radiance_per_unit_wavelength
+          counts:
+            standard_name: counts
```
sfinkens: Although it is used in the same way by other readers, `counts` is actually not a valid CF standard name

satpy/readers/avhrr_l1b_gaclac.py
```diff
@@ -163,8 +164,19 @@ def get_dataset(self, key, info):
                        'Solar contamination of blackbody in channels 3',
                        'Solar contamination of blackbody in channels 4',
                        'Solar contamination of blackbody in channels 5']
+        elif key.calibration == 'counts':
+            # if we ask for uncalibrated data we will get all channels back
+            data = self.reader.get_counts()
```
sfinkens: I think it would make sense to cache the computed counts

satpy/etc/readers/avhrr_l1b_gaclac.yaml
helgaweb: Good Point and yes, I basically copied it from here /seviri_l1b_native.yaml. But this means all .yaml reader files with  `counts` would be need to be changed, no?

sfinkens: Probably yes. But it's not a big deal I think. I'll file a satpy issue for that.

sfinkens:
@mraspaud Ready for final review if you have time (not urgent)

satpy/readers/avhrr_l1b_gaclac.py
```diff
@@ -163,10 +165,13 @@ def get_dataset(self, key, info):
                        'Solar contamination of blackbody in channels 3',
                        'Solar contamination of blackbody in channels 4',
                        'Solar contamination of blackbody in channels 5']
-        else:
-            data = self._get_channel(key.name)
+        elif key.name.upper() in self.chn_dict:
+            # Read and calibrate channel data
+            data = self._get_channel(name=key.name, calibration=key.calibration)
```
mraspaud: You could just pass the key object to the function, maybe it'll need eg. the resolution in the future ?

satpy/readers/avhrr_l1b_gaclac.py
```diff
@@ -299,6 +310,18 @@ def _strip_invalid_lat(self):
             self.first_valid_lat, self.last_valid_lat = start, end
         return self.first_valid_lat, self.last_valid_lat
 
+    def _update_attrs(self, res):
+        """Update dataset attributes."""
+        for attr in self.reader.meta_data:
+            res.attrs[attr] = self.reader.meta_data[attr]
+        res.attrs['platform_name'] = self.reader.spacecraft_name
+        res.attrs['orbit_number'] = self.filename_info['orbit_number']
```
mraspaud: what if the filepattern doesn't provide an orbit ? Should we have some default value ?

helgaweb: @mraspaud would you have an example for that?

satpy/readers/avhrr_l1b_gaclac.py
mraspaud: yes

satpy/readers/avhrr_l1b_gaclac.py
mraspaud: I'm just thinking of the case where the user has rename its file and there is no orbit number anymore.
so something like `res.attrs['orbit_number'] = self.filename_info.get('orbit_number', None)` ?

satpy/readers/avhrr_l1b_gaclac.py
```diff
@@ -286,16 +288,16 @@ def _get_qual_flags(self):
...
+        """Get angles and buffer results."""
         if self.angles is None:
             sat_azi, sat_zenith, sun_azi, sun_zenith, rel_azi = self.reader.get_angles()
             self.angles = {'sensor_zenith_angle': sat_zenith,
                            'sensor_azimuth_angle': sat_azi,
                            'solar_zenith_angle': sun_zenith,
                            'solar_azimuth_angle': sun_azi,
                            'sun_sensor_azimuth_difference_angle': rel_azi}
-        return self.angles[name]
+        return self.angles
```
mraspaud: Shouldn't this be `self.angles[key.name]` ?

satpy/readers/avhrr_l1b_gaclac.py
helgaweb: I had something different in mind - thanks :)

satpy/readers/avhrr_l1b_gaclac.py
helgaweb: Can be both, no? But I can change it.

satpy/readers/avhrr_l1b_gaclac.py
```diff
@@ -263,28 +265,39 @@ def _slice(self, data):
...
+        """Get angles and buffer results."""
         if self.angles is None:
             sat_azi, sat_zenith, sun_azi, sun_zenith, rel_azi = self.reader.get_angles()
             self.angles = {'sensor_zenith_angle': sat_zenith,
                            'sensor_azimuth_angle': sat_azi,
                            'solar_zenith_angle': sun_zenith,
                            'solar_azimuth_angle': sun_azi,
                            'sun_sensor_azimuth_difference_angle': rel_azi}
-        return self.angles[name]
+        return self.angles[key.name]
```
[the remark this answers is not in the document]
mraspaud: yes, you are passing key to the function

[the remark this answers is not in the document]
helgaweb: Sorry, happend accidently.

mraspaud:
This looks good to me now, thanks for all the work. However, the tests are failing. Anything you can fix ?

mraspaud:
@helgaweb Did have the time to take a look at the failing tests ?

helgaweb:
@mraspaud Bussy week. Let's check if we can close this PR ;)

helgaweb:
@mraspaud Thanks for the approval! Is it ready to be merged?

mraspaud:
@helgaweb Looks good, sorry for not checking in earlier. Merging.

helgaweb:
@mraspaud @sfinkens Thanks for the support!

PR #1053, opened 2020-01-30
Add masking method to datasets

johnlavelle:
- [ ] Closes #xxxx 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 


mraspaud:
Thanks for implementing the accessors! It is a nice way to provide satpy features directly into the xarray.DataArray objects.
At the last pytroll community meeting, we decided we will start adding accessors.

Regarding the mask, the `wqsf` dataset is specific to sentinel 3, ie it's called something else for the viirs or modis instruments for example, we maybe should explore the quality flags using ancillary variables and standard names so that we don't have to look for a specific name. Also the bit flags seem to be implemented in different ways in different readers, maybe we should factorize these ?

djhoese:
It looks like us changing the master branch to main allowed github to detect that your fork doesn't exist anymore. Let me know if that is not true and we can try to find a way to get this PR back. As of right now, I'm unable to reopen this.

johnlavelle:
Hello David,

I'm sorry but I somehow missed your email that you sent on the 6th of July
2020, regarding the accessors!

I deleted the branch earlier this year, as I was not using it. I didn't
realize that it would have this effect.

I have since simplified the implementation of the accessors, so I don't
have touch scene.py, and I can keep the code separate from the rest of
Satpy. I will send and an example and will help with implementing it in
Satpy, if I can be of help.

Thanks,
John

PR #1061, opened 2020-02-06
Add support for NUCAPS Science EDRs

tommyjasmin:
- [ ] Closes #xxxx 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/reader_tests/test_nucaps.py
+++ b/satpy/tests/reader_tests/test_nucaps.py
@@ -139,9 +139,15 @@ class FakeNetCDF4FileHandler2(FakeNetCDF4FileHandler):
         file_content[k + '/attr/_FillValue'] = -9999.
 
         attrs = ('_FillValue', 'flag_meanings', 'flag_values', 'units')
+        cris_fors_dim_name = 'Number_of_CrIS_FORs'
+        pressure_levels_dim_name = 'Number_of_P_Levels'
+        """Version 1 EDRs have different dimension names"""
+        if ('_v1' in filename):
+            cris_fors_dim_name = 'number_of_FORs'
+            pressure_levels_dim_name = 'number_of_p_levels'
         convert_file_content_to_data_array(
             file_content, attrs=attrs,
-            dims=('z', 'number_of_FORs', 'number_of_p_levels'))
+            dims=('z', cris_fors_dim_name, pressure_levels_dim_name))
         return file_content
 
 
--- a/satpy/tests/reader_tests/test_nucaps.py
+++ b/satpy/tests/reader_tests/test_nucaps.py
@@ -88,12 +88,8 @@ class FakeNetCDF4FileHandler2(FakeNetCDF4FileHandler):
             ('H2O_MR', 'g/g', ''),
             ('O3', '1', ''),
             ('O3_MR', '1', ''),
-            # These two are standard EDR only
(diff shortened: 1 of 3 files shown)


satpy/tests/reader_tests/test_nucaps.py
```diff
@@ -337,11 +335,195 @@ def test_load_pressure_levels_single_and_pressure_levels(self):
         self.assertTupleEqual(pl_ds.shape, (1,))
 
 
+# For testing Science EDRs
```
djhoese: Redundant?

satpy/tests/reader_tests/test_nucaps.py
```diff
@@ -24,10 +24,7 @@
...
 from satpy.tests.utils import convert_file_content_to_data_array
 
-if sys.version_info < (2, 7):
-    import unittest2 as unittest
-else:
-    import unittest
+import unittest
 
 try:
     from unittest import mock
```
djhoese: If you have to make other changes you could remove this try/except and only do `from unittest import mock`.

satpy/tests/reader_tests/test_nucaps.py
tommyjasmin: Wow, yes.

tommyjasmin:
> Nice job. For the science edr tests, could you make a separate test dataset that has the same dimension names as your files? Might require copying the `FakeNetCDF4FileHandler2` object. Otherwise I had a couple other small comments. Thanks for doing this.

But the dimension names on Science EDRs are already just a subset of those in standard EDRs.

satpy/tests/reader_tests/test_nucaps.py
tommyjasmin: Ah, yeah you mentioned that.

djhoese:
The dimension names in the tests I wrote use `number_of_FORs` iirc. Your new files have the `Number_Of_CrIS_FORs` naming or whatever it is.

tommyjasmin:
> The dimension names in the tests I wrote use `number_of_FORs` iirc. Your new files have the `Number_Of_CrIS_FORs` naming or whatever it is.

Then we should just change the single ref to check the file name for version num and label the dimension name  `number_of_FORs` only when appropriate.

djhoese:
You mean where the test generates the fake data? Or in the actual reader's file handler? :+1: to the former, not sure on the latter.

tommyjasmin:
> You mean where the test generates the fake data? Or in the actual reader's file handler? 👍 to the former, not sure on the latter.

I was thinking this at the end of the FakeN4 file handler:

`        """Version 1 EDRs have different dimension names"""
        if ('_v1' in filename):
            cris_fors_dim_name = 'number_of_FORs'
            pressure_levels_dim_name = 'number_of_p_levels'
        convert_file_content_to_data_array(
            file_content, attrs=attrs,
            dims=('z', cris_fors_dim_name, pressure_levels_dim_name))
        return file_content`

djhoese:
That works for me.

tommyjasmin:
> Great! Looks good code-wise. What did we decide for the docstring on the reader module to talk about the differences in the types of files and where they come from?

We didn't yet Dave :-)
I'm meeting with Brad for a bit tomorrow and I'll get some verbiage straight we can add use for docstring.

PR #1064, opened 2020-02-10
Adjust default blending in DayNightCompositor

pnuu:
This PR adjusts the default Sun zenith angle (SZA) interval where the overlapping day and night products are blended from `85.0 ... 95.0` to `85.0 ... 88.0`.

The default limit of SZA correction in Satpy is the same 88.0 degrees. If the upper limit is higher, there will be a dark band between the products.

The original version:

natural_with_colorized_ir_clouds_old_20200210_082000

Proposed version:

natural_with_colorized_ir_clouds_new_20200210_082000

 - [x] Tests adjusted
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Documentation updated

### DIFF
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -105,7 +105,7 @@ zone can be defined when initializing the compositor (default values
 shown in the example below).
 
     >>> from satpy.composites import DayNightCompositor
-    >>> compositor = DayNightCompositor("dnc", lim_low=85., lim_high=95.)
+    >>> compositor = DayNightCompositor("dnc", lim_low=85., lim_high=88.)
     >>> composite = compositor([local_scene['true_color'],
     ...                         local_scene['night_fog']])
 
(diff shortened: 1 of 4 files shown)


pnuu:
Yes, this just shifts the night time product a bit into the day side. I've made 11 separate DNC composites that I've passed to our forecasters for reviews, and all of them look nicer with these new limits. Here's a composite with true color and ash RGBs.

crop_truecolor_with_ash_20200210_082000

djhoese:
Good image. That's what I was hoping for; an image with a very dark image on one side and a bright image on the other. Looks good.

mraspaud:
How does it look like if you take a night-only composite on the night side, eg night_microphysics, does it still look ok ?

pnuu:
Here's one with night microphysical:
crop_natural_with_night_microphysical_20200210_082000

I'll still run one test with higher SZA limit for the day side.

pnuu:
And here are the three above composites with `low_lim=87.0` and `high_lim=88.0`, so only one degree SZA range where the blending is done.

Natural color / colorized IR clouds:
crop_natural_with_colorized_ir_clouds_20200210_082000_low87

True color with ash:
crop_truecolor_with_ash_20200210_082000_low87

Natural color with night microphysical:
crop_natural_with_night_microphysical_20200210_082000_low87

mraspaud:
Looks really good @pnuu !

PR #1065, opened 2020-02-10
Add AHI airmass, ash, dust, fog, and night_microphysics RGBs

djhoese:
The RGBs added in this pull request are common EUMETSAT recipes that are available for most other sensors. However, AHI doesn't have a 10.8um or a 12.0um channel (or channels whose ranges cover those wavelengths). This is my attempt at trying to balance all the information available for what the _best_ AHI RGB recipes are. I've mostly taken this from various slides by JMA found by googling, from opinions by @ScottLindstrom at the SSEC which agreed with the JMA slides, and from existing ABI recipes. See https://github.com/ssec/polar2grid/issues/258 for more discussion on this and why I'm doing it (to add them to CSPP Geo Geo2Grid).

The worst thing is I can't find a definitive guide for AHI recipes. The slides from conferences that I've found (2016 and 2017) are completely different from the training materials (2015?) which are also different from ABI recipes out there.

Note: None of these recipes were loadable for AHI data before this PR so this shouldn't be changing any existing functionality for users.

 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 

# Examples

## Airmass

airmass_20181022_030019

## Ash

ash_20181022_030019

## Dust

dust_20181022_030019

## Fog

fog_20181022_030019

## Night Microphysics

night_microphysics_20181022_030019

### DIFF
--- a/satpy/etc/enhancements/generic.yaml
+++ b/satpy/etc/enhancements/generic.yaml
@@ -143,16 +143,6 @@ enhancements:
         min_stretch: [-25, -40, 243]
         max_stretch: [0, 5, 208]
 
-  airmass_abi:
-    standard_name: airmass
-    sensor: abi
-    operations:
-    - name: stretch
-      method: *stretchfun
-      kwargs:
-        stretch: crude
-        min_stretch: [-26.2, -43.2, 243.9]
-        max_stretch: [0.6, 6.7, 208.5]
   green_snow_default:
     standard_name: green_snow
     operations:
--- a/satpy/etc/composites/ahi.yaml
+++ b/satpy/etc/composites/ahi.yaml
@@ -28,6 +28,78 @@ composites:
       modifiers: [sunz_corrected]
     standard_name: toa_bidirectional_reflectance
 
+  airmass:
+    # PDF slides: https://www.eumetsat.int/website/home/News/ConferencesandEvents/DAT_2833302.html
+    # Under session 2 by Akihiro Shimizu (JMA)
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+      - compositor: !!python/name:satpy.composites.DifferenceCompositor
+        prerequisites:
+          - name: B08
+          - name: B10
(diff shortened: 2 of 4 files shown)


satpy/etc/enhancements/generic.yaml
```diff
@@ -143,16 +143,6 @@ enhancements:
...
-  airmass_abi:
-    standard_name: airmass
-    sensor: abi
-    operations:
-    - name: stretch
-      method: *stretchfun
-      kwargs:
-        stretch: crude
-        min_stretch: [-26.2, -43.2, 243.9]
-        max_stretch: [0.6, 6.7, 208.5]
```
mraspaud: Isn't this PR about AHI ?

djhoese: Yes, but this chunk of YAML was in the generic.yaml and *only* applies to ABI. It makes more sense to me to move this to the ABI specific enhancement config; especially since I copied it for the AHI specific config.

djhoese:
Here are some night examples. I noticed some odd artifacts in the night_microphysics but I *think* that is just in the data. Note these pictures are 10% the resolution of the original so the effect is made more apparent. I'll investigate this next.

## Airmass

airmass_20181022_170019

## Ash

ash_20181022_170019

## Dust

dust_20181022_170019

## Fog

fog_20181022_170019

## Night Microphysics

night_microphysics_20181022_170019

djhoese:
Yep this is the B07 having bad data in it:

Screenshot_20200212_121627

@mraspaud If the night_microphysics looks OK to you otherwise I think this is good to go.

ScottLindstrom:
That doesn't look like bad B07 data to me, rather it's a manifestation of the poor precision of B07 at very cold cloud top temperatures.  So a single radiance value might span several tenths of degrees at cold temperatures.  Thus you get speckled values.  All of the RGBs look reasonable to me, although I don't see a lot of obvious fog in the NightTime microphysics but that might be a function of the weather on that day.

djhoese:
I got approval from @mraspaud on slack. Merging now...

PR #1067, opened 2020-02-10
Add GOES-17 support to the 'geocat' reader

djhoese:
The 'geocat' group at the SSEC is now working on producing files operationally with GOES-17 data. I had hacked something together for GOES-16, but forgot to include GOES-17. This extends the hack to support GOES-17 until geocat adds projection information to their files (possibly in the future?).

 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``


djhoese:
Got slack approval from @mraspaud. Merging...

PR #1103, opened 2020-03-09
Add a reader for FY-4A LMI level 2 data

zxdawn:
This PR adds a reader for the FY-4A LMI (Lightning Mapping Imager) L2 data format which are NetCDF files.

 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/lmi_l2.py
+++ b/satpy/readers/lmi_l2.py
@@ -115,6 +115,10 @@ class LMIL2FileHandler(NetCDF4FileHandler):
         """Get metadata."""
         metadata = {}
         metadata.update(data.attrs)
+        units = data.attrs['units']
+        # fix the wrong unit "uJ/m*m/ster"
+        if not units.isascii():
+            metadata['units'] = b'\xc2\xb5J/m*m/ster'
         metadata.update(ds_info)
         metadata.update({
             'platform_shortname': self.filename_info['platform_id'],
--- a/satpy/readers/lmi_l2.py
+++ b/satpy/readers/lmi_l2.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2019 Satpy developers
+#
+# This file is part of satpy.
+#
+# satpy is free software: you can redistribute it and/or modify it under the
+# terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# satpy is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
+"""Lightning Mapping Imager Dataset reader
+
+The files read by this reader are described in:
+
+    http://fy4.nsmc.org.cn/data/en/data/realtime.html
+
+"""
+
+from satpy.readers.netcdf_utils import NetCDF4FileHandler, netCDF4
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class LMIL2FileHandler(NetCDF4FileHandler):
+    """File handler for LMI L2 netCDF files."""
+
+    @property
+    def start_time(self):
+        return self.filename_info['start_time']
+
+    @property
+    def end_time(self):
+        return self.filename_info['end_time']
+
+    @property
+    def platform_shortname(self):
+        return self.filename_info['platform_id']
+
+    @property
+    def sensor(self):
+        return self.filename_info['instrument'].lower()
+
+    def available_datasets(self, configured_datasets=None):
+        """Automatically determine datasets provided by this file."""
+        logger.debug("Available_datasets begin...")
+
+        # Determine shape of the geolocation data (lat/lon)
+        lat_shape = None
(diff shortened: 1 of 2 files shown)


satpy/readers/lmi_l2.py
```diff
@@ -0,0 +1,148 @@
...
+        return self.filename_info['instrument'].lower()
+
+    def available_datasets(self, configured_datasets=None):
+        """Automatically determine datasets provided by this file."""
+        logger.debug("Available_datasets begin...")
+
+        # Determine shape of the geolocation data (lat/lon)
+        lat_shape = None
+        for var_name, _val in self.file_content.items():
+            # Could probably avoid this hardcoding, will think on it
```
djhoese: If you know there is only one latitude dataset then you could load only the `LAT` *or* you could look through the configured datasets for one with `standard_name: latitude`. Just an idea.

zxdawn: Embed this into the part of reading configured datasets

zxdawn: @djhoese I tested this again and found a bug: some data like this `{'name': 'EOT', 'file_key': 'EOT', 'file_type': 'lmi', 'resolution': '7800m'}` don't have the `standard_name` key in the `ds_info` dict, because we don't add it to the `yaml` file. So, I switched to use `name` for this purpose.

zxdawn:
@djhoese I have to assign the `y` coords, although the units isn't meters. Otherwise, `MultiScene` won't work. Since there're `LON` and `LAT` variables, is the wrong units not matter?

BTW, if I assign the `time` coords which have the same dimension of `y` coords, the dimensions of `y` corrds at different times are still different. That's why `MultiScene` doesn't work.

satpy/readers/lmi_l2.py
```diff
@@ -0,0 +1,154 @@
...
+        logger.debug("Getting data for: %s", ds_id.name)
+        file_key = ds_info.get('file_key', ds_id.name)
+        data = self[file_key]
+        data.attrs = self.get_metadata(data, ds_info)
+        # rename coords
+        data = data.rename({'x': 'y'})
+        # assign 'y' coords which is useful for multiscene,
+        # although the units isn't meters
+        len_data = data.coords['y'].shape[0]
+        data.coords['y'] = np.arange(len_data)
```
djhoese, line 143: Why do you need to assign these? I don't think you need `y` coordinates here. Maybe I'm wrong and naming it `y` is a bad idea, but I think that is consistent with other readers and the rest of Satpy. The dimension should be `y`, but no `coords` for it. My thought was instead that you add a `data.coords['time'] = [... series of times ...]` then the MultiScene `timeseries` function could be updated to look at that coordinate. So `time` would be the only coordinate you add in the reader, the base reader will then add `lon` and `lat` for you.

zxdawn: @djhoese 

Got it! Here's the method:
```
        tseries = pd.DatetimeIndex(np.repeat(data.attrs['start_time'], data.shape[0]))
        data = data.expand_dims({'time': tseries})
        data = data.rename({'x': 'y'})
```
Example:
```
<xarray.DataArray 'ER' (time: 31, y: 31)>
dask.array<where, shape=(31, 31), dtype=float32, chunksize=(31, 31), chunktype=numpy.ndarray>
Coordinates:
  * time     (time) datetime64[ns] 2019-07-25T05:25:10 ... 2019-07-25T05:25:10
Dimensions without coordinates: y
```
But, I will get this error when using `MultiScene` by `mscn = MultiScene.from_files(filenames, reader='lmi_l2', group_keys=('start_time', 'subtask_num'))` and `imgs = mscn.blend(blend_function=timeseries)`:
```
ValueError: arguments without labels along dimension 'y' cannot be aligned because they have different dimension sizes: {8, 74, 44, 45, 86, 55, 24, 60, 93, 31}
```

The modification of `MultiScene` is mentioned in #1173.

zxdawn:
Here's an example of plotting the number of events on 7800m resolution.
It looks summed values are correct.
But, I'm not sure how to create the correct 7800m `AreaDefinition`. I can't find the explanation of this resolution ... So, is it suitable to create the correct `AreaDefinition` based on the sub-satellite point and same information of FY4A AGRI?

<details><summary>Code</tt></summary>

```
mscn = MultiScene.from_files(filenames, reader='lmi_l2', group_keys=('start_time', 'subtask_num'))
mscn.load(['ER'])

# resample to 7800m (resolution of LMI)
mscn = mscn.resample('china_7.8km', resampler='bucket_count')

# blend by time dim
imgs = mscn.blend(blend_function=timeseries)

# sum by time dim
sum_num = imgs['ER'].sum('time')

# copy attrs
sum_num.attrs = imgs['ER'].attrs

crs = sum_num.attrs['area'].to_cartopy_crs()
ax = plt.axes(projection=crs)
ax.coastlines()
ax.gridlines()
ax.set_global()
plt.imshow(sum_num, transform=crs, extent=crs.bounds, origin='upper')
cbar = plt.colorbar()
```

</details>

<details><summary>Figures</tt></summary>

2019-07-25 05:15:10 to 05:34:49
lmi_events

2019-07-25 05:15:10 to 05:24:49
lmi_events_1

2019-07-25 05:25:10 to 05:34:49
lmi_events_2

</details>

djhoese:
> But, I'm not sure how to create the correct 7800m `AreaDefinition`. I can't find the explanation of this resolution ... So, is it suitable to create the correct `AreaDefinition` based on the sub-satellite point and same information of FY4A AGRI?

That seems reasonable. I think it depends on what you are doing and what you hope to get out of it.

PR #1106, opened 2020-03-10
Scale valid range if available.

adybbroe:
This is to fix the coloring of the NWCSAF cloud parameters. It is not clear from documentation if the color palette applies to the scaled or unscaled values. To get the right coloring, corresponding to the nwcsaf.org examples, we need to scale the valid-range.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/nwcsaf_nc.py
+++ b/satpy/readers/nwcsaf_nc.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2017-2019 Satpy developers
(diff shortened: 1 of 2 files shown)


adybbroe:
Before reviewing I guess I just want to see that also the PPS products are colored correctly.

adybbroe:
Before reviewing I guess I just want to see that also the PPS products are colored correctly.

adybbroe:
As far as I can see nothing is broken. It works for both PPS and NWCSAF/Geo

PR #1107, opened 2020-03-10
Raise an error if data and angle shapes don't match in NIRReflectance

pnuu:
This is a fix to a minor bug that I encountered when using `nir_reflectance` modifier with VIIRS SDR data. By default Satpy loads the highest resolution (`I` channel) datasets, and when this modifier is used in a composite using `M`-channel data there's a crash because the arrays have different shapes.

Naturally, the correct way to define the composite is to use the `nir_reflectance_lowres` modifier where the proper resolution is defined.

 - [ ] Tests passed 
 - [x] Passes ``flake8 satpy``


djhoese:
This should probably use `projectables = self.match_data_arrays(projectables)` instead to match the other compositors/modifiers and handle cases other than the area/shape differences.

djhoese:
Oh there are a couple compositors that don't use this. Thoughts on adding this to all of them?

pnuu:
Oh, that was a new method to me! I've changed `NIRReflectance` and the emissive modifiers. I wasn't sure what to do when optional prereqs are used, like in `PSPAtmosphericalCorrection`.

PR #1118, opened 2020-03-20
Fix supported Python version in devguide

gerritholl:
In the developer guide, update the supported Python versions.  Python 2
is no longer supported.  For maintainability, don't mention the specific
Python versions that are supported but refer the reader to ``setup.py``.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented 

### DIFF
--- a/doc/source/dev_guide/index.rst
+++ b/doc/source/dev_guide/index.rst
@@ -23,8 +23,9 @@ and all code should follow the
 `PyTroll coding guidelines and best
 practices <http://pytroll.github.io/guidelines.html>`_.
 
-Satpy currently supports Python 2.7 and 3.4+. All code should be written to
-be compatible with these versions.
+Satpy is now Python 3 only and it is no longer needed to support Python 2.
+Check ``setup.py`` for the current Python versions any new code needs
+to support.
 
 Development installation
 ========================


doc/source/dev_guide/index.rst
```diff
@@ -23,8 +23,9 @@ and all code should follow the
 `PyTroll coding guidelines and best
 practices <http://pytroll.github.io/guidelines.html>`_.
 
-Satpy currently supports Python 2.7 and 3.4+. All code should be written to
-be compatible with these versions.
+Satpy is now Python 3 only and it is no longer needed to support Python 2.
+Check ``setup.py`` for the current Python versions any new code needs
+to support.
```
djhoese: What are your thoughts about making the `setup.py` a link to the file on the master branch of github, specifically pointing to the `python_requires` line? Or maybe mention the `python_requires`? Or maybe we point instead to the main index page where we are more likely to keep this up to date?

gerritholl: I don't know how to specifically point to the `python_requires` line.  I know how to point to a specific line number, but that would get out of date as soon as any lines above are added or removed.

I'm not sure what you mean by the main index page.  There is no mention of supported Python versions at https://satpy.readthedocs.io/en/latest/index.html

djhoese: You're right about the line number. I realized it right after I mentioned it. But maybe that is the best we can do?

> There is no mention of supported Python versions at

You're right. Oops. I guess that isn't an option.

PR #1128, opened 2020-04-02
Add tm5_constant_a and tm5_constant_b for tropomi_l2

zxdawn:
Because of different shape, `tm5_constant_a` and `tm5_constant_b` are ignored by the `tropomi_l2.py` reader.
I added them in `tropomi_l2.yaml` now.

### DIFF
--- a/satpy/etc/readers/tropomi_l2.yaml
+++ b/satpy/etc/readers/tropomi_l2.yaml
@@ -54,4 +54,14 @@ datasets:
         file_type: tropomi_l2
         file_key: 'PRODUCT/time'
         standard_name: ref_time
+    tm5_constant_a:
+        name: 'tm5_constant_a'
+        file_type: tropomi_l2
+        file_key: 'PRODUCT/tm5_constant_a'
+        standard_name: tm5_constant_a
+    tm5_constant_b:
+        name: 'tm5_constant_b'
+        file_type: tropomi_l2
+        file_key: 'PRODUCT/tm5_constant_b'
+        standard_name: tm5_constant_b
 


djhoese:
CC @tommyjasmin for opinions.

tommyjasmin:
Hi all - @zxdawn Xin - can you explain the use of these constants and which products they are expected to be present in?  In the current YAML, we reference them directly under the PRODUCT group:

`'PRODUCT/tm5_constant_a'`

This appears to be correct for NO2 data, which I presume you are working with. 

But with a very brief look, I do **not** see these constants in the CO sample data I have.  I **do** see these constants in my SO sample data.  However, for those files they are present in a support data subgroup:

`'PRODUCT/SUPPORT_DATA/INPUT_DATA/tm5_constant_a'`

So we can't assume these constants are always present in the top level PRODUCT group.

zxdawn:
@tommyjasmin Hi Tommy, Yes, I tested it for NO2 files.

The constants are used to calculate the TM5 model pressure levels which are used in the NO2 retrieval, like this:
```
P = A + B*PSFC
```

I will fix this and test all kinds of tropomi products soon. Thanks.

zxdawn:
@tommyjasmin I've tested the newest YAML and it works well for NO2, SO2, O3 and CO data. Because the `var_name` in YAML is tested whether exists in the file:
https://github.com/pytroll/satpy/blob/2ce07c0b1f477e9cd400a4e9b7504f36a08cc1cd/satpy/readers/tropomi_l2.py#L98

So, is it OK to just keep these two variables in the YAML? @mraspaud @djhoese

PR #1141, opened 2020-04-13
Add night lights composites for ABI, AHI and AMI

simonrp84:
This PR adds true color composites that use the VIIRS night lights as a background for night-time pixels. No code changes have been made, it simply re-uses the existing SEVIRI night lights composite (`natural_color_with_night_ir`). In the case of the new instruments, however, I chose to use `true_color` instead of `natural_color` to reflect the capabilities of the instruments.

 - [x] Tests passed

### DIFF
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -134,6 +134,12 @@ composites:
     - C14
     standard_name: overview
 
+  colorized_ir_clouds:
+    compositor: !!python/name:satpy.composites.SingleBandCompositor
+    prerequisites:
+      - name: 'C13'
+    standard_name: colorized_ir_clouds
+
   airmass:
     compositor: !!python/name:satpy.composites.GenericCompositor
     prerequisites:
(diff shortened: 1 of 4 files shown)


simonrp84:
I've just pushed a few more composites to this, which add a `_nocorr` version of `true_color` and `natural_color` for SEVIRI, ABI, AMI and AHI.
These versions (i.e: `true_color_nocorr`) disable both the solar zenith and rayleigh scattering corrections, and hence produce data as it looked to the satellite sensor.

satpy/etc/composites/ahi.yaml
```diff
@@ -229,6 +258,12 @@ composites:
     - wavelength: 1.6
     standard_name: cloud_phase_distinction
 
+  colorized_ir_clouds:
+    compositor: !!python/name:satpy.composites.SingleBandCompositor
+    prerequisites:
+      - name: '10.4'
```
pnuu, line 264: This should either be the channel name `B13` or `10.4` (a float).

satpy/etc/composites/ami.yaml
```diff
@@ -49,6 +65,12 @@ composites:
       - 10.4
     standard_name: overview
 
+  colorized_ir_clouds:
+    compositor: !!python/name:satpy.composites.SingleBandCompositor
+    prerequisites:
+      - name: '10.4'
```
pnuu, line 71: This should either be the channel name B13 or 10.4 (a float).

simonrp84:
I've now, after a few failures, updated the YAML with @pnuu's request - the channels needed for the new composites are now specified by name.

satpy/etc/composites/ami.yaml
pnuu:
```suggestion
      - name: B13
```

satpy/etc/composites/abi.yaml
```diff
@@ -134,6 +162,12 @@ composites:
     - C14
     standard_name: overview
 
+  colorized_ir_clouds:
+    compositor: !!python/name:satpy.composites.SingleBandCompositor
+    prerequisites:
+      - name: 'C13'
```
pnuu, line 168:
```suggestion
      - name: C13
```

PR #1152, opened 2020-04-17
amsr2 reader for l2 ssw product

### LINKED ISSUE #1151: amsr2 l1b reader also match amsr2 l2 products
the amsr2 l1b reader also match amsr2 l2 products

l1b filename: GW1AM2_202004171159_065B_L1SNBTBR_2220220.h5
l2  filename:  GW1AM2_202004161432_090B_L2SNSSWLB3300300.h5

I'm looking into add reading of the L2 SSW product and this needs some small updates to the l1b reader to make it work.

Should I update the l1b reader or should I create a amsr2_l2 reader mostly the same as the l1b reader?

If I create a new reader we need to update the l1b reader to only match L1 data

TAlonglong:
Add reader for amsr2 l2 ssw product

 - [x] Closes #1151 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [ ] Fully documented

### DIFF
--- a/satpy/etc/readers/amsr2_l2.yaml
+++ b/satpy/etc/readers/amsr2_l2.yaml
@@ -5,26 +5,26 @@ reader:
   sensors: [amsr2]
 
 file_types:
-  amsr2_l2:
+  amsr2_l2_ssw:
     file_reader: !!python/name:satpy.readers.amsr2_l2.AMSR2L2FileHandler
-    file_patterns: ['{platform_shortname:3s}{instrument_shortname:3s}_{start_time:%Y%m%d%H%M}_{path_number:3d}{orbit_direction:1s}_L2{process_kind:2s}{product_id:3s}{resolution_id:1s}{dev_id:1s}{product_version:1s}{algorithm_version:3d}{parameter_version:3d}.h5']
+    file_patterns: ['{platform_shortname:3s}{instrument_shortname:3s}_{start_time:%Y%m%d%H%M}_{path_number:3d}{orbit_direction:1s}_L2{process_kind:2s}SSW{resolution_id:1s}{dev_id:1s}{product_version:1s}{algorithm_version:3d}{parameter_version:3d}.h5']
 
 datasets:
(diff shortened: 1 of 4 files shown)


satpy/etc/readers/amsr2_l2.yaml
```diff
@@ -0,0 +1,41 @@
...
+    file_reader: !!python/name:satpy.readers.amsr2_l2.AMSR2L2FileHandler
+    file_patterns: ['{platform_shortname:3s}{instrument_shortname:3s}_{start_time:%Y%m%d%H%M}_{path_number:3d}{orbit_direction:1s}_L2{process_kind:2s}{product_id:3s}{resolution_id:1s}{dev_id:1s}{product_version:1s}{algorithm_version:3d}{parameter_version:3d}.h5']
+
+datasets:
+  longitude_ssw:
+    name: longitude_ssw
+    file_type: amsr2_l2
+    standard_name: longitude
+    resolution: 15000
+    units: degree
```
djhoese, line 18: Should be `degrees_east` and latitude units should be `degrees_north`.

satpy/etc/readers/amsr2_l2.yaml
```diff
@@ -0,0 +1,41 @@
+reader:
+  name: amsr2_l2
+  description: GCOM-W1 AMSR2 instrument HDF5 reader
+  reader: !!python/name:satpy.readers.yaml_reader.FileYAMLReader
+  sensors: [amsr2]
+
+file_types:
+  amsr2_l2:
```
djhoese, line 8: If each product is in a separate file then we should either have a separate file type for each product.

TAlonglong: ok

satpy/etc/readers/amsr2_l2.yaml
TAlonglong: will do

PR #1155, opened 2020-04-18
Lowercase sensor of tropomi_l2

### LINKED ISSUE #1153: tropomi reader: scene attributes and data array attributes are different
**Describe the bug**
The scene attributes and data array attributes are different. This is a problem concerning the sensor name. It's lowercase` 'tropomi'` for `scn.attrs['sensor']` but uppercase `'TROPOMI'` for `data_arr.attrs['sensor']`. 
This is a problem for adding composites as the composite .yaml file must be `tropomi.yaml`, but the enhancements file must be `TROPOMI.yaml`.

@djhoese wrote to this problem: _"I think this is a bug in the tropomi reader. What is happening is that the file handler doesn't define a sensor_names property. That means that for scn.attrs['sensor'] it is defaulting to what is in the YAML file for the reader.
So that's why the composites are lowercase.
The data_arr.attrs['sensor'] is being added by the file handler from the file attribute which is uppercase. That's what is being used to get the enhancement config name."_

Please fix sensor name to be equal concerning upper-/lowercase to make composites/enhancements .yaml file combinations work again.

zxdawn:
- [x] Closes #1153 

Make the `sensor` lowercase for DataArray in tropomi_l2 variables.

### DIFF
--- a/satpy/readers/tropomi_l2.py
+++ b/satpy/readers/tropomi_l2.py
@@ -62,8 +62,13 @@ class TROPOMIL2FileHandler(NetCDF4FileHandler):
         """Get sensor."""
         res = self['/attr/sensor']
         if isinstance(res, np.ndarray):
-            return str(res.astype(str))
-        return res
+            return str(res.astype(str)).lower()
+        return res.lower()
+
+    @property
+    def sensor_names(self):
+        """Get sensor set."""
+        return {self.sensor}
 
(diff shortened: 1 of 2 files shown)


pnuu:
The unit test for the reader needs to be updated for the lowercase naming: https://travis-ci.org/github/pytroll/satpy/jobs/676587111#L1981

PR #1157, opened 2020-04-20
Add a function to Scene that can compute the pixel size in degrees

simonrp84:
This PR enables the user to compute the size of each pixel in an image on a per-pixel basis. It does this by:
1) Computing lat/lon from the area definition
2) Rolling the lat/lon by one pixel and subtracting from the original pixel
3) square-rooting the sum of squared differences.
4) Creating a new dataset in the scene (`PixSize`) to store the output.

This is useful for geostationary satellites where the pixel size is *not* constant across the scene. It is not designed for use with polar sensors. The end-goal of adding this function is to update multiscene so that, rather than just stacking scenes on top of each other, it can optionally compute the ideal scene to use for a given pixel by minimising the pixel size on a per-pixel basis.

It can be called with: `scn.get_pixsize_from_area(my_dataset)`

 - [ ] Tests added 
 - [ ] Tests passed
 - [ ] Passes ``flake8 satpy``
 - [ ] Fully documented

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -276,8 +276,6 @@ class Scene(MetadataObject):
 
         Args:
             dataset: A dataset whose area you wish to examine.
-        Returns:
-            diff: A numpy array of pixel sizes.
         """
         in_area = self[dataset].attrs.get('area')
         lons, lats = in_area.get_lonlats()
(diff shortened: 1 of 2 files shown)


satpy/scene.py
```diff
@@ -271,6 +271,30 @@ def max_area(self, datasets=None):
...
+        # This does not account for differences in one direction or another, but
+        # is close enough for most purposes.
+        lonoff = lons - np.roll(lons, 1, axis=1)
+        latoff = lats - np.roll(lats, 1, axis=0)
+
+        # Now compute the total difference, in degrees.
+        diff = np.sqrt(lonoff*lonoff + latoff*latoff)
+
+        self['PixSize'] = self[dataset].copy()
+        self['PixSize'].attrs['units'] = 'Decimal degrees'
```
djhoese, line 298: Is this a CF unit?

simonrp84: Changed it to `degrees`, which is in the CF standard.

djhoese:
It looks like this pull request was using a branch from an old fork of yours @simonrp84. Not sure we can get it back.

PR #1175, opened 2020-05-04
Add more flexible masking

### LINKED ISSUE #1168: Add more versatile options for masking datasets
## Feature Request

Currently we have `MaskingCompositor` that can be used to mask data with categorical data from NWC SAF PPS and GEO Cloud Type data. There is also `CloudCompositor` that does masking by creating the mask from the data it self (and adds some transitional smoothing for the transparency).

We should have a more flexible and generic way of masking data, e.g. using quality flags, thresholds and even combination of several things.

A generic way for describing different masking operations in the composite YAML files. Either as a modifier, or as a compositor.

This feature would simplify the workflow for masking data in multiple use cases.

**Additional context**
There are two approaches from which to choose:

-  a new compositor
-  a new modifier

The compositor approach would mean "wrapping" all the desired composites within another compositor. The modifier should be much easier to apply.

But the thing I need input for (I'll add something below on a new comment):
1. use cases
2. ideas on how the configuration should look like for different cases

pnuu:
This PR adds a more flexible way to apply masking to composites.

 - [x] Closes #1168 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1608,11 +1608,14 @@ class MaskingCompositor(GenericCompositor):
                                settings.
 
         Each condition in *conditions* consists of of three items:
-        - `method`: the Numpy-equivalent method name for <, >, <=, ==, >= or >
+        - `method`: the Numpy-equivalent method name for the following
+          opertions: <, >, <=, ==, >=, >, `isnan`, `isinf`
         - `value`: threshold value of the *mask* applied with the
           operator. Can be a string, in which case the corresponding
           value will be determined from `flag_meanings` and
           `flag_values` attributes of the mask.
+          NOTE: the `value` should not be given to 'isnan` or `isfinite`
+          methods.
         - `transparency`: transparency from interval [0 ... 100] used
           for the method/threshold. Value of 100 is fully transparent.
 
@@ -1623,7 +1626,9 @@ class MaskingCompositor(GenericCompositor):
                             {'method': 'greater_equal', 'value': 1,
                              'transparency': 80},
                             {'method': 'greater_equal', 'value': 2,
-                             'transparency': 0}]
+                             'transparency': 0},
+                            {'method': 'isnan',
+                             'transparency': 100}]
           >>> compositor = MaskingCompositor("masking compositor",
                                              transparency=transparency)
           >>> result = compositor([data, mask])
@@ -1632,7 +1637,8 @@ class MaskingCompositor(GenericCompositor):
         the `mask` dataset.  Locations where `mask` has values of `0`
         will be fully transparent, locations with `1` will be
         semi-transparent and locations with `2` will be fully visible
-        in the resulting image.  All the unlisted locations will be
+        in the resulting image.  In the end all `NaN` areas in the mask are
+        set to full transparency.  All the unlisted locations will be
         visible.
 
         The transparency is implemented by adding an alpha layer to
@@ -1680,7 +1686,7 @@ class MaskingCompositor(GenericCompositor):
 
         for condition in self.conditions:
             method = condition['method']
-            value = condition['value']
+            value = condition.get('value', None)
             if isinstance(value, str):
                 value = _get_flag_value(mask_in, value)
             transparency = condition['transparency']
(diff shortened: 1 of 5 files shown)


pnuu:
It seems that my initial idea of adding a modifier isn't feasible, as the modifier can't be used to add alpha channel. Thus I went with yet another compositor.

The first test image looks like this:
ir_test_20200417_060000

Here brightness temperature values above 270 K are fully transparent, and values from 260 to 270 are with 50 % transparency and the values are inverted before saving.

Corresponding composite definition:
```yaml
  ir_test:
    compositor: !!python/name:satpy.composites.MaskCompositor
    conditions:
    # operator, limit/threshold/value, transparency
    # the below is equivalent to
    #    alpha_val = 1. - transparency / 100.
    #    data = data.where(np.greater_equal(mask, 260), alpha_val, 1.0)`
    - [greater_equal, 260, 50]
    - [greater_equal, 270, 100]
    prerequisites:
    - name: IR_108
    - name: IR_108
    standard_name: ir_test
```

and enhancement:
```yaml
  ir_test:
    standard_name: ir_test
    operations:
    - name: invert
      method: *stretchfun
      kwargs:
        stretch: crude
        # Note: also the alpha-channel limits need to be defined!
        min_stretch: [300, 0]
        max_stretch: [215, 1]
```

pnuu:
I'm not sure what causes the weird pattern on the right of the image. It's not visible when I open the original GeoTIFF image in GIMP, but it is visible when viewed with `qiv`.

pnuu:
As a side note: if I save directly to PNG the image or omit the `tiled=True` kwarg to the `lcl.save_datasets()` the image is fine.

pnuu:
Would it be more clear to have
```yaml
    - method: greater_equal
      value: 260
      transparency: 50
```
instead of
```yaml
    - [greater_equal, 260, 50]
```
?
Or maybe there are better names for the "dictionary" version above, or another completely different way to give these information. Suggestions are wellcome! I'm going to adapt for the `method`/`value`/`transparency` approach in the meanwhile, it seems much more clear.

pnuu:
The composite configuration is now like this:
```yaml
  ir_test:
    compositor: !!python/name:satpy.composites.MaskCompositor
    conditions:
    # operator, limit/threshold/value, transparency
    # the below is equivalent to
    #    alpha_val = 1. - transparency / 100.
    #    data = data.where(np.greater_equal(mask, 260), alpha_val, 1.0)`
    - method: greater_equal
      value: 260
      transparency: 50
    - method: greater_equal
      value: 270
      transparency: 100
    prerequisites:
    - name: IR_108
    - name: IR_108
    standard_name: ir_test
```

pnuu:
In the end I'll be merging this to the existing `MaskingCompositor` which was created to handle specifically masking with categorical cloud type data from NWC SAF PPS/GEO.

pnuu:
Now also the existing composites that were using `MaskingCompositor` are converted to use the new syntax, and the functionalities are merged in to one compositor.

pnuu:
I added one new composite that uses this feature: `hrv_severe_storms_masked`. It also uses a new feature, where `np.isnan` can be used to create a mask. Otherwise the data outside `HRV` channel would get a non-transparent background in the alpha channel.

pnuu:
crop_hrv_severe_storms_masked_20200417_060000

This is a crop of the new `hrv_severe_storms_masked` composite around central Africa.

pnuu:
I can't make Codebeat any happier. The `MaskingCompositor.__init__()` needs all the arguments, and the other complaints are to files that I didn't touch.

satpy/composites/__init__.py
```diff
@@ -1647,35 +1669,81 @@ def __call__(self, projectables, *args, **kwargs):
...
         return res
+
+    def _get_mask(self, method, value, mask_data):
+        """Get mask array from *mask_data* using *method* and threshold *value*.
+
+        The *method* is the name of a numpy function.
+
+        """
+        try:
+            func = getattr(np, method)
```
mraspaud, line 1714: So in principle this means any numpy method can be called. Just for security reasons, would you mind adding a check that `method` is not outside what we expect ?

pnuu: I added a list of accepted methods, and an unit test for trying to use something else. I hope I listed all the currently relevant methods.

gerritholl:
Documentation on `hrv_severe_storms`: https://www.slideserve.com/yepa/applications-of-meteosat-second-generation-msg

PR #1178, opened 2020-05-04
Optimize readers searching for matching filenames

### LINKED ISSUE #1172: find_files_and_readers is slow
**Describe the bug**

The function `satpy.readers.find_files_and_readers` is slow.  This is particularly clear for instruments with many files and many file patterns. For ABI L1b, an hour of M6 full disk data consists of 96 files, for M6 CONUS data this is even 196 files.  There are 16 file patterns defined in the yaml file.  It appears that `find_files_and_readers` scales poorly with both the number of files and the number of patterns.  With 125 files, `find_files_and_readers` takes nearly 3 seconds, even if all actual disk I/O is filtered out.

**To Reproduce**

This reproduction is based on current master but with the branch for PR #1169 merged, so that I can implement a dummy filesystem class to exclude I/O from the performance measurement.  In this example, `glob` always returns a list of the same 125 files instantly:

```python
import datetime
import timeit
import satpy.readers
pat = "noaa-goes16/ABI-L1b-Rad{m:s}/2020/{d:>03d}/{h:>02d}/OR_ABI-L1b-RadC-M3C{ch:>02d}_G16_s2020{d:>03d}{h:>02d}00000_e2020{d:>03d}{h:>02d}10000_c2020{d:>03d}{h:>02d}15000.nc"                                                                                                        
ref = [pat.format(m="F", ch=ch, d=d, h=h)
        for ch in range(5)
        for d in range(5)
        for h in range(5)]

class DummyFS:
    def glob(self, _):
        return ref

satpy.readers.find_files_and_readers(
        base_dir="/it/doesnt/matter",
        fs=DummyFS(),
        reader="abi_l1b",
        start_time=datetime.datetime(2020, 1, 1, 1, 30))
```

Running this with the Python profiler:

```
python -m cProfile -o mwe37b.prof mwe37.py
```

gives us a profile file that we can visualise with [snakeviz](https://jiffyclub.github.io/snakeviz/).

**Expected behavior**

I would expect this to run quickly.  Sure, it needs to test 125 files up to 16 times, but a single call to `fnmatch` only lasts 689 ns ± 45 ns:

```python
In [25]: %timeit fnmatch.fnmatch('noaa-goes16/ABI-L1b-RadF/2020/000/00/OR_ABI-L1b-RadC-M3C00_G16_s20200000000000_e20200000010000_c20200000015000.nc', '??_???-L1b-???*-??C01_???_s?????????????*_e?????????????*_c?????????????*.nc*')
689 ns ± 44.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Even when multiplied by 125*16 this makes just 1.378 milliseconds, 3 orders of magnitude less than `find_files_and_readers`, so the overhead is somewhere else than in matching the filenames.  I think it should be reasonable to expect `find_files_and_readers` to run at least 10 times faster than it does now and probably in less than 100 ms for this particular case.

**Actual results**

When running without the profiler, the function takes 2.4 seconds.  When running with the profiler, it takes about 6.1-6.3 seconds. The profiling results (see above) reveals that cumulatively, 5.12 seconds are spent in `globify`, which is apparently called 50020 times.  The full profile:

[mwe37b.prof.gz](https://github.com/pytroll/satpy/files/4559299/mwe37b.prof.gz)

**Screenshots**

Screenshot of the "Icicle" type visualisation in snakeviz, revealing 6.13 seconds for `find_files_and_readers` including a cumulative 5.12 seconds in `globify`:

grafik

When sorting the table of statistics per function by cumulative time, we can see the number of calls for the most time-consuming parts of the code:

grafik

Apparently, there are 20 calls to `match_filenames`, that's OK, but then 50020 calls to `globify`, which seems excessive.

**Environment Info:**
 - OS: openSUSE 15.0
 - Satpy Version: Feature branch, branche off master at `v0.21.0-14-g65cd83f9` with both #1165 and #1169 merged. 
 - Trollsift version: 0.3.4

**Additional context**

This may seem such a big deal compared to the I/O involved with reading or downloading the data, nor with the total processing time.  However, in [fogtools](https://github.com/gerritholl/fogtools/) it is considerably slowing down unit tests that simulate downloading of ABI data including file selection for different parameters, which is where I notice this most.

djhoese:
This addresses some of the performance issues mentioned in #1172. Thanks to @gerritholl's investigating, it was discovered that the base reader's functions for searching for file was taking a long time. We've tracked it down to a few key parts, the main one being that globify was called thousands of times for only hundreds of files. The changes in this PR seem to make a big difference.

Here is a PyCharm profiling graph showing the most called or longest running functions in the call graph of my test script:

globify_orig

With a total run time of ~16.5s. Here's what it looks like after this PR (globify is not even listed):

globify_2

With a total run time of ~4.7s.

 - [x] Closes #1172 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -199,9 +199,11 @@ class AbstractYAMLReader(metaclass=ABCMeta):
     def select_files_from_pathnames(self, filenames):
         """Select the files from *filenames* this reader can handle."""
         selected_filenames = []
+        filenames = set(filenames)  # make a copy of the inputs
 
         for pattern in self.file_patterns:
             matching = match_filenames(filenames, pattern)
+            filenames -= set(matching)
             for fname in matching:
                 if fname not in selected_filenames:
                     selected_filenames.append(fname)
@@ -398,11 +400,14 @@ class FileYAMLReader(AbstractYAMLReader):
     @staticmethod
     def filename_items_for_filetype(filenames, filetype_info):
         """Iterate over the filenames matching *filetype_info*."""
+        if not isinstance(filenames, set):
+            # we perform set operations later on to improve performance
+            filenames = set(filenames)
         matched_files = []
         for pattern in filetype_info['file_patterns']:
-            for filename in match_filenames(filenames, pattern):
-                if filename in matched_files:
-                    continue
+            matches = match_filenames(filenames, pattern)
+            filenames -= matches
+            for filename in matches:
                 try:
                     filename_info = parse(
                         pattern, get_filebase(filename, pattern))
@@ -500,6 +505,9 @@ class FileYAMLReader(AbstractYAMLReader):
 
     def filter_selected_filenames(self, filenames):
         """Filter provided files based on metadata in the filename."""
+        if not isinstance(filenames, set):
+            # we perform set operations later on to improve performance
+            filenames = set(filenames)
         for _, filetype_info in self.sorted_filetype_items():
             filename_iter = self.filename_items_for_filetype(filenames,
                                                              filetype_info)
@@ -511,6 +519,9 @@ class FileYAMLReader(AbstractYAMLReader):
 
     def new_filehandlers_for_filetype(self, filetype_info, filenames, fh_kwargs=None):
         """Create filehandlers for a given filetype."""
+        if not isinstance(filenames, set):
+            # we perform set operations later on to improve performance
+            filenames = set(filenames)
         filename_iter = self.filename_items_for_filetype(filenames,
                                                          filetype_info)
         if self.filter_filenames:
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -406,7 +406,7 @@ class FileYAMLReader(AbstractYAMLReader):
         matched_files = []
         for pattern in filetype_info['file_patterns']:
             matches = match_filenames(filenames, pattern)
-            filenames -= matches
+            filenames -= set(matches)
             for filename in matches:
                 try:
                     filename_info = parse(
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -188,9 +188,12 @@ class AbstractYAMLReader(metaclass=ABCMeta):
         filenames = set()
         if directory is None:
             directory = ''
-        for pattern in self.file_patterns:
-            matching = glob.iglob(os.path.join(directory, globify(pattern)))
-            filenames.update(matching)
+        # all the glob patterns that we are going to look at
+        all_globs = set(os.path.join(directory, globify(pattern))
+                        for pattern in self.file_patterns)
+        # get all files matching these patterns
+        for glob_pat in all_globs:
+            filenames.update(glob.iglob(glob_pat))
         return filenames
 
     def select_files_from_pathnames(self, filenames):
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -64,7 +64,7 @@ def listify_string(something):
         return list()
 
 
-def get_filebase(path, pattern):
+def _get_filebase(path, pattern):
     """Get the end of *path* of same length as *pattern*."""
     # convert any `/` on Windows to `\\`
     path = os.path.normpath(path)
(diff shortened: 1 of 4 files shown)


djhoese:
I think I'd like to look at caching these globified patterns so in MultiScene cases they are only globified once. That's my next step although I don't expect it to make a big difference with my current test case.

mraspaud:
If globify isn't listed anymore, then maybe the optimization should happen somewhere else now ?

djhoese:
Did a quick MultiScene test trying to create 60 Scenes using `MultiScene.from_files` with the current changes from this PR:

satpy_multiscene_2

Takes 37 seconds to create 60 Scenes. Not sure that is good or bad.

djhoese:
We could think about using `asv` to perform some benchmark tests: https://tomaugspurger.github.io/maintaing-performance.html

If we wanted this semi-automated we could add it to the integration tests on Jenkins (running on the SSEC's bumi server).

djhoese:
Ok, after answering messages and taking a break for a bit, I started playing around with `lru_cache` (from `functools` in the standard library) and added it to some key points where patterns were being used. It cut the timing down to 2.8s total.

globify_3

On slack @mraspaud had brought up making `parser` faster, but I wasn't sure that was possible since it is a lower-level function. This caching shows it is possible. We just have to control it well so we don't fill up people's disk or hold on to too much (the default is 128 items I think).

I'd still like to restructure some of the stuff in Satpy as it is needlessly calling some of these functions. I'll see what I can do.

djhoese:
Ok I think I'm done with all of the satpy-specific optimizations, but I have a related refactor I want to do. @mraspaud what are your thoughts on renaming some of the internal functions so they have `_` at the front? @gerritholl has asked if some functions are considered public interfaces. Some of them are, sure, but others shouldn't be. I'd like to do prefix some of these with `_` so I can more easily justify require a `set` object as input rather than converting to a `set` first everywhere.

djhoese:
See https://github.com/pytroll/trollsift/pull/25 for trollsift optimization.

djhoese:
Here is the pycharm profiling graph after both these sets of changes (total run time ~2.3s):

globify_4

gerritholl:
I tried this with the latest trollsift master and also #1169 merged.  With the same test script as before.  New times are:

* 125 fake files: 40 calls to `globify`, 0.14 seconds for `find_files_and_readers` (previously 5.12 seconds).
* 1000 fake files: 40 calls to `globify`, 0.28 seconds for `find_files_and_readers`.
* 60000 fake files: 40 calls to `globify`, 12.4 seconds for `find_files_and_readers`.

satpy/readers/yaml_reader.py
```diff
@@ -184,20 +185,25 @@ def select_files_from_directory(self, directory=None):
...
-        filenames = []
+        filenames = set()
         if directory is None:
             directory = ''
-        for pattern in self.file_patterns:
-            matching = glob.iglob(os.path.join(directory, globify(pattern)))
-            filenames.extend(matching)
+        # all the glob patterns that we are going to look at
+        all_globs = set(os.path.join(directory, globify(pattern))
+                        for pattern in self.file_patterns)
```
gerritholl, line 193: Why `set(x for ...)` rather than `{x for ...}`?  It probably makes no difference in practice (the latter is a bit faster but this can't possibly be a bottleneck), just wondering.

satpy/readers/yaml_reader.py
```diff
@@ -407,6 +415,7 @@ def filename_items_for_filetype(filenames, filetype_info):
                     continue
                 matched_files.append(filename)
                 yield filename, filename_info
+            filenames -= set(matched_files)
```
gerritholl, line 418: Why is `matched_files` first a list (line 407), then turned into a set here, rather than a set all the time?

satpy/readers/yaml_reader.py
djhoese: Oh probably just from typing it out and not switching it. I guess this started as a generator expression and then I just decided to wrap it in something. I'll switch it since it is such a small thing for better performance.

satpy/readers/yaml_reader.py
djhoese: Good point. I think I was worried about if matched_files needed to be a certain order for the best performance. I think you're right though, I should just make everything a set instead of copying so many times.

djhoese:
I think I've addressed everything @gerritholl mentioned and I'm ready for a re-review and merge assuming the tests pass.

gerritholl:
Excellent work, thanks.  All good as far as I can tell :)

PR #1196, opened 2020-05-08
Fix the geolocation of the HRV channel in the SEVIRI native reader

ameraner:
This PR fixes a shift in the SEVIRI native reader HRV channel geolocation. 

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``


ameraner:
A shift was observed in the geolocation of the native HRV channel.

This is caused by the wrong assumption that the projection center point for HRV is placed in the middle of the central pixel (as for all other channels). Instead, for HRV it is two pixels down and two pixels left from the center. See excerpt from the Image Format Description:
image


**Notice how the center is in the pixel with coordinates 5566 (_and not image_size/2=11136/2=5568_).** 


Test of current implementation:
image

Test after this PR:
image

satpy/readers/seviri_l1b_native.py
```diff
@@ -329,7 +329,7 @@ def get_area_extent(self, dataset_id):
 
         if dataset_id.name == 'HRV':
             grid_origin = data15hd['ImageDescription']['ReferenceGridHRV']['GridOrigin']
-            center_point = HRV_NUM_COLUMNS / 2
+            center_point = (HRV_NUM_COLUMNS-2) / 2
```
sjoro, line 332: with 11136 columns the `center_point` with this formula is 5567. according to the documentation the center point should be 5566. should this be `center_point = (HRV_NUM_COLUMNS / 2) - 2`? or am i missing something here?

ameraner: You are right, thanks for spotting this. Sorry, dumb mistake. I wonder when area extents will stop confusing me....

ameraner:
@sjoro found a mistake in my fix, this is the image after the correction:
fixhrv

mraspaud:
Thanks for fixing this, merging now.

PR #1199, opened 2020-05-11
Cache dimension per variable

### LINKED ISSUE #1198: Let NetCDF4FileHandler cache variable dimension names
## Feature Request

In #720, I would like to access the dimension names for variables prior to reading them, in order to decide whether or not to return the variable from the `available_datasets()` module.  Currently I can access the shape for each data variable and the dictionary of dimensions for the entire file, but not the dimension names belonging to a particular variable.

I'd like that as part of the metadata stored for each variable:

https://github.com/pytroll/satpy/blob/cf9aab464517b3af66f7de1699974bb310d68b0e/satpy/readers/netcdf_utils.py#L147-L152

there is also a storage of the dimension names, such as currently done globally:

https://github.com/pytroll/satpy/blob/cf9aab464517b3af66f7de1699974bb310d68b0e/satpy/readers/netcdf_utils.py#L157-L159

No changes in reasonable circumstances.

**Additional context**

Currently I'm reading the variable and then accessing the attributes, but this is inefficient.  Moreover, it's harder to test as the testing approach in readers using `NetCDF4FileHandler` has been to return fake data as an xarray.DataArray in `NetCDF4FileHandler.file_content, whereas the real reader puts `netCDF4._netCDF4.Variable` instances in this dictionary, so the mock diverges from the real case which complicates testing.

gerritholl:
In the NetCDF4FileHandler class, cache the tuple of dimension names for
each variable processed.

 - [x] Closes #1198
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -149,6 +149,7 @@ class NetCDF4FileHandler(BaseFileHandler):
             self.file_content[var_name] = var_obj
             self.file_content[var_name + "/dtype"] = var_obj.dtype
             self.file_content[var_name + "/shape"] = var_obj.shape
+            self.file_content[var_name + "/dimensions"] = var_obj.dimensions
             self._collect_attrs(var_name, var_obj)
         self._collect_attrs(name, obj)
 
--- a/satpy/tests/reader_tests/test_netcdf_utils.py
+++ b/satpy/tests/reader_tests/test_netcdf_utils.py
@@ -54,6 +54,7 @@ class FakeNetCDF4FileHandler(NetCDF4FileHandler):
             - '/attr/global_attr'
             - 'dataset/attr/global_attr'
             - 'dataset/shape'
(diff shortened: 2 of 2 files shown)


gerritholl:
Does "codebeat pending" mean I have to resolve the one introduced issue before it approves the PR?  I'm not sure what it's doing otherwise and I don't know how to solve the assignment branch condition issue I have apparently introduced.

djhoese:
Pending is supposed to mean it is still processing. In this case I think the github notification/event got hung up coming from codebeat so github is still waiting for it to finish. Even more odd, codebeat has the message of how many issues were introduced/resolved but didn't say it is finished.

If you click on help on codebeat for that issue it will tell you why it was introduced. In this case I think it is caused by too many assignments in one function. I personally don't think it is worth trying to fix it. codebeat is too strict in some cases.

PR #1211, opened 2020-05-19
Fix "rows_per_scan" not being available from VIIRS SDR readers

djhoese:
The "rows_per_scan" piece of metadata used to be in this reader (I'm 90% sure), but somewhere along the line it disappeared. This PR adds it back in. It's also possible it is in viirs_l1b and maybe the compact reader but never got added to this reader. Anyway...should be fixed now with tests to check for it.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy`` 
 - [ ] Fully documented 

I checked the flake8 line because I didn't get a warning from pre-commit, but I'm also pretty sure some of the docstrings are wrong. Maybe it isn't checking the full file anymore?

### DIFF
--- a/satpy/readers/viirs_compact.py
+++ b/satpy/readers/viirs_compact.py
@@ -155,6 +155,7 @@ class VIIRSCompactFileHandler(BaseFileHandler):
         else:
             m_data = self.read_geo(key, info)
         m_data.attrs.update(info)
+        m_data.attrs['rows_per_scan'] = self.scan_size
         return m_data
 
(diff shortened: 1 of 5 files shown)


djhoese:
Ah turns out I didn't have pre-commit installed on my current environment. Will try now.

djhoese:
Done

djhoese:
Got the OK from Martin on slack. Merging (appveyor is passing but the notification didn't make it all the way to github apparently).

PR #1227, opened 2020-06-01
Delete kdtree after saving cache

pnuu:
This simple change (delete the kdtree after saving the cache) makes it possible to use `dask.distributed` with `nearest` resampling. The only requirement is that `cache_dir` needs to be used, so this will not be very useful for other than geostationary satellites.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``


pnuu:
The computations appear to happen automatically in the local thread when saving the cache, so doesn't need manual adjustment for the scheduler. At least with the currend Dask version :grimacing:

PR #1188, opened 2020-05-06
Add areas for FCI

gerritholl:
Add two areas 'fci_0deg_1km' and 'fci_0deg_2km', corresponding to the
full disk FCI view.  I created those areas using:

```python
print(sc["vis_08"].attrs["area"].create_areas_def())
```

and

```python
print(sc["ir_123"].attrs["area"].create_areas_def())
```

with subsequent manual editing of name and description.

 - [x] Closes #1132  
 - [ ] Tests passed

PR #1228, opened 2020-06-03
Correct FCI area definitions in areas.yaml

ameraner:
This PR corrects the new area definitions for the FCI full disk added to areas.yaml by #1188 .

On the side, it also removes the duplicated `worldeqc30km` areadef. 

 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/etc/areas.yaml
+++ b/satpy/etc/areas.yaml
@@ -126,18 +126,6 @@ SouthAmerica_flat:
     lower_left_xy: [-8326322.82790897, -4609377.085697311]
     upper_right_xy: [-556597.4539663679, 1535833.8895192828]
     units: m
-worldeqc30km:
-  description: World in 3km, platecarree
-  projection:
-    proj: eqc
-    ellps: WGS84
-  shape:
-    height: 410
-    width: 820
-  area_extent:
-    lower_left_xy: [-20037508.3428, -10018754.1714]
-    upper_right_xy: [20037508.3428, 10018754.1714]
-    units: m
 south_america:
   description: south_america, platecarree
   projection:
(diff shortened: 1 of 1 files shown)


ameraner:
The area definitions in this PR are created by taking the values from:

```
print(scn["vis_08"].attrs["area"].create_areas_def())
print(scn["ir_123"].attrs["area"].create_areas_def())
```

and by switching the line values (y) between `lower_left_xy` and `upper_right_xy`, effectively flipping the image upside down.

satpy/etc/areas.yaml
```diff
@@ -126,18 +126,6 @@ SouthAmerica_flat:
...
-  projection:
-    proj: eqc
-    ellps: WGS84
-  shape:
-    height: 410
-    width: 820
-  area_extent:
-    lower_left_xy: [-20037508.3428, -10018754.1714]
-    upper_right_xy: [20037508.3428, 10018754.1714]
-    units: m
```
gerritholl, line 140: Looks like you deleted the world.

ameraner: Yes since it was duplicated, see: https://github.com/pytroll/satpy/blob/a761d23d1e2839b35fa1053a63a4186367c57b17/satpy/etc/areas.yaml#L201-L212

Pycharm was complaining that the key is defined twice, that's why I noticed. There's a small remark in the PR description

gerritholl: Ah, so it is.  I missed that you mentioned it in the PR, sorry.

ameraner: No problem :)

PR #1233, opened 2020-06-07
Support emtpy datasets in yaml reader

zxdawn:
- [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 

Because some L2 and L3 data have different varnames in different products, we need to support empty `datasets` in yaml reader.

### DIFF
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -240,49 +240,52 @@ class AbstractYAMLReader(metaclass=ABCMeta):
     def load_ds_ids_from_config(self):
         """Get the dataset ids from the config."""
         ids = []
-        for dataset in self.datasets.values():
-            # xarray doesn't like concatenating attributes that are lists
-            # https://github.com/pydata/xarray/issues/2060
-            if 'coordinates' in dataset and \
-                    isinstance(dataset['coordinates'], list):
-                dataset['coordinates'] = tuple(dataset['coordinates'])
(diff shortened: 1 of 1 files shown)


djhoese:
Could you provide more detail and some examples of what you want to work and what isn't currently working? The handling of YAML datasets is complex and fairly flexible so I'm sure there is a way to support what you need without changing how this works. For example, if L2 and L3 are different file types then the variables can be defined with per-file_type parameters. I'm 90% sure this is allowed and will work.

zxdawn:
Here's the example of four kinds of GPM L2 data:
image


The group name and root varnames are different. I don't know the suitable method of adding some to `datasets` in yaml.

Besides, the 2D lon and lat have different shapes under different groups in the same file:
image
I don't know how to set the coordinates which can be used in resample in the reader.

djhoese:
Sometimes it seems like file designers don't want people to easily use their files.

So this PR is needed because in your PR (#1234) you are defining all datasets with `available_datasets` right? I don't understand why this doesn't work already. Were you getting some kind of error before? In your other PR you have `datasets: ~`. What if you do `datasets: {}`?

zxdawn:
If I change `datasets: ~` to `datasets: {}`, it works well. Shall we close this PR?

mraspaud:
If you're happy with `datasets: {}`, please close this :)

PR #1244, opened 2020-06-29
Bugfix (FY4A AGRI): Correct the type of orbital_parameters to float

### LINKED ISSUE #1243: Wrong data type of orbital_parameters in FY4A AGRI reader
**Describe the bug**
The elements of `orbital_parameters` should be number instead of array.
Otherwise, it can cause the error when it's saved to netcdf file:
```
  File "/yin_raid/xin/miniconda3/envs/tobac/lib/python3.7/site-packages/satpy/writers/cf_writer.py", line 318, in default
    return [self.default(item) for item in obj]
TypeError: iteration over a 0-d array
```

**Expected behavior**
```
orbital_parameters: {'satellite_nominal_latitude': 0.0, 'satellite_nominal_longitude': 104.69999694824219, 'satellite_nominal_altitude': 35786000.0}
```

**Actual results**
```
orbital_parameters: {'satellite_nominal_latitude': array(0., dtype=float32), 'satellite_nominal_longitude': array(104.7, dtype=float32), 'satellite_nominal_altitude': array(35786000., dtype=float32)}
```

zxdawn:
- [x] Closes #1243 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/tests/reader_tests/test_agri_l1.py
+++ b/satpy/tests/reader_tests/test_agri_l1.py
@@ -292,7 +292,7 @@ class Test_HDF_AGRI_L1_cal(unittest.TestCase):
             else:
                 self.assertEqual('K', res[band_name].attrs['units'])
 
-        # check the data type of orbital_parameters
+        # check whether the data type of orbital_parameters is float
         orbital_parameters = res[band_names[0]].attrs['orbital_parameters']
         for attr in orbital_parameters:
             self.assertEqual(type(orbital_parameters[attr]), float)
--- a/satpy/tests/reader_tests/test_agri_l1.py
+++ b/satpy/tests/reader_tests/test_agri_l1.py
@@ -184,10 +184,16 @@ class FakeHDF5FileHandler2(FakeHDF5FileHandler):
     def get_test_content(self, filename, filename_info, filetype_info):
         """Mimic reader input file content."""
         global_attrs = {
-            '/attr/NOMCenterLat': 0.0, '/attr/NOMCenterLon': 104.7, '/attr/NOMSatHeight': 3.5786E7,
-            '/attr/dEA': 6378.14, '/attr/dObRecFlat': 298.257223563,
-            '/attr/OBIType': 'REGC', '/attr/RegLength': 2.0, '/attr/RegWidth': 5.0,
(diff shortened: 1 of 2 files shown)


satpy/tests/reader_tests/test_agri_l1.py
```diff
@@ -286,6 +292,11 @@ def test_fy4a_all_resolutions(self):
             else:
                 self.assertEqual('K', res[band_name].attrs['units'])
 
+        # check whether the data type of orbital_parameters is float
+        orbital_parameters = res[band_names[0]].attrs['orbital_parameters']
+        for attr in orbital_parameters:
+            self.assertEqual(type(orbital_parameters[attr]), float)
```
sfinkens, line 298: This would be a nice opportunity to also test the contents of the `orbital_parameters` dictionary here ;)

zxdawn: OK. I switched to check the exact values instead of type.

sfinkens: I meant testing both type and value, sorry if I wasn't clear on that.

zxdawn: Em ... If the type is `array`, then `assertEqual` would give `False`. So, this statement actually checks both type and value, right?

sfinkens: I didn't check `assertEqual`, but at least this evaluates to `True`:

```python
In [11]: np.array(0) == 0.                                                                                                                                                                                                                   
Out[11]: True
```

zxdawn: Oh, my fault! I tested it like:

```
a = [0.]
a == 0.
```

Forgot to add `np.array()` ...

sfinkens: Oh, in `get_test_content` you should probably set 
```
'/attr/NOMCenterLat': np.array(0.0),
```
instead of 
```
'/attr/NOMCenterLat': np.array([0.0]),
```
to trigger the original bug

zxdawn: Thanks, hope I solved all of them.

sfinkens:
Looks good to me now!

PR #1245, opened 2020-06-29
cf_writer supports array type in attrs

zxdawn:
If the attribute type is array, the cf_writer could cause the error which is similar to #1243 .

 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy``

### DIFF
--- a/satpy/writers/cf_writer.py
+++ b/satpy/writers/cf_writer.py
@@ -314,6 +314,8 @@ class AttributeEncoder(json.JSONEncoder):
             for key, val in obj.items():
                 serialized[key] = self.default(val)
             return serialized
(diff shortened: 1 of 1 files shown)


satpy/writers/cf_writer.py
```diff
@@ -314,6 +314,8 @@ def default(self, obj):
             for key, val in obj.items():
                 serialized[key] = self.default(val)
             return serialized
+        elif isinstance(obj, np.ndarray) and not obj.shape:
+            return float(obj)
```
djhoese, line 318: Isn't there a `.item()` method on numpy arrays to get the scalar value?

zxdawn: Thanks, modified.

PR #1260, opened 2020-07-15
Add missing docs with sphinx-apidoc

gerritholl:
Added missing documentations by re-running
``sphinx-apidoc -f -T -o source/api ../satpy ../satpy/tests``

 - [x] Fully documented

### DIFF
--- a/doc/source/api/satpy.rst
+++ b/doc/source/api/satpy.rst
@@ -5,6 +5,7 @@ Subpackages
 -----------
 
 .. toctree::
+   :maxdepth: 4
 
    satpy.composites
    satpy.demo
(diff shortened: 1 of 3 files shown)


doc/source/api/satpy.rst
```diff
@@ -5,6 +5,7 @@ Subpackages
 -----------
 
 .. toctree::
+   :maxdepth: 4
```
djhoese, line 8: What is the difference in the result between the default and 4?

gerritholl: No clue, all I did was run `sphinx-apidoc` as documented.

djhoese: Ah ok. So this is a generated change. Nevermind then.

PR #1262, opened 2020-07-17
Fix handling of HRV channel navigation for RSS data in seviri_l1b_native reader

ameraner:
This PR fixes the navigation of the HRV channel from Rapid Scanning Service data in the SEVIRI native format reader.

The issue was caused by the wrong assumption that the HRV RSS data is formatted the same way as HRV ROI data. This was causing wrong navigations, with stretched and shifted images. 

Instead, the RSS HRV is handled similarly to the full-disk case; the navigation still needs to be done using the "`Actual`" lines and columns positions stored in the trailer information. The difference w.r.t. the full-disk case is that only one area is present, instead of two separate ones. 

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/readers/seviri_l1b_native.py
+++ b/satpy/readers/seviri_l1b_native.py
@@ -47,7 +47,6 @@ from satpy.readers.seviri_l1b_native_hdr import (GSDTRecords, native_header,
                                                  native_trailer)
 from satpy.readers._geos_area import get_area_definition
 
-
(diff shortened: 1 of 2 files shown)


ameraner:
image

vis_sharpened_ir images
Left: current master, Right: with this PR

Notice the mismatch of the coloring, extracted from the IR channels, and the HRV background in the left image.

PR #1280, opened 2020-07-29
read specific area from fulldisk GOES16

jhbravo:
Commonly for generate the composite on a **Region of Interest** it is necessary read the entire data file, and then re-sample to particular region. 
based on examples:

- https://github.com/pytroll/pytroll-examples/blob/master/satpy/HRIT%20AHI%2C%20Hurricane%20Trami.ipynb

- https://github.com/pytroll/pytroll-examples/blob/master/satpy/ahi_true_color_pyspectral.ipynb

- https://github.com/pytroll/pytroll-examples/blob/master/satpy/FY4A_agri_introduction(EN).ipynb

It is not a problem, for this particular case using GOES16 data, when is used a region sector like CONUS or MESO, but for Fulldisk this may take a little longer time.

My proposal is define areas of interest, country's for example, for read only this section area instead of fulldisk, with some like this using Guatemala for read this country as specific region:

> from satpy import Scene
debug_on()
from glob import glob
pth = "/path/to/your/sat/data/G16-F-L1b-Rad-i/*.nc"
urls = glob(pth)
scn1 = Scene(reader='abi_l1b', filenames=urls, reader_kwargs={'sub_area': 'Guatemala'})
scn1.load(['C{:02d}'.format(j) for j in range(1,6)])
ahkm = scn1['C02'].area
rgbim = 'true_color'
lscn = scn1.resample(ahkm)
lscn.load([rgbim])
lscn.show(rgbim)

Captura de pantalla de 2020-07-29 14-24-54

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 


djhoese:
In general this shouldn't be necessary with dask. You should be able to use the `Scene.crop` method or slice the DataArrays yourself (`Scene` can be sliced too). With dask the data shouldn't be loaded from disk until you actually compute it, even if resampling is involved.

djhoese:
Do you have example timings to show the difference between this PR and using something like slicing or cropping as I described?

jhbravo:
> Do you have example timings to show the difference between this PR and using something like slicing or cropping as I described?

@djhoese just now, I finished doing some test, and I think in both cases the time is similar

> own -> 0:00:04.813529
> cropping -> 0:00:04.676791

really never before I had tried to use `crop`, and I think is a better solution than my proposal
however I will try to do some extra testing
thanks

PR #1282, opened 2020-07-30
Add parallax correction modifier

zxdawn:
This PR adds the **parallax correction** modifier which is useful for high/convective clouds.
This modifier needs the input of the IR channel L1 data and the cloud top height L2 data.

 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -862,21 +862,6 @@ class ParallaxCorrector(CompositeBase):
         # get the corrected lon and lat based on cloud top height (km)
         lat_corr, lon_corr = self._parallax_corr(optional_datasets, sat_lon, sat_lat, sat_h, lon, lat)
 
-        import matplotlib.pyplot as plt
-        f, axs = plt.subplots(nrows=1, ncols=2)
-
-        ax = axs[0]
-        m = ax.imshow(lon)
-        plt.colorbar(m, ax=ax, cmap='viridis', shrink=0.5)
-        ax.set_title('original lon')
-
-        ax = axs[1]
-        m = ax.imshow(np.rad2deg(lon_corr))
-        plt.colorbar(m, ax=ax, cmap='viridis', shrink=0.5)
-        ax.set_title('parallax_corrected lon')
-        f.savefig('lon_parallax.png')
-        print('-'*10)
-
         # get indices for the pixels for the original position
         (proj_x, proj_y) = band.attrs['area'].get_proj_coords()
         (x, y) = band.attrs['area'].get_xy_from_proj_coords(proj_x, proj_y)
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -27,6 +27,7 @@ import dask.array as da
 import numpy as np
 import xarray as xr
 import yaml
+from numpy import sin, cos, tan, arctan, arctan2, sqrt
 
 try:
     from yaml import UnsafeLoader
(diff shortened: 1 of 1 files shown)


zxdawn:
This is still under development ...

Here's the bug in this stage:
I add these lines for checking the corrected longitude after `lat_corr, lon_corr = self._parallax_corr(.............)`:

```
        import matplotlib.pyplot as plt
        f, axs = plt.subplots(nrows=1, ncols=2)

        ax = axs[0]
        m = ax.imshow(lon)
        plt.colorbar(m, ax=ax, cmap='viridis', shrink=0.5)
        ax.set_title('original lon')

        ax = axs[1]
        m = ax.imshow(np.rad2deg(lon_corr))
        plt.colorbar(m, ax=ax, cmap='viridis', shrink=0.5)
        ax.set_title('parallax_corrected lon')
        f.savefig('lon_parallax.png')
```

The result is wrong:
Screenshot from 2020-07-30 15-02-27

I followed the method in [ORAC correct_parallax](https://github.com/ORAC-CC/orac/blob/master/post_processing/correct_parallax.F90) written by @simonrp84 in Fortran. Not sure where is wrong ...

zxdawn:
## Method

This is the method to call this modifier:

```
from glob import glob
from satpy import Scene

f_l1 = glob('FY4A/L1/DISK/FY4A-_AGRI*_2019080706*.HDF')
f_l2 = glob('FY4A/L2/CTH/FY4A-_AGRI*2019080706*NC')
scn = Scene({"agri_l1": f_l1, "agri_l2": f_l2})

ds_id = DatasetID(name='C12', modifiers=('parallax_corrected',))
scn.load([ds_id])
```

mraspaud:
@zxdawn Great work so far! Regarding the multiple possibilities for the cth, could you use only optional prerequisites and raise an error if you get none?

zxdawn:
@mraspaud Thanks. Regarding using only optional prerequisites, do you have any example? I don't get the exact point ...

mraspaud:
For example, the sun zenith correction doesn't have any hard prerequisites: https://github.com/pytroll/satpy/blob/master/satpy/composites/__init__.py#L381-L418
You could do the same here and just take the cth that is available

zxdawn:
@mraspaud  Updated. Hope that's the correct method?

satpy/composites/__init__.py
```diff
@@ -743,6 +744,178 @@ def __call__(self, projectables, optional_datasets=None, **info):
...
+
+        # corrected surface coordinates
+        xcorr = xsurf + (x/axdiff)*xdiff
+        ycorr = ysurf + (x/axdiff)*ydiff
+        zcorr = zsurf + (x/axdiff)*zdiff
+
+        # convert back to latitude and longitude
+        latcorr = arctan(zcorr/sqrt(xcorr**2 + ycorr**2))
+        latcorr = np.deg2rad(arctan(tan(latcorr)*radius_ratio**2))
+        loncorr = np.deg2rad(arctan2(ycorr, xcorr))
```
simonrp84, line 806: I think these two lines should use `np.rad2deg` instead.

satpy/composites/__init__.py
```diff
@@ -743,6 +744,178 @@ def __call__(self, projectables, optional_datasets=None, **info):
...
+        paper: 10.3390/rs12030365 (doi)
+    """
+
+    def _parallax_corr(self, optional_datasets,
+                       sat_lon, sat_lat, sat_h,
+                       lon, lat):
+        """Applying Parallax Correction"""
+        # varius earth radius information
+        radius_eq = 6378.077
+        radius_pole = 6356.577
```
simonrp84, line 761: Can we have an option to take these from the area def, if possible? The GEOsats should all define these in their projection description so it'd be better to use those values, then fall back to the hardcoded values above for satellites that do not have the values in their metadata.

satpy/composites/__init__.py
simonrp84, line 806: I think these two lines should use `np.rad2deg` instead.

satpy/composites/__init__.py
```diff
@@ -743,6 +744,178 @@ def __call__(self, projectables, optional_datasets=None, **info):
...
+
+        # angle conversion to radians
+        sat_lat = np.deg2rad(sat_lat)
+        sat_lon = np.deg2rad(sat_lon)
+        lat = np.deg2rad(lat)
+        lon = np.deg2rad(lon)
+
+        # cartesian coordinates for the satellite
+        # satlat_geod is the geodetic satellite latitude
+        sat_lat_geod = arctan(tan(sat_lat)*radius_ratio**2)
```
simonrp84, line 772: In my code I'm dividing by `radius_ratio` rather than multiplying. Not sure who is right here. ;)
If you notice errors, try changing this to a division.

satpy/composites/__init__.py
```diff
@@ -743,6 +744,178 @@ def __call__(self, projectables, optional_datasets=None, **info):
...
+        xsat = sat_h * cos(sat_lat_geod) * cos(sat_lon)
+        ysat = sat_h * cos(sat_lat_geod) * sin(sat_lon)
+        zsat = sat_h * sin(sat_lat_geod)
+
+        # cartesian coordinates of the surface point
+        lat_geod = arctan(tan(lat)*radius_ratio**2)
```
simonrp84, line 778: Same here, I'm doing a division.

zxdawn:
Thanks for the review @simonrp84 .
All your comments are right and the result looks well now:
image

I will pick one convection case and plot that specific region later.

satpy/composites/__init__.py
zxdawn: Here's the projection of AGRI:

```
Projection: {'a': '6378140', 'h': '35786000', 'lon_0': '104.699996948242', 'no_defs': 'None', 'proj': 'geos', 'rf': '298.257223562999', 'type': 'crs', 'units': 'm', 'x_0': '0', 'y_0': '0'}
```

Just have `a` ....

How to deal with this issue

mraspaud:
About the semi axis, look here https://github.com/pytroll/satpy/issues/1278#issuecomment-665616648

zxdawn:
@mraspaud Thanks, Updated.

jleinonen:
Some comments from reading the code:

1. You have this requirement in `__call__`:
```python
        if len(projectables) != 2:
            raise ValueError("Expected 2 datasets \
                              (first: dataset to be corrected, \
                              second: IR band dataset), \
                              got %d" %
                             (len(projectables), ))
```
but later in this function only `projectables[0]` is used so why is the second element needed?

2. This check:
```python
        if da_name not in ['CTH', 'CLTH']:
            raise ValueError("Missing cloud top height inputs")
```
seems to check for valid cloud top height variable but in GOES-R data, for instance, this is called `HT` so it would fail here.

3. What happens in cloud-free areas? Probably the most appropriate thing to do for parallax correction is to set cloud top height to 0 there, but for example in GOES-R cloud top height product these pixels are set to `nan`.

4. What happens if the corrected longitude and latitude are outside the `area`?

zxdawn:
@jleinonen 

1. At the initial stage, I want to estimate the CTH by the IR dataset.
After a discussion with @simonrp84 , we decide to force users use the CTH data.
I will delete this condition statement.

2. Yes, that's where I'm not satisfied with ... Any suggestion of checking the CTH whether available?

3. Nothing is changed for the cloud-free area:
```
        # create the mask based on cth
        cth = optional_datasets[0]
        cth = cth.where(cth >= 0).to_masked_array(copy=False)

        # copy to the new data
        new_proj = xr.full_like(projectables[0], np.nan)

        # copy clear pixel
        ind = np.where(cth.mask == True)
        ind_x = xr.DataArray(x[ind])
        ind_y = xr.DataArray(y[ind])
        new_proj[ind_x, ind_y].values = projectables[0].isel(x=ind_x, y=ind_y)
```

4. Interesting question! I haven't thought of this before. How did you deal with this in your script?

simonrp84:
I don't think it's possible for the corrected lat/lon to be outside the area. 
By definition, the corrected position will always be closer to the subsatellite point than the original position and will hence be inside the area.

gerritholl:
> 
> 
> I don't think it's possible for the corrected lat/lon to be outside the area.
> By definition, the corrected position will always be closer to the subsatellite point than the original position and will hence be inside the area.

Is the correction applied before or after resampling?  If I have resampled to an area that does not include the equator I can see how the most equatorward pixels drop out of the area after correcting the position.

jleinonen:
As far as I can see, the correction is done after resampling so you can end up with some points outside the area. In the solution I was working on, the correction is done at the projection level so the shift simply causes the resampler to fetch points from outside the original box.

PR #1285, opened 2020-07-31
[WIP] Add MIRS file reader

katherinekolman:
This adds a reader for files from MIRS.

 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 


joleenf:
@djhoese In this commit, the file metadata is taking precedence over metadata in the yaml for the metop file used for testing.  (IMG_SX.M2.D17037.S1601.E1607.B0000001.WE.HR.ORB.nc). Therefore, standard_name is added when reading the file metadata. I thought this line in the [available_datasets](https://satpy.readthedocs.io/en/latest/api/satpy.readers.file_handlers.html#satpy.readers.file_handlers.BaseFileHandler.available_datasets) meant that it is possible to append new information to the metadata:

> This method is not guaranteed that it will be called before any other file type’s handler. The availability “boolean” not being None does not mean that a file handler called later can’t provide an additional dataset, but it must provide more identifying (DataID) information to do so and should yield its new dataset in addition to the previous one.
 
Can the boolean be True, False or None?  Or is it only either True or None?  I think the available_dataset() information above is saying, "Even if the availability "boolean" is True rather than None, a file handler could provide additional datasets." After that line, what is the more identifying DataID? Is this referring to the more definitive DataIDs like Brightness Temperature data defined by frequency and polarization "btemp_23v" versus the wider scope of BT? If so, then that is working correctly in this reader, because both the BT and the definitive brightness temperature channels are being yielded. Is there a way to keep variable metadata from a yaml and use metadata from a file, or is it either/or?  
Also, in "Migrating to Xarray" the [example](https://satpy.readthedocs.io/en/stable/dev_guide/xarray_migration.html#id1) shows using the "coord" attribute. Is assign_coords the only concern in this [issue report](https://github.com/pydata/xarray/issues/3068)?

joleenf:
@djhoese Previous comment was supposed to be under #1511

djhoese:
Replaced by #1511

PR #1310, opened 2020-08-10
Add pad_data functionality to FCI L1c FDHSI reader

ameraner:
This PR plugs the `fci_l1c_fdhsi` reader into the `GEOSegmentYAMLReader`. With this, missing chunks (aka segments) are padded with nans using the `scn.load([ds_name], pad_data=True)` mechanism.

After switching to the  `GEOSegmentYAMLReader`, the FCI tests were failing due to issues with the "hackyness" in the `TestGEOSegmentYAMLReader`. This was solved using the same unittest patch methods implemented in #1194.  

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/readers/yaml_reader.py
+++ b/satpy/readers/yaml_reader.py
@@ -259,7 +259,7 @@ class AbstractYAMLReader(metaclass=ABCMeta):
                 if val_type is not None and issubclass(val_type, tuple):
                     # special case: wavelength can be [min, nominal, max]
                     # but is still considered 1 option
-                    id_kwargs.append((val, ))
+                    id_kwargs.append((val,))
                 elif isinstance(val, (list, tuple, set)):
                     # this key has multiple choices
                     # (ex. 250 meter, 500 meter, 1000 meter resolutions)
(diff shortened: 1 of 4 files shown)


satpy/readers/yaml_reader.py
```diff
@@ -1235,3 +1260,47 @@ def _find_missing_segments(file_handlers, ds_info, dsid):
         slice_list.append(None)
 
     return counter, expected_segments, slice_list, failure, projectable
+
+
+def _get_new_area_extent_heights(previous_area, previous_seg_size, segment_n, padding_fci_scene):
+    """Get the area extent heights in meters (projection coordinates) and pixels."""
```
pnuu, line 1266: Technically the extents can also be scan angle (like ABI), feet, or something else, so maybe just say `in projection coordinates` instead of specifying the unit which isn't enforced in the code in any case.

ameraner: Right, thanks! I'll also rename the function to be `_get_new_areadef_heights`, since the pixel height is not part of the area extent.

ameraner:
Thanks a lot for the reviews and the merge!

PR #1366, opened 2020-09-15
Fixing a few typos in slstr_l1b yaml reader

ninahakansson:
Fixing a few typos in slstr_l1b yaml reader. Not sure if wavelength for S9 is correct.
I used the width and center from https://sentinel.esa.int/web/sentinel/user-guides/sentinel-3-slstr/resolutions/radiometric.


mraspaud (approved):
LGTM

PR #1370, opened 2020-09-17
Add support for reading AAPP level-1c MHS/AMSU-B data

adybbroe:
Adding support for reading the MHS/AMSU-B AAPP level-1c formatted data

Some more generalized support for handling MW type of data have been added. The use case has been the data from the Passive Micorwave sensing instrument MHS on board the Metop and NOAA-POES satellites.

For MHS we use two kind of frequency channel representations. One is the `FrequencyRange`, and the other is the `FrequencyDoubleSideBand` type. The former is the more simple one, used for the window channels, and the latter is more complex and used for humidity type sounding bands. See figure below.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented 

Below a schematic sketch of the MHS type of humidity sounding bands. There are two "double-side-bands" and a "normal" `FrequencyRange` type band to the right of the WV absorption maximum.

double_side_band_mhs

Here a small example use case loading one MHS band, specifying the approximate band frequency:
```
In [1]: from satpy.dataset import DataQuery
In [2]: from satpy import Scene
In [4]: scn = Scene(filenames=[FILENAME, ], reader='mhs_l1c_aapp')
In [5]: frq_156 = DataQuery(frequency_range=156)
In [6]: scn.load([frq_156])
In [7]: print(scn[frq_156].attrs['frequency_range'])
157.0 GHz (2.8 GHz)
```

And here is an RGB with the three 183 GHz bands:
```
from satpy import Scene

scn = Scene(filenames=[FILENAME, ], reader='mhs_l1c_aapp')
scn.load(["mw183_humidity"])
local = scn.resample('euro4')
local.show("mw183_humidity")
```

mhs_rgb_color1_20200917_1006_euro4


adybbroe:
@mraspaud and @djhoese and others. I might need a little help or guidance here.
I would like to be able to load the MHS channel one like this (by specifying the frequency):
```
from satpy import Scene
debug_on()
FILENAME = "/path/to/my/data/mhsl1c_metop01_20200917_1006_41509.l1c"
scn = Scene(filenames=[FILENAME, ], reader='mhs_l1c_aapp')
scn.load([89.])
```

But right now, with my small additions of the yaml reader config and the `FrequencyRange` class it seems to expext a wavelength:
```
KeyError: 'Unknown datasets: {DataQuery(wavelength=89.0)}'
```

What am I missing?

mraspaud:
Hmm, that's because at the moment requesting a number results in a wavelength query (that's hardcoded). It's rests from before, when we had fixed DatasetIDs.
A solution here would be to choose which parameter a number-query maps to dynamically, based on the reader's id-keys. But it might be a bit tricky.

adybbroe:
Ok, thanks @mraspaud that was also my thought, "choose which parameter a number-query maps to dynamically, based on the reader's id-keys", but I would have (at least yet) no clue how to do that.
So, for this PR, your suggestion is leaving that out?

mraspaud:
yes, that doesn't belong in this PR. For now, just be explicit in the queries.

adybbroe:
By the way @mraspaud I took and merged your Frequency Range PR into this one. Perhaps not the most clean solution? Or actually you could merge yours first and then we take this one?

mraspaud:
Doesn't really matter, as long as it works :)

djhoese:
pre-commit.ci autofix

satpy/dataset/dataid.py
```diff
@@ -72,6 +72,246 @@ def __repr__(self):
...
+            a tuple equal to self, False otherwise.
+
+        """
+        if other is None:
+            return False
+        elif isinstance(other, numbers.Number):
+            return other in self
+        elif isinstance(other, (tuple, list)) and len(other) == 2:
+            return self[:2] == other
+        return super().__eq__(other)
```
djhoese, line 255: Not sure why your local line numbers were messed up, but here's the unnecessary set of `elif`s. Each one can be a regular `if`.

adybbroe: Great, I will have a look and fix!

adybbroe:
I see now that Codebeat have some good points on reusing existing code. Should I refactor and make a `AAPPL1BaseFileHandler` perhaps?
@mraspaud @pnuu @djhoese ?

djhoese:
It is up to you, but if there is really that much reused code then a base class would be nice.

adybbroe:
Ok @djhoese I was about to refactor, but found out that I hadn't added tests for the MHS reader. I guess I need to go back and fix that first at least!? :-(
I was tempted leaving it for another PR to refactor. But I think tests should be added first at least.
I have not done much testing for readers before. But I will look at the aapp_l1b reader tests and see what I can do...

mraspaud:
Yes, tests first, then refactor.

satpy/readers/aapp_mhs_l1c.py
```diff
@@ -0,0 +1,222 @@
...
+
+class MHSAAPPL1CFile(AAPPL1BaseFileHandler):
+    """Reader for AMSU-B/MHS L1C files created from the AAPP software."""
+
+    def __init__(self, filename, filename_info, filetype_info):
+        """Initialize object information by reading the input file."""
+        super(MHSAAPPL1CFile, self).__init__(filename, filename_info,
+                                             filetype_info)
+
+        self.channels = {i: None for i in MHS_CHANNEL_NAMES_SET}
```
mraspaud, line 69: why does this need to be a set?

satpy/readers/aapp_mhs_l1c.py
```diff
@@ -0,0 +1,222 @@
...
+                  18: 'NOAA-18',
+                  19: 'NOAA-19',
+                  1: 'Metop-B',
+                  2: 'Metop-A',
+                  3: 'Metop-C',
+                  4: 'Metop simulator'}
+
+MHS_PLATFORMS = ['Metop-A', 'Metop-B', 'Metop-C', 'NOAA-18', 'NOAA-19']
+
+MHS_ANGLES = ['sensor_zenith_angle', 'sensor_azimuth_angle',
```
mraspaud, line 57: Should this be called `MHS_ANGLE_NAMES` for consistency?

satpy/readers/aapp_mhs_l1c.py
```diff
@@ -0,0 +1,222 @@
...
+              'solar_zenith_angle', 'solar_azimuth_difference_angle']
+
+
+class MHSAAPPL1CFile(AAPPL1BaseFileHandler):
```
mraspaud, line 61: I think both this class and the file names should be change to reflect the fact that amsu can also be read here.

satpy/dataset/dataid.py
```diff
@@ -72,6 +72,246 @@ def __repr__(self):
         return '<' + str(self) + '>'
 
 
+class FrequencyDoubleSideBandBase(NamedTuple):
```
mraspaud, line 75: This class and the next three are really specific to MHS/AMSU, should we put them in the mhs reader for now instead?

satpy/readers/aapp_mhs_l1c.py
```diff
@@ -0,0 +1,222 @@
...
+    """Calibrate channel data.
+
+    *calib_type* in brightness_temperature.
+
+    """
+    if calib_type not in ['brightness_temperature']:
+        raise ValueError('Calibration ' + calib_type + ' unknown!')
+
+    channel = da.from_array(data["btemps"][:, :, chn] / 100., chunks=(LINE_CHUNK, 90))
+    mask &= channel != 0  # What does this do? FIXME!
```
mraspaud, line 157: This updates the mask array to mark 0 values as invalid

mraspaud (changes requested):
Don't forget to add a line to the reader table in the documentation! Nice job, just some comments.

satpy/dataset/dataid.py
```diff
@@ -221,47 +461,47 @@ def __hash__(self):
...
+        'radiance',
+        'counts'
+    ],
+    'transitive': True,
+},
+    'modifiers': {
+    'default': ModifierTuple(),
+    'type': ModifierTuple,
+},
+}
```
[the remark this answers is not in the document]
adybbroe: That's really strange, this was not me! It must be an auto-correct thingy that has gone hay-wire?

satpy/etc/composites/mhs.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - name: '5'
+    standard_name: mw183_humidity
+
+  mw183_humidity_surface:
+    compositor: !!python/name:satpy.composites.RGBCompositor
+    prerequisites:
+    - name: '1'
+    - name: '2'
+    - name: '3'
+    standard_name: mw183_humidity_surface
```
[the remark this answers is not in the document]
adybbroe: No. I think we are on uncharted territory here? I just made these two simple ones and included them here. It might require some more contemplation what is meaningful to do here...

satpy/readers/aapp_l1b.py
```diff
@@ -63,55 +65,29 @@ def create_xarray(arr):
...
-                         (11, '3a'),
-                         (10, '3b'),
-                         (9, '4'),
-                         (8, '5'))
-        activated = dict()
-        for bit, channel_name in bits_channels:
-            activated[channel_name] = bool(status >> bit & 1)
-        return activated
+        self._scan_type = np.dtype([("siteid", "<i2")])
+        self._header_type = np.dtype([("siteid", "<i2")])
```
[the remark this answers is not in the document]
adybbroe: Was unsure what to do for a dummy/default. This base class is not used on its own, so these are supposed to be overwritten

satpy/readers/aapp_l1b.py
```diff
@@ -41,6 +41,8 @@
 logger = logging.getLogger(__name__)
 
 CHANNEL_NAMES = ['1', '2', '3a', '3b', '4', '5']
+AVHRR_CHANNEL_NAMES = ("1", "2", "3a", "3b", "4", "5")
```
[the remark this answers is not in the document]
adybbroe: Very good question! :-)
I shall fix...

satpy/readers/aapp_mhs_l1c.py
adybbroe: Yes, I will fix...

satpy/readers/aapp_mhs_l1c.py
adybbroe: I does not. Fixed.

satpy/dataset/dataid.py
adybbroe: Hmmm, you have a point. But I was hoping to add MWS support soon, and perhaps also ATMS. And AWS (same file format as MWS).

adybbroe:
All right, @mraspaud and @djhoese and others interested. I am satisfied, are you?
Maybe only two (small) outstanding issues left:

- the codebeat issue - which I think is on the edge
- and if I should keep the new Frequency classes to the MHS reader, untill we have more readers using these types?

satpy/readers/aapp_mhs_l1c.py
adybbroe: Done

satpy/readers/aapp_mhs_l1c.py
adybbroe: Done

mraspaud (approved):
LGTM

adybbroe:
mhs_amsub_satpy_support

I added MHS as `Nominal` as I think most of what you would expect is there and it is tested on real MHS data. AMSU-B is here `Beta`, as the yaml file for AMSU-B was added, but I have not tested it on actual data.

Maybe this is not the right way to classify?

The file formats of MHS and AMSU-B are the same, it is just the yaml file that needs to be slightly different, as the cahnnels are not exactly the same.

adybbroe:
I think #1912 should be closed/removed/rejected when this on is merged. Will you @mraspaud ?

PR #1376, opened 2020-09-21
Switch to travis for Windows testing

djhoese:
The ci-helpers project is deprecating appveyor soon. This PR removes appveyor and instead uses travis CI for our Windows environment. Includes enabling mamba package manager in ci-helpers.

Replaces #1361 which tried to switch to conda-lock files but I noticed even worse CI performance. Theoretically it will be faster in the future, but given the added maintenance cost and worse performance for using conda-lock files I don't think we'll switch to them right now.


djhoese:
The failures here are related to https://github.com/astropy/ci-helpers/issues/460.

This was just working not that long ago so this is something in ci-helpers that isn't set up properly to use mamba when dependencies can't be met.

mraspaud (approved):
LGTM, feel free to merge when the tests pass.

djhoese:
Waiting for this https://github.com/conda-forge/pylibtiff-feedstock/pull/3 which should let me switch back to depending on pylibtiff from conda-forge and workaround some of the ci-helpers mamba issues.

PR #1382, opened 2020-09-24
Fixed bug getting name to the calibration in mitiff writer

TAlonglong:
Fixed bug getting name from the prerequisites when the prereq consist of DataQuery objects

 - [X] Tests added 
 - [x] Passes ``flake8 satpy``


mraspaud (approved):
LGTM, nice with the real testing.

PR #1390, opened 2020-10-05
Add support to Pyspectral NIRReflectance masking limit

pnuu:
This is adds support to `masking_limit` keyword argument for `NIRReflectance` modifier added in https://github.com/pytroll/pyspectral/pull/113

 - [x] Tests added 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented 

I'm not sure where to document this feature, but the usage shown below.

This disables the masking altogether. See https://github.com/pytroll/pyspectral/pull/113 for an example image. This is the old default behaviour, which was changed in Pyspectral May (https://github.com/pytroll/pyspectral/commit/e9a234f317afd56121bb4539b40e225ae53a073c#diff-2a5232e6ece84bcd6cc7c97d82a7a574).
```yaml
  nir_reflectance_no_masking:
    compositor: !!python/name:satpy.composites.NIRReflectance
    prerequisites:
      - name: IR_108
    optional_prerequisites:
    - solar_zenith_angle
    - IR_134
    masking_limit: null
```

And definition of a specific masking limit:

```yaml
  nir_reflectance_no_masking:
    compositor: !!python/name:satpy.composites.NIRReflectance
    prerequisites:
      - name: IR_108
    optional_prerequisites:
    - solar_zenith_angle
    - IR_134
    masking_limit: 88.0
```

The default is to mask with the default limit defined in `pyspectral.near_infrared_reflectance.TERMINATOR_LIMIT`, which is also the current behaviour.

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -60,8 +60,6 @@ NEGLIBLE_COORDS = ['time']
 MASKING_COMPOSITOR_METHODS = ['less', 'less_equal', 'equal', 'greater_equal',
                               'greater', 'not_equal', 'isnan', 'isfinite',
                               'isneginf', 'isposinf']
-NIR_REFLECTANCE_TERMINATOR_LIMIT = 85.0
-NIR_REFLECTANCE_MASKING_LIMIT = 88.0
 
 
 class IncompatibleAreas(Exception):
@@ -622,8 +620,11 @@ class PSPRayleighReflectance(CompositeBase):
 class NIRReflectance(CompositeBase):
     """Get the reflective part of NIR bands."""
 
-    def __init__(self, sunz_threshold=NIR_REFLECTANCE_TERMINATOR_LIMIT,
-                 masking_limit=NIR_REFLECTANCE_MASKING_LIMIT, **kwargs):
+    TERMINATOR_LIMIT = 85.0
+    MASKING_LIMIT = 88.0
+
+    def __init__(self, sunz_threshold=TERMINATOR_LIMIT,
+                 masking_limit=MASKING_LIMIT, **kwargs):
         """Collect custom configuration values.
 
         Args:
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -42,10 +42,10 @@ from satpy.utils import sunzen_corr_cos, atmospheric_path_length_correction, get
 from satpy.writers import get_enhanced_image
 
 try:
-    from pyspectral.near_infrared_reflectance import Calculator, TERMINATOR_LIMIT
+    from pyspectral.near_infrared_reflectance import Calculator
 except ImportError:
     Calculator = None
-    TERMINATOR_LIMIT = 85.0
+
 try:
     from pyorbital.astronomy import sun_zenith_angle
 except ImportError:
@@ -60,6 +60,8 @@ NEGLIBLE_COORDS = ['time']
 MASKING_COMPOSITOR_METHODS = ['less', 'less_equal', 'equal', 'greater_equal',
                               'greater', 'not_equal', 'isnan', 'isfinite',
                               'isneginf', 'isposinf']
+NIR_REFLECTANCE_TERMINATOR_LIMIT = 85.0
+NIR_REFLECTANCE_MASKING_LIMIT = 88.0
 
 
 class IncompatibleAreas(Exception):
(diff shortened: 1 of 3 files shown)


pnuu:
Before I forget. [This](https://github.com/pytroll/satpy/blob/master/satpy/tests/test_composites.py#L656-L666) test tested that the *default* keyword argument should not be `None`, but at the same time explicitly passed `None` as the value. I changed this to do what the method name said.

mraspaud:
@pnuu this is the reason for that test: https://github.com/pytroll/satpy/commit/53620fdaadee933e2319e37f4a0360aa64bfe9bb#diff-c3f713f166b7940f26ca41873bcd1890R650-R655
So the objective was to check that the default value from the *calculator* was use if None was passed.

satpy/composites/__init__.py
```diff
@@ -700,12 +703,9 @@ def _init_reflectance_calculator(self, metadata):
...
 
-        if self.sun_zenith_threshold is not None:
-            reflectance_3x_calculator = Calculator(metadata['platform_name'], metadata['sensor'], metadata['name'],
-                                                   sunz_threshold=self.sun_zenith_threshold)
-        else:
-            reflectance_3x_calculator = Calculator(metadata['platform_name'], metadata['sensor'], metadata['name'])
-            self.sun_zenith_threshold = reflectance_3x_calculator.sunz_threshold
+        reflectance_3x_calculator = Calculator(metadata['platform_name'], metadata['sensor'], metadata['name'],
+                                               sunz_threshold=self.sun_zenith_threshold,
+                                               masking_limit=self.masking_limit)
```
mraspaud, line 708: I'm not sure this is right. When `self.sun_zenith_threshold` is None, we want to replace it with the default value from the calculator.
This was what #1319 was about.

pnuu: Oh. That sounds wrong. I'd expect `None` to mean "do not apply SZA correction in the Calculator". I'd just fix the documentation to say `If not explicitly specified, the default value defined in Pyspectral will be used.`

But anyway, the behaviour on the Satpy side doesn't change: the default value is used if the user doesn't override it. If the user decides to override the default with `None`, they'll probably know what they are doing (which is, crash the code :grin:).

mraspaud: ok, just change the `__init__` docstring then :)

pnuu: Docstring updated. Good thing, as I didn't remember to add the new kwarg there before.

pnuu:
> @pnuu this is the reason for that test: [53620fd#diff-c3f713f166b7940f26ca41873bcd1890R650-R655](https://github.com/pytroll/satpy/commit/53620fdaadee933e2319e37f4a0360aa64bfe9bb#diff-c3f713f166b7940f26ca41873bcd1890R650-R655)
> So the objective was to check that the default value from the _calculator_ was use if None was passed.

... which was because Satpy didn't respect the default and overrode it with `None`. This PR also fixes that by importing the defaults from Pyspectral.

mraspaud:
Last question (maybe :) ): Should we use the pyspectral default or prevent potential problems if pyspectral is modified and hardcode a satpy-specific set of limits?

pnuu:
I thought about that. Now that you raise the point again, I think that would make more sense than importing from Pyspectral. It would also make testing more explicit. I'll add Satpy-internal thresholds for these two arguments. Any suggestions what the default for `masking_limit` should be? Based on my experience with Day/Night composites I'd say `masking_limit=None` (no masking) is the best choice, but for "standalone" composites the "strict" limit might be better :thinking:

mraspaud:
ok, what if the masking limit is beyond the sunz_threshold, would that have the same effect as `masking_limit` set to None for Day/Night composites?

pnuu:
It behaves the same as long as there's overlap between the two composites. It seems the default blending in `DayNightCompositor` is from `85.0` to `88.0` degrees SZA. So beyond `88.0` the `NIRReflectance` shouldn't matter at all. I'll try that limit and use it if seems to work as expected.

pnuu:
Yep, The results are identical. Below are images with the `88.0` degree limit, first the single composite and then with `night_microphysical` on the night side. I'll update the tests and push the updates in a bit.

day_microphysical_20200320_180000

day_microphysical_with_night_microphysical_20200320_180000_88

satpy/composites/__init__.py
```diff
@@ -59,6 +60,8 @@
 MASKING_COMPOSITOR_METHODS = ['less', 'less_equal', 'equal', 'greater_equal',
                               'greater', 'not_equal', 'isnan', 'isfinite',
                               'isneginf', 'isposinf']
+NIR_REFLECTANCE_TERMINATOR_LIMIT = 85.0
+NIR_REFLECTANCE_MASKING_LIMIT = 88.0
```
mraspaud, line 64: Just to keep things gathered in the same place, would it be worth have these at *class* attributes to `NIRReflectance`?

mraspaud (approved):
LGTM, just one stylistic question

satpy/composites/__init__.py
pnuu: Will do. Update follows in a minute or three.

pnuu: Code and tests updated, tested with real data.

pnuu:
Just as a reminder: the Pyspectral PR https://github.com/pytroll/pyspectral/pull/113 needs to be merged (and released) first for this to be usable.

mraspaud:
Probably one more thing to do here is to set the minimal version of pyspectral in setup.py

pnuu:
Agreed. That'll need the new Pyspectral release to see the actual version number.

pnuu:
Pyspectral version requirement updated to the release made yesterday.

pnuu:
This PR isn't blocked anymore by the Pyspectral PR which has been merged and released.

PR #1411, opened 2020-10-21
Added MERSI-2 file pattern for data from NMSC

ninahakansson:
Added MERSI-2 file pattern for data from NMSC

 - [x] Passes ``flake8 satpy``


mraspaud (changes requested):
Thanks for adding this pattern. Could you place it before the generic one? It is meant as a catchall, even if in this case the pattern wouldn't catch the nmsc pattern.

mraspaud (approved):
LGTM

PR #1416, opened 2020-10-28
Update GRIB reader for greater flexibility.

simonrp84:
Currently, the grib reader expects certain keys to be present in the grib file, [for example](https://github.com/pytroll/satpy/blob/master/satpy/readers/grib.py#L236-L254):
```
ds_info.update({
            'filename': self.filename,
            'shortName': msg['shortName'],
            'long_name': msg['name'],
            'pressureUnits': msg['pressureUnits'],
            'typeOfLevel': msg['typeOfLevel'],
            'standard_name': msg['cfName'],
            'units': msg['units'],
            'modelName': msg['modelName'],
            'model_time': model_time,
            'centreDescription': center_description,
            'valid_min': msg['minimum'],
            'valid_max': msg['maximum'],
            'start_time': start_time,
            'end_time': end_time,
            'sensor': msg['modelName'],
            # National Weather Prediction
            'platform_name': 'unknown',
        })
```

For some grib files, such as those from the ERA5 and CAMS datasets, not all keys are present. `modelName` is not present in ERA5 single level datasets.

This PR adds some flexibility to the grib reader and will now set missing keys to `unknown` in `ds_info` rather than raising an error.
I have only tested this with ECMWF/ERA/GFS data but it seems to work well with those.

 - [ ] Tests passed
 - [ ] Passes ``flake8 satpy``

### DIFF
--- a/satpy/modifiers/atmosphere.py
+++ b/satpy/modifiers/atmosphere.py
@@ -22,6 +22,7 @@ from weakref import WeakValueDictionary
 
 import numpy as np
 import dask.array as da
+import xarray as xr
 
 from satpy.modifiers import ModifierBase
 from satpy.utils import get_satpos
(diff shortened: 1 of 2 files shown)


simonrp84:
Deepcode is complaining about line 258, this is a false positive and `.keys()` is required here.

PR #1420, opened 2020-10-31
Add support for Near-realtime VIIRS L1b data.

simonrp84:
Currently, the `viirs_l1b` reader only supports archived VIIRS data from NASA. This PR adds support for the near realtime data, which has a slightly different filename structure.


djhoese (approved):
I'm fine with this, but wondering if it could be more generic? Any idea if anyone uses something besides `_NRT`? Maybe a wildcard `{processing_suffix}` would be good? I know @mraspaud didn't like the idea of this for the ABI reader in a recent PR I made or at least not as an all-in-one file pattern.

Eh on second though maybe it is better to be explicit.

djhoese:
Where does the NRT data come from? Is there a URL we could/should add to a docstring somewhere?

simonrp84:
The NRT data comes from the same source as the 'normal' data that's already supported by this reader, they're both NASA products and are available from the same places (earthdata, f.ex).

PR #1422, opened 2020-11-04
Fix HDF5 utility file handler not decoding byte arrays consistently

djhoese:
A couple users noticed that some readers are now causing errors when multiple granule's metadata dictionaries are combined. You'll see errors like:

```
  File "/home/cpeters/anaconda2/envs/satpy/lib/python3.7/site-packages/numpy/core/numeric.py", line 2287, in isclose
    xfin = isfinite(x)
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

I've tracked this down to at least one specific issue: some readers are putting byte arrays (`ex. b''"'`) in numpy arrays in their `DataArray.attrs` metadata. When attempting to combine metadata satpy sees these as arrays and attempts to combine them. This fails with the above error.

Both readers that have encountered this are using the base HDF5 utility file handler. I discovered that although the file handler is converting attributes to regular strings when possible, it is only doing this for attributes accessed through the string form `my_handler['/my_data/attr/some_attr']` but *NOT* for the DataArray that is produced so doing `my_handler['my_data'].attrs['some_attr]` will have the unconverted data.

Additionally this fixes the unstable travis environment failures in the HY2 reader.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Passes ``flake8 satpy``

### DIFF
--- a/satpy/readers/mersi2_l1b.py
+++ b/satpy/readers/mersi2_l1b.py
@@ -62,8 +62,8 @@ class MERSI2L1B(HDF5FileHandler):
         return sensor
 
     def _get_single_slope_intercept(self, slope, intercept, cal_index):
-        # sometimes slope has multiple elements
         try:
+            # convert scalar arrays to scalar
             return slope.item(), intercept.item()
         except ValueError:
             # numpy array but has more than one element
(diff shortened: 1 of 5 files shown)


satpy/readers/mersi2_l1b.py
```diff
@@ -61,18 +61,28 @@ def sensor_name(self):
...
-            # sometimes slope has multiple elements
-            if hasattr(slope, '__len__') and len(slope) == 1:
-                slope = slope[0]
-                intercept = intercept[0]
-            elif hasattr(slope, '__len__'):
-                slope = slope[cal_index]
-                intercept = intercept[cal_index]
+            slope, intercept = self._get_single_slope_intercept(
+                slope, intercept, cal_index)
+            # if hasattr(slope, '__len__') and len(slope) == 1:
```
TAlonglong, line 80: Maybe drop these comments

TAlonglong (approved):
LGTM

satpy/readers/mersi2_l1b.py
djhoese: :man_facepalming: yep that was left over. Thanks.

mraspaud (approved):
Thanks for taking care of this, LGTM

PR #1423, opened 2020-11-05
read goes-r data from https aws, gcp or azure (it uses netcdf4)

jhbravo:
Hello, @djhoese 
As I wrote in the slack, I just uploaded the methods to read data from https, it may not be the best method but it has worked for me
this uses netcdf4, and is adapted of this ipynb:
https://github.com/HamedAlemo/visualize-goes16/blob/master/visualize_GOES16_from_AWS.ipynb

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/abi_base.py
+++ b/satpy/readers/abi_base.py
@@ -22,6 +22,8 @@ from datetime import datetime
 
 import numpy as np
 import xarray as xr
+import requests
+import netCDF4
 
 from pyresample import geometry
 from satpy.readers.file_handlers import BaseFileHandler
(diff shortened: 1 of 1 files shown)


djhoese:
Thanks for making these PRs. Nice to see some alternative solutions. We've done some experimenting with other tools. Some of this work isn't on github yet but we've had some discussions in #1169 and #1321 and #1349. We're hoping to come up with a solution that works for most readers (at least the ones that use xarray directly) without having to modify individual readers/file handlers. We will definitely use what you've done as inspiration. If we can't come up with a generic solution then we will likely merge what you've done here. For now, if it is OK with you, I'd like to wait and see what generic solution we can come up with. Thanks again.

PR #1435, opened 2020-11-13
Fix tests for GEOFlippableFileYAMLReader after AreaDefinition.area_extent being immutable

ameraner:
This PR fixes the `TestGEOFlippableFileYAMLReader` tests that were broken by https://github.com/pytroll/pyresample/pull/290 . The old tests were changing `area_def.area_extent` in-place, which is not possible anymore. 
```bash
>       dummy_ds_xr.attrs['area'].defs[0].area_extent = (1500, -1000, -1500, 1000)
E       AttributeError: can't set attribute
```

The new tests now use `area_def.copy(area_extent=new_area_extents)`.

 - [x] Tests added 
 - [x] Passes ``flake8 satpy``

### DIFF
--- a/satpy/tests/test_yaml_reader.py
+++ b/satpy/tests/test_yaml_reader.py
@@ -626,7 +626,7 @@ class TestGEOFlippableFileYAMLReader(unittest.TestCase):
 
         # check wrong input
         with self.assertRaises(ValueError):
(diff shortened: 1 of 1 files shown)


djhoese (approved):
LGTM. Thanks for the quick fix.

PR #1438, opened 2020-11-16
Full disk padding feature for SEVIRI Native data

### LINKED ISSUE #1440: Error reading SEVIRI native file from EUMETSAT API
I downloaded a seviri file from eumetsat api in nat format  but when I tried to save the HRV channel   in netcdf format I got a satpy error :

ValueError: Chunks do not add up to shape. Got chunks=((3711, 3711, 3714),), shape=(8640,)

Please do you have any idea about this error, Thank you.

strandgren:
This PR adds the possibility to load any native SEVIRI data as a full-disk array, padded with no-data where necessary, by providing the `fill_disk` as True in the `reader_kwargs` . This is especially useful for the HRV channel in FES (full Earth Scanning) mode, as it contains data from two different windows, but can also be used for VISIR and HRV RSS (Rapid Scanning Service) and ROI (Region of Interest) data. By default the original, unpadded, data are loaded (fill_disk=False).

The PR also includes the harmonization of the area id.

Wrt. the documentation I updated the status of the reader to "Nominal", but as far as I could see there is no specific documentation to update for the native reader.

 - [x] Closes #1440 
 - [x] Tests added 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented 

### DIFF
--- a/satpy/readers/seviri_base.py
+++ b/satpy/readers/seviri_base.py
@@ -432,6 +432,15 @@ def calculate_area_extent(area_dict):
     return tuple(aex)
 
 
+def get_service_mode(ssp_lon):
+    """Get information about whether we're dealing with data from the FES, RSS or IODC service mode."""
+    seviri_service_modes = {'0.0': {'name': 'FES', 'desc': 'Full Earth scanning service'},
+                            '9.5': {'name': 'RSS', 'desc': 'Rapid scanning service'},
+                            '41.5': {'name': 'IODC', 'desc': 'Indian Ocean data coverage service'}}
+
+    return seviri_service_modes['%.1f' % ssp_lon]
+
+
 def get_padding_area(shape, dtype):
     """Create a padding area filled with no data."""
     if np.issubdtype(dtype, np.floating):
--- a/satpy/readers/seviri_base.py
+++ b/satpy/readers/seviri_base.py
@@ -432,31 +432,37 @@ def calculate_area_extent(area_dict):
     return tuple(aex)
 
 
-def pad_data_ew(data, final_size, east_bound, west_bound):
+def get_padding_area(shape, dtype):
+    """Create a padding area filled with no data."""
+    if np.issubdtype(dtype, np.floating):
+        init_value = np.nan
+    else:
+        init_value = 0
+
+    padding_area = da.full(shape, init_value, dtype=dtype, chunks=CHUNK_SIZE)
+
+    return padding_area
+
+
+def pad_data_horizontally(data, final_size, east_bound, west_bound):
     """Pad the data given east and west bounds and the desired size."""
     nlines = final_size[0]
     if west_bound - east_bound != data.shape[1] - 1:
         raise IndexError('East and west bounds do not match data shape')
-    padding_east = da.zeros((nlines, east_bound - 1),
-                            dtype=data.dtype, chunks=CHUNK_SIZE)
-    padding_west = da.zeros((nlines, (final_size[1] - west_bound)),
-                            dtype=data.dtype, chunks=CHUNK_SIZE)
-    if np.issubdtype(data.dtype, np.floating):
-        padding_east = padding_east * np.nan
-        padding_west = padding_west * np.nan
+
+    padding_east = get_padding_area((nlines, east_bound - 1), data.dtype)
+    padding_west = get_padding_area((nlines, (final_size[1] - west_bound)), data.dtype)
+
     return np.hstack((padding_east, data, padding_west))
 
 
-def pad_data_sn(data, final_size, south_bound, north_bound):
+def pad_data_vertically(data, final_size, south_bound, north_bound):
     """Pad the data given south and north bounds and the desired size."""
     ncols = final_size[1]
     if north_bound - south_bound != data.shape[0] - 1:
         raise IndexError('South and north bounds do not match data shape')
-    padding_south = da.zeros((south_bound - 1, ncols),
-                             dtype=data.dtype, chunks=CHUNK_SIZE)
-    padding_north = da.zeros(((final_size[0] - north_bound), ncols),
-                             dtype=data.dtype, chunks=CHUNK_SIZE)
-    if np.issubdtype(data.dtype, np.floating):
-        padding_south = padding_south * np.nan
-        padding_north = padding_north * np.nan
+
+    padding_south = get_padding_area((south_bound - 1, ncols), data.dtype)
+    padding_north = get_padding_area(((final_size[0] - north_bound), ncols), data.dtype)
+
     return np.vstack((padding_south, data, padding_north))
--- a/satpy/readers/seviri_base.py
+++ b/satpy/readers/seviri_base.py
@@ -30,6 +30,7 @@ import dask.array as da
 from satpy.readers.utils import apply_earthsun_distance_correction
 from satpy.readers.eum_base import (time_cds_short,
                                     issue_revision)
+from satpy import CHUNK_SIZE
 
 PLATFORM_DICT = {
     'MET08': 'Meteosat-8',
@@ -50,6 +51,7 @@ C2 = 1.43877523
 VISIR_NUM_COLUMNS = 3712
 VISIR_NUM_LINES = 3712
 HRV_NUM_COLUMNS = 11136
+HRV_NUM_LINES = 11136
 
 CHANNEL_NAMES = {1: "VIS006",
                  2: "VIS008",
@@ -428,3 +430,33 @@ def calculate_area_extent(area_dict):
     aex = np.array([ll_c, ll_r, ur_c, ur_r]) * area_dict['resolution']
 
     return tuple(aex)
+
+
+def pad_data_ew(data, final_size, east_bound, west_bound):
+    """Pad the data given east and west bounds and the desired size."""
+    nlines = final_size[0]
+    if west_bound - east_bound != data.shape[1] - 1:
+        raise IndexError('East and west bounds do not match data shape')
+    padding_east = da.zeros((nlines, east_bound - 1),
+                            dtype=data.dtype, chunks=CHUNK_SIZE)
+    padding_west = da.zeros((nlines, (final_size[1] - west_bound)),
+                            dtype=data.dtype, chunks=CHUNK_SIZE)
+    if np.issubdtype(data.dtype, np.floating):
+        padding_east = padding_east * np.nan
+        padding_west = padding_west * np.nan
+    return np.hstack((padding_east, data, padding_west))
+
+
+def pad_data_sn(data, final_size, south_bound, north_bound):
+    """Pad the data given south and north bounds and the desired size."""
+    ncols = final_size[1]
+    if north_bound - south_bound != data.shape[0] - 1:
+        raise IndexError('South and north bounds do not match data shape')
+    padding_south = da.zeros((south_bound - 1, ncols),
+                             dtype=data.dtype, chunks=CHUNK_SIZE)
+    padding_north = da.zeros(((final_size[0] - north_bound), ncols),
+                             dtype=data.dtype, chunks=CHUNK_SIZE)
+    if np.issubdtype(data.dtype, np.floating):
+        padding_south = padding_south * np.nan
+        padding_north = padding_north * np.nan
+    return np.vstack((padding_south, data, padding_north))
--- a/satpy/readers/seviri_l1b_hrit.py
+++ b/satpy/readers/seviri_l1b_hrit.py
@@ -168,8 +168,8 @@ from satpy.readers.hrit_base import (HRITFileHandler, ancillary_text,
                                      image_data_function)
 from satpy.readers.seviri_base import (CALIB, CHANNEL_NAMES, SATNUM,
                                        VIS_CHANNELS, SEVIRICalibrationHandler,
-                                       chebyshev, get_cds_time,  HRV_NUM_COLUMNS,
-                                       pad_data_ew)
+                                       chebyshev, get_cds_time, HRV_NUM_COLUMNS,
+                                       pad_data_horizontally)
 from satpy.readers.seviri_l1b_native_hdr import (hrit_epilogue, hrit_prologue,
                                                  impf_configuration)
 from satpy.readers._geos_area import get_area_extent, get_area_definition
@@ -735,18 +735,18 @@ class HRITMSGFileHandler(HRITFileHandler, SEVIRICalibrationHandler):
         data_list = list()
         if upper_south_line > 0:
             # we have some of the lower window
-            data_lower = pad_data_ew(res[:upper_south_line, :].data,
-                                     (upper_south_line, HRV_NUM_COLUMNS),
-                                     bounds['LowerEastColumnActual'],
-                                     bounds['LowerWestColumnActual'])
+            data_lower = pad_data_horizontally(res[:upper_south_line, :].data,
+                                               (upper_south_line, HRV_NUM_COLUMNS),
+                                               bounds['LowerEastColumnActual'],
+                                               bounds['LowerWestColumnActual'])
             data_list.append(data_lower)
 
         if upper_south_line < nlines:
             # we have some of the upper window
-            data_upper = pad_data_ew(res[upper_south_line:, :].data,
-                                     (nlines - upper_south_line, HRV_NUM_COLUMNS),
-                                     bounds['UpperEastColumnActual'],
-                                     bounds['UpperWestColumnActual'])
+            data_upper = pad_data_horizontally(res[upper_south_line:, :].data,
+                                               (nlines - upper_south_line, HRV_NUM_COLUMNS),
+                                               bounds['UpperEastColumnActual'],
+                                               bounds['UpperWestColumnActual'])
             data_list.append(data_upper)
         return xr.DataArray(da.vstack(data_list), dims=('y', 'x'))
 
--- a/satpy/readers/seviri_l1b_hrit.py
+++ b/satpy/readers/seviri_l1b_hrit.py
@@ -689,9 +689,7 @@ class HRITMSGFileHandler(HRITFileHandler, SEVIRICalibrationHandler):
         res = super(HRITMSGFileHandler, self).get_dataset(key, info)
         res = self.calibrate(res, key['calibration'])
         if key['name'] == 'HRV' and self.fill_hrv:
-            attrs = res.attrs
             res = self.pad_hrv_data(res)
-            res.attrs = attrs
 
         res.attrs['units'] = info['units']
         res.attrs['wavelength'] = info['wavelength']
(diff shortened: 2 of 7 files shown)


satpy/readers/seviri_l1b_native.py
```diff
@@ -373,9 +396,18 @@ def get_area_extent(self, dataset_id):
 
             # get actual navigation parameters from trailer data
             data15tr = self.trailer['15TRAILER']
-            HRV_bounds = data15tr['ImageProductionStats']['ActualL15CoverageHRV']
-
-            # lower window
+            HRV_bounds = data15tr['ImageProductionStats']['ActualL15CoverageHRV'].copy()
+            if self.fill_hrv:
+                HRV_bounds['UpperEastColumnActual'] = 1
+                HRV_bounds['UpperWestColumnActual'] = 11136
```
sjoro, line 402: the hard coded value 11136 should probably defined and imported from `seviri_base.py`. it has `HRV_NUM_COLUMNS` defined already, the same should be defined for `HRV_NUM_ROWS`

satpy/readers/seviri_l1b_native.py
```diff
@@ -480,6 +516,46 @@ def get_dataset(self, dataset_id, dataset_info):
 
         return dataset
 
+    def pad_hrv_data(self, dataset):
+        """Pad HRV data with empty pixels."""
+        logger.debug('Padding HRV data to full disk')
+        nlines = int(self.mda['hrv_number_of_lines'])
+        nlines_pad = 11136
```
sjoro, line 523: same as above. line 402.

satpy/readers/seviri_l1b_native.py
```diff
@@ -480,6 +516,46 @@ def get_dataset(self, dataset_id, dataset_info):
 
         return dataset
 
+    def pad_hrv_data(self, dataset):
```
sjoro, line 519: the areas (row & column numbers) low lower and upper HRV windows differ depending on whether we're dealing with FES or RSS data. this method should probably be divided in two, one method for FES one for RSS so that the implementation could be achieved with stacking different arrays together (instead of item assignment). see comment below.

satpy/readers/seviri_l1b_native.py
```diff
@@ -480,6 +516,46 @@ def get_dataset(self, dataset_id, dataset_info):
...
+        """Pad HRV data with empty pixels."""
+        logger.debug('Padding HRV data to full disk')
+        nlines = int(self.mda['hrv_number_of_lines'])
+        nlines_pad = 11136
+        ncols_pad = 11136
+        HRV_bounds = self.trailer['15TRAILER']['ImageProductionStats']['ActualL15CoverageHRV']
+
+        data_pad = np.zeros((nlines_pad, ncols_pad), dtype=dataset.dtype)
```
sjoro, line 527: item assignment not supported by `dask`. maybe a solution with stacking?

sjoro (changes requested):

satpy/readers/seviri_l1b_native.py
sjoro: fix this in `seviri_hrit`-reader in the same PR, too?

satpy/readers/seviri_l1b_native.py
```diff
@@ -480,6 +516,46 @@ def get_dataset(self, dataset_id, dataset_info):
...
+        upper_south_line = HRV_bounds['UpperSouthLineActual']
+        upper_east_col = HRV_bounds['UpperEastColumnActual']
+        upper_west_col = HRV_bounds['UpperWestColumnActual']
+
+        if upper_north_line > upper_south_line:  # we have some of the upper HRV window
+            if upper_west_col - upper_east_col != dataset.data.shape[1] - 1:
+                raise IndexError('East and west bounds of upper HRV window do not match expected data shape')
+
+            data_pad[upper_south_line - 1:upper_north_line, upper_east_col - 1:upper_west_col] = \
+                dataset[upper_south_line - nlines_pad + nlines - 1: upper_north_line - nlines_pad + nlines, :].data
```
mraspaud, line 555: This is twice the same code basically. Could you extract a new method out of this and use it twice instead?

satpy/readers/seviri_l1b_native.py
```diff
@@ -279,20 +293,29 @@ def get_area_def(self, dataset_id):
...
                 area = geometry.StackedAreaDefinition(lower_area, upper_area)
                 area = area.squeeze()
+
+            elif is_rapid_scan:
+                # handle RSS HRV data separately in case of padding to full disk (ncols -> 11136)
+                [lower_area_extent, lower_nlines, lower_ncols] = self.get_area_extent(dataset_id)
+                pdict['a_desc'] = 'SEVIRI high resolution channel, lower window (rss)'
+                pdict['nlines'] = lower_nlines
+                pdict['ncols'] = lower_ncols
+                area = get_area_definition(pdict, lower_area_extent)
```
mraspaud, line 317: This is duplicated code from a few lines above, can you refactor it?

mraspaud: you could even take in the the upper window code if you make the new method a bit more generic.

mraspaud (changes requested):
Nice feature, thanks for adding it. Does this take care of the cases when we just have an extract region instead of the full disk?

satpy/readers/seviri_l1b_native.py
strandgren: This is now fixed, also for the hrit-reader. I also made sure to add the attributes that went lost in case of hrv padding in the hrit-reader.

satpy/readers/seviri_l1b_native.py
strandgren: As already discussed I now use dask arrays instead. This in combination with adding also the possibility to pad region of interest data (see Martin's comment), lead quite some differences in the implementation though. There is still only primary function to pad  all data types, but in case of standard FES/RSS data, the relevant lines and columns are read from the trailer, whereas for ROI data this information is read from the secondary header. There are, however, two padding functions in seviri_base now, one for padding in east-west (the function that was/is used in the hrit reader), and another for padding in south-north.

satpy/readers/seviri_l1b_native.py
strandgren: The code segment has now been optimized and there should no longer any duplicated code.

satpy/readers/seviri_l1b_native.py
```diff
@@ -366,58 +368,93 @@ def get_area_extent(self, dataset_id):
...
+        #   3. Region Of Interest (ROI) mode: data for one area subset defined by the user with its area extent stored
+        #      in the secondary header.
+        if dataset_id['name'] == 'HRV':
+            area_extent = list()
+            nlines = list()
+            ncolumns = list()
+            window = list()
+
+            if not self.is_roi():
+                # If we're dealing with stadnard FES or RSS HRV data we use the actual navigation parameters
```
mraspaud, line 385:
```suggestion
                # If we're dealing with standard FES or RSS HRV data we use the actual navigation parameters
```

satpy/readers/seviri_l1b_native.py
strandgren: The initial implementation did in fact not pad region of interest (ROI) data. This has now been implemented with the latest commits. Since the information about the lines and columns of the data boundaries is stored differently for the ROI data, this lead to quite a different implementation. To do this, I added a function that checks whether we're dealing with standard FES/RSS data or ROI data (self.is_roi()). In case of standard FES/RSS data, the relevant lines and columns are read from the trailer, whereas for ROI data this information is read from the secondary header. As a result of the many changes, the code segment above has changed considerable, but in general I have tried to remove all instances of duplicate code and make it as generic as possible.

I would be happy to discuss any questions or possible improvements.

mraspaud: Sounds good! I'll have another look :)

satpy/readers/seviri_l1b_native.py
```diff
@@ -366,58 +368,93 @@ def get_area_extent(self, dataset_id):
...
-            )
-
-            upper_nlines = upper_north_line - upper_south_line + 1
-            upper_ncols = upper_west_column - upper_east_column + 1
-
-            return [upper_area_extent, lower_area_extent, upper_nlines, upper_ncols, lower_nlines, lower_ncols]
-
-        # If the data was ordered in a defined ROI, the area extent is in one piece, the corner points are
-        # the same as for VISIR channels, and the HRV channel is having three times the amount of columns and rows.
+        # If we're dealing with HRV data, three different configurations of the data must be considered:
```
mraspaud, line 371: Alright, so this function is waaay too long, more that 200 lines :) Not that I count lines, but too many things are happening here. So I can see that two methods could be extracted from here, one for HRV data, and another for the other channels. The HRV one could then be subdivided further between the FES/RSS case and the the ROI case. What do you think?

satpy/readers/seviri_l1b_native.py
```diff
@@ -480,6 +521,85 @@ def get_dataset(self, dataset_id, dataset_info):
...
+        return not self.mda['is_full_disk'] and not (is_rapid_scan and is_top3segments)
+
+    def pad_hrv_data(self, dataset):
+        """Pad HRV data with empty pixels."""
+        logger.debug('Padding HRV data to full disk')
+
+        nlines = int(self.mda['hrv_number_of_lines'])
+        ncols = int(self.mda['hrv_number_of_columns'])
+
+        if not self.is_roi():
```
mraspaud, line 550: This could also be subdivided further to two method, one for ROI and one for the other cases.

satpy/readers/seviri_l1b_native.py
strandgren: Yeah, you're right! I'll fix this. Could I then create functions like 'get_area_extent_visir',  'get_area_extent_hrv' and 'get_area_extent_roi', just as an example? Maybe I could even split FES and RSS, I'll see when I get back to the coding.

satpy/readers/seviri_l1b_native.py
strandgren: Yes, good point! I'll split this in a similar manner as discussed above :)

satpy/readers/seviri_l1b_native.py
mraspaud: sounds good. I would suggest you prefix these with `_`, like `_get_area_extent_visir` which indicate private methods (as I don't think these should be part of the public API, right?)

strandgren: ok, that's probably the case. Should these methods, as well as the padding methods, still be part of the NativeMSGFileHandler class, or shall I put them separately at the end of the reader?

sjoro: i would think private methods within the file handler-class, unless there's a use for those methods in other classes (hrit)? in that case -> `seviri_base.py`

strandgren: Alright, thanks!

strandgren:
So this is the latest stand of the PR including the refactoring and cleaning of the code as well as corresponding unit tests. Attached is a document showing several plots using different types of data with and without data padding.
[test_summary_seviri_l1b_native_padding.pdf](https://github.com/pytroll/satpy/files/5642475/test_summary_seviri_l1b_native_padding.pdf)

satpy/readers/seviri_l1b_native.py
```diff
@@ -526,138 +518,134 @@ def calibrate(self, data, dataset_id):
...
+        we start by extracting the first (only) elements of the image boundary lists. We then pad the data
+        both horizontally and vertically to form a full disk array.
+        """
+        south_bound, north_bound, east_bound, west_bound = [v[0] for v in self._img_bounds.values()]
+        nlines = north_bound - south_bound + 1
 
-        # Pad data in south-north direction
-        south, north = self._boundary['south'][0], self._boundary['north'][0]
-        padded_data = pad_data_sn(padded_data, final_shape, south, north)
+        padded_data = pad_data_horizontally(dataset.data, (nlines, self._final_shape[1]), east_bound, west_bound)
```
mraspaud: If I'm reading this correctly, these 3 lines could (almost) be replaced with
```
padded_data = self._pad_fes_hrv_data(dataset)
```
In which case `_pad_fes_hrv_data` should be renamed

(yeah, sorry, I don't like duplicated code :) )

mraspaud (commented):
This PR is coming along really nicely! I have a just couple of minor comments, and codecov complains about a few place that aren't tested in the pad functions, but after that I think it's good to merge.
I see that codecov also complains about the get_dataset function, but we'll leave that for another PR.

satpy/readers/seviri_l1b_native.py
```diff
@@ -526,138 +518,134 @@ def calibrate(self, data, dataset_id):
...
+            south_bound, north_bound, east_bound, west_bound = self._get_selected_img_bounds(dataset_id)
 
-        return {'south': south, 'north': north, 'east': east, 'west': west}
+        return {'south_bound': south_bound, 'north_bound': north_bound,
+                'east_bound': east_bound, 'west_bound': west_bound}
 
-    def _get_hrv_actual_image_bounds(self):
-        """Get the actual HRV image line and column boundaries for a given HRV window."""
+    def _get_hrv_actual_img_bounds(self):
+        """Get HRV (if not ROI) image boundaries from the ActualL15CoverageHRV infroamtion stored in the trailer."""
```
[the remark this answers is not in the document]
strandgren: Fixed

satpy/readers/seviri_l1b_native.py
```diff
@@ -526,138 +518,134 @@ def calibrate(self, data, dataset_id):
...
-        south = int(sec15hd['SouthLineSelectedRectangle']['Value'])
-        east = int(sec15hd['EastColumnSelectedRectangle']['Value'])
+    def _get_hrv_img_shape(self):
+        nlines = int(self._mda['hrv_number_of_lines'])
+        ncolumns = int(self._mda['hrv_number_of_columns'])
+        return nlines, ncolumns
 
-        north = south + nlines - 1
-        west = east + ncolumns - 1
+    def _get_visir_img_shape(self, ):
```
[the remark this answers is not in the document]
strandgren: fixed

satpy/readers/seviri_l1b_native.py
strandgren: I found another way to do the p[adding more concise, but for this solution the bots now complain about possibly undefined variables. Is that acceptable? Because in practice I don't see how they would be undefined. But it's Friday evening and I might very well miss something ;)

satpy/readers/seviri_l1b_native.py
```diff
@@ -288,14 +288,16 @@ def get_area_def(self, dataset_id):
...
             pdict['p_id'] = 'seviri_hrv'
         else:
-            pdict['a_name'] = 'geos_seviri_visir'
-            pdict['a_desc'] = 'SEVIRI low resolution channel area'
+            res = 3.0
             pdict['p_id'] = 'seviri_visir'
 
+        service_mode = get_service_mode(pdict['ssp_lon'])
+        pdict['a_name'] = 'msg_seviri_%s_%.0fkm' % (service_mode['name'], res)
+        pdict['a_desc'] = 'SEVIRI %s area definition with %.0f km resolution' % (service_mode['desc'], res)
```
ameraner, line 299: Thanks for contributing to https://github.com/pytroll/satpy/issues/1248 ! We could have the formatting of the name and description standardized in a separate function in `seviri_base`, to be accessible by all readers. But that's for another PR :)

sfinkens (commented):
Good job! Nice approach with separate bounds and padder classes. And thanks for the comprehensive tests. I only have a couple of questions and comments.

satpy/readers/seviri_l1b_hrit.py
```diff
@@ -689,7 +689,9 @@ def get_dataset(self, key, info):
         res = super(HRITMSGFileHandler, self).get_dataset(key, info)
         res = self.calibrate(res, key['calibration'])
         if key['name'] == 'HRV' and self.fill_hrv:
+            attrs = res.attrs
             res = self.pad_hrv_data(res)
+            res.attrs = attrs
```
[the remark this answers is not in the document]
strandgren: That looks like a neat solution, but I did not get it working. I think the reason is that the returned xr-array from `pad_hrv_data` is not a modified version of `res`, but a newly defined xr-array defined from a set of stacked dask-arrays that use information from the input array `res`. Or what do you think, should the attributes still be passed on?

satpy/readers/seviri_l1b_native.py
```diff
@@ -363,123 +369,107 @@ def get_area_extent(self, dataset_id):
...
+            'projection_latitude': 0.,
+            'projection_altitude': self.mda['projection_parameters']['h']}
+
+        if self.fill_disk and not (dataset_id['name'] != 'HRV' and self.mda['is_full_disk']):
+            attrs = dataset.attrs
+            padder = Padder(dataset_id,
+                            self.image_boundaries.get_img_bounds(dataset_id, self.is_roi()),
+                            self.mda['is_full_disk'])
+            dataset = padder.pad_data(dataset)
+            dataset.attrs = attrs
```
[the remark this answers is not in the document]
strandgren: See comment above

satpy/readers/seviri_base.py
```diff
@@ -428,3 +430,48 @@ def calculate_area_extent(area_dict):
     aex = np.array([ll_c, ll_r, ur_c, ur_r]) * area_dict['resolution']
 
     return tuple(aex)
+
+
+def get_service_mode(ssp_lon):
```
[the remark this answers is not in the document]
strandgren: That's a good question! I though that if e.g. MSG-3 takes over the operational 0.0E service (but from 9.5E), the data would still be provided at the nominal 0.0E grid, but this is just some vague memory I have. @sjoro or @ameraner might now?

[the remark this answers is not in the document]
sjoro: yes, if MSG-3 takes over FES, the satellite will be located on 9.5E, but the images will be rectified to 0.0E.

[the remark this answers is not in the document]
sfinkens: Oh right, this is actually the projection longitude here. Then it's fine, thanks for the clarification!

satpy/readers/seviri_l1b_hrit.py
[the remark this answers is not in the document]
sfinkens: Ok I see. Then what about moving that functionality to the padding methods? For example

```
def pad(dataset):
    ...
    return xr.DataArray(padded_data, dims=('y', 'x'), attrs=dataset.attrs)
```
Then the level of abstraction in the calling method would be equalized a bit. But that's details really :)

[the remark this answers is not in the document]
strandgren: That's a much better solution, I just pushed a new version fixing this. Thanks!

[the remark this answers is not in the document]
sfinkens: Very good, you're welcome!

sfinkens (approved):

satpy/readers/seviri_l1b_native.py
mraspaud: Care to make an issue for this?

mraspaud (approved):
LGTM, thanks for taking the time to clean up the code so well!

PR #1448, opened 2020-11-24
DOC: add explanation to the way x and y work in aggregate

raybellwaves:
@mraspaud , pointed me to this in the slack chat. Improved the wording of the example.

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -656,8 +656,8 @@ class Scene:
             xarray.DataArray.coarsen
 
         Example:
-            `scn.aggregate(func='min', x=2, y=2)` will aggregate 2x2 pixels by
-            applying the `min` function.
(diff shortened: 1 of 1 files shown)


djhoese (commented):
Admittedly as someone who doesn't do a lot of this, this change made it harder for me to understand. The "to half the pixel size" was not what I expected to read and through me off. That said, I don't have a better suggestion at the moment.

raybellwaves:
> half the pixel size" was not what I expected to read and through me off.

Agreed. Half the pixel size is weird. What it was missing was a mention that it's a window of size 2. Had to see how they wrote it in xarray http://xarray.pydata.org/en/stable/generated/xarray.DataArray.coarsen.html#xarray-dataarray-coarsen.

Kudos to Martin for putting a see also in the doc string!

mraspaud (approved):
LGTM

PR #1468, opened 2020-12-03
Harmonize SEVIRI auxiliary data

sfinkens:
This PR makes the following auxiliary data available in the Native and netCDF readers:

- Mean scanline acquisition time as dependent dataset coordinate (`acq_time`)
- Actual and nominal satellite position in dataset attributes (`orbital_parameters`)
- Raw L1.5 header in dataset attributes (`raw_metadata`). This is for HRIT and Native only, because in netCDF files that information is scattered among multiple variables).

Before, this information was only provided by the HRIT reader. Along the way this also became a bigger refactoring of the corresponding unit tests:

- Reduce amount of mocking and increase amount of test data, so that the new functionality can be tested by calling public methods only.
- This was a good opportunity to add a small test for `NativeMSGFileHandler.get_dataset` without a lot of mocking. Would be nice to extend this in the future (HRV channel, padding).
- Move HRIT test setup to a separate file

Documentation has been updated as well. Furthermore I performed tests with real data: Same scan in three different formats, results are identical, also for the HRV channel. But since this is quite a big change, I'd be grateful if you could also test this PR in your environment.

 - [X] Tests added 
 - [X] Passes ``flake8 satpy`` 
 - [X] Fully documented


satpy/readers/seviri_l1b_native.py
```diff
@@ -511,6 +517,81 @@ def _get_calib_coefs(self, channel_name):
...
+        tline2 = tline[:, 2]
+        return da.stack((tline0, tline1, tline2), axis=1).reshape(
+            self.mda['hrv_number_of_lines']).compute()
+
+    def _get_acq_time_visir(self, dataset_id):
+        """Get raw acquisition time for VIS/IR channels."""
+        # Check if there is only 1 channel in the list as a change
+        # is needed in the arrray assignment ie channl id is not present
+        if len(self.mda['channel_list']) == 1:
+            return self.dask_array['visir']['acq_time']
```
ameraner, line 543: The other two `_get_acq_time*` returns are computed, but this one isn't, is this intended?

ameraner (approved):
LGTM, thanks for the great and useful work! I've had colleagues asking how to get the satellite position from native files, now I can tell them it's very easy :)

Just one question on a line, see below.

satpy/readers/seviri_l1b_native.py
sfinkens: Oh, good catch! That should indeed be computed as well!

satpy/readers/seviri_l1b_hrit.py
```diff
@@ -476,7 +368,7 @@ def _get_header(self):
         self.mda['projection_parameters']['SSP_latitude'] = 0.0
 
         # Orbital parameters
-        actual_lon, actual_lat, actual_alt = self.prologue_.get_satpos()
+        actual_lon, actual_lat, actual_alt = self.prologue_.satpos
```
mraspaud, line 371: I understand from the code that `satpos` will be `(None, None, None)` there is not valid polynomial, since the original `get_satpos` raises an exception which is intercepted by `get_satpos_safe`.

Now, I would rather remove `get_satpos_safe` and just get the exception.
In this `__init__` for example, that would mean catching the exception instead of checking for `actual_*` being `None`, which I think would explain more why the `orbital_parameters` aren't set, especially since the name of the exception is so explicit (`NoValidOrbitParams`)

mraspaud (commented):
Great work, thanks for putting in the effort. I just have two comments, one with a high level of code-cleanliness nittpicking :)

satpy/readers/seviri_l1b_hrit.py
mraspaud: This is highly subjective of course, feel free to argue.

sfinkens: Yes, that's a good solution! The `get_satpos_safe` didn't feel right, but I didn't come up with a better solution (to avoid catching and logging the exception in each reader). But since I check for `None` type satellite positions in each reader anyway, it's more clear to catch the exception instead. Thanks for the hint!

satpy/readers/seviri_base.py
```diff
@@ -582,6 +623,194 @@ def chebyshev(coefs, time, domain):
...
+        self.end_time = end_time
+
+    def evaluate(self, time):
+        """Get satellite position in earth-centered cartesion coordinates.
+
+        Args:
+            time: Timestamp where to evaluate the polynomial
+
+        Returns:
+            x, y, z in meters
```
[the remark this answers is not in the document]
sfinkens: Good point, I'll add the coordinate system.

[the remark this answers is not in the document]
sfinkens: @mraspaud Just noticed that the coordinate system is already mentioned in the first line of the docstring. I guess that's what you meant? :slightly_smiling_face:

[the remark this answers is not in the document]
mraspaud: haha, yes, but it's better to repeat it I think

mraspaud (approved):
LGTM!

sfinkens:
Retry DeepCode

mraspaud:
Deepcode has been broken for a while now, let's just ignore it.
Anything you want to do for the codebeat comments?

sfinkens:
I merged the x/y/z coefficients into one argument. Regarding polynomial evaluation: I think it would be nice to pass a 3D array to the `chebyshev` method, but numpy doesn't support this. And splitting the interval finder wouldn't improve readability I think.

sfinkens:
Ok wait, I can do that :)

sfinkens:
Ok, did my best, only one codebeat issue left :slightly_smiling_face:

mraspaud:
Nice! for `_get_closest_interval`, you could split it in two:
```python
def _get_closest_interval(self, time):
    """Find interval closest to the given timestamp."""
    intervals_centre = self.valid_from + 0.5 * (
            self.valid_to - self.valid_from
    )
    diffs_us = (time - intervals_centre).astype('i8')
    closest_match = np.argmin(np.fabs(diffs_us))
    distance = abs(intervals_centre[closest_match] - time)
    return closest_match, distance


def _get_closest_interval_within(self, time, threshold):
    closest_match, distance = self._get_closest_interval(time)
    threshold_diff = np.timedelta64(threshold, 'h')
    if distance < threshold_diff:
        return closest_match
    raise NoValidOrbitParams(
        'Unable to find orbit coefficients valid for {} +/- {}'
        'hours'.format(time, threshold)
```

PR #1472, opened 2020-12-04
Return FSFile instances from find_files_and_readers if passed fs instance

### LINKED ISSUE #1471: Allow find_files_and_readers to return FSFile instances
## Feature Request

#1169 added fsspec support to `find_files_and_readers`, allowing users to search inside archives or on remote file systems.  #1439  added an `FSFile` class that follows the `PathLike` protocol and can be used by selected readers.  However, currently the user still needs to manually produce such `FSFile` instances from the output by `find_files_and_readers` before they can make use of the new `FSFile`/`fsspec` support in satpy.

I would like that if the user passes a `file_system` instance to `find_files_and_readers`, that this function no longer returns simple strings but rather instances of the `FSFile` class, so that they can pass this directly to the `Scene` class and to the reader instead of simply filenames.

Users who are currently passing a `file_system` to `find_files_and_readers` will observe that it returns an `FSFile` instance instead of a string for the filenames.  That means that if they are currently doing string operations, those will fail.  If they pass them directly to the `Scene` class they will not fail.

As far as I'm aware, few users are currently passing `file_system` to `find_files_and_readers`, so I think this backward compatibility issue should be acceptable.

**Additional context**

Users could continue to produce their own `FSFile` instances.  However, as #1169 is still quite new and #1439 is very new, I think no users are currently doing so, and the added convenience proposed in this issue would be welcome.

gerritholl:
If `find_files_and_readers` is passed a file_system object (via the `fsspec` protocol), it will return `FSFile` instances that can be passed directly to `Scene.__init__`. 

 - [x] Closes #1471 
 - [x] Tests added 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented


gerritholl:
Converting back to draft because I'm apparently using `FSFile` wrongly, passing a `fsspec.implementations.local.LocalFileOpener` instead of a `fsspec.OpenFile`.

gerritholl:
Although I'm apparently using `FSFile` wrongly, the unit tests pass and this works:

```
import s3fs, satpy, satpy.readers, datetime, fsspec.implementations.cached
fs_s3 = s3fs.S3FileSystem(anon=True)

cache_storage_block = "/data/gholl/cache/fscache/blockcache"

fs_block = fsspec.implementations.cached.CachingFileSystem(
        fs=fs_s3,
        cache_storage=cache_storage_block,
        cache_check=600,
        check_files=False,
        expiry_times=False,
        same_names=False)

files = satpy.readers.find_files_and_readers(
        base_dir="s3://noaa-goes16/ABI-L1b-RadM/2019/321/14/",
        fs=fs_block,
        reader="abi_l1b",
        start_time=datetime.datetime(2019, 11, 17, 14, 40),
        end_time=datetime.datetime(2019, 11, 17, 14, 42),
        use_fsfile=True,
        fs_open_args={"block_size": 65536})

print("found", len(files["abi_l1b"]), "files")

sc = satpy.Scene(filenames=files)
sc.load(["C01", "C02", "C03"])
print(sc["C03"][100:120, 100:120])
```

gerritholl:
The test is failing on Windows because:

```
>       assert str(ri["abi_l1b"][0]) == str(p)
E       AssertionError: assert 'C:/Users/run...2950029000.nc' == 'C:\\Users\\r...2950029000.nc'
E         Skipping 67 identical trailing characters in diff, use -v to show
E         - C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\test_fsspec_fsfile0\OR_ABI-L1
E         ?   ^     ^           ^       ^     ^    ^                     ^        ^                   ^
E         + C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_fsspec_fsfile0/OR_ABI-L1
E         ?   ^     ^           ^       ^     ^    ^                     ^        ^                   ^
```

gerritholl:
This needs a different approach akin to the comment in #1470.  To be continued later.

gerritholl:
Related: #1475

PR #1474, opened 2020-12-04
Fix EUMGACFDR reader so that all datasets can be read.

ninahakansson:
Fixes so that all datasets can be loaded. All datasets need file_type

 - [X] Passes ``flake8 satpy``


mraspaud (approved):
LGTM

PR #1476, opened 2020-12-06
Add support for multiple values in the DecisionTree used for enhancements

djhoese:
The current `DecisionTree` class, which is used as the base class for the `EnhancementDecisionTree`, currently has issues when a DataArray has multiple sensors specified for `.attrs['sensor']`. It currently causes a `TypeError` with a semi-cryptic debug log message produced:

```
Strange stuff happening in decision tree for %s: %s
```

This is because the DecisionTree expects a single string value for the "sensor" value, but is getting a set (ex. `{'abi', 'glm'}`). What this means is that you can never actually configure an enhancement for a DataArray with multiple sensors as it will fail when trying to find the match. This PR is meant to allow for you to both configure an enhancement for `sensor: [abi, glm]` to match it exactly, or for the provided DataArray with `.attrs['sensor'] = {'abi', 'glm'}` to match either `sensor: abi` or `sensor: glm`.

My one concern about this is that specifying `sensor: [abi, glm]` in the YAML config will make people think that that matches either `abi` or `glm` when it will actually only match the exact pair `{'abi', 'glm'}`.

 - [x] Tests added 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented

### DIFF
--- a/satpy/tests/test_writers.py
+++ b/satpy/tests/test_writers.py
@@ -111,53 +111,12 @@ class TestEnhancer(unittest.TestCase):
             ValueError, Enhancer, enhancement_config_file="is_not_a_valid_filename_?.yaml")
 
 
-class TestEnhancerUserConfigs(unittest.TestCase):
-    """Test `Enhancer` functionality when user's custom configurations are present."""
+class _BaseCustomEnhancementConfigTests:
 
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM

PR #1481, opened 2020-12-09
Remove x/y coordinates in mviri_l1b_fiduceo_nc

sfinkens:
Remove x/y coordinates from `mviri_l1b_fiduceo_nc` datasets as they only contain image rows/columns and prevent satpy from adding projection coordinates (based on the area definition).

 - [X] Tests added 
 - [x] Passes ``flake8 satpy``


mraspaud (approved):
LGTM!

PR #1486, opened 2020-12-15
Add MIRS reader

joleenf:
- [ ] Work in Progress PR takes over for #1285.  
 - [ ] Minor adjustments have been made to add coordinates, rename the dimensions and update code for retrieving sensor and platform names.  
 - [ ]  Brightness Temperature dataset ID's are available in dataset ID list.  Started separating get_dataset call into data that can be loaded without modification and the 3D sets that need work.  The 3D section is not complete and not correct since no limb correction is being applied.  
 - [ ] Tests remain the same as Katherine Kolman's work at this time.  
 - [ ] Passes ``flake8 satpy``


satpy/etc/readers/mirs_l2_nc.yaml
```diff
@@ -0,0 +1,62 @@
...
+  long_name: MiRS Level 2 Swath Product Reader (NetCDF4)
+  reader: !!python/name:satpy.readers.yaml_reader.FileYAMLReader
+  sensors: [amsu-mhs, atms, ssmis, gmi]
+
+file_types:
+  mirs_atms:
+    file_reader: !!python/name:satpy.readers.mirs_l2_nc.MIRSHandler
+    file_patterns:
+      - 'NPR-MIRS-IMG_v{version}_{platform_shortname}_s{start_time:%Y%m%d%H%M%S}{extra_num1}_e{end_time:%Y%m%d%H%M%S}{extra_num2}_c{creation_time:%Y%m%d%H%M%S}{extra_num3}.nc'
+      - 'NPR-MIRS-SND_v{version}_{platform_shortname}_s{start_time:%Y%m%d%H%M%S}{extra_num1}_e{end_time:%Y%m%d%H%M%S}{extra_num2}_c{creation_time:%Y%m%d%H%M%S}{extra_num3}.nc'
```
djhoese, line 14: I don't think this file pattern is needed as this reader can't actually use these files if I'm not mistaken. It at least shouldn't be needed for the related Polar2Grid work.

djhoese (commented):
We'll have to be careful since it looks like the LUTs were added to the git repository. When this is all done the "proper" way we'll have to remove them from the git history (otherwise they will take up space for no reason).

joleenf:
Oops. The LUT data in the code was just a placeholder that I was going to fix later, I didn't realize I added the LUTs to the repo.  Sorry about that.  I was going to talk with you about where to put the LUTs  tomorrow.  For the file pattern, let's talk about that at the meeting.  I think I am misunderstanding the big picture of how satpy actually runs with the glue script in P2G.

satpy/readers/mirs_l2_nc.py
```diff
@@ -0,0 +1,506 @@
...
+import os
+import logging
+import datetime
+import numpy as np
+import xarray as xr
+import dask.array as da
+from pyproj import CRS
+from satpy.readers.file_handlers import BaseFileHandler
+
+LOAD_CHUNK_SIZE = int(os.getenv('PYTROLL_LOAD_CHUNK_SIZE', -1))
```
djhoese, line 29: This should be taken `from satpy import CHUNK_SIZE` so it can be updated more easily in the future. The environment variable is `PYTROLL_CHUNK_SIZE` that controls that value.

satpy/readers/mirs_l2_nc.py
```diff
@@ -0,0 +1,506 @@
...
+        for fov_idx in range(N_FOV):
+            new_ds[:, fov_idx] = da.map_blocks(self.coeff_cums_calc,
+                                               datasets, channel_idx,
+                                               fov_idx, dmean, coeffs,
+                                               amean, nchx,
+                                               nchanx, dtype=ds.dtype)
+
+        return new_ds
+
+    @staticmethod
```
djhoese, line 211: There are some cases (multi-process scheduler) where dask doesn't always work well with functions that aren't globally defined. I'm not sure if staticmethods have this problem, but what are your thoughts on moving this staticmethod (and any other delayed/map_blocks'd methods) to be functions in the global scope?

djhoese (changes requested):
A couple quick questions/changes. Hopefully can answer your questions from slack after a little more discussion.

joleenf:
Hi @djhoese, I have moved the limb_correction code to a global function.  CHUNK_SIZE is being used.  I believe I have removed all of the correction files that I have inadvertently committed.  The correction is currently not active due to the fact that we need to figure out where those correction files will be stored.

joleenf:
@djhoese I probably should have left this as a draft considering the question of coeffcient files.

djhoese:
Hm it seems you have confused github (6,859 commits). Yay git! It looks like the commands you ran to remove the files from your commits must have gone through and edited  almost all Satpy commits (master has ~8k commits). If you can't figure it out, it may be worth copying your changed files to a new branch and making a new PR. Sorry.

djhoese:
Additionally it looks like one of the test environments failed.

joleenf:
Oh. No.

PR #1491, opened 2020-12-23
Fix missing coordinates for bucket resamplers

### LINKED ISSUE #1488: Resampling with bucket resamplers drops coords from xr.DataArray
**Describe the bug**
Using bucket resamplers drops the coords from the `xr.DataArray` of the Scene datasets.

**To Reproduce**

```python
scn = Scene(reader="seviri_l1b_nc", filenames=["PATH_TO_SEVIRI_NC_FILE"])
scn.load(["IR_039"])
scn = scn.resample("EuropeCanary", resampler="bucket_avg")
scn["IR_039"].coords

```

**Expected behavior**
The coordinates of the `xr.DataArrays` should be preserved.

**Environment Info:**
 - OS: Linux
 - Satpy Version: 0.24.0 and older versions
 - PyResample Version: 1.17.0

BENR0:
Bucket resamplers update DataArray coords after resampling now.

 I fixed the tests by adding necessary return values to the `target_geo_def` mocks. For `crs` I just added `None` and within the resamplers no tests if the coords are updated are actually performed. I think  this is ok since the change is only that the call to `update_resampled_coords` is added to the resample method and tests for the update function are already covered.

 - [x] Closes #1488  
 - [x] Tests added 
 - [x] Passes ``flake8 satpy``


djhoese (approved):
Nice job. Looks simple enough. I know this wasn't your doing but I really dislike the mocking of AreaDefinition objects. They aren't that hard to create and mocking them just makes us less aware of any breakage between pyresample and satpy.

Any idea how hard it would be to replace the source and target geo defs with real AreaDefinitions?

BENR0:
Well I guess it wouldn't be extremely complex. There already is a helper function to create test data at the beginning of the file, but I think there would be some amount of refactoring involved. I am really not that familiar with the code and the tests but from a quick look I think if the mocking of area definitions in the bucket resampler tests get refactored maybe the tests could also be extended. Currently the shape, dims and if attributes are kept is tested. Shouldn't also be tested if the additional attributes depending on the resampler are added?

djhoese:
Don't worry about it. Let's just merge this. Thanks again for the fixes.

PR #1498, opened 2021-01-04
Fix: bucket resampler changing dtype

BENR0:
Fixes the dtype of the data returned from bucket resampler being different from the input data dtype.

The `get_sum`, `get_average` and `get_fractions` called in the `compute` method of the resampler use `dask.array.where` which does not preserve dtype.

 - [x] Tests added 
 - [ ] Passes ``flake8 satpy``

### DIFF
--- a/satpy/resample.py
+++ b/satpy/resample.py
@@ -1105,7 +1105,7 @@ class BucketResamplerBase(BaseResampler):
             attrs['calibration'] = ''
             attrs['standard_name'] = 'number_of_observations'
 
-        result = xr.DataArray(result, dims=dims, coords=coords,
+        result = xr.DataArray(result.astype(data.dtype.type), dims=dims, coords=coords,
                               attrs=attrs)
 
         return update_resampled_coords(data, result, self.target_geo_def)
--- a/satpy/tests/test_resample.py
+++ b/satpy/tests/test_resample.py
@@ -749,6 +749,7 @@ class TestBucketAvg(unittest.TestCase):
     @mock.patch('pyresample.bucket.BucketResampler')
     def test_resample(self, pyresample_bucket):
         """Test bucket resamplers resample method."""
+        import numpy as np
         import xarray as xr
         import dask.array as da
         self.bucket.resampler = mock.MagicMock()
@@ -766,12 +767,13 @@ class TestBucketAvg(unittest.TestCase):
         self.assertTrue('bar' in res.attrs)
         self.assertEqual(res.attrs['bar'], 'baz')
 
-        # 2D input data
-        data = xr.DataArray(da.ones((5, 5)), dims=('foo', 'bar'))
+        # 2D input data with dtype
+        data = xr.DataArray(da.ones((5, 5), dtype=np.float32), dims=('foo', 'bar'))
         self.bucket.compute.return_value = da.ones((5, 5))
         res = self.bucket.resample(data)
         self.assertEqual(res.shape, (5, 5))
         self.assertEqual(res.dims, ('y', 'x'))
+        self.assertEqual(res.dtype, data.dtype)
 
         # 3D input data with 'bands' dim
         data = xr.DataArray(da.ones((1, 5, 5)), dims=('bands', 'foo', 'bar'),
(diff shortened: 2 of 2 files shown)


djhoese:
Could you point me to the specific `where` you are talking about? Recent versions of xarray have made huge improvements in handling of dtypes as long as the type of the fill (replacement) value is the right dtype.

BENR0:
I am refering to lines:

- https://github.com/pytroll/pyresample/blob/270b5f080f26ac5f7a28b5e37d93e1fc622181cb/pyresample/bucket/__init__.py#L168
- https://github.com/pytroll/pyresample/blob/270b5f080f26ac5f7a28b5e37d93e1fc622181cb/pyresample/bucket/__init__.py#L246
- https://github.com/pytroll/pyresample/blob/270b5f080f26ac5f7a28b5e37d93e1fc622181cb/pyresample/bucket/__init__.py#L283

As far as I know xarray will preserve dtype when `xarray.DataArray.where` (or Dataset) is used which does not allow an alternative value. I think the `dask.array.where` is similar to `xarray.where` and `np.where` and those are not preserving dtype.

djhoese:
I think this needs to be fixed in the various resamplers you've pointed to. CC @pnuu for ideas and concerns. If all of the internals of the resamplers are using floats then fine, but if they are using the original data values then `np.isnan` isn't going to be correct for a lot of these. Also `get_fractions` isn't in same units as the original data is it? It is a percentage/ratio I think, right?

BENR0:
Ok I guess that would be better since, as you pointed out, not all results should necessarily have the same dtype as the input. I just came across this for the `bucket_avg` and didn't think about the fractions variant. As far as I see it does not hurt to have the same dtype as the input data for `get_fractions` but I guess it could be cast down.

djhoese:
I would say if `get_fractions` is the fraction of the whole image that is being used (or something similar) then it doesn't make sense for this to be anything except a float. The sum and average make more sense although a "category" product (ex. cloud_type) doesn't really make sense for either of these resamplers anyway. :thinking:

BENR0:
:-D I think I should take a break. Obviously I haven't thought this through. My assumption was that the input is always float which is not justified and of course fractions should always be float.

djhoese:
@BENR0 Let me know what you think on this PR now that it has been a few months. Should we close it? Do you feel like changes similar to this are still needed or would be helpful?

BENR0:
@djhoese yes I think so. Obviously the situation is more complex than I thought but never the less I think dtypes should be preserved as much as possible. Right now due to `dask.array.where` not preserving dtypes even if the input is `np.float32` for example the result will be `np.float64`. Apart from being unexpected from the user point of view this might also have memory implications.

djhoese:
I think in recent versions of numpy + dask + xarray this situation has been improved where it doesn't magically upcast data if it doesn't need to. Any chance you could explore if this is still a problem with newer versions of these libraries?

BENR0:
Ok that would be even better. I probably will have a chance to look at it next week.

djhoese:
Had a chance to look into this at all?

BENR0:
Sorry totally forgot about this. I looked into it now. I used numpy 1.20.0, xarray 0.16.2 and dask 2.30.0 and can confirm that indeed dtypes are preserved if possible using `*.where`. But for example the bucket average resampler still does not preserve the dtype. The reason in this case seems to be the division by counts which is of dtype `int64` which leads to the result being upcast to `float64` (see: https://github.com/pytroll/pyresample/blob/270b5f080f26ac5f7a28b5e37d93e1fc622181cb/pyresample/bucket/__init__.py#L243-L245).

I did not check all bucket resamplers in detail but from a quick look there might be some lines where data is simply cast to `float` or `int` which always returns the default 64 bit dtype.
Obviously, as discussed above, what I proposed in this PR does not cut it but I think it would still be nice if results did not get upcast unnecessarily.

djhoese:
I think I kind of expressed this idea above, but to me the average is producing something new. It isn't actually the original data in a strict sense so preserving the dtype seems unnecessary in my opinion. But that's just my opinion. @pnuu if you're OK with it, I'd be OK with doing a check and casting `nan` to a specific floating point type here:

https://github.com/pytroll/pyresample/blob/270b5f080f26ac5f7a28b5e37d93e1fc622181cb/pyresample/bucket/__init__.py#L245

I would think that would set the dtype for the rest of the calculations done.

BENR0:
True the data may not be the same but I would argue that by averaging data the needed precision will not increase, so if the input is `float32` I don't see why the averaged data should be `float64`. I the user really would like to use `float64` he can still upcast the data before. I hope that makes sense.

djhoese:
Makes sense. It is OK with me.

gerritholl:
@BENR0 What is the status of this PR?  I read through the comments and I'm not sure if this is still relevant?  In any case, it has merge conflicts…

PR #1499, opened 2021-01-04
Fix: clipping of values for geotiff writer if enhance=False and no dtype is specified

BENR0:
When writing datasets to geotiff with `enhance=False` and no explicit dtype is given using the `dtype` parameter the dtype of the dataset is now used by default instead of `uint8` which resulted in clipping before.

 - [x] Passes ``flake8 satpy``


doc/source/enhancements.rst
```diff
@@ -1,3 +1,5 @@
+.. _chapter_enhancements:
+
```
djhoese, line 2: This shouldn't be needed. I think you can do `:doc:'enhancements'` and it will serve the same purpose.

satpy/writers/geotiff.py
```diff
@@ -186,8 +192,10 @@ def save_image(self, img, filename=None, dtype=None, fill_value=None,
             fill_value = self.info.get('fill_value')
 
         dtype = dtype if dtype is not None else self.dtype
-        if dtype is None:
+        if (dtype is None) and (self.enhancer is not False):
```
djhoese, line 195: The parentheses aren't necessary. Please remove them.

satpy/writers/geotiff.py
```diff
@@ -34,7 +34,12 @@ class GeoTIFFWriter(ImageWriter):
 
         >>> scn.save_datasets(writer='geotiff')
 
-    Un-enhanced float geotiff with NaN for fill values:
+    By default the writer will use the :class:`~satpy.writers.Enhancer` class to
+    linear stretch the data (see :ref:`Enhancements <chapter_enhancements>`).
+    To get Un-enhanced images ``enhance=False`` can be specified which will
+    write a geotiff with the data type of the dataset with NaN for
+    fill values by default. If a geotiff with a certain datatype is desired
```
djhoese, line 41: Where is this `NaN` set? It would probably be best to use the `_FillValue` attribute if it is defined and the DataArray is an integer (this is common practice in Satpy) otherwise `NaN`. Obviously, NaN can't be used for integer types so forcing the NaN fill value would cast the data no matter what. If it is possible to not alter the fill value at all that would be the best.

djhoese (changes requested):
I like the idea, but I'm hoping it can be more clearly defined. One small style change requested.

doc/source/enhancements.rst
BENR0: Good to know. Will change it.

satpy/writers/geotiff.py
BENR0: I did not change anything regarding the setting of the fill value. This was documented before I just added a little more context to the documentation. So I guess if this was default behaviour in Satpy it should still be.

djhoese: How about editing the last part of that sentence "with NaN for fill values by default" to "with an appropriate fill value". Or you could be clearer about it, but that may be more documentation than is needed.

satpy/writers/geotiff.py
```diff
@@ -34,7 +34,13 @@ class GeoTIFFWriter(ImageWriter):
 
         >>> scn.save_datasets(writer='geotiff')
 
-    Un-enhanced float geotiff with NaN for fill values:
+    By default the writer will use the :class:`~satpy.writers.Enhancer` class to
+    linear stretch the data (see :doc:`../enhancements`).
+    To get Un-enhanced images ``enhance=False`` can be specified which will
+    write a geotiff with the data type of the dataset. The fill value defaults
+    to the the datasets ``"fill_value"`` attribute if not ``None`` and no value is
```
djhoese, line 41: `_FillValue` not `fill_value`, it's a CF convention thing.

BENR0: True for CF conventions its `_FillValue` but as far as I can see, at least for the geotiff writer, neither the `_FillValue` nor `fill_value` attribute is used. But I think that could easily be implemented.

djhoese: See https://github.com/pytroll/trollimage/blob/master/trollimage/xrimage.py#L970-L995

BENR0: Ok I see. I didn't think of looking in xrimage since for floats the NaN gets set in the geotiff writer if `fill_value` is not passed and I didn't find any reference to `_FillValue` around there. Just one question why isn't `_FillValue` checked for float data? Wouldn't it be more consistent to default to NaN for float data only if `_FillValue` is None or non existent and no `fill_value` parameter is passed similar to the integer data case?

djhoese: I suppose it could be but in most cases a float having the `_FillValue` attribute is usually a mistake in the reader code where it wasn't removed. The `_FillValue` usually corresponds to the in-file integer fill value and not the in-memory unscaled version of the data. I guess I wouldn't be against it, but I also don't think it would be a very common case. It is also very likely that any xarray operations will put NaNs into the float data for invalid values.

BENR0: Ok yes from a "complete" Satpy workflow point of view starting with a reader this makes totally sense since `_FillValue` should indeed be removed. I guess I thought it would in general be easier to understand for the user how fill values area handled if the order of what is checked is similar between data types. But I agree this is probably not a very common case (especially for possible workflows I have in mind) and in any case would also just shift the "problem" from the user supplying ``fill_value`` parameter to setting the `_FillValue` attribute and I guess in any case, as most of the time, the user should know what he's doing anyway. I updated the documentation and hope it is clearer now. Sorry for all the back and forth about this rather "unimportant" detail.

djhoese:
@BENR0 I was looking over this and I think it is good to go except for a test. Do you think you could add one?

djhoese (approved):
This looks good to me. Thanks for adding a test. In the future we should probably make it easier to check geotiff files that are written to disk. Given how the other tests are written I think your added test is consistent so I'm OK with it. Thanks.

PR #1503, opened 2021-01-05
Fix issue with reading MSG GRIB products from the eumetsat datastore

### LINKED ISSUE #1477: seviri l2 grib add file names from Eumetsat datastore
The seviri_l2_grib reader can read l2 grib files from the archive as far as I can understand.

With the eumetsat datastore a user can download the same data but with a slightly different filename.

Need to add this file name to be read.

Add after: https://github.com/pytroll/satpy/blob/master/satpy/etc/readers/seviri_l2_grib.yaml#L16 something like this:

`- '{spacecraft:4s}-SEVI-MSGCLMK-0100-0100-{start_time:%Y%m%d%H%M%S}.000000000Z-NA.grb'`

Also consider the other products in the reader if applicable.

ColinDuff:
…etsat datastore and added AES, CTH, MPE and ClearSkyMap GRIB products

Resolves Issue 1477 by adding the EUMETSAT Datastore file pattern

Also further Seviri Grib products were added

A line of code in the reader was also fixed as it didnt seem to be working correctly

Closes #1477 
Test script ran and no problems were observed
Passes ``flake8 satpy``

### DIFF
--- a/satpy/etc/readers/seviri_l2_grib.yaml
+++ b/satpy/etc/readers/seviri_l2_grib.yaml
@@ -14,6 +14,7 @@ file_types:
         - 'CLMEncProd_{start_time:%Y%m%d%H%M%S}Z_00_{server:8s}_{spacecraft:5s}_{scan_mode:3s}_{sub_sat:5s}'
         - '{spacecraft:4s}-SEVI-MSGCLMK-0100-0100-{start_time:%Y%m%d%H%M%S}.000000000Z-{product_creation_time:%Y%m%d%H%M%S}-{ord_num:7s}'
         - '{spacecraft:4s}-SEVI-MSGCLMK-0100-0100-{start_time:%Y%m%d%H%M%S}.000000000Z-{product_creation_time:%Y%m%d%H%M%S}-{ord_num:7s}.grb'
+        - '{spacecraft:4s}-SEVI-MSGCLMK-0100-0100-{start_time:%Y%m%d%H%M%S}.000000000Z-NA.grb'
 
   # EUMETSAT MSG SEVIRI L2 Optimal Cloud Analysis files in GRIB format
   grib_seviri_oca:
@@ -22,6 +23,7 @@ file_types:
         - 'OCAEncProd_{start_time:%Y%m%d%H%M%S}Z_00_{server:8s}_{spacecraft:5s}_{scan_mode:3s}_{sub_sat:5s}'
         - '{spacecraft:4s}-SEVI-MSGOCAE-0100-0100-{start_time:%Y%m%d%H%M%S}.000000000Z-{product_creation_time:%Y%m%d%H%M%S}-{ord_num:7s}'
         - '{spacecraft:4s}-SEVI-MSGOCAE-0100-0100-{start_time:%Y%m%d%H%M%S}.000000000Z-{product_creation_time:%Y%m%d%H%M%S}-{ord_num:7s}.grb'
+        - '{spacecraft:4s}-SEVI-MSGOCAE-0100-0100-{start_time:%Y%m%d%H%M%S}.000000000Z-NA.grb'
 
   # EUMETSAT MSG SEVIRI L2 Active Fire Monitoring files in GRIB format
   grib_seviri_fir:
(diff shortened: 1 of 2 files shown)


ColinDuff:
@mraspaud @djhoese 
can you have a look at this PR - mainly yaml file updates 
does it look ok to you

satpy/readers/seviri_l2_grib.py
```diff
@@ -153,7 +153,7 @@ def get_dataset(self, dataset_id, dataset_info):
 
                 # Retrieve values and metadata from the GRIB message, masking the values equal to missing_value
                 xarr = self._get_xarray_from_msg(gid)
-                xarr.where(xarr.data == missing_value, np.NaN)
+                xarr.data = xr.where(xarr.data == missing_value, np.nan, xarr.data)
```
djhoese, line 156: This seems a little off to me. Since you are accessing `xarr.data` then it would be clearer to me to use dask's `da.where`. As far as I can tell though the original line should have been `xarr = xarr.where(xarr == missing_value, np.nan)`. Is there a reason that won't work here?

satpy/etc/readers/seviri_l2_grib.yaml
```diff
@@ -152,3 +187,123 @@ datasets:
...
+    parameter_number: 12
+    units: "%"
+    long_name: nir_refl_39
+
+  num_accumulations:
+    name: num_accumulations
+    resolution: 3000
+    file_type: grib_seviri_crm
+    parameter_number: 6
+    units: ""
```
djhoese, line 284: I think it is semi-standard in Satpy to use `"1"` for unitless fields.

djhoese (changes requested):
Two small comments inline otherwise it looks pretty good. If there are a lot of fields in these files it may be time to try to replace the long set of yaml datasets with the `available_datasets` method in the reader to dynamically determine what's available from the files provided.

satpy/readers/seviri_l2_grib.py
ColinDuff: hi, thanks for the comments. In the original code it didnt set the missing value data points to nan and i couldnt figure out why so i used a similar line of code i had in another reader and this worked ok. 
i can update the reader to try the line of code you mention

satpy/etc/readers/seviri_l2_grib.yaml
ColinDuff: ok will update

satpy/readers/seviri_l2_grib.py
ColinDuff: do you have an example of how you would use the available_datasets method - not sure how we would use this

djhoese: I'm not saying you should do it in this PR necessarily unless you want to explore it as a simpler version of the YAML code. There are a couple readers who use it currently, but the best documentation is currently in the base class itself:

https://github.com/pytroll/satpy/blob/93f0233f9e4675d516de70dd36183dc6320fa82e/satpy/readers/file_handlers.py#L163-L248

The only time this doesn't work is if you want to define additional metadata in the YAML that isn't available in the file (or can't be easily determined in code).

satpy/etc/readers/seviri_l2_grib.yaml
ColinDuff: ok, done

satpy/readers/seviri_l2_grib.py
ColinDuff: hi, the amendment to the original line of code, masking out the missing values doesnt work. The returned array still includes missing values thus my data values are incorrect. 
I updated the new line of code to use dask array where and this works as expected.
any thoughts

djhoese: My original suggestion was wrong, it should have been `xarr != missing_value`. What you have now is fine, but a little "indirect" since there is a method that is more commonly used in Satpy readers and is supposed to work just fine. If what you have now works then leave it (it is at least consistent with its use of dask and xarray objects).

djhoese (approved):
Thanks for the updates. LGTM

mraspaud (approved):
LGTM

PR #1511, opened 2021-01-22
Add MiRS reader

joleenf:
This PR is a replacement for PRs #1486 and #1285.  This MiRS reader loads the level 2 EDR IMG swath files produced by the Microwave Integrated Retrieval System.  

- The location of coefficient files still needs to be determined and code needs to be updated as appropriate.


djhoese:
I've updated the title just because we use it in our release notes so didn't want to clutter it with the PR numbers.

satpy/readers/mirs.py
```diff
@@ -0,0 +1,502 @@
...
+
+        """
+        data = self.nc[item]
+        attrs = data.attrs.copy()
+        data, attrs = self._scale_data(data, attrs)
+        data, attrs = self._fill_data(data, attrs)
+
+        # handle coordinates (and recursive fun)
+        new_coords = {}
+        # 'time' dimension causes issues in other processing
```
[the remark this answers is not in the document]
joleenf: Oops.  Same problem reference to wrong dimension from another dataset.

satpy/readers/mirs.py
```diff
@@ -0,0 +1,502 @@
...
+        sea_bt = apply_atms_limb_correction(bt_data, idx, *sea)
+
+        land = read_atms_limb_correction_coeffs(LIMB_LAND_FILE)
+        land_bt = apply_atms_limb_correction(bt_data, idx, *land)
+
+        LOG.info("Finishing limb correction")
+        is_sea = (surf_type_mask == 0)
+        bt_data = bt_data[idx, :, :]
+        new_data = (sea_bt.squeeze() * [is_sea] +
+                    land_bt.squeeze() * [~is_sea])
```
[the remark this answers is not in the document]
joleenf: Only in setting up the dims and coords for the new product array.  That can be handled differently.

satpy/readers/mirs.py
```diff
@@ -0,0 +1,502 @@
...
+
+        # for consistency when not doing limb correction, return a dask.array
+        new_data = da.from_array(new_data.squeeze())
+
+        bt_data = xr.DataArray(new_data.squeeze(), dims=bt_data.dims,
```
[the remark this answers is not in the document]
joleenf: None are necessary with da.where.

satpy/etc/readers/mirs.yaml
```diff
@@ -5,6 +5,11 @@ reader:
   long_name: MiRS Level 2 Swath Product Reader (NetCDF4)
   reader: !!python/name:satpy.readers.yaml_reader.FileYAMLReader
   sensors: [amsu, amsu-mhs, atms, ssmis, gmi]
+  data_files:
+     - url: "https://zenodo.org/record/4472664/files/limbcoef_atmsland_noaa20.txt"
```
djhoese, line 9:
```suggestion
     - url: "https://zenodo.org/record/4472664/files/limbcoef_atmsland_noaa20.txt"
       known_hash: "sha256:08a3b7c1594a963610dd864b7ecd12f0ab486412d35185c2371d924dd92c5779"
```

Generated by doing `openssl sha256 <file>` and copying the produced hash and preprending `sha256:`. Could you add these to all the files?

satpy/readers/mirs.py
```diff
@@ -18,16 +18,20 @@
...
 import numpy as np
 import xarray as xr
 import dask.array as da
 from satpy import CHUNK_SIZE
 from satpy.readers.file_handlers import BaseFileHandler
+from satpy.aux_download import find_registerable_files, retrieve
 
 LOG = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG)
+SATPY_DATA_DIR = appdirs.user_data_dir("satpy")
```
djhoese, line 34: I don't think you're using this anymore or appdirs (imported above). If you use `retrieve` you shouldn't need the data directory anyway. `retrieve` should give you the full path.

satpy/readers/mirs.py
```diff
@@ -256,6 +258,28 @@ def end_time(self):
...
+        coeff_fn = {'sea': None, 'land': None}
+
+        for filename in platform_files:
+            if "atmssea" in filename:
+                coeff_fn['sea'] = retrieve(filename)
+            elif "atmsland" in filename:
+                coeff_fn['land'] = retrieve(filename)
+            else:
+                LOG.warning('Unknown coefficient filename {}', filename)
+        return coeff_fn
```
djhoese, line 281: I'd prefer if these filenames were hard-coded rather than the filter and regex stuff. You as the developer shouldn't need to use `find_registerable_files`. If you have ideas for how I can clear this up in the documentation let me know.

I was hoping it would be something like:

```
if self.platform_name == "noaa-20":
    land_fn = retrieve('limbcoef_atmsland_noaa20.txt')
coeff_fn['land'] = land_fn
```

satpy/etc/readers/mirs.yaml
```diff
@@ -0,0 +1,64 @@
+reader:
+  description: NetCDF Reader for the Microwave Integrated Retrieval System Level 2 swath products
+  name: mirs
+  short_name: MiRS Level 2 NetCDF4
+  long_name: MiRS Level 2 Swath Product Reader (NetCDF4)
+  reader: !!python/name:satpy.readers.yaml_reader.FileYAMLReader
+  sensors: [amsu, amsu-mhs, atms, ssmis, gmi]
+  data_files:
+     - url: "https://zenodo.org/record/4472664/files/limbcoef_atmsland_noaa20.txt"
+       file: "limbcoef_atmsland_noaa20.txt"
```
djhoese, line 10: The parameter is `filename` (or at least it should be). It technically isn't necessary here as it should end up being the same as the filename in the URL.

satpy/readers/mirs.py
```diff
@@ -0,0 +1,486 @@
...
+    return all_dmean, all_coeffs, all_amean, all_nchx, all_nchanx
+
+
+def apply_atms_limb_correction(datasets, channel_idx, dmean, coeffs, amean, nchx, nchanx):
+    """Apply the atms limb correction to the brightness temperature data."""
+    datasets = datasets.persist()
+    ds = datasets[channel_idx]
+    new_ds = np.zeros(ds.shape, dtype=ds.dtype)
+    for fov_idx in range(N_FOV):
+        new_ds[:, fov_idx] = da.map_blocks(coeff_cums_calc,
```
djhoese, line 140: I could be wrong, but doesn't this force the dask arrays to be computed (writing a dask array to a numpy array)? What if you got a list of all of the dask arrays and then concatenated them together along the proper axis?

satpy/etc/readers/mirs.yaml
joleenf: I will pull the "file" lines.

satpy/readers/mirs.py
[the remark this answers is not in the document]
joleenf: The da.where change eliminated all of the squeeze commands.

satpy/readers/mirs.py
```diff
@@ -0,0 +1,502 @@
...
+            return self.filename_info[key]
+
+    def force_time(self, key):
+        """Force datetime.time for combine."""
+        if isinstance(self.filename_info.get(key), datetime.datetime):
+            return self.filename_info.get(key).time()
+        else:
+            return self.filename_info.get(key)
+
+    def limb_correct_atms_bt(self, bt_data, ds_info):
```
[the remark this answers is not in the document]
joleenf: Not exactly.  I started pulling the lines that use self out of this subroutine and it created an incredibly long function call.  If that is preferred, I can still make the change.   The way the coefficient filenames are stored would have to change, but I am not sure if they necessarily have to be part of the dask Dataset attributes.

satpy/readers/mirs.py
```diff
@@ -0,0 +1,502 @@
...
+                do_not_apply = True
+            else:
+                do_not_apply = True   # TODO:  Change this to false when coeff loc is resolved.
+
+            if self.sensor.lower() != "atms" or do_not_apply:
+                LOG.info("Limb Correction will not be applied to non-ATMS BTs")
+                data = data[:, :, idx]
+            else:
+                data = self.limb_correct_atms_bt(data, ds_info)
+                self.nc = self.nc.merge(data)
```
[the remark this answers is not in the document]
joleenf: Yes

satpy/readers/mirs.py
```diff
@@ -0,0 +1,502 @@
...
+                continue
+
+            if isinstance(val, xr.DataArray):
+                # get specific brightness temperature band products
+                if (var_name == 'BT' and
+                        self.filetype_info['file_type'] in 'mirs_atms'):
+                    freq = self.nc.coords.get('Freq', self.nc.get('Freq'))
+                    polo = self.nc['Polo']
+                    from collections import Counter
+                    c = Counter()
```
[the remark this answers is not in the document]
joleenf: Changed to ch_total (tally of same channel/polarization combinations) and ch_cnt, which of any duplicate channel/polarization combinations is currently having metadata created for it.

satpy/readers/mirs.py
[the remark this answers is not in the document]
djhoese: I'm not sure I understand. I mean that this method `limb_correct_atms_bt` is an instance method and has its first argument as `self` however it never actually uses `self` inside the method (from what I can tell). This means it could be made a `@staticmethod` or the method could be moved outside of the class to be a function.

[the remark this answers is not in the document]
joleenf: The method itself doesn't need it, you are correct.  I'll fix it.

satpy/readers/mirs.py
joleenf: No.  You are correct.

joleenf: updated in f526e996

joleenf: f526e99 is wrong though.  It is the version which sends dask arrays to a function called by map_blocks.  ac4bcccb uses numpy input arrays and the function returns a numpy array.

satpy/readers/mirs.py
```diff
@@ -173,10 +173,10 @@ def limb_correct_atms_bt(bt_data, surf_type_mask, coeff_fns, ds_info):
     is_sea = (surf_type_mask == 0)
     new_data = da.where(is_sea, sea_bt, land_bt)
 
-    bt_corrected = xr.DataArray(new_data, dims=bt_data[idx, :, :].dims,
-                                coords=bt_data[idx, :, :].coords,
+    bt_corrected = xr.DataArray(new_data, dims=("y", "x"),
+                                coords=surf_type_mask.coords,
```
djhoese, line 177: What are the coords this variable has?

joleenf: I was thinking about this and am not sure the best way to add the coords.  I was considering adjusting the way the method new_coords works at line 211 to make this a more obvious way of assigning longitude and latitude coordinates.

joleenf: Actually was thinking about that same thing for the chunk size for the  map_blocks.  If I added a general self.nc.shape so that I can access the scanlines/fov without slicing the bt_data.

djhoese: Let's talk about it at the meeting today *BUT* you shouldn't need to add lon/lat/x/y coordinates. The base reader class will do that for you based on the `coordinates:` defined in the YAML (or your available_datasets method).

djhoese:
> @djhoese In this commit, the file metadata is taking precedence over metadata in the yaml for the metop file used for testing. (IMG_SX.M2.D17037.S1601.E1607.B0000001.WE.HR.ORB.nc). Therefore, standard_name is added when reading the file metadata. I thought this line in the available_datasets meant that it is possible to append new information to the metadata:
> 
>> This method is not guaranteed that it will be called before any other file type’s handler. The availability “boolean” not being None does not mean that a file handler called later can’t provide an additional dataset, but it must provide more identifying (DataID) information to do so and should yield its new dataset in addition to the previous one.
>
> Can the boolean be True, False or None? Or is it only either True or None? I think the available_dataset() information above is saying, "Even if the availability "boolean" is True rather than None, a file handler could provide additional datasets." After that line, what is the more identifying DataID? Is this referring to the more definitive DataIDs like Brightness Temperature data defined by frequency and polarization "btemp_23v" versus the wider scope of BT? If so, then that is working correctly in this reader, because both the BT and the definitive brightness temperature channels are being yielded. Is there a way to keep variable metadata from a yaml and use metadata from a file, or is it either/or?
Also, in "Migrating to Xarray" the example shows using the "coord" attribute. Is assign_coords the only concern in this issue report?

I think it can be True/False/None. I'm not sure I know or remember exactly what you're talking about and am missing some of the context of this comment/discussion. I think you have it right though. The `available_datasets` method is not supposed to stop/limit/remove datasets that have been "reported" by other file handlers. There is also no order to which file handler's `available_datasets` is called first so that why it has to be very passive.

Regarding coords, it isn't how they are assigned, it is what is being assigned (a 2D dask array). Until we can get more confidence that xarray won't "accidentally" compute these, it would be best not to include them as `.coords`.

joleenf:
I am sorry Dave, I meant to put a link for the [available_datasets](https://satpy.readthedocs.io/en/stable/api/satpy.readers.html#satpy.readers.file_handlers.BaseFileHandler.available_datasets) section and did not realize I hadn't done that.  Sorry for the confusion.   I am pretty sure that any file metadata in the mirs reader supersedes metadata read from the yaml.  Perhaps I have missed a basic concept in building the [datasets](https://satpy.readthedocs.io/en/stable/dev_guide/custom_reader.html?highlight=available_datasets#dynamic-dataset-configuration) through the configuration file and dynamic datasets.

djhoese:
I think you based this reader off of the GAASP reader which probably makes some assumptions about where dataset definitions are coming from. I think you are right, but technically the behavior is probably "undefined" when two datasets are yielded with the same DataID. I think you'll either need to keep a "cache" of what DataIDs have already been yielded by the file handler and not re-yield/override it. I think I've used a list/set called `_handled` in other reader's file handlers. Or pass each sub-generator (the 3 methods you call and `yield from` in your `available_datasets`) to the next so you can do more checks *but* it may be more efficient to move the initial `for` loop outside these generators.

Obviously, making it work first is more important.

joleenf:
> I think you based this reader off of the GAASP reader

Correct, I based this on the GAASP reader and it does make assumptions about the location of Metdata, mainly, I don't think it expects any of the metadata to be coming from the yaml.  Currently, that is also how the mirs reader is working.  Initially, I thought I would use the yaml to add metadata that was missing, but then I realized, that could be a tricky thing.  I would then want to make sure that the file metadata would always override metadata from the yaml like units.  I am comfortable with the idea that all the metadata comes from the file and adding things like long_name and standard_name with in the reader.  
I have yet another question about the xarray coords attribute.  If I am using open_dataset, doesn't this assign the coords to the xarray.DataArray just as if I had used assign_coords or even built an xarray.DataArray using a coords attribute?  So, we are not clear from the the [unexpected computation issue](https://github.com/pydata/xarray/issues/3068)?

djhoese:
Yes, it should be adding coords as it (xarray) sees fit. You could `.reset_coords` if you want to force the removal of them. :thinking: Just leave the coords as they are then. We can't know what issues they will cause unless we keep them there so maybe keeping them there is the best choice here.

satpy/readers/mirs.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+def get_coeff_by_sfc(coeff_fn, bt_data, idx):
+    """Read coefficients for specific filename (land or sea)."""
+    sfc_coeff = read_atms_limb_correction_coefficients(coeff_fn)
+    # transpose bt_data for correction
+    bt_data = bt_data.transpose("Channel", "y", "x")
+    c_size = bt_data[idx, :, :].chunks
+    correction = da.map_blocks(apply_atms_limb_correction,
+                               bt_data.values, idx,
```
djhoese, line 158: The `.values` should be removed here, right?

djhoese (approved):
This looks good enough to me. Thanks for the patience on getting things in order and waiting on the aux data download stuff.

PR #1514, opened 2021-01-28
Correct the pdict a_name of agri_l1 reader

zxdawn:
Fix the simple type error.


mraspaud (approved):
LGTM. Please consider adding a test for this in the future.

PR #1519, opened 2021-01-29
Allow to pass pathlike-objects to FSFile

### LINKED ISSUE #1516: FSFile should support any PathLike objects
## Feature Request

When the `FSFile` constructor is passed anything that isn't a string, the `FSFile.__str__` method triggers a `TypeError`:

```python
import pathlib
from fsspec.implementations.local import LocalFileSystem
fsf = FSFile(pathlib.Path("/tmp/abc"), LocalFileSystem())
print(str(fsf))
```

which gives:

```
  File "mwe/mwe002.py", line 5, in <module>
    print(str(fsf))
TypeError: __str__ returned non-string (type PosixPath)
```

This is not a bug because the `FSFile` docstring does state it expects a string, but it would be nice if other pathlike objects could be passed as well.

I would like that the `FSFile` constructor supports [the file system protocol](https://www.python.org/dev/peps/pep-0519/), such that the code above would work as intended.

None reasonably expected.

**Additional context**

It would save me a call to `__fspath__` or `str(...)` prior to creating the `FSFile` objects.

gerritholl:
When constructing an FSFile object, allow to pass not only strings, but
any object meeting the os.PathLike protocol.

 - [x] Closes #1516 
 - [x] Tests added 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/__init__.py
+++ b/satpy/readers/__init__.py
@@ -546,10 +546,14 @@ class FSFile(os.PathLike):
     def __init__(self, file, fs=None):
         """Initialise the FSFile instance.
 
-        *file* can be string, an object implementing the PathLike protocol, or
-            an fsspec.OpenFile instance. In the latter case, the follow argument
-            *fs* has no effect.
-        *fs* can be None or a fsspec filesystem instance.
+        Args:
+            file (str, Pathlike, or OpenFile):
+                String, object implementing the `os.PathLike` protocol, or
+                an `fsspec.OpenFile` instance.  If passed an instance of
+                `fsspec.OpenFile`, the following argument ``fs`` has no
+                effect.
+            fs (fsspec filesystem, optional)
+                Object implementing the fsspec filesystem protocol.
         """
         try:
             self._file = file.path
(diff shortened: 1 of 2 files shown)


gerritholl:
As often, the codecov report makes no sense.  I've only added test code and test coverage has gone down?

satpy/readers/__init__.py
```diff
@@ -546,7 +546,9 @@ class FSFile(os.PathLike):
     def __init__(self, file, fs=None):
         """Initialise the FSFile instance.
 
-        *file* can be string or an fsspec.OpenFile instance. In the latter case, the follow argument *fs* has no effect.
+        *file* can be string, an object implementing the PathLike protocol, or
+            an fsspec.OpenFile instance. In the latter case, the follow argument
+            *fs* has no effect.
```
djhoese, line 551: This should just be switched to an `Args:` list. I also prefer two backticks around literals over the asterisks for bold (I know you didn't do it, but still).

djhoese (changes requested):
Thanks for fixing it. One small docstring change please.

satpy/readers/__init__.py
gerritholl: Fixed.

djhoese (approved):
Thanks!

mraspaud (approved):
LGTM, thanks for adding this functionality

PR #1528, opened 2021-02-02
Add external image to composites docs

### LINKED ISSUE #1527: Add images to documentation
## Feature Request

Satpy is a python library for reading, manipulating, **visualising**, and writing data from remote-sensing earth-observing meteorological satellite instruments.  Yet there is not a single image in the documentation!  There are great examples hidden in notebooks (where images do not always appear) and pull requests (where they do, but which few users will find).  Documentation for software used to visualise satellite data without any images is boring.

Pictures!

nice satellite picture  
(Source: #804)

another one  
(Source: #1163)

and another one  
(Source: #1065)

would be great to have in the docs (just an example).

Documentation will be more fun to read.  More users will read documentation.  Less users will run to slack without reading documentation.  Core developers and their children will be happier.

**Additional context**

I opened #1206 a while ago for a gallery of areas, and was thinking of a gallery of composites, but realised that not only do we have neither, we have no pictures at all.

Pictures take up space but they'd be in the docs just for illustration purposes, so they could be very small (<100 kB each), so space is probably not too much of a problem.  Maybe they can even directly deep-link/include the ones already uploaded to Github as part of PRs :)

gerritholl:
Add an external image to the composite documentation, directly linked from the Github PR where the code to generate the image was introduced.  It works locally, let's see if it works on RTD.

 - [ ] Closes #1527 
 - [x] Fully documented

### DIFF
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -407,6 +407,9 @@ image) for both of the static images::
           min_stretch: [0, 0, 0]
           max_stretch: [255, 255, 255]
 
+.. image:: https://user-images.githubusercontent.com/500246/106626015-8b72c400-6577-11eb-87b7-5bf32e45d5e1.png
+   :width: 600
+
 .. _enhancing-the-images:
 
 Enhancing the images


djhoese:
I'm not sure we should depend on github's in-comment image URLs staying static. We should probably copy those somewhere more static if we can. If I need to host them on some SSEC servers I can. We should put an estimate together to see how much space would be needed. Of course if we are only doing thumbnails (and name them accordingly) then keeping things down to 400-600px wide might not be too bad.

PR #1534, opened 2021-02-03
Clarify usage of config 'config_path' option

### LINKED ISSUE #1530: Improve documentation/handling of string input for config_path
## Feature Request

When setting up at runtime the configuration parameter `config_path`, this must be defined as a list containing strings:
```python
satpy.config.set(config_path=["/my/new/config/path"])
```
The other parameter `cache_dir` and `data_dir`, in contrary, must be a string. Although this difference is described in https://github.com/pytroll/satpy/blob/master/doc/source/config.rst, it is easy to oversee that specific passage if the page is not read carefully.
Moreover, if the user inputs a string, the code executes without complaints, but the config path will not be handled properly and Satpy will fail to find the configuration files. It will not be very easy for a user to find out why this is happening.

I propose to highlight the fact that  `config_path` must be a list more clearly in the documentation, possibly adding an explicit code snipped for this specific case. There is now only an example for `cache_dir`, which might induce a reader to assume that the other parameters all work the same.
Additionally, the code could be adapted to handle a string and emit a warning if the user passes a string instead of a list, or throw an exception with a meaningful message.

None, just clearer configuration setup instructions.

djhoese:
This addresses the confusion in #1530. I've added more documentation on setting the config_path and also added a "safe" get utility function which should 

 - [x] Closes #1530 
 - [x] Tests added 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented

### DIFF
--- a/doc/source/config.rst
+++ b/doc/source/config.rst
@@ -123,6 +123,8 @@ Note that this value must be a list. In Python, this could be set by doing:
 
 As mentioned above, if setting an environment variable then it must be a
 colon-separated string and must be set **before** calling/importing Satpy.
+If the environment variable is a single path it will be converted to a list
+when Satpy is imported.
 
 .. code-block:: bash
 
--- a/doc/source/config.rst
+++ b/doc/source/config.rst
@@ -103,9 +103,14 @@ Component Configuration Path
 * **Default**: ``[]``
 
 Base directory, or directories, where Satpy component YAML configuration files
-are stored. Directories provided will typically include subdirectories for
-each component that is being configured (ex. ``readers``, ``writers``, etc).
-This option replaces the legacy ``PPP_CONFIG_DIR`` environment variable.
+are stored. Satpy expects configuration files for specific component types to
+be in appropriate subdirectories (ex. ``readers``, ``writers``, etc), but
+these subdirectories should not be included in the ``config_path``.
+For example, if you have custom composites configured in
+``/my/config/dir/etc/composites/visir.yaml``, then ``config_path`` should
+include ``/my/config/dir/etc`` for Satpy to find this configuration file
+when searching for composites. This option replaces the legacy
+``PPP_CONFIG_DIR`` environment variable.
 
 Note that this value must be a list. In Python, this could be set by doing:
 
--- a/doc/source/config.rst
+++ b/doc/source/config.rst
@@ -107,21 +107,7 @@ are stored. Directories provided will typically include subdirectories for
 each component that is being configured (ex. `readers`, `writers`, etc).
 This option replaces the legacy ``PPP_CONFIG_DIR`` environment variable.
 
-Satpy will always include the builtin configuration files that it
-is distributed with regardless of this setting. When specified as an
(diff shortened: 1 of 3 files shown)


doc/source/config.rst
```diff
@@ -115,6 +115,21 @@ configuration files, they are merged in reverse order. This means "base"
 configuration paths should be at the end of the list and custom/user paths
 should be at the beginning of the list.
 
+Note that this value must be a list. In Python, this could be set by doing:
+
+.. code-block:: python
+
+    satpy.config.set(config_path=['/path/custom1', '/path/custom2'])
+
+As mentioned above, if setting an environment variable then it must be a
```
ameraner, line 124: Thank you, this is already much clearer!
To keep it a bit shorter and avoid repeating the informations, I would propose to delete the "As mentioned above, " here, and rather move the paragraph above to after the code snippets, like:
```
If setting an environment variable then it must be a colon-separated string and must be set **before** calling/importing Satpy. If the environment variable is a single path it will be converted to a list
when Satpy is imported.

.. code-block:: bash
    export SATPY_CONFIG_PATH="/path/custom1:/path/custom2"
    
Satpy will always include the builtin configuration files that it
is distributed with regardless of this setting. When a component supports merging of
configuration files, they are merged in reverse order. This means "base"
configuration paths should be at the end of the list and custom/user paths
should be at the beginning of the list.
```

ameraner (commented):
Thanks for taking care of this so fast, I think this is already much better! Just a minor proposal to make the text a bit more linear.

doc/source/config.rst
djhoese: Sounds good and makes sense. I'll edit that after I finish up some other local changes. Thanks.

djhoese: Ok I just pushed a commit with these changes. Do you think the `Note that this value must be a list` part at the bottom is still useful? Is it in a good place?

ameraner: Yes I think it's definitely still useful, I would even move it one or two paragraphs higher (before `If setting an environment variable...`), so that you first show the two examples, and then at the end add the note on the order of the paths (`Satpy will always include...`).

djhoese: Done, but now I'm wondering if the sentence:

> Directories provided will typically include subdirectories for
each component that is being configured (ex. ``readers``, ``writers``, etc).

Is confusing. It sounds like "you should include `/readers` when you specify this, but I'm trying to say the opposite.

gerritholl: Probably best clarified with an example, also to illustrate that the `etc/` should be included with `SATPY_CONFIG_PATH`; that may not be immediately obvious to all.

ameraner: Indeed it is a bit confusing. It's also probably good to explicitly give the list of subdirectory names that will be automatically detected and understood by Satpy if contained in one of the paths of `config_path`.

djhoese: Ok I updated it one more time. Let me know what you guys think.

ameraner: Looks good, thank you!

ameraner (approved):

PR #1539, opened 2021-02-08
Fix args of bucket_sum and bucket_avg resampler

ameraner:
This PR fixes the `bucket_sum` and `bucket_avg` resampler in Satpy, adapting for the changes in pyresample's PR https://github.com/pytroll/pyresample/pull/319 .

On the side, it also fixes a small typo in the readers docs.

 - [x] Tests added 
 - [x] Passes ``flake8 satpy`` 
 - [x] Fully documented


ameraner:
The changes in Pyresample that call for this PR have not been released yet. Should we wait to merge this PR until the new Pyresample release comes out, or how do we deal with these situations?

djhoese (changes requested):
Could you accept the `mask_all_nans` kwarg and if provided raise an exception to update pyresample. Or you could map it to the new kwarg and issue a warning. I'll leave that up to you.

ameraner:
Hi @djhoese thanks for the review, I added a DeprecationWarning for the old kwarg. 

There is still a bit of an issue with the Satpy/Pyresample versions here due to the renaming of the arguments: 
The current Satpy master works with the latest Pyresample release, but not with the current Pyresample master.
After this PR, Satpy will work with the current Pyresample master, but not the latest Pyresample release.

So if we want Satpy to work with the current and older Pyresample releases, we would need to catch the TypeError (or inspect the function arguments) and pass the old kwarg with warnings. Otherwise we wait to merge this PR until the next Pyresample release and put a requirement. I suppose the first is preferred?

djhoese:
Yeah not great is it. I was going to say let's leave it because I'm not a fan of inspecting functions for their arguments *BUT* what if you checked the version of pyresample at the top of the module? Like `_pr_use_skipna = pyresample.__version__ >= 1.18.0`? And then use that near the warning. If True then do the warning and reassign `skipna = mask_all_nan` and if False then only do the warning?

ameraner:
@djhoese thanks for the hint, that sounds like a good idea! I tried to implement it in the last commit. To compare the versions I had to import `packaging`, is that fine or is there a better way? Hope the code didn't get too ugly now. Are tests required for these changes?

satpy/resample.py
```diff
@@ -1046,6 +1050,23 @@ def compute(self, data, expand=True, **kwargs):
         return update_resampled_coords(data, new_data, target_geo_def)
 
 
+def _get_arg_to_pass_for_skipna_handling(skipna, **kwargs):
+    """Determine if skipna can be passed to the compute functions for the average and sum bucket resampler."""
+    # FIXME this can be removed once Pyresample 1.18.0 is a Satpy requirement
+    if 'mask_all_nan' in kwargs:
+        if PR_USE_SKIPNA:
+            warnings.warn('Argument mask_all_nan is deprecated. Please use skipna for missing values handling. '
+                          'Continuing with default skipna=True, if not provided differently.', DeprecationWarning)
```
djhoese, line 1059: You could add a line here for `skipna = kwargs['mask_all_nan']` followed by `return `{'skipna': skipna}`, right?

Then in the `else` below this do `return kwargs`...I think. Basically I think you can combine these if statements.

Additionally, I think you need to update the `kwargs` dictionary instead of only returning a specific dictionary. Otherwise you lose any additional arguments, right?

djhoese: Maybe the arguments to this function should just be `**kwargs`?

djhoese (changes requested):
I think the way you did it is fine. I personally think I would have done the "more wrong" way of just string comparisons.

I had one issue with your new function and commented below.

satpy/resample.py
ameraner: Thanks for the review!
I rearranged a bit the logic and now I'm handling the kwargs only, I think it makes sense too.
I also rearranged the if statements a bit and avoided one `else` now. I think this cannot be reduced further, if we want to issue the warnings only if the `mask_all_nan` was passed by the user. I also added tests for this logic.

satpy/resample.py
```diff
@@ -1161,22 +1199,29 @@ class BucketSum(BucketResamplerBase):
     ----------
     fill_value : float (default: np.nan)
         Fill value for missing data
-    mask_all_nans : boolean (default: False)
-        Mask all locations with all-NaN values
+    skipna : boolean (default: True)
+        If True, skips NaN values for the sum calculation
+                    (similarly to Numpy's `nansum`). Buckets containing only NaN are set to zero.
```
djhoese, line 1204: Maybe it is just github but the spacing on this docstring seems really weird. Everything should be aligned with the `If True` in the first sentence.

satpy/resample.py
```diff
@@ -1046,6 +1050,26 @@ def compute(self, data, expand=True, **kwargs):
...
+            warnings.warn('Argument mask_all_nan is deprecated. Please use skipna for missing values handling. '
+                          'Continuing with default skipna=True, if not provided differently.', DeprecationWarning)
+            kwargs.pop('mask_all_nan')
+    else:
+        if 'mask_all_nan' in kwargs:
+            warnings.warn('Argument mask_all_nan is deprecated.'
+                          'Please update Pyresample and use skipna for missing values handling.',
+                          DeprecationWarning)
+        kwargs.setdefault('mask_all_nan', False)
+        kwargs.pop('skipna')
```
djhoese, line 1068: Is skipna always provided? Otherwise, you might need a `, None)` at the end of the pop. Otherwise I think it raises a KeyError if it is missing.

djhoese (changes requested):
Looks like you have some linting/website issues which I've comment on, but otherwise I think the logic looks good. Thanks for looking it over again.

satpy/resample.py
ameraner: Yes, it's always provided, so  it can stay like this.

satpy/resample.py
ameraner: Yep, I removed the indents and now the website builds.

djhoese (approved):
LGTM but I also never use the bucket resampler. If @pnuu could review before a merge that would be appreciated.

pnuu (approved):
LGTM.

PR #1542, opened 2021-02-09
Fix compression not being applied in awips_tiled writer

djhoese:
User discovered that the files weren't being compressed. I had assumed setting the global encoding of `zlib=True` was good enough, but it needs to actually be set on every variable.

 - [x] Tests added 
 - [x] Passes ``flake8 satpy``


djhoese:
Tests pass. Coverage is...misinformed. Not sure what's going on there but it isn't my fault :wink: 

Merging...

PR #1553, opened 2021-02-19
Remove unnecessary use of numpy masked arrays in awips_tiled writer

djhoese:
This is a small wasteful piece of code I found when debugging something in my own software. The old version of the `awips_tiled` writer used numpy masked arrays, but I removed all usage of that when I switched to dask and xarray. At least I don't intend to be benefiting from it. This PR removes the small usage I had of `np.ma.*`.


djhoese:
Due to the nature of this PR I'm not going to wait for a review. If you don't approve of that, please talk to me on slack.

PR #1559, opened 2021-02-22
Fix geotiff writer FAQ link

gerritholl:
Fix geotiff writer FAQ link to point to Satpy FAQ rather than to xarray
FAQ.

Add a gitignore for automatically generated RST sources by sphinx-apidoc.

 - [x] Fully documented

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,4 @@ doc/source/_build/*
 # this should be generated automatically when installed
 satpy/version.py
 doc/api/*.rst
+doc/source/api/*.rst
--- a/.gitignore
+++ b/.gitignore
@@ -73,5 +73,4 @@ doc/source/_build/*
 # setuptools_scm files
 # this should be generated automatically when installed
 satpy/version.py
-doc/api/*.rst
 doc/source/api/*.rst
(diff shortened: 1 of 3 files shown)


djhoese (changes requested):
That `.gitignore` shouldn't be needed. It should be covered by the `<root>/.gitignore`.

djhoese:
https://github.com/pytroll/satpy/blob/1aa9c5f6a501319c60009853e0c7eec15cffcbfe/.gitignore#L76

djhoese:
Oh! It needs to be `doc/source/api/*.rst`.

gerritholl:
> That `.gitignore` shouldn't be needed. It should be covered by the `<root>/.gitignore`.

Without it I get lots of "untracked files" notifications".  The root `.gitignore` does not include any `doc/source/api/*.rst`, but I'm happy to add one if you prefer it there rather than here.

djhoese:
Could you remove the line above yours in the `.gitignore`. That was supposed to be the one catching the api files.

djhoese (approved):
Thanks!

PR #1560, opened 2021-02-22
Allow custom dataset names in 'generic_image' reader and fix nodata handling

ch-k:
This PR is about two issues. Sorry about mixing it, but it concerns the same file.

1. 

At the moment it is not possible to use the GenericImageFileHandler for reader configs that define multiple dataset names. This is an example:
```
reader:
    name: world_generic_image
    description: generic image reader
    sensors: [world-images]
    default_channels: [image_goes16, image_goes17]

datasets:
  image_goes16:
    name: image_goes16
    file_type: graphic_goes16
  image_goes17:
    name: image_goes17
    file_type: graphic_goes17
  image_zds:
    name: image_zds
    file_type: graphic_zds
  image_iodc:
    name: image_iodc
    file_type: graphic_iodc
  image_himawari8:
    name: image_himawari8
    file_type: graphic_himawari8

file_types:
    graphic_goes16:
        file_reader: !!python/name:satpy.readers.generic_image.GenericImageFileHandler
        file_patterns:
         - 'goes16_{filename}{start_time:%Y%m%d%H%M}.tif'
         - 'goes16_{filename}.tif'
    graphic_goes17:
        file_reader: !!python/name:satpy.readers.generic_image.GenericImageFileHandler
        file_patterns:
         - 'goes17_{filename}{start_time:%Y%m%d%H%M}.tif'
         - 'goes17_{filename}.tif'
    graphic_zds:
        file_reader: !!python/name:satpy.readers.generic_image.GenericImageFileHandler
        file_patterns:
         - 'ZDS_{filename}{start_time:%Y%m%d%H%M}.tif'
         - 'ZDS_{filename}.tif'
    graphic_iodc:
        file_reader: !!python/name:satpy.readers.generic_image.GenericImageFileHandler
        file_patterns:
         - 'IODC_{filename}{start_time:%Y%m%d%H%M}.tif'
         - 'IODC_{filename}.tif'
    graphic_himawari8:
        file_reader: !!python/name:satpy.readers.generic_image.GenericImageFileHandler
        file_patterns:
         - 'himawari8_{filename}{start_time:%Y%m%d%H%M}.tif'
         - 'himawari8_{filename}.tif'

```
That doesn't work because a hardcoded name `image` (https://github.com/pytroll/satpy/blob/master/satpy/readers/generic_image.py#L85) is used to _set_ the data and the name used in the reader config is used to _get_ the data. Therefore only one dataset with name `image` is possible. Otherwise the reader complains that the dataset is not available.

2.
The fill value of GeoTIFF files is not used to mask the data. With this PR, `fill value` pixels are set to nan as it is already the case when there are alpha channels present.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 


mraspaud:
Could you explain the motivation behind this change?

ch-k:
> Could you explain the motivation behind this change?

Sorry, should be a draft only.  My primary objectiv was to create the DWD world composite with a new trollflow2/satpy chain. I have geotiff files of all satellites on disk and segment gatherer should be used to publish an event when everything is available. Then trollflow2/satpy should use the generic image reader to read all images together into the same scene object. I've created a custom reader.yml there I specified each satellites' image as channel. Afterwards I'll try to create a new compositor to stich the images. To read all images together, this patch is necessary. I can provide the reader.yml when I'm back in office.

djhoese:
Ah I see, the main problem is:

```
self.file_content['image'] = data
```

This doesn't use the key provided in the YAML at all. Maybe it would be better to copy the `file_content['image']` DataArray and reassign the `.attrs['name']` to the provided key's name. Maybe there should be a key in the `info` (that comes from the YAML) that specifies the `file_key`. This could default to `image` and would provide flexibility in the future if we wanted to make it possible to load other things besides the image (the alpha channel maybe, I don't know).

Just some brainstorming.

ch-k:
Here is an example of the reader config that I'd like to use to read the images into the same scene (GOES16 + 17 is only the first step):
```yaml
reader:
    name: world_generic_image
    description: generic image reader
    sensors: [images]
    default_channels: [image_goes16, image_goes17]

datasets:
  image_goes16:
    name: image_goes16
    file_type: graphic_goes16
  image_goes17:
    name: image_goes17
    file_type: graphic_goes17

file_types:
    graphic_goes16:
        file_reader: !!python/name:satpy.readers.generic_image.GenericImageFileHandler
        file_patterns:
         - 'goes16_{filename}{start_time:%Y%m%d%H%M}.tif'
         - 'goes16_{filename}.tif'
    graphic_goes17:
        file_reader: !!python/name:satpy.readers.generic_image.GenericImageFileHandler
        file_patterns:
         - 'goes17_{filename}{start_time:%Y%m%d%H%M}.tif'
         - 'goes17_{filename}.tif'
```

ch-k:
> Ah I see, the main problem is:
> 
> ```
> self.file_content['image'] = data
> ```
> 
> This doesn't use the key provided in the YAML at all. Maybe it would be better to copy the `file_content['image']` DataArray and reassign the `.attrs['name']` to the provided key's name. Maybe there should be a key in the `info` (that comes from the YAML) that specifies the `file_key`. This could default to `image` and would provide flexibility in the future if we wanted to make it possible to load other things besides the image (the alpha channel maybe, I don't know).
> 
> Just some brainstorming.
At first, I tried to change the static 'image' to something like
```python
    def read(self):
        """Read the image."""
        dataset = rasterio.open(self.finfo['filename'])
...
        self.file_content[key['name']] = data
```
... until I noticed that it's not available in this stage. I've thought about moving the whole reading stuff to `get_dataset` but I'm unsure if that's reasonable.

djhoese:
Yeah hard to say. @pnuu I think wrote this reader originally and mind have ideas for what can and can't be done on `__init__` of the file handler. I don't think my idea or your idea would be bad. No reason we couldn't do both (from the little I know about this reader). We definitely want to support your use case as I think this was kind of the intention of making this functionality a reader in the first place.

pnuu:
Sorry, this might not be very helpful in respect to the changes needed to do this purely in Satpy.

> My primary objectiv was to create the DWD world composite with a new trollflow2/satpy chain. I have geotiff files of all satellites on disk and segment gatherer should be used to publish an event when everything is available.

This sounds very much like `create_global_mosaic.py` we *had* in `pytroll-collectors` still few years back, but didn't want to keep it working with the current Satpy and decided to remove it all together. If you want to have a look, see [this](https://github.com/pytroll/pytroll-collectors/tree/v0.8.4/bin) and [this](https://github.com/pytroll/pytroll-collectors/blob/v0.8.4/examples/global_mosaic.yaml_template) and [this](https://github.com/pytroll/pytroll-collectors/blob/v0.8.4/pytroll_collectors/global_mosaic.py).

It collected the images similar to segment gatherer, then cut the input images very naïvely at the GEO satellites' mid-point longitudes and put the images together. It required the input images to be in latlon grid, didn't do any blending, and had some other simplifications.

My current solution is to rely on putting the images (in what ever projection they are generated in) to WMS and let the client software do the requests so that the images are in correct projection, and let it also handle (possible) blending, arrangement and everything else.

ch-k:
> Sorry, this might not be very helpful in respect to the changes needed to do this purely in Satpy.
> 
> > My primary objectiv was to create the DWD world composite with a new trollflow2/satpy chain. I have geotiff files of all satellites on disk and segment gatherer should be used to publish an event when everything is available.
> 
> This sounds very much like `create_global_mosaic.py` we _had_ in `pytroll-collectors` still few years back, but didn't want to keep it working with the current Satpy and decided to remove it all together. If you want to have a look, see [this](https://github.com/pytroll/pytroll-collectors/tree/v0.8.4/bin) and [this](https://github.com/pytroll/pytroll-collectors/blob/v0.8.4/examples/global_mosaic.yaml_template) and [this](https://github.com/pytroll/pytroll-collectors/blob/v0.8.4/pytroll_collectors/global_mosaic.py).
> 
> It collected the images similar to segment gatherer, then cut the input images very naïvely at the GEO satellites' mid-point longitudes and put the images together. It required the input images to be in latlon grid, didn't do any blending, and had some other simplifications.
> 
> My current solution is to rely on putting the images (in what ever projection they are generated in) to WMS and let the client software do the requests so that the images are in correct projection, and let it also handle (possible) blending, arrangement and everything else.

@pnuu I remember that we discussed about it in the past. Currently we still have a similar solution running. It uses some reading and writing functions provided by good old mpop. It has to be migrated. I hope to use as much satpy/trollflow2 functionality as possible. It should be parametrized as all the other stuff. The core mosaic creation would be done in a new satpy compositor. Geoserver is not an option for our use case.

pnuu:
So would the use be like this?
```python
fnames = ["goes16_true_color_202102241200.tif", "goes17_true_color_202102241200.tif"]
scn = Scene(reader="world_generic_reader", filenames=fnames)
scn.load(["image_goes16", "image_goes17"])
```
And another question: how are you planning to handle resampling so that the intertwined pixels do not cause weird effects where the images overlap? The plain nearest neighbour resampling will just blindly use the closest pixel to each target location, so the illumination, parallax and (very slightly) calibration will cause artifacts on the overlapping areas. Maybe `MultiScene` (several separate scenes) and `MultiScene.blend()` would handle this readily in a nice way?

ch-k:
> So would the use be like this?
> 
> ```python
> fnames = ["goes16_true_color_202102241200.tif", "goes17_true_color_202102241200.tif"]
> scn = Scene(reader="world_generic_reader", filenames=fnames)
> scn.load(["image_goes16", "image_goes17"])
> ```
Yes, exactly. But we start with single channel images..
> 
> And another question: how are you planning to handle resampling so that the intertwined pixels do not cause weird effects where the images overlap? The plain nearest neighbour resampling will just blindly use the closest pixel to each target location, so the illumination, parallax and (very slightly) calibration will cause artifacts on the overlapping areas. Maybe `MultiScene` (several separate scenes) and `MultiScene.blend()` would handle this readily in a nice way?
In our use case, no resampling is necessary. The input images are already resampled to the target area:
goes16_IR_1035_ir_wwwWcm40km_202102181630
goes17_IR_1035_ir_wwwWcm40km_202102181530

ch-k:
> Maybe `MultiScene` (several separate scenes) and `MultiScene.blend()` would handle this readily in a nice way?
Does trollflow2 work with `MultiScene`?

pnuu:
> Does trollflow2 work with `MultiScene`?

No, not at the moment. Creating a plugin that supports the usage of it should be pretty simple.

ch-k:
I've added some code to handle the fill_value when reading GeoTIFFs. 
Sorry for mixing it with the other issue. It's difficult to manage changes in different branches when I need all of them to get my feature running (world comp). 

> @pnuu I think wrote this reader originally and mind have ideas for what can and can't be done on __init__ of the file handler. I don't think my idea or your idea would be bad. No reason we couldn't do both (from the little I know about this reader). We definitely want to support your use case as I think this was kind of the intention of making this functionality a reader in the first place.

@pnuu: What do think? I see my PR more as a fix than an enhancement.

satpy/readers/generic_image.py
```diff
@@ -117,4 +117,9 @@ def mask_image_data(data):
                                 for i in range(data.shape[0])])
         data.data = masked_data
         data = data.sel(bands=BANDS[data.bands.size - 1])
+    elif hasattr(data, 'nodatavals') and data.nodatavals:
+        data = data.astype(np.float64)
```
djhoese, line 121: Do you have to convert to float here? If we can keep the initial data type and use a fill value dependent on that data type then it may work for more situations. What if you copied the nodatavals to `.attrs['_FillValue']` for integer types, but replaced them with `np.nan` for floats? This would match what we do in other readers for "category" products.

satpy/readers/generic_image.py
```diff
@@ -102,8 +102,8 @@ def end_time(self):
 
     def get_dataset(self, key, info):
         """Get a dataset from the file."""
-        logger.debug("Reading %s.", key)
-        return self.file_content[key['name']]
+        logger.debug("Reading 'image' (ignoring provided key %s).", key)
+        return self.file_content['image']
```
djhoese, line 106: How about you either:

1. Copy `self.file_content['image']` (with `.copy()` method) and update the attributes with the `info` dictionary. Like `img_copy.attrs.update(info)`. I think this should update the `name` and any other `DataID` information that someone might include in the YAML.
2. Keep the old way of doing `key['name']` but update the `read` method to write the file content to the right `file_content` key.

Option 1 is easiest. Option 2 provides the most flexibility in the future. @pnuu let us know if there is a reason not to move `.read` to the `get_dataset` call as that also seems like it could help here.

djhoese (changes requested):
Couple suggestions to make this more flexible...I think.

satpy/readers/generic_image.py
ch-k: I'm unsure what's better option. Unfortunately, NaN is only possible with floats and I have the feeling that most of the compositors only work with NaNs as masks. It seems that only ColormapCompositor cares about "_FillValue".  That's why I've done the float conversation as it is already the case if alpha channels are present.
Maybe I should introduce a new reader parameter and let the user decide how to handle nodatavals?

satpy/readers/generic_image.py
ch-k: Let's wait for @pnuu 's input and I'll change it accordingly. Option 2 seems nicer to me.

pnuu: The `'image'` is there just to make things work with the existing way Satpy is designed to load data with a named dataset. In an image, there's only one dataset and `scn.load()` should probably just load the image. But yeah, when loading data with multiple readers that'd not work.

I think option 2 is the better way.

pnuu:
Could you add a description )to the description field) of what this PR really is about? Looking at the the code, I'm also unsure about the title of the PR. Maybe clarify the title too?

ch-k:
> 
> 
> Could you add a description )to the description field) of what this PR really is about? Looking at the the code, I'm also unsure about the title of the PR. Maybe clarify the title too?

I've added a description text (sorry for that). But I don't have a better title..

djhoese:
I changed the title...maybe it's better?

satpy/readers/generic_image.py
ch-k: I'm going to remove the `read()` completely as it is only used in the ctor and move the reading stuff to `get_dataset()`. Then there's also no need to keep an attribute `self.file_content` any longer. The "How-To for custom readers" (https://satpy.readthedocs.io/en/stable/dev_guide/custom_reader.html) only demands implementing `get_dataset()`. 
And I couldn't find any other readers that use `GenericImageFileHandler` as base class.
Keeping `self.file_content` seems only necessary for caching reasons. But I think there are other mechanisms in satpy that care about this, right?
Okay?

djhoese: Regarding moving the functionality: sounds good as long as the code is clean.

Regarding file_content, that is likely just a common practice used from other readers. It also makes writing tests easier, at least in other readers, because complex NetCDF/HDF files can be described by a dictionary and put in `file_content`. For this reader, maybe that isn't necessary. As long as the tests can be made to work with whatever you do then :+1:

ch-k: I've decided to lease reading in ctor as it is because otherwise `get_area_def()` needs special treatment. The area is set during reading. Although it seems that  `get_area_def()`  is called after `get_dataset()` and therefore area should be already available in my tests - it don't know if it's the case in all situations..
The latest commit introduces a class attribute `dataset_name` that is set to default `image` in the read function. In derived fake readers you can provide own file_content and leave `ds_name` empty to work with multiple datasets - if somebody really wants to do this..

satpy/readers/generic_image.py
ch-k: I just stumbled across the following paragraph in the satpy docs (https://satpy.readthedocs.io/en/latest/dev_guide/custom_reader.html)

> Single precision floats (np.float32) is a good compromise, as it has 23 significant bits (mantissa) and can thus represent 16 bit integers exactly, as well as keeping the memory footprint half of a double precision float.
> One commonly used method in readers is xarray.DataArray.where() (to mask invalid data) which can be coercing the data to np.float64. To ensure for example that integer data is coerced to np.float32 when xarray.DataArray.where() is used, you can do ..
So converting to floats to mask data with NaN seems a common practice. As noted in the docs, float32 would be better. Conversion to floats is already done few lines above when alpha channels are handled.

satpy/readers/generic_image.py
djhoese: FYI there was a semi-recent change in the base reader that changed the order of when the dataset was loaded and when the area was loaded. It used to be that the area was always loaded first (if I remember correctly), but now it is loaded after.

satpy/readers/generic_image.py
djhoese: I think that is more specifically talking about integer data that is in the file as integers and needs to be scaled (ex. `data * scale_factor + add_offset`) to floats, but also has a fill value. In that case you'd need to do the `.where` call to mask out the floats with some other value. In Satpy that is NaN for floating point data and any `_FillValue` for integer data. In older versions of numpy (I think newer versions have fixed this) if you do `.where` with `np.nan` it would cast 32-bit floats to 64-bit floats. Now it knows that you really mean 32-bit float -> 32-bit float NaN. For integers, they would always be cast to 64-bit floats so the note in the documentation is trying to say "there is no reason for that so force 32-bit float conversion".

The note should probably be updated to refer to integer fields that are meant to stay integers like cloud type products or in this case 8-bit unsigned integer RGB images. If an RGB is being read in by Satpy and is provided as floating point data (instead of as an integer) then it should probably be normalized to a 0-1 range as that is the assumption by the trollimage package when saving images.

djhoese: Things like scale factor and add offset in a geotiff are not standard so I don't think we should add any support for that. Support for that should be added in a separate reader (a custom reader created by the user).

Geotiffs, at least those created by GDAL/rasterio, have `nodata` values set that should be masked. That should be considered the same as `_FillValue` in my opinion. In the future I'd like this reader to use the `rioxarray` library if possible to read GDAL datasets as xarray DataArrays.

As for not knowing if something is a category, assume that if it is an integer that there is a `nodata` value that can be used as `_FillValue`. If it integer and has an alpha band or has no `nodata` value specified, then don't set `_FillValue` and leave it as is.

In other readers, things like `units` are just metadata to fill in the space of missing information in the file. In those cases, the reader is specific to a particular implementation of a data file (VIIRS SDR in the HDF5 format versus generic HDF5 files) or are representing a standard (CF compliant NetCDF files). This generic image reader should try to use standard geotiff features.

pnuu: My original use for which I wrote this reader was to read any image as an image and use it as a background for `BackgroundCompositor` and to make it possible to resample them to a specific area. So in a way, I never considered it for anything else than categorical data.

ch-k: > Yeah, sure, but the question is what is the point of this pull request and how much work do you want to include in it? I am totally fine with the python code allowing for all sorts of flexibility as long as it is documented and that the defaults work. This is the "generic" image reader so none of this extra functionality should be required to get _something_ out of the reader. Adding options to a static list of datasets in the YAML sounds fine.
> 
> I think this reader should fix the dataset naming so it can be overridden in the YAML. If it includes some minimal nodata handling for geotiffs, then great. Anything beyond that should probably go in a separate PR, right?

My latest commit should address this points. There's now an option to control the handling of nodata. Default is to do nothing with the data but copying the nodata value (first element) to dataset.attrs['_FillValue']. 
Beside that you can add the option "nodata_handling" to the readers yaml dataset sections. If it is set to "nan_mask", the data is converted to float32 and masked with np.nan.

>  Anything beyond that should probably go in a separate PR, right?

Absolutely right. Anything else should be in another PR. But I need this together do get my stuff working and it's hard to handle multiple PRs addressing the same file. 

If you have any proposals for better naming of this option - you're welcome :-)
Hope this commit fulfills the demands.

ch-k: > My original use for which I wrote this reader was to read any image as an image and use it as a background for `BackgroundCompositor` and to make it possible to resample them to a specific area. So in a way, I never considered it for anything else than categorical data.

It's good to have such reader. Everything starts small. I hope that your needs are also covered with this modifications and the code still doesn't get too complex. These are only small changes and it may be more confusing to implement this in a separate reader.

satpy/readers/generic_image.py
```diff
@@ -20,6 +20,11 @@
 
 Returns a dataset without calibration.  Includes coordinates if
 available in the file (eg. geotiff).
+If nodata values are present (and rasterio is able to read them), it
+will be preserved as attribute ``_FillValue`` in the returned dataset.
+In case that nodata values should be used to mask pixels (that have
+equal values) with np.nan, it has to be enabled in the reader yaml
+file (key ``nodata_handling`` per dataset with value ``nan_mask``).
```
djhoese, line 27:
```suggestion
file (key ``nodata_handling`` per dataset with value ``"nan_mask"``).
```

satpy/readers/generic_image.py
```diff
@@ -75,14 +84,9 @@ def read(self):
...
 
-        # Mask data if alpha channel is present
-        try:
-            data = mask_image_data(data)
-        except ValueError as err:
-            logger.warning(err)
-
         data.attrs = attrs
-        self.file_content['image'] = data
+        self.dataset_name = 'image'
```
djhoese, line 88: Doesn't this mean that this will *always* be set to `"image"` which means the `ds_name = self.dataset_name if self.dataset_name else key['name']` on line 109 will always be `"image"`?

satpy/readers/generic_image.py
```diff
@@ -102,12 +106,27 @@ def end_time(self):
...
+            data = mask_image_data(data, info)
+        except ValueError as err:
+            logger.warning(err)
+
+        return data
 
 
-def mask_image_data(data):
-    """Mask image data if alpha channel is present."""
+def mask_image_data(data, info=None):
```
djhoese, line 122: When is `info` ever not provided?

satpy/readers/generic_image.py
```diff
@@ -117,4 +136,24 @@ def mask_image_data(data):
                                 for i in range(data.shape[0])])
         data.data = masked_data
         data = data.sel(bands=BANDS[data.bands.size - 1])
+    elif hasattr(data, 'nodatavals') and data.nodatavals:
+        try:
+            nodata_handling = info['nodata_handling'] if info else NODATA_HANDLING_FILLVALUE
+        except KeyError:
+            nodata_handling = NODATA_HANDLING_FILLVALUE
```
djhoese, line 143: Assuming `info` is never `None` as it is always passed, then this block could be simplified to:

```suggestion
        nodata_handling = info.get('nodata_handling', NODATA_HANDLING_FILLVALUE)
```

satpy/readers/generic_image.py
```diff
@@ -117,4 +136,24 @@ def mask_image_data(data):
...
+                                    for i, nodataval in enumerate(data.nodatavals)])
+            data.data = masked_data
+            data.attrs['_FillValue'] = np.nan
+        elif nodata_handling == NODATA_HANDLING_FILLVALUE:
+            # keep data as it is but set _FillValue attribute to provided
+            # nodatavalue (first one as it has to be the same for all bands at least
+            # in GeoTiff, see GDAL gtiff driver documentation)
+            fill_value = data.nodatavals[0]
+            if np.issubdtype(data.dtype, np.integer):
+                fill_value = int(fill_value)
```
djhoese, line 157: Is the nodatavals ever not an integer? What case is this if statement handling? Mostly just curious if it is needed and could be removed.

satpy/tests/reader_tests/test_generic_image.py
```diff
@@ -154,6 +170,18 @@ def test_geotiff_scene(self):
...
+        scn.load(['image'])
+        self.assertEqual(scn['image'].shape, (1, self.y_size, self.x_size))
+        self.assertEqual(np.sum(scn['image'].data[0][:10, :10].compute()), 0)
+
+        fname = os.path.join(self.base_dir, 'test_l_nan_nofillvalue.tif')
+        scn = Scene(reader='generic_image', filenames=[fname])
+        scn.load(['image'])
+        self.assertEqual(scn['image'].shape, (1, self.y_size, self.x_size))
+        self.assertTrue(np.all(np.isnan(scn['image'].data[0][:10, :10].compute())))
+
```
djhoese, line 184: Would you mind separating these out into their own test functions. I think we've generally switched to not combining a ton of test cases into one function whenever possible. It makes it easier to debug.

djhoese (changes requested):
Looks really good. Thanks for the work. I have a couple suggestions and changes I'd like to talk about that I've made comments about.

satpy/readers/generic_image.py
ch-k: I don't know. This function is used in `test_generic_image.py`. That's why I fear it is part of the public interface and used by somebody somewhere..

satpy/readers/generic_image.py
ch-k: Yes, would be much nicer. If we consider `mask_image_data` as private function and info as not none, I can change this.

satpy/readers/generic_image.py
ch-k: Yes, seems that rasterio returns this as float - at least in my setup (nodata value 0 was returns as `[0.0,]`).

djhoese: Ok. The only other issue I could see this having is if `int(fill_value)` is producing a 64-bit integer and is used in later numpy/xarray operations if it ends up upcasting the array data to 64-bit. I don't think this is true, but want to mention it here in case you run into it in your future changes.

satpy/tests/reader_tests/test_generic_image.py
ch-k: Done

satpy/readers/generic_image.py
pnuu: Whet Dave said. And indeed, this ought to be private.

satpy/readers/generic_image.py
ch-k: Yes, absolutetly. This should just address the testing think discussed before:

> Regarding file_content, that is likely just a common practice used from other readers. It also makes writing tests easier, at least in other readers, because complex NetCDF/HDF files can be described by a dictionary and put in file_content. For this reader, maybe that isn't necessary

Someone can inherit this reader and override the ctor() and/or read() to provide own test data.

djhoese: But for your use case, where you want the YAML to configure a new dataset name, this pull request no longer handles that case right? This `self.dataset_name` is always set to `'image'` and the `get_dataset` method never actually uses `key['name']` to overwrite the `.attrs['name']`, right? Actually it looks like `get_dataset` never uses the provided `info` dictionary to update the metadata of the image (`.attrs` should include the information provided in `info`).

Does that make sense? I guess I'm just confused at how this works because at first glance it doesn't seem like it should.

ch-k: Hmm, I don't really know what for `attrs['name']` is used in satpy, but in my case it obviously doesn't matter. The problem was that in `get_dataset()` the access to `file_content` was with a different key than the data was put into it. And that is solved here (key is always `image`).

satpy/readers/generic_image.py
ch-k: Done

satpy/readers/generic_image.py
```diff
@@ -102,19 +106,50 @@ def end_time(self):
...
+
+        # Mask data if necessary
+        try:
+            data = self.__mask_image_data(data, info)
+        except ValueError as err:
+            logger.warning(err)
+
+        return data
+
+    def __mask_image_data(self, data, info):
```
djhoese, line 121: Why did this become a class method? Couldn't/shouldn't this remain a function outside the class? Also, please use a single underscore. Double underscore has special handling in python and isn't needed here.

ch-k: Sorry, too much java in the last months. I'll fix it tomorrow morning.

ch-k: Fixed it.

satpy/readers/generic_image.py
djhoese: @ch-k If you load a `Scene` using this reader, what does `print(scn.keys())` show you? The `.attrs['name']` is used by Satpy to actually provide the name that Satpy uses. Satpy is actually using `name` and other pieces of metadata to create a `DataID` object to fully/uniquely identify a DataArray. If the `.attrs` for the `DataArray` don't include what the YAML has then it is likely that multiple datasets will be loaded as a single item. Based on what I understand of this code, even if your YAML said it had "image1" and "image2", they would both be returned with "image" as their name.

It is possible the base reader is overwriting the attributes with the information from the YAML so that this doesn't matter, but I'd prefer the readers be explicit if possible. Sorry, if this doesn't make much sense.

satpy/tests/reader_tests/test_generic_image.py
```diff
@@ -180,10 +211,75 @@ def test_GenericImageFileHandler(self):
...
-        # Test masking of floats
+    def test_GenericImageFileHandler_masking_only_integer(self):
+        """Test direct use of the reader."""
+        from satpy.readers.generic_image import GenericImageFileHandler
+
+        class FakeGenericImageFileHandler(GenericImageFileHandler):
+
+            def __init__(self, filename, filename_info, filetype_info, file_content, **kwargs):
+                """Get fake file content from 'get_test_content'."""
+                if GenericImageFileHandler is object:
```
djhoese, line 222: When will this import ever return `object`? Also, does this class need to be inside the test function?

satpy/readers/generic_image.py
ch-k:
```
scn.load(['image_goes16', 'image_goes17'])
print(scn.keys())
```
> [DataID(name='image_goes16', modifiers=()), DataID(name='image_goes17', modifiers=())]

and after loading the derived composite:

```
scn.load(['dwd_world_comp'])
print(scn.keys())
```
> [DataID(name='dwd_world_comp'), DataID(name='image_goes16', modifiers=()), DataID(name='image_goes17', modifiers=())]

Looks fine.

satpy/tests/reader_tests/test_generic_image.py
ch-k: It's basically a copy of an other fake reader (from file test_hdf5_utils.py). I supposed it is best practice t odo so (without knowning why..). Of cource I can remove it. I've already noted that some the checkers marked this as issue.
It is an inner class because all the other test routines in this file do their imports in the method itself. I thought, it would be better to do the same here too.

satpy/readers/generic_image.py
ch-k: I digged a little bit more in satpy code. It seems that the DataID is derived from the reader yaml. This id is used to call the readers `get_dataset()` method and this id is also used as key to fill a DatasetDict (in yaml_reader.py `load()`). Couldn't really find a reader that explicitly sets `attrs['name']` (Just want to see how it should work).

satpy/tests/reader_tests/test_generic_image.py
djhoese: Yeah I think now that we are using pytest for tests it isn't as important to have the various imports in the test functions themselves. @mraspaud thoughts?

satpy/readers/generic_image.py
djhoese: What you should find is file handlers that do `.attrs.update(info)` or something like that (ex. `DataArray(..., attrs=new_attrs)`). The `info` passed to the file handler is meant to be a dictionary of the info provided in the YAML file including DataID-like parameters like `name`, `calibration`, etc.

ch-k: Alright. Just pushed this change.

satpy/readers/generic_image.py
```diff
@@ -102,12 +106,29 @@ def end_time(self):
 
     def get_dataset(self, key, info):
         """Get a dataset from the file."""
-        logger.debug("Reading %s.", key)
-        return self.file_content[key['name']]
+        ds_name = self.dataset_name if self.dataset_name else key['name']
```
djhoese, line 109: Just to make sure I understand, the idea here is that if a developer wants to, they could redefine the `read` method, set `dataset_name` to something else (or not at all) and then load more than just a basic image from an input dataset, right? Put another way, it gives people subclassing this file handler the flexibility to do more complex stuff?

ch-k: Yes, like the fake file handler in the test. Though I didn't use different names there.

satpy/tests/reader_tests/test_generic_image.py
ch-k: Just removed `if GenericImageFileHandler is object`.

mraspaud: I tihnk it's fine without the check

djhoese:
Last request: Could you look at fixing some of the style issues from codebeat and deepcode? It looks like one of the deepcode ones will be unfixable (the one regarding `super`) as I think what you've done is intended.

ch-k:
> 
> 
> Last request: Could you look at fixing some of the style issues from codebeat and deepcode? It looks like one of the deepcode ones will be unfixable (the one regarding `super`) as I think what you've done is intended.

Fixed some of the issues. But I don't really know how to make codebeat happy. These two remaining issues are classified as info. I can't imagine that further reducing of ACB or branch nesting makes the code more readable and maintainable.

djhoese (approved):
Looks good. Thanks.

djhoese:
Thanks for making the attempt. And yeah, removing all the issues identified by these tools is just something you get used to and readability is arguable.

PR #1564, opened 2021-02-24
Fix DifferenceCompositor ignoring YAML metadata

djhoese:
I needed to be able to set a `standard_name` for my difference and noticed that it was being ignored. This pull request fixes that.

 - [x] Tests added


djhoese:
Retry DeepCode

PR #1565, opened 2021-02-24
Rename min_area() and max_area() methods

pnuu:
To clarify naming, this PR renames the `Scene.min_area()` and `Scene.max_area()` to `.coarsest_area()` and `.finest_area()`, respectively. The original naming was causing confusion what was meant: is `min_area()` the "geographically smallest area in the scene", "area with smallest pixels (highest resolution)" or "area of the data with smallest resolution (highest resolution)"? The new naming introduced in this PR got consensus in Slack, so lets use these. Warnings about the naming changes have been added.

 - [x] Tests adjusted

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -244,7 +244,8 @@ class Scene:
                                  current Scene. Defaults to all datasets.
 
         """
-        warnings.warn("'max_area' is deprecated, use 'finest_area' instead.")
+        warnings.warn("'max_area' is deprecated, use 'finest_area' instead.",
+                      warnings.DeprecationWarning)
         return self.finest_area(datasets=datasets)
 
     def coarsest_area(self, datasets=None):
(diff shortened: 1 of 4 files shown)


mraspaud (commented):
LGTM.

Should we start using deprecation warnings? (#1554)

pnuu:
Sure, I can do that. But as those are not shown _by_ _default_, practically no-one using this feature would ever see the warning.

satpy/scene.py
```diff
@@ -233,7 +234,20 @@ def max_area(self, datasets=None):
...
+        """Get highest resolution area for the provided datasets. Deprecated.
+
+        Args:
+            datasets (iterable): Datasets whose areas will be compared. Can
+                                 be either `xarray.DataArray` objects or
+                                 identifiers to get the DataArrays from the
+                                 current Scene. Defaults to all datasets.
+
+        """
+        warnings.warn("'max_area' is deprecated, use 'finest_area' instead.")
```
djhoese, line 247: I think @gerritholl wants these types of warnings to be `DeprecationWarning` now instead of the default `UserWarning`.

djhoese (approved):
Besides maybe using a DeprecationWarning, this looks good to me. Now I just need to update my tutorials to use this once this is released.

mraspaud (approved):
LGTM

PR #1582, opened 2021-03-03
Ensure FSFile objects compare equal when they should

### LINKED ISSUE #1581: FSFile object compares unequal when all properties equal
**Describe the bug**

An `FSFile` object compares unequal against a different `FSFile` object representing the same file at the same file system.

**To Reproduce**

```python
print(FSFile("tofu") == FSFile("tofu"))
```

**Expected behavior**

```python
True
```

**Actual results**

```python
False
```

**Environment Info:**

 - Satpy Version: 0.25.2.dev218+ge344613f

gerritholl:
Ensure that FSFile objects compare equal when they refer to the same file on the same file system, and unequal otherwise.

 - [x] Closes #1581
 - [x] Tests added
 - [x] Fully documented

### DIFF
--- a/satpy/readers/__init__.py
+++ b/satpy/readers/__init__.py
@@ -588,6 +588,12 @@ class FSFile(os.PathLike):
         """Implement ordering."""
         return os.fspath(self) < os.fspath(other)
 
+    def __eq__(self, other):
+        """Implement equality comparisons."""
+        return (isinstance(other, FSFile) and
+                self._file == other._file and
+                self._fs == other._fs)
+
 
 def open_file_or_filename(unknown_file_thing):
     """Try to open the *unknown_file_thing*, otherwise return the filename."""
--- a/satpy/readers/__init__.py
+++ b/satpy/readers/__init__.py
@@ -585,11 +585,19 @@ class FSFile(os.PathLike):
             return open(self._file)
 
     def __lt__(self, other):
-        """Implement ordering."""
+        """Implement ordering.
+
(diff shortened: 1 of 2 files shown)


gerritholl:
Open questions:

- Ordering considers only the filename, whereas equality comparisons also consider the filesystem object.  That leads to surprising results (but not to any failing tests at the moment):

```python
In [22]: a = FSFile("a")

In [23]: b = FSFile("a", LocalFileSystem())

In [24]: a < b
Out[24]: False

In [25]: a == b
Out[25]: False

In [26]: a > b
Out[26]: True
```

What I suspect is happening here is that the `@total_ordering` decorator decides that if not `a < b` and not `a == b` then we must have `a > b`.  This problem already exists in current master, though:

```
In [3]: a = FSFile("a")

In [4]: b = FSFile("a")

In [5]: a < b
Out[5]: False

In [6]: a == b
Out[6]: False

In [7]: a > b
Out[7]: True
```

so I think it is OK to leave this open problem to another PR.

- `FSFile("tofu")' and 'FSFile("tofu", LocalFileSystem())` refer to the same file, but do not have the same filesystem object.  Should they compare equal or not?

gerritholl:
I can't tell why the test is failing, as the linked page is freezing my browser.

djhoese:
Almost froze my browser too. The conda HTTP connections were hanging. I've restarted the jobs.

mraspaud (approved):
LGTM, I can live with this for now :smile:

Issue #1571, opened 2021-02-26
Satpy test suite raises 1313 warnings when no filters


gerritholl:
**Describe the bug**

When running the Satpy test suite, many many warnings are raised (currently 1313 when I tested it).

**To Reproduce**

```bash
pytest -W always
```

**Expected behaviour**

In a perfect world, there would be no warnings issued at all.  This is not very realistic, in particular not for warnings raised by Satpy dependencies.  But maybe it could be less than 1313.

**Actual results**

```
=================== 1165 passed, 2 skipped, 1313 warnings in 125.61s (0:02:05) ====================
```

**Environment Info:**
 - OS: openSUSE 15.0
 - Satpy Version: 0.25.2.dev50+g905197a3
 - PyResample Version: 1.17.0+57.gc440d1b

djhoese:
Last I checked these were mostly:

1. pyproj warnings about proj_dict/proj_str. See #1552 and https://github.com/pytroll/pyresample/pull/332
2. recent numpy no long wanting you to use np.int_ or np.bool_ (or whatever) and instead just use the builtin int and bool.

djhoese:
#1552 is now merged. Hopefully this number will go more when the pyresample PR is merged but we're at 996 now so much better. I'll see if I can throw some of the numpy fixes into another PR today.

gerritholl:
Addressed by https://github.com/pytroll/satpy/pull/3195  but I don't know if it can be closed yet.

djhoese:
Thanks for reminding me of this issue. The test warnings in CI are now very low. The majority are caused by a warning from pyresample that is fixed in pyresample's `main` but not released (I'm waiting on pandas Python 3.14 support to release pyresample with 3.14 wheels).

I would consider this issue close-able when we add `"error"` to the pytest warning filters so that any unexpected warnings cause test failures.

PR #1585, opened 2021-03-03
Cleanup various warnings encountered when running Satpy tests

djhoese:
Addresses some of the warnings encountered when running Satpy tests. This is discussed a little in #1571. Along with my pyresample PR this gets me to <700 warnings. The warnings fixed here:

1. Numpy dtype warnings: we used `np.int/float/bool` which is now deprecated as they mean the same thing as Python's builtin `int/float/bool`.
2. A few pyproj deprecations
3. Some tests that were explicitly testing deprecated Satpy features.

The majority that remain include:

1. Numpy warnings in other libraries including some pytroll libraries like pyorbital
2. A few remaining pyproj warnings
3. Dask and numpy warnings about either empty slices or NaN/invalid values in computations. Not sure what we can do about these.


pnuu (approved):
LGTM and still works.

mraspaud (approved):
LGTM

PR #1586, opened 2021-03-05
Update GRIB reader for greater flexibility.

simonrp84:
Currently, the grib reader expects certain keys to be present in the grib file, [for example](https://github.com/pytroll/satpy/blob/master/satpy/readers/grib.py#L236-L254):
```
ds_info.update({
            'filename': self.filename,
            'shortName': msg['shortName'],
            'long_name': msg['name'],
            'pressureUnits': msg['pressureUnits'],
            'typeOfLevel': msg['typeOfLevel'],
            'standard_name': msg['cfName'],
            'units': msg['units'],
            'modelName': msg['modelName'],
            'model_time': model_time,
            'centreDescription': center_description,
            'valid_min': msg['minimum'],
            'valid_max': msg['maximum'],
            'start_time': start_time,
            'end_time': end_time,
            'sensor': msg['modelName'],
            # National Weather Prediction
            'platform_name': 'unknown',
        })
```

For some grib files, such as those from the ERA5 and CAMS datasets, not all keys are present. `modelName` is not present in ERA5 single level datasets.

This PR adds some flexibility to the grib reader and will now set missing keys to `unknown` in `ds_info` rather than raising an error.
I have only tested this with ECMWF/ERA/GFS data but it seems to work well with those.

 - [ ] Tests passed
 - [ ] Passes ``flake8 satpy``

### DIFF
--- a/satpy/readers/grib.py
+++ b/satpy/readers/grib.py
@@ -271,25 +271,33 @@ class GRIBFileHandler(BaseFileHandler):
             center_description = msg['centreDescription']
         except (RuntimeError, KeyError):
             center_description = None
+
+        key_dicts = {
+            'shortName': 'shortName',
+            'long_name': 'name',
+            'pressureUnits': 'pressureUnits',
+            'typeOfLevel': 'typeOfLevel',
+            'standard_name': 'cfName',
+            'units': 'units',
+            'modelName': 'modelName',
+            'valid_min': 'minimum',
+            'valid_max': 'maximum',
+            'sensor': 'modelName'}
+
         ds_info.update({
             'filename': self.filename,
-            'shortName': msg['shortName'],
-            'long_name': msg['name'],
-            'pressureUnits': msg['pressureUnits'],
-            'typeOfLevel': msg['typeOfLevel'],
-            'standard_name': msg['cfName'],
-            'units': msg['units'],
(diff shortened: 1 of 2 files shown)


satpy/readers/grib.py
```diff
@@ -271,25 +271,33 @@ def get_metadata(self, msg, ds_info):
...
             'end_time': end_time,
-            'sensor': msg['modelName'],
-            # National Weather Prediction
-            'platform_name': 'unknown',
-        })
+            'platform_name': 'unknown'})
+
+        for key in key_dicts:
+            if key_dicts[key] in msg.keys():
+                ds_info.update({key: msg[key_dicts[key]]})
```
djhoese, line 297:
```suggestion
                ds_info[key] = msg[key_dicts[key]]
```

And same for the line below. Thoughts?

mraspaud (approved):
LGTM

satpy/readers/grib.py
simonrp84: Changed this and the other line, don't have any opinions tbh but if this is a better way then the original then that's all good with me.

djhoese (approved):
Thanks for the fixes. Will merge when tests pass. We should make sure all the other PRs are closed when it is merged.

PR #1589, opened 2021-03-09
Delete unnecessary coordinates in tropomi reader

zxdawn:
The coordinates of `longitude` and `latitude` shouldn't be linked to themselves again.

And this PR can fix this error of `scn.save_datasets()` when loading `longitude` or `latitude`:

```
  File "test_save.py", line 77, in <module>
    scn.save_datasets(filename='./output/test.nc')
  File "/public/home/zhangxin/new/miniconda3/envs/s5p_tobac/lib/python3.8/site-packages/satpy/scene.py", line 1065, in save_datasets
    return writer.save_datasets(dataarrays, compute=compute, **save_kwargs)
  File "/public/home/zhangxin/new/miniconda3/envs/s5p_tobac/lib/python3.8/site-packages/satpy/writers/cf_writer.py", line 721, in save_datasets
    res = dataset.to_netcdf(filename, engine=engine, group=group_name, mode='a', encoding=encoding,
  File "/public/home/zhangxin/new/miniconda3/envs/s5p_tobac/lib/python3.8/site-packages/xarray/core/dataset.py", line 1689, in to_netcdf
    return to_netcdf(
  File "/public/home/zhangxin/new/miniconda3/envs/s5p_tobac/lib/python3.8/site-packages/xarray/backends/api.py", line 1107, in to_netcdf
    dump_to_store(
  File "/public/home/zhangxin/new/miniconda3/envs/s5p_tobac/lib/python3.8/site-packages/xarray/backends/api.py", line 1142, in dump_to_store
    variables, attrs = conventions.encode_dataset_coordinates(dataset)
  File "/public/home/zhangxin/new/miniconda3/envs/s5p_tobac/lib/python3.8/site-packages/xarray/conventions.py", line 806, in encode_dataset_coordinates
    return _encode_coordinates(
  File "/public/home/zhangxin/new/miniconda3/envs/s5p_tobac/lib/python3.8/site-packages/xarray/conventions.py", line 765, in _encode_coordinates
    written_coords.update(attrs["coordinates"].split())
AttributeError: 'list' object has no attribute 'split'
```

PR #1590, opened 2021-03-09
convert attributes coords to str

zxdawn:
Fix the CF_writer problem as mentioned in #1589 .

 - [ ] Tests added

### DIFF
--- a/satpy/writers/cf_writer.py
+++ b/satpy/writers/cf_writer.py
@@ -556,6 +556,10 @@ class CFWriter(Writer):
         if 'prerequisites' in new_data.attrs:
             new_data.attrs['prerequisites'] = [np.string_(str(prereq)) for prereq in new_data.attrs['prerequisites']]
 
+        # convert coordinates into string
+        if 'coordinates' in new_data.attrs and not isinstance(new_data.attrs['coordinates'], str):
+            new_data.attrs['coordinates'] = ' '.join(new_data.attrs['coordinates'])
+
         # Flatten dict-type attributes, if desired
         if flatten_attrs:
             new_data.attrs = flatten_dict(new_data.attrs)


zxdawn:
`xarray` keeps them as string: https://github.com/pydata/xarray/blob/d2582c2f8811a3bd527d47c945b1cccd4983a1d3/xarray/conventions.py#L522-L530.

Yes, Satpy removes the attrs `coordinates` if it successfully assigns them to `.coords`.


Could you give me some hints of editing lines of `yaml_reader.py`? I don't know where it tries to load coordinate variables.

djhoese:
I'm having trouble figuring out where the removal of the `coordinates` happens. Here is where it gets the coordinates from the YAML:

https://github.com/pytroll/satpy/blob/f56a089ab7ad515473699b4d719795c645122804/satpy/readers/yaml_reader.py#L742

But this shouldn't include the coordinates *from* the file handler unless available_datasets is defined in the file handler (which I believe the tropomi reader does). You'll also see that the base reader calls this function for adding crs/x/y coordinates:

https://github.com/pytroll/satpy/blob/f56a089ab7ad515473699b4d719795c645122804/satpy/resample.py#L260

Neither of these pieces removes the coordinates from what I can tell. If I had to guess, the tropomi reader must be doing something special (or not special) with the coordinates attribute. Maybe it is including them in the `.attrs` when other readers aren't. Maybe the attributes are defined in the files and the reader isn't modifying that (a combination of the file format and xarray and the satpy file handler all causing the overall behavior).

zxdawn:
I find the definition of `coordinates` in tropomi reader's `available_datasets`: https://github.com/pytroll/satpy/blob/f56a089ab7ad515473699b4d719795c645122804/satpy/readers/tropomi_l2.py#L139-L148

Maybe, this causes the problem ...

zxdawn:
I come up with three solutions:

1. The current method: converting the coordinates to string in the `cf_writer`.
2. Don't assign `longitude` and `latitude` coordinates to `longitude` or `latitude` variables in the tropomi.yaml. This has been pulled in #1589 .
3. Make Satpy using the same routine (using string instead of list) as xarray. This may be the best solution? But, it seems to need more modifications ....

PR #1592, opened 2021-03-10
Fix tests where xarray was unable to guess backend engine

djhoese:
Our unstable CI tests started failing because https://github.com/pydata/xarray/pull/4989 was merged. Xarray was no longer able to guess the backend engine for our tests that weren't using the `.nc` suffix even though the tests were created/using netcdf files. This PR fixes these tests.

 - [ ] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM thanks for the quick reaction!

PR #1597, opened 2021-03-12
add file_patterns in file_types with resolution type for satpy_cf_nc reader

TAlonglong:
If a instrument provide different spatial resolution, like viirs with M, I and DN Bands this can be stored in one file per resolution type.
This PR adds a file_pattern to match this.


mraspaud (approved):
LGTM, I'm good with that.

PR #1599, opened 2021-03-15
Add MiRS and mimicTPW2_comp readers to sphinx documentation

joleenf:
Adds the mirs and mimic readers to the index.rst list for documentation.


doc/source/index.rst
```diff
@@ -272,6 +272,12 @@ the base Satpy installation.
     * - AMSR2 level 2 GAASP in NetCDF4 format
       - `amsr2_l2_gaasp`
       - Beta
+    * - MiRS level 2 Precipitation and Surface Products (IMG) in NetCDF4 format
+      - `mirs`
+      - Beta
+    * - MIMIC Total Precipitable Water Product Reader in NetCDF format
+      - mimic_TPW2_nc
```
djhoese, line 279: So it looks like the name is actually `mimicTPW2_comp` but the python module is `mimic_TPW2_nc`. We don't need to change it, but this name is ugly. I didn't think we were going to allow uppercase letters in a reader name. Well here we are I guess.

djhoese (changes requested):
Looks like the mimic reader name is wrong. Otherwise, thanks for updating this.

doc/source/index.rst
joleenf: I will fix it in the index.  I can change the uppercase for the reader too, it is up to you.  I might be the only one using the mimic reader.

djhoese: Maybe change the name in another PR. Let's see what @mraspaud and others think about on slack. What does `comp` mean by the way?

joleenf: ```comp``` means that I was just getting started with satpy and did not know what to call it.  :)  Originally, the reader name matched the name ```mimicTPW2_comp``` in the yaml file.  I changed the reader name when I reread the documentation, but forgot to change it within the yaml file.  FWIW, comp was meant to signal that this is a composite of tpw from the MiRS suite.  Someone else mentioned that the sensor and platform names are well named at all and the word product should appear somewhere.  This is the opportunity to fix many issues.  sensor name, platform name, navigation description and reader name.

joleenf: I can pull the mimic information from index.rst until that is resolved, but the MiRS information can remain in the index.rst for the PR.

djhoese: Keep both, but just fix the name of the mimic reader in the table so it matches what the reader is actually called (in the YAML, not the python module).

djhoese (approved):
LGTM

PR #1602, opened 2021-03-16
Fix missing VIIRS SDR DNB solar and lunar azimuth angle datasets

djhoese:
I'd like to use these datasets but they weren't configured in the `viirs_sdr` and `viirs_l1b` readers.

 - [x] Tests added 
 - [x] Fully documented


mraspaud (approved):
LGTM! why not add it to the compact viirs reader too?

djhoese:
I don't know anything about the viirs compact format so I wasn't sure if it even existed, how to add it, or how to test it.

PR #1603, opened 2021-03-18
Update slstr_l2.yaml

JohannesSMHI:
Replacing:
- "dt1" with "start_time"
- "dt2" with "end_time"

Using start_time and end_time so that ```satpy.readers.find_files_and_readers``` can filter out datafiles based on timestamps.

 - [ ] Tests added 
 - [ ] Fully documented


djhoese (approved):
Thanks! I'd still like someone who has more experience with this reader to double check that this isn't a problem. @mraspaud ?

mraspaud (approved):
LGTM, thanks!

PR #1610, opened 2021-03-22
Fix auxiliary download script not using provided data directory

djhoese:
Small typo that stopped the auxiliary download script from customizing the location of data to download. This script was supposed to be so simple it didn't need testing... :facepalm: 

 - [ ] Closes #xxxx 
 - [x] Tests added


mraspaud (approved):
LGTM, thanks for adding the test.

PR #1620, opened 2021-03-25
Add feature to StaticImageCompositor to allow filenames relative to Satpy 'data_dir'

yufeizhu600:
- [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented 

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1059,8 +1059,6 @@ class StaticImageCompositor(GenericCompositor, DataDownloadMixin):
 
     @staticmethod
     def _get_cache_filename_and_url(filename, url):
-        data_dir = satpy.config.get('data_dir')
-
         if filename:
             filename = os.path.expanduser(os.path.expandvars(filename))
 
@@ -1068,10 +1066,11 @@ class StaticImageCompositor(GenericCompositor, DataDownloadMixin):
             url = os.path.expandvars(url)
             if not filename:
                 filename = os.path.basename(url)
-        else:
-            if filename and not os.path.isabs(filename):
-                if os.path.exists(os.path.join(data_dir, filename)):
-                    filename = os.path.join(data_dir, filename)
+        elif filename and not os.path.isabs(filename):
+            data_dir = satpy.config.get('data_dir')
+            path = os.path.join(data_dir, filename)
+            if os.path.exists(path):
+                filename = path
 
         if (url and os.path.isabs(filename)) or \
            (not url and (not filename or not os.path.isabs(filename))):
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1028,12 +1028,12 @@ class StaticImageCompositor(GenericCompositor, DataDownloadMixin):
         Use cases:
             1. url + no filename:
                Satpy determines the filename based on the filename in the URL,
-               then downloads the URL,  and saves it to <data_dir>/<filename>.
+               then downloads the URL, and saves it to <data_dir>/<filename>.
                If the file already exists and known_hash is also provided, then the pooch
                library compares the hash of the file to the known_hash. If it does not
                match, then the URL is re-downloaded. If it matches then no download.
             2. url + relative filename:
-               Same as 1 but filename is already provided so download goes to
+               Same as case 1 but filename is already provided so download goes to
                <data_dir>/<filename>. Same hashing behavior. This does not check for an
                absolute path.
             3. No url + absolute filename:
@@ -1059,19 +1059,18 @@ class StaticImageCompositor(GenericCompositor, DataDownloadMixin):
 
     @staticmethod
     def _get_cache_filename_and_url(filename, url):
+        if filename:
+            filename = os.path.expanduser(os.path.expandvars(filename))
+
         if url:
             url = os.path.expandvars(url)
             if not filename:
                 filename = os.path.basename(url)
-        else:
-            if filename:
-                filename = os.path.expanduser(os.path.expandvars(filename))
-
-                if not os.path.isabs(filename):
-                    data_dir = satpy.config.get('data_dir')
-                    path = os.path.join(data_dir, filename)
-                    if os.path.exists(path):
-                        filename = path
+        elif filename and not os.path.isabs(filename):
+            data_dir = satpy.config.get('data_dir')
+            path = os.path.join(data_dir, filename)
+            if os.path.exists(path):
+                filename = path
 
         if (url and os.path.isabs(filename)) or \
            (not url and (not filename or not os.path.isabs(filename))):
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1058,18 +1058,18 @@ class StaticImageCompositor(GenericCompositor, DataDownloadMixin):
         self._cache_key = cache_keys[0]
 
     @staticmethod
-    def _get_cache_filename_and_url(filename, url):
-        def _check_relative_filename(filename):
-            data_dir = satpy.config.get('data_dir')
-            path = os.path.join(data_dir, filename)
+    def _check_relative_filename(filename):
+        data_dir = satpy.config.get('data_dir')
+        path = os.path.join(data_dir, filename)
 
-            return path if os.path.exists(path) else filename
+        return path if os.path.exists(path) else filename
 
+    def _get_cache_filename_and_url(self, filename, url):
         if filename:
             filename = os.path.expanduser(os.path.expandvars(filename))
 
             if not os.path.isabs(filename) and not url:
-                filename = _check_relative_filename(filename)
+                filename = self._check_relative_filename(filename)
 
         if url:
             url = os.path.expandvars(url)
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1059,14 +1059,6 @@ class StaticImageCompositor(GenericCompositor, DataDownloadMixin):
 
     @staticmethod
     def _get_cache_filename_and_url(filename, url):
-        """ """
-        def _check_valid_filename_url(filename, url):
-            if (url and os.path.isabs(filename)) or \
-               (not url and (not filename or not os.path.isabs(filename))):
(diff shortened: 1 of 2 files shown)


satpy/composites/__init__.py
```diff
@@ -1041,12 +1042,22 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
     def _get_cache_filename_and_url(filename, url):
         if filename is not None:
             filename = os.path.expanduser(os.path.expandvars(filename))
+
+            if not os.path.isabs(filename):
+                data_dir = satpy.config.get('data_dir')
+                filename = os.path.join(data_dir, filename)
+                if not os.path.exists(filename):
+                    raise ValueError('Can not find image {} for '
+                                     'StaticImageCompositor'.format(filename))
```
djhoese, line 1051: I think this changes the overall behavior of how this is supposed to work. The `filename` is meant to allow the user to customize how the file is stored in the data_dir. If you add the data_dir here and it doesn't exist, it should still allow the `url` to be used in `register_data_files`. I think if the absolute path with data_dir doesn't exist then the original filename should be used and we should depend on the error message at the end of this method.

Thoughts?

satpy/composites/__init__.py
```diff
@@ -1041,12 +1042,22 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
...
+
         if url is not None:
             url = os.path.expandvars(url)
             if filename is None:
                 filename = os.path.basename(url)
+
         if url is None and (filename is None or not os.path.isabs(filename)):
             raise ValueError("StaticImageCompositor needs a remote 'url' "
+                             "or just a 'filename' without path, "
```
djhoese, line 1060:
```suggestion
            raise ValueError("StaticImageCompositor needs a remote 'url', "
                             "or an existing 'filename' relative to Satpy's 'data_dir', "
```

satpy/composites/__init__.py
yufeizhu600: Updated the logic according to the discussion of use cases in slack.

satpy/composites/__init__.py
yufeizhu600: Done. Thanks

satpy/composites/__init__.py
```diff
@@ -1023,6 +1025,24 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
             area (str): Name of area definition for the image.  Optional
                 for images with built-in area definitions (geotiff).
 
+        Use cases:
+            1. url + no filename:
+               Satpy determines the filename based on the filename in the URL,
+               then downloads the URL,  and saves it to <data_dir>/<filename>.
```
djhoese, line 1031:
```suggestion
               then downloads the URL, and saves it to <data_dir>/<filename>.
```

satpy/composites/__init__.py
```diff
@@ -1039,15 +1059,26 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
...
+        if url:
             url = os.path.expandvars(url)
-            if filename is None:
+            if not filename:
                 filename = os.path.basename(url)
-        if url is None and (filename is None or not os.path.isabs(filename)):
-            raise ValueError("StaticImageCompositor needs a remote 'url' "
-                             "or absolute path to 'filename'.")
+        else:
+            if filename:
```
djhoese, line 1067: You could combine `else:` and `if filename:` into one `elif filename:`. That would help with the codebeat branching.

satpy/composites/__init__.py
```diff
@@ -1039,15 +1059,26 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
...
+                filename = os.path.expanduser(os.path.expandvars(filename))
+
+                if not os.path.isabs(filename):
+                    data_dir = satpy.config.get('data_dir')
+                    path = os.path.join(data_dir, filename)
+                    if os.path.exists(path):
+                        filename = path
```
djhoese, line 1074: You could put this chunk of code into a separate method (ex. `def _check_relative_filename(filename):`, could even be a staticmethod I think). That would make codebeat happier too.

satpy/composites/__init__.py
```diff
@@ -1039,15 +1059,26 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
...
+
+        if (url and os.path.isabs(filename)) or \
+           (not url and (not filename or not os.path.isabs(filename))):
+            raise ValueError("StaticImageCompositor needs a remote 'url', "
+                             "or absolute path to 'filename', "
+                             "or an existing 'filename' relative to Satpy's 'data_dir'.")
```
djhoese, line 1080: This could also got into its own method like `def _check_valid_url_filename(url, filename):`.

satpy/composites/__init__.py
```diff
@@ -1039,15 +1059,26 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
...
     @staticmethod
     def _get_cache_filename_and_url(filename, url):
-        if filename is not None:
-            filename = os.path.expanduser(os.path.expandvars(filename))
-        if url is not None:
+        if url:
             url = os.path.expandvars(url)
-            if filename is None:
+            if not filename:
                 filename = os.path.basename(url)
```
djhoese, line 1065: By switching the order that url and filename are checked, I think you've made it so `filename` doesn't have environment variables expanded when `url` is provided. That's not a huge issue, but I think it is an unnecessary limitation here. I think the code can be cleaner if you swap them back (check filename first, then check URL). You'd just need to change `if not os.path.isabs(filename):` to `if not os.path.isabs(filename) and not url:`. Thoughts?

djhoese (changes requested):
I like where this is going. I have a few suggestions to make codebeat happier and one concern about how you flipped the order that url and filename are checked.

satpy/composites/__init__.py
```diff
@@ -1039,15 +1059,28 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
 
     @staticmethod
     def _get_cache_filename_and_url(filename, url):
-        if filename is not None:
+        def _check_relative_filename(filename):
+            data_dir = satpy.config.get('data_dir')
+            path = os.path.join(data_dir, filename)
+
+            return path if os.path.exists(path) else filename
```
djhoese, line 1066: Could you make this a separate method on the class? If you have to make the calling method not static that's fine. You could also do `StaticImageCompositor._check_relative_filename` but maybe that's too ugly.

satpy/composites/__init__.py
yufeizhu600: I did not realized the use case of expandable filename passed to URL. I have switched code back. Thanks.

satpy/composites/__init__.py
yufeizhu600: Sure, if you are fine with that I change the calling method to non static, I can move it.

satpy/composites/__init__.py
djhoese: It is an edge case, but not something I see as impossible.

satpy/composites/__init__.py
yufeizhu600: done.

satpy/composites/__init__.py
```diff
@@ -1038,16 +1058,29 @@ def __init__(self, name, filename=None, url=None, known_hash=None, area=None,
...
+        path = os.path.join(data_dir, filename)
+
+        return path if os.path.exists(path) else filename
+
+    def _get_cache_filename_and_url(self, filename, url):
+        if filename:
             filename = os.path.expanduser(os.path.expandvars(filename))
-        if url is not None:
+
+            if not os.path.isabs(filename) and not url:
```
djhoese, line 1071: So codebeat is still complaining about this method. I bet if you move this if statement into the `_check_relative_filename` method it might be a little happier...maybe.

yufeizhu600: I am not that comfortable to bring the 'url' parameter into `_check_relative_filename` method, it is better just check`filename`. Moreover, the codebeat `issue` is actually an `info` that the ABC metric is above 10. It maybe because the condition of the exception raising is too long.

djhoese: Understandable. Perhaps the `if not url and ...:` could be replaced by an `elif ...:`, connecting it to the `if url:` block above it.

satpy/tests/test_composites.py
```diff
@@ -990,6 +989,18 @@ def load(self, arg):
...
+        comp = StaticImageCompositor("name", url=remote_tif, filename="bar.tif")
+        self.assertEqual(comp._url, remote_tif)
+        self.assertEqual(comp._cache_filename, "bar.tif")
+
+        # No URL, filename without absolute path, use default data_dir from config
+        with mock.patch('os.path.exists') as exists:
+            exists.return_value = True
+            comp = StaticImageCompositor("name", filename="foo.tif")
+            self.assertEqual(comp._url, None)
+            self.assertEqual(comp._cache_filename, "/path/to/image/foo.tif")
```
djhoese, line 1002: The tests are failing on Windows. I think this has to be `os.path.join('path', 'to', 'image', 'foo.tif')`, but I'm not sure how the root directory is handled.

yufeizhu600: Thanks, Dave. I have fixed the test case, added an `os.path.sep` as root directory.

satpy/composites/__init__.py
yufeizhu600: Done.

djhoese (approved):
Codebeat is still complaining but I think this looks pretty good. There are maybe some other ways you could refactor the method, but I'm not sure how strict we want to be about that. @mraspaud ?

gerritholl:
Thanks!  This will be good for fogpy.

djhoese:
@yufeizhu600 Is there anything else needed on this? Or are you ready for a re-review and a merge?

yufeizhu600:
I think it is good to merge.

djhoese (approved):
Thanks!

PR #1621, opened 2021-03-25
MiRS Reader:  Add limb_correction keyword

### LINKED ISSUE #1601: Allow MiRS reader to apply limb correction optionally
## Feature Request

Currently, the MiRS reader [automatically limb corrects ATMS data](https://github.com/pytroll/satpy/blob/master/satpy/readers/mirs.py#L354).

This may not be suitable for some users, who would prefer the original, raw, data in order to apply their own corrections. Therefore, it would be good to have an option to enable / disable this correction.

A reader option that allows the user to disable the correction. Something like:
```
scn = Scene(my_files, reader='mirs', reader_kwargs={'do_correction': False})
```
Having the correction enabled by default would be good, as this is likely what most users would prefer.

No changes unless user wishes to switch off the correction.

**Additional context**
In addition, should the `MiRS` reader be renamed? `mirs.py` -> `mirs_l1b.py` to make it more consistent with other readers?

joleenf:
Addresses #1601 by adding a limb_correction keyword argument to the reader.  
Default:  limb_correction for atms sensors is on, but will not be applied when keyword is set to false.
```python
    from satpy import Scene, find_files_and_readers

    filenames = find_files_and_readers(base_dir, reader="mirs")
    scene = Scene(filenames, reader_kwargs={'limb_correction': False})```

 - Closes #1601 
 - Tests added:
 asserts check for limb correction if the sensor is atms and for lack of limb correction if the sensor is not atms.

### CODE REVIEW

File: satpy/readers/mirs.py [hunk 1]
```diff
@@ -190,13 +190,28 @@ def limb_correct_atms_bt(bt_data, surf_type_mask, coeff_fns, ds_info):
...
+    sensors.  For the ATMS sensors, a limb correction
+    is applied by default.  In order to change that
+    behavior, use the keyword argument limb_correction=False
+
+
+    from satpy import Scene, find_files_and_readers
+
+    filenames = find_files_and_readers(base_dir, reader="mirs")
+    scene = Scene(filenames, reader_kwargs={'limb_correction': False})
+
```
djhoese, line 205:
```suggestion
    behavior, use the keyword argument ``limb_correction=False``::

        from satpy import Scene, find_files_and_readers

        filenames = find_files_and_readers(base_dir, reader="mirs")
        scene = Scene(filenames, reader_kwargs={'limb_correction': False})

```

djhoese (changes requested):
Looks good except for the docstring needs some small changes to get formatted correctly as a block of code.

djhoese:
Oh and do you have plans for a test with this keyword argument?

joleenf:
I will work on the test for the argument and fix the docstring.

djhoese (commented):
Nice job. Would it be possible to combine the two tests on the bottom into one that is parametrized on `limb_correction` (you can have multiple parametrize decorators if needed)?

satpy/tests/reader_tests/test_mirs.py
```diff
@@ -255,28 +257,45 @@ def _check_attrs(data_arr, platform_name):
...
                 self._check_attrs(data_arr, platform_name)
+
+                if reader_kw and not reader_kw['limb_correction']:
+                    fd.assert_not_called()
+                else:
+                    sensor = data_arr.attrs['sensor']
+                    if sensor == 'atms':
+                        fd.assert_called()
+                    else:
+                        fd.assert_not_called()
```
djhoese, line 301:
```suggestion
                sensor = data_arr.attrs['sensor']
                if reader_kw.get('limb_correction', True) and sensor == 'atms':
                    fd.assert_called()
                else:
                    fd.assert_not_called()
```

satpy/tests/reader_tests/test_mirs.py
```diff
@@ -259,24 +261,41 @@ def _check_attrs(data_arr, platform_name):
...
+
             for data_id, data_arr in loaded_data_arrs.items():
+                sensor = data_arr.attrs['sensor']
+
                 if data_id['name'] not in ['latitude', 'longitude']:
                     self._check_area(data_arr)
                 self._check_fill(data_arr)
                 self._check_attrs(data_arr, platform_name)
+
+                sensor = data_arr.attrs['sensor']
```
djhoese, line 297: This is already set above (since your second to last commit). I think one of them can be removed.

djhoese (approved):

djhoese:
Thank you!

PR #1637, opened 2021-04-13
Change default filename for cf writer to be compatible with satpy_cf_nc reader

BENR0:
The default filename pattern of the cf writer is changed to the format the `satpy_cf_nc` reader expects. Also the "instrument" attribute which the `satpy_cf_nc` reader expects is now added by the reader.

I have been trying out the `satpy_cf_nc` reader and stumbled over some things.
1. The reader expects a certain filename format which is documented. While this is not a big point never the less I think to be consistent the cf writer should by default use the same filename format because as a user I would expect to be able to read a file written with the default settings.
2. The reader also expects a global attribute `instrument` in the `get_sensor` function of the reader. This is currently not documented and is not added by the cf writer but needs to be added by the user by supplying the `header_attrs` parameter. With the same argument I think the reader should add this.

Apart from this I am wondering why the attribute is named `instrument` in the `satpy_cf_nc` reader while everywhere else it seems to be named `sensor`? Maybe this should be renamed?
Additionally I was wondering why this is retrieved from the global attributes since the reader relies on dynamic datasets anyway and retrieves some attributes anyway in ` _assign_ds_info` which could also be used to get the sensor. I guess the sensor information might not be available in all readers?
However this is not to say that global attributes should not be used. Actually maybe the cf writer should write attributes common to all datasets to the global attributes similar to the ` to_xarray_dataset` scene method?

Note: a lot of the cf writer tests are failing due to the `instrument` attribute addition. I wanted to hear your thoughts on this before fixing these.


djhoese:
I think the main purpose of this PR has been handled now in some other PR. You can see the current file types here:

https://github.com/pytroll/satpy/blob/153d8685699f546e6a4f91d06a2327b8efed651d/satpy/etc/readers/satpy_cf_nc.yaml#L14-L15

I think we can close this? If you don't think so, let me know and we can reopen it.

BENR0:
Well I would have to see the PR. The file type patterns above still don't match the patterns the cf writer uses by default. Additionally the cf reader needs the `instrument` attribute in the metadata of the netcdf which also does not get written by default. So I think it would be good to keep this open.

djhoese:
Here's what your (this) PR has:

```
'{platform_name}-{sensor}-{start_time:%Y%m%d%H%M%S}-{end_time:%Y%m%d%H%M%S}.nc'
```

And here's what is there currently:

```
'{platform_name}-{sensor}-{start_time:%Y%m%d%H%M%S}-{end_time:%Y%m%d%H%M%S}.nc' 
```

I think those are exactly the same, right?

BENR0:
:-D yes exactly the filename pattern from the reader the and writer should be the same. With my changes they are. Currently the filename pattern of the writer is:
https://github.com/pytroll/satpy/blob/62c0e83e40cef01e3d9a566e76e25b05a08e6ba6/satpy/etc/writers/cf.yaml#L5

djhoese:
Oh! :facepalm: Your changes are to the writer. @TAlonglong @martin @sfinkens what do you think? Should the CF writer's default output filename be the more complex name in this PR (see my comment above) or should the CF **reader** be updated to also match the current simpler writer default filename (see above).

mraspaud:
Yes, the more complete name is better imo

sfinkens:
Thanks for looking into this @BENR0! I agree that the more complete name is better.

Regarding the attribute name: `sensor` is certainly more consistent in the Satpy world. You could argue that the netCDF file is for people from the CF community. And they are probably more familiar with `instrument`. At least that is what's recommended by [ACDD](https://wiki.esipfed.org/Attribute_Convention_for_Data_Discovery_1-3). I have a tendency to keep the `instrument` name.

---

Thinking a bit further: There's actually a standardized vocabulary for platforms, instruments etc ([GCMD](https://gcmd.earthdata.nasa.gov/static/kms/)). For example you would specify

```
platform = "Earth Observation Satellites > METEOSAT > METEOSAT-9"
instrument = "Earth Remote Sensing Instruments > Passive Remote Sensing > Spectrometers/Radiometers > Imaging Spectrometers/Radiometers > SEVIRI > Spinning Enhanced Visible and Infrared Imager"
instrument_vocabulary = "GCMD Instruments, Version 8.1"
platform_vocabulary = "GCMD Platforms, Version 8.1"
```
I think it would be nice if all the readers followed that convention, if possible. But that's clearly beyond the scope of this PR.

BENR0:
@sfinkens thanks for the info.

 Some more thoughts from my side about the `instrument` attribute. In this PR I just added writing of this attribute to the cf writer because the cf reader complains otherwise. That said from what I grasp the cf reader does not really depend on that information itself but just for the `sensor_names` property. So maybe the cf reader should be changed to not need the `instrument` attribute?
If the writer should add that attribute it would be necessary to ensure that every dataset has it (I am not sure if currently it is possible that a dataset does not have it?).
Furthermore in this PR the `sensor` attribute of the first dataset is used. It might be necessary to refactor that to account for cases where not all datasets in the scene are from the same sensor.

sfinkens:
>  In this PR I just added writing of this attribute to the cf writer because the cf reader complains otherwise

Yes that's good, reader and writer should be consistent!

> So maybe the cf reader should be changed to not need the instrument attribute?

The `sensor_names` property is prescribed by the base class so we better keep it :)

> Furthermore in this PR the sensor attribute of the first dataset is used. It might be necessary to refactor that to account for cases where not all datasets in the scene are from the same sensor.

I think that's fine for now.

BENR0:
@djhoese @mraspaud @sfinkens working on #1695 I stumbled about this problem again and I think this is still relevant. Especially the `instrument` attribute is not very intuitive to users. For the default filenames in the reader I think it would also be good to even add `{filename}.nc` to the list like for the `generic_image` reader since a user might just want to save to a custom name which he can not read back in since the cf reader does not find any files.

BENR0:
@djhoese @mraspaud @sfinkens this still is an issue. I have merged the current main again and pushed to this branch but since this PR got closed it does not show up. Should I open another PR?

satpy/writers/cf_writer.py
```diff
@@ -822,6 +822,17 @@ def save_datasets(self, datasets, filename=None, groups=None, header_attrs=None,
                 header_attrs = flatten_dict(header_attrs)
             root.attrs = encode_attrs_nc(header_attrs)
 
+        # Add header attributes needed for satpy_cf_reader
+        root.attrs["instrument"] = datasets[0].attrs["sensor"]
```
djhoese, line 826: This `instrument` thing should be fixed by #2176. Could you remove it from this PR and we'll try a merge of 2176 and remerge master into this PR.

satpy/writers/cf_writer.py
```diff
@@ -822,6 +822,17 @@ def save_datasets(self, datasets, filename=None, groups=None, header_attrs=None,
...
+
+        _history_create = 'Created by pytroll/satpy on {}'.format(datetime.utcnow())
+        if 'history' in root.attrs:
+            if isinstance(root.attrs['history'], list):
+                root.attrs['history'] = ''.join(root.attrs['history'])
+            root.attrs['history'] += '\n' + _history_create
+        else:
+            root.attrs['history'] = _history_create
+
```
djhoese, line 835: I think this is fixed in `main`, right?

https://github.com/pytroll/satpy/blob/acd074530f1eeaa860c1d18de22ecbfd75be8166/satpy/writers/cf_writer.py#L581-L588

satpy/etc/readers/satpy_cf_nc.yaml
```diff
@@ -17,3 +17,4 @@ file_types:
         file_patterns:
          - '{platform_name}-{sensor}-{resolution_type}-{start_time:%Y%m%d%H%M%S}-{end_time:%Y%m%d%H%M%S}.nc'
          - '{platform_name}-{sensor}-{start_time:%Y%m%d%H%M%S}-{end_time:%Y%m%d%H%M%S}.nc'
+         - '{filename}.nc'
```
djhoese, line 20: From the little I gathered of the previous conversation the change to the default name in the writer was supported. I'm a little scared how much things will break for users with this change...but yeah.

So with that change included, do we *need* this generic catch-all netcdf pattern?

djhoese (changes requested):
Thanks for working on updating this again. I've re-opened the PR. Looks like I closed it and forgot I had done that even though the conversation clearly led to it needing to be reopened.

I've made a few comments and things I'm questioning. There are few things in the writer that I don't think are needed anymore. Sorry for taking so long to get to this PR that part of it were handled by others.

satpy/writers/cf_writer.py
BENR0: Yep true. I must have missed this while resolving the merge conflicts.

satpy/etc/readers/satpy_cf_nc.yaml
BENR0: The reason I also added the catch-all pattern was that if the user decides to use his own file naming pattern he won't be able to read them again because the reader does not find any files. So without it the writer/reader loose some flexibility. I personally would be in favour of it since I do like to be able to use my own custom file names. But that is subjective. I guess that is something to be discussed. If it is decided that it should not be included this is definitely something that needs to be documented maybe even with a warning or info box so users are aware of this.

djhoese: Very good point. This seems like a good idea. Thanks.

djhoese:
Not sure how but it seems the tests are failing because `pygrib` is missing?

djhoese (approved):

BENR0:
@djhoese @mraspaud just looking through some PR's to clean up some open stuff. Is there something else than merging main again which needs to be done here or can this be merged?

gerritholl:
> @djhoese @mraspaud just looking through some PR's to clean up some open stuff. Is there something else than merging main again which needs to be done here or can this be merged?

It's marked as draft.  Draft pull requests cannot be merged.

satpy/etc/readers/satpy_cf_nc.yaml
gerritholl, line 20: For consistency with the GRIB2 reader, could this be `{stem}.nc`?  `stem` would also be more accurate, as `.nc` is part of the filename.

BENR0: @gerritholl I would be ok with that. I just added `filename` based on the `generic_image` reader config.

mraspaud: fine by me.

mraspaud (commented):
Thanks for this PR, I just have a question regarding the group file test.

satpy/tests/test_readers.py
```diff
@@ -751,7 +751,7 @@ def test_no_reader(self):
 
         # without files it's going to be an empty result
         assert group_files([]) == []
-        groups = group_files(self.g16_files)
+        groups = group_files(self.g16_files, reader='abi_l1b')
```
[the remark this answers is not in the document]
mraspaud: @BENR0 can you comment on this?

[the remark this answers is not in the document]
BENR0: @mraspaud what you are saying makes sense to me. I honestly didn't check what that test does I just made this change based on @djhoese suggestion so maybe he can shed more light on this.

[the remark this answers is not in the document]
djhoese: Whoa I did not realize what the test was actually trying to accomplish. This is one downside of the generic filename in the reader. It matches everything and has the change to "steal" matches from other readers. Either:

1. This test gets updated to expect 1 group of files (not great).
2. The reader gets updated with group keys to sort by start time...which might work and fix this test.
3. Remove the default file pattern from the reader

satpy/etc/readers/satpy_cf_nc.yaml
djhoese: I vote for `stem` also. The generic_image can be updated later.

satpy/tests/test_readers.py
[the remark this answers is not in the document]
BENR0: Well obviously I am against option 3. :-D. I think removing a "feature" because a test fails is not the way to go. I am not sure which of the remaining options is the best way to move forward.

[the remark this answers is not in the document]
gerritholl: Satpy needs a strategy to choose a reader if multiple reader match a file pattern.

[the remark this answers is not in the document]
djhoese: Reader order may come down to:

https://github.com/pytroll/satpy/blob/c0b69beb9173ded3f4dea13893a7c64b7e5620f2/satpy/readers/__init__.py#L328

in most cases. There may be others. It may get even more complicated when we look at that `glob_configs` function since it is searching not only Satpy's configs but any user configured paths too:

https://github.com/pytroll/satpy/blob/c0b69beb9173ded3f4dea13893a7c64b7e5620f2/satpy/_config.py#L156-L184

Either way, it clearly uses `iglob` which isn't going to be sorted.

As for my option 3 @BENR0, it isn't that the test fails, it is that the failing test represents a use case (as @gerritholl says): which reader gets chosen when a user doesn't specify a reader (they don't know what one they want). With this PR they seem to always get the CF reader even though there is a much better choice.

@gerritholl besides alphabetical order, I'm not sure we want to go down the performance-lossy road of checking all readers for a match against a filename. We *could* implement something that knows that the CF reader and the generic image reader should be last resorts:

```
# pseudo code
generic_choice = None
for reader, reader_info in ...:
    if reader.matches(filename):
        if not reader_info["generic"]:
            return reader
        generic_choice = reader
return generic_choice
```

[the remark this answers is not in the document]
BENR0: I understand. This discussion reminded me of the conversation regarding overlapping file patterns  in #913 which might be related.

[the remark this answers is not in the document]
djhoese: Similar, but they are separate bits of code. The overlapping file patterns will likely always be a little difficult. This reader choice one can have a solution, but it might slow some operations down.

satpy/etc/writers/cf.yaml
```diff
@@ -2,6 +2,6 @@ writer:
   name: cf
   description: Generic netCDF4/CF Writer
   writer: !!python/name:satpy.writers.cf_writer.CFWriter
-  filename: '{name}_{start_time:%Y%m%d_%H%M%S}.nc'
+  filename: '{platform_name}-{sensor}-{start_time:%Y%m%d%H%M%S}-{end_time:%Y%m%d%H%M%S}.nc'
```
djhoese, line 5: Not sure it was mentioned in the comments, but does anyone know/remember where this `-` separate filename came from? As you mentioned it seems to have existed in the reader first, but it isn't clear to me why/where from.

BENR0:
I removed the `stem` from the filename patterns for the reader again because at this point it is causing to much trouble and should be solved in a separate PR. I will add an issue. This is ready to go then.

satpy/etc/writers/cf.yaml
djhoese: I think my question here still stands, but not sure we're going to get anyone who has a real answer.

djhoese (approved):
LGTM

PR #1639, opened 2021-04-13
Fix MultiScene writer handling of multiple delayed objects

BENR0:
To support groups the `cf_writer` was implemented in a way that even if no groups were specified two delayeds were returned. This made it impossible to use distributed writing with multiscenes.
Now if now groups are used a single delayed is returned since all datasets can be written at once without appending to the file.

 - [ ] Tests added


djhoese (commented):
I don't know much about the internals of the CF writer, but what about having multiple delayed objects returned stops `MultiScene` from working?

BENR0:
@djhoese scratch that. Sorry for the confusion. You are quite right in general there is no problem with multiple delayed objects hence the for loop. I experimented with this a while ago and I fixed something at the wrong point (fixed the cf writer to return only one delayed which in turn needed removal of the loop). What I think actually needs fixing is: https://github.com/pytroll/satpy/blob/2245c9a8330946810977e0441941c056a4f04396/satpy/multiscene.py#L348
While this catches situations with writers returning (source, target) combinations it also catches the cf writer case because it always returns a list of at least two delayed objects. This if clause should be more specific so it does not apply to the cf writer.

djhoese:
Ah yes, that does seem like too simplistic of a condition. There may be utilities in `satpy/writers/__init__.py` that could be used to "parse" the return values from the writers.

BENR0:
I checked `satpy/writers/__init__.py` for something useful but could not find something directly applicable. But if I see it correctly if a reader has `sources` they are always the first item so adding ``and isintance(delayed[0], da.Array)`` to the condition would solve the issue I think.

The changes to the reader are strictly not necessary then. The only "advantage" being that in the case with no groups only one delayed will be returned and therefore only one file access is needed but I am not sure if that has any relevant benefit. So I would propose to add the additional condition in the distributed writing and roll back the changes in the writer and maybe keep some of the code simplifying changes only.

djhoese:
The MultiScene could use `split_results`: https://github.com/pytroll/satpy/blob/398445320973c64477e94e3f6dee43d5924d0a55/satpy/writers/__init__.py#L487-L514 and if there are any sources then raise the error. Although, if this function is used here it should maybe be renamed to something like `sort_writer_results`.

BENR0:
Yes that's were I got the idea from. I am not really that familiar with all the writers and how their return values look like. It is very possible that what I proposed is not covering it and `split_writer_results` should be used. I was just thinking that it over complicates things if the small addition to the condition will fix the issue.

If using `split_writer_results` is preferred I will change it and push it again.

djhoese:
I don't think there is anything logically wrong with what you're suggesting, but if you use `split_writer_results` then we will catch issues like this sooner if writer results are changed in the future. And when developers need to change split_writer_results they will see it is being used in multiscene.py and that it also needs to be updated.

satpy/multiscene.py
```diff
@@ -345,7 +345,8 @@ def load_data(q):
 
         for scene in scenes_iter:
             delayed = scene.save_datasets(compute=False, **kwargs)
-            if isinstance(delayed, (list, tuple)) and len(delayed) == 2:
+            sources, targets, _ = split_results(delayed)
```
djhoese, line 348: Shouldn't the `_` be overwriting `delayed`? This way, in the future, if the MultiScene can handle source/target combinations and delayed objects the code won't have to be changed. Right?

BENR0: Good point.

djhoese: Targets should never be returned without a source. They are source + target pairs. The simple image writer should return delayed objects. The geotiff writer returns source (dask array) + target (output file-like object). If you have another example, then I still propose that `_` be replaced with `delayed`.

BENR0: Ok I see now I got fooled by the test which returned targets without sources from `split_results` for the simple image writer because the mock did not pass as a delayed object. I changed the code to overwrite delayed now and also fixed the test. Additionally I added a test for writers with source/target combinations to fail.

djhoese:
I re-titled this after it seemed like most of the work was in getting the MultiScene to work. However, I realize you still did a lot of work on the CF writer. Is this work still needed? If so, would you mind merging with the pytroll main branch and resolve the conflicts in the CF writer?

satpy/multiscene.py
```diff
@@ -341,17 +341,20 @@ def load_data(q):
 
         input_q = Queue(batch_size if batch_size is not None else 1)
         load_thread = Thread(target=load_data, args=(input_q,))
+        # set threads to daemon so they are killed if error is raised from main thread
+        load_thread.daemon = True
```
djhoese, line 345: This can be set in the `Thread` init right?

satpy/tests/test_multiscene.py
```diff
@@ -453,6 +453,8 @@ def test_save_datasets_distributed(self):
         client_mock.compute.side_effect = lambda x: tuple(v for v in x)
         client_mock.gather.side_effect = lambda x: x
         future_mock = mock.MagicMock()
+        from dask.delayed import Delayed
+        future_mock.__class__ = Delayed
```
djhoese, line 457: I'm guessing this is needed for the split_results? Could you move the import higher up?

satpy/multiscene.py
BENR0: Yes indeed. Changed it.

satpy/tests/test_multiscene.py
BENR0: Done.

djhoese (approved):
LGTM thanks!

PR #1649, opened 2021-04-22
Fix requiring newer numpy just for typing annotations

djhoese:
The `numpy.typing` subpackage was added in numpy 1.20 it seems. Satpy should still be able to support older versions of numpy. This PR adds a try/except around the `numpy.typing` that I added in #1646 for these older versions.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/_compat.py
+++ b/satpy/_compat.py
@@ -26,3 +26,10 @@ except ImportError:
     def cached_property(func):
         """Port back functools.cached_property."""
         return property(lru_cache(maxsize=None)(func))
+
+
+try:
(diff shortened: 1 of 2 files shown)


mraspaud (commented):
How about putting this in _compat.py?

djhoese:
Done

mraspaud (approved):
LGTM

ch-k:
Works for me with (numpy 1.19.5).

PR #1655, opened 2021-04-30
Apply valid_range in MiRS reader when present

joleenf:
This fixes a bug in which the valid range was not used.  Fix checks for a valid range in the attributes and applies the range if present.  There is a new check in the test code which determines if valid_range is removed from the attributes after loading the data.


satpy/readers/mirs.py
```diff
@@ -356,6 +356,17 @@ def _fill_data(self, data_arr, attrs):
             data_arr = data_arr.where(data_arr != fill_value, fill_out)
         return data_arr, attrs
 
+    def _get_valid_range(self, data_arr, attrs):
```
djhoese, line 359:
```suggestion
    def _apply_valid_range(self, data_arr, attrs):
```

Really nitpicking here. Not necessary unless you make other changes.

satpy/readers/mirs.py
```diff
@@ -356,6 +356,17 @@ def _fill_data(self, data_arr, attrs):
...
+        # handle valid_range
+        valid_range = attrs.pop('valid_range', None)
+        if isinstance(valid_range, np.ndarray):
```
djhoese, line 362: What are the other options for what this could be? Would this catch more cases if it was `valid_range is not None`?

djhoese (commented):
I had a couple questions. Also, how hard would it be to check the data in the tests to make sure the valid range was applied?

satpy/readers/mirs.py
joleenf: valid_range is not None is better.

djhoese (approved):
LGTM

PR #1666, opened 2021-05-11
Ensure that orbital parameters are in a dict

pnuu:
The `'cf'` writer converts orbital parameter dictionaries to dicts using `json.dumps()`. When reading this data back, the string representation isn't converted back to a dictionary. This PR adds the conversion to `file_handlers.py` where the orbital data are handled.

I guess the more correct place for this fix is in the `satpy_cf_nc` reader, but that would complicate both the implementation and testing significantly.

 - [x] Tests added

### DIFF
--- a/satpy/readers/file_handlers.py
+++ b/satpy/readers/file_handlers.py
@@ -18,7 +18,6 @@
 """Interface for BaseFileHandlers."""
 
 from abc import ABCMeta
-import json
 
 import numpy as np
 from pyresample.geometry import SwathDefinition
@@ -64,7 +63,6 @@ class BaseFileHandler(metaclass=ABCMeta):
     @staticmethod
     def _combine(infos, func, *keys):
         res = {}
-        infos = [_str2dict(i) for i in infos]
         for key in keys:
             if key in infos[0]:
                 res[key] = func([i[key] for i in infos])
@@ -110,7 +108,7 @@ class BaseFileHandler(metaclass=ABCMeta):
             # Collect all available keys
             orb_params_comb = {}
             for d in orb_params:
-                orb_params_comb.update(_str2dict(d))
+                orb_params_comb.update(d)
 
             # Average known keys
             keys = ['projection_longitude', 'projection_latitude', 'projection_altitude',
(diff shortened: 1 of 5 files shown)


djhoese:
> Actually, the file_handlers.py probably is the correct place for this, as that's the only place where the .attrs['orbital_parameters'] are explicitly handled.

Yeah, but this will only ever be triggered (out of the readers that exist currently) by the `satpy_cf_nc` reader.

pnuu:
@sfinkens this **is** about the `satpy_cf_nc` reader. I don't know about the raw metadata. There was a report in Slack that reading back AHI HSD data from Satpy-saved NetCDF files were failing, and this is a fix specifically for the orbital parameters that were returned as a string.

And now the fix is in the `satpy_cf_nc` reader.

sfinkens (commented):
@pnuu Ok, I see! What about converting any string type attribute starting with a `{` to dict? This should also work with raw metadata, see https://github.com/pytroll/satpy/blob/main/satpy/writers/cf_writer.py#L72. And since `satpy.dataset.metadata.combine_metadata` ignores them, that should not cause further problems.

pnuu:
Yeah, that would be an easy addition. But I can't get the example you linked to work, the `'raw_metadata'` attribute isn't saved. This is what I get with the branch of this PR, which is essentially the current `'main'`.

```python
glbl = Scene(reader='seviri_l1b_hrit', filenames=fnames_seviri)
glbl.load(['IR_108', 'VIS006'])
glbl.save_datasets(filename="/tmp/seviri_test.nc", writer='cf', datasets=['VIS006', 'IR_108'])
ds = xr.open_dataset("/tmp/seviri_test.nc")
print(ds['IR_108'].attrs.keys())
dict_keys(['calibration', 'end_time', 'georef_offset_corrected', 'grid_mapping', 'modifiers', 'orbital_parameters', 'platform_name', 'resolution', 'satellite_altitude', 'satellite_latitude', 'satellite_longitude', 'sensor', 'standard_name', 'start_time', 'units', 'wavelength'])
```

I suggest we add the "parse all strings starting with '{'" to a later PR.

sfinkens (approved):
Agreed. Apparently my recent attempt to fix the combination of raw metadata actually removes them entirely. Will look into that.

PR #1671, opened 2021-05-13
Improve MiRS reader handling of missing metadata

joleenf:
- adds metadata to the yaml file for files which do not have internal metadata
- combines metadata that may be in file to yaml attributes if they are different but file takes precedence.   This is meant to address the loss of the units from the final metadata in the previous version
- applies valid_range, fill_data and scale before limb correction
- all mirs data files will get more information about lat/lon coordinates from the yaml
- adjusts "Kelvin" to "K" for consistency within the project
- adds tests for units, and fill_value
- add coordinates to fake test data set to match how actual files look to open_dataset.


satpy/readers/mirs.py
```diff
@@ -433,6 +465,7 @@ def _available_btemp_datasets(self):
                 'name': new_name,
                 'description': desc_bt,
                 'units': 'K',
+                'scale_factor': self.nc['BT'].attrs['scale_factor'],
```
djhoese, line 468: Why only `scale_factor` and not `add_offset`?

satpy/etc/readers/mirs.yaml
```diff
@@ -25,4 +25,105 @@ file_types:
...
+    units: degrees
+    standard_name: latitude
+  rain_rate:
+    name: RR
+    description: Rain Rate
+    long_name: rain_rate
+    file_key: RR
+    file_type: metop_amsu
+    scale_factor: 0.1
+    _FillValue: -99.9
```
djhoese, line 54: Both scale_factor *and* _FillValue aren't provided in the files?

satpy/readers/mirs.py
```diff
@@ -334,70 +327,122 @@ def _nan_for_dtype(data_arr_dtype):
...
+        handled_vars = set()
         for is_avail, ds_info in (configured_datasets or []):
             if is_avail is not None:
                 # some other file handler said it has this dataset
                 # we don't know any more information than the previous
                 # file handler so let's yield early
                 yield is_avail, ds_info
                 continue
+
+            if self.file_type_matches(ds_info['file_type']):
```
djhoese, line 438: Is it possible to restructure this block of code so `file_type_matches` doesn't have to be called twice?

djhoese (approved):
LGTM. Thanks

djhoese:
FYI, I squashed all your commits into a single commit and merged it. Looking at your fork, I think your main has commits that it shouldn't so your pull requests are getting out of sync with what github expects for going in the `main` branch. You may want to do some git magic to get your main branch to the proper commit.

joleenf:
What do you want me to do with my main branch?

joleenf:
I should start with what I think I understand.  I have too many commits that are more intermediate/confused and are not needed.  This is causing problems with the main branch of this repo when trying to merge my fork into  pytroll/satpy main.  How do I know that my main branch is at the proper commit?  What do you mean by git magic?

djhoese:
Let's talk in the CSPP slack.

PR #1674, opened 2021-05-17
Feature add support for AHI True Color Reproduction

mherbertson:
This PR adds support for True Color Imagery Rendering for Himawari-8 based on the CIE XYZ Color System.

 - [x] Tests added 
 - [x] Fully documented 


test.test
```diff
@@ -0,0 +1 @@
+test
```
mraspaud, line 1: Could you remove this file?

satpy/enhancements/ahi.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python
+# Copyright (c) 2019 Satpy developers
```
mraspaud, line 2:
```suggestion
# Copyright (c) 2021 Satpy developers
```

satpy/composites/ahi.py
```diff
@@ -26,18 +26,23 @@
...
 
-        new_green = green * self.fractions[0] + nir * self.fractions[1]
-        new_green.attrs = green.attrs.copy()
+        for band in self.match_data_arrays(projectables):
+            if new_green is None:
+                new_green = band * self.fractions[fraction_index]
+                new_green.attrs = band.attrs.copy()
+            else:
+                new_green += band * self.fractions[fraction_index]
+            fraction_index += 1
```
mraspaud, line 47: This could be replaced with
```python
new_green = sum(fraction * value for fraction, value in zip(fractions, projectables))
new_green.attrs = combine_metadata(*projectables)
```

mraspaud (commented):
Thanks for adding this composite to satpy!
I have a couple of comment inline.
Otherwise, don't forget to add you name to the author list (alphabetical order) and to add a couple of tests especially for the refactored green corrector.
Look good so far!

satpy/composites/ahi.py
mherbertson: Knew there'd be a neater way

satpy/composites/ahi.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2015-2017 Satpy developers
+# Copyright (c) 2019 Satpy developers
```
mraspaud, line 3:
```suggestion
# Copyright (c) 2015-2021 Satpy developers
```

mherbertson:
The compositor is now passing the default test after the kwarg edit commit. I'm in the process of designing new tests.

This is currently how the image looks with this compositor, note there's no stretching so the image appears flat but I think the base colours match up.

satpy_tcr

This is the JMA one on their website, not the more vibrant colours, could this just be down to stretching/brightness and any after processing? 

jma_tcr

Any observations?

satpy/etc/enhancements/ahi.yaml
```diff
@@ -9,3 +9,11 @@ enhancements:
...
           min_stretch: [-26.2, -43.2, 243.9]
           max_stretch: [0.6, 6.7, 208.5]
+
+  true_color_reproduction:
+    standard_name: true_color_reproduction
+    operations:
+      - name: contrast
+        method: !!python/name:satpy.enhancements.ahi.jma_true_color_reproduction
+      - name: cira_stretch
+        method: !!python/name:satpy.enhancements.cira_stretch
```
mraspaud, line 19: Instead of the cira stretch, you could have a try with the reinhard_to_srgb stretch for a slightly brighter image overall
```suggestion
      - name: reinhard_stretch
        method: !!python/name:satpy.enhancements.reinhard_to_srgb
```

mraspaud:
@mherbertson looking really good with the colors!
as suggested inline, you could try the reinhard enhancement to make the image pop out a bit more. It even has two arguments that you can play with, saturation and white, see here https://github.com/pytroll/satpy/blob/main/satpy/enhancements/__init__.py#L215-L250

simonrp84:
Thanks for doing this @mherbertson. I'll give it a test tomorrow.
Regarding comparison to the images on the Himawari website: How does it look around sunset? When I implemented something similar the big differences I saw were around 12:00 UTC when the sun is setting on the Indian Ocean - the Himawari website pics look much better than anything I produced, so it'd be good to see how your implementation does at this time.

mherbertson:
Actually I did notice the sunset displayed the most color difference, I will do some experimenting too, thank you both for the feedback!

mherbertson:
Here's images close to sunset. I tried the reinhard_stretch and it's close but I feel the CIRA stretch is the closest match, I will experiment further, I think we can get it closer to JMA.

reinhard stretch

satpy_reinhard_stretch

JMA

jma

mraspaud:
Nice! looks like they have the clouds more saturated, have you tried using the `white` parameter of the reinhard stretch? Maybe test with 80?

mherbertson:
Ah good point, I'll try some more additional settings to see what I can get!

mherbertson:
A crude stretch helps with the overall image. In the JMA image oceans are far more vibrant and areas of high sza are less green/red. This could be down to different rayleigh scattering or sunz correction. This is especially noticeable at high sza as @simonrp84 mentioned. There's a possibility of experimenting with different blending at high sza.

Here's aus area:

Satpy:

satpy_true_color_reproduction_202105290010-aus-cropped

JMA:

jma_true_color_reproduction_202105290010-aus

Full Disk

Satpy:

satpy_true_color_reproduction_202105290010-fldk

JMA:

jma_true_color_reproduction_202105290010-fldk

mherbertson:
Difference in coloring at high sza. Will experiment further.

Satpy:

true_color_reproduction_20210525_1010_full

JMA:

fd__trm_1010

mherbertson:
I've managed to get as close as I can to JMA style images. I've added stretching and blend atmospherically uncorrected data at high sza. Aerosol is now `rayleigh_only`. Looking at ways of moving forward towards blending out rayleigh correction.

Satpy:

satpy_tcr_20210525_1010_fldk

JMA: 

fd__trm_1010

mraspaud:
@mherbertson I looks great!

One solution to avoid having to blend two composites would be to set a gradient on the corrections, both sza correction and rayleigh scattering correction.

The sza correction seems to have possibilities for this already, see here: https://github.com/pytroll/satpy/blob/main/satpy/modifiers/geometry.py#L118-L123 right @pnuu ?

as for the rayleigh correction, it's not implemented yet, but it shouldn't be too hard as the correction could be lower in the twilight zone: https://github.com/pytroll/satpy/blob/main/satpy/modifiers/atmosphere.py#L113
Maybe the same parameters as the sza correction could be implemented, correction_limit and max_sza?

mherbertson:
The JMA style requires two seperate images (one with with no rayleigh correction) with different SZA calculations to get the desired imagery therefore my rayleigh blending on SZA would not work. The processing time was increased due to extra SZA correction and linear stretching.

Using a custom cruide strech this is as good as I can get and I feel is a good overall compromise on quality and performance for the end user.  They can then blend out the high SZA as they wish.

I'll begin working on some tests!

true_color_reproduction_202108260300

true_color_reproduction_202108251000

true_color_reproduction_day_night_202108251000

true_color_reproduction_202108252200

true_color_reproduction_day_night_202108252200

mherbertson:
Not sure if documentation is required for this. I've added the references to the yaml from the JMA website for users to view for more information.

mraspaud (approved):
LGTM! Gorgeous renderings, great work @mherbertson

simonrp84:
Hi, @mherbertson, I've just tried this out and unfortunately have some trouble with the results.

This is the view from the Himawari NICT website:


And this is how it looks from Satpy with this PR:
AHI_true_color_reproduction_20210906_1800_BIG

Both images are 18:00 UTC yesterday, 6th Sept. The `true_color_reproduction` version seems very overexposed compared to the NICT version...

mherbertson:
Thanks for checking out the composite @simonrp84

From memory NICT is uncorrected imagery. 

You'll get more closer results on [https://www.data.jma.go.jp/mscweb/data/himawari/index.html](url) and selecting True Color Reproduction Image as the band. Unfortunately FLDK is very low resolution.

Here is the equivalent image from JMA for that timestamp. [https://www.data.jma.go.jp/mscweb/data/himawari/img/fd_/fd__trm_1800.jpg](url)

fd__trm_1800

There will be brightness issues as the crude stretch gave the best results at full light, again this was a compromise for more efficient processing. You could try a custom linear stretch for your particular use case.

Let me know if I can help!

simonrp84:
The NICT imagery is corrected, but using a slightly different version of the true color representation method. As mentioned by @djhoese on slack regarding GOES/ABI, adjusting the Rayleigh scattering using the IR brightness temperatures will help a little bit here. I'm investigating that now.

mherbertson:
Revisiting this again.

Comparing the CIRA Geocolor (as far as I know where some of this composite is based off) they're both washed out in the same area.  I think to get the same visuals as NICT would need a whole other composite or some tweaking of the other base enhancements of rayleigh/sunz/IR temp. Thoughts?

Geocolor:
202111031730_geo_color

This composite:
202111031730_true_color_reproduction

satpy/composites/ahi.py
```diff
@@ -27,17 +28,17 @@
 class GreenCorrector(GenericCompositor):
     """Corrector of the AHI green band to compensate for the deficit of chlorophyll signal."""
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, fractions=(0.85, 0.15), **kwargs):
```
mraspaud, line 31: looks much better with fractions in the constructor signature! thanks

satpy/etc/composites/ahi.yaml
```diff
@@ -14,6 +14,19 @@ modifiers:
     - solar_azimuth_angle
     - solar_zenith_angle
 
+  true_color_reproduction_rayleigh_corrected:
```
mraspaud, line 17: This is probably not specific to the true color reproduction composite, so could it be rename to eg: `no_aerosol_rayleigh_corrected` ?

satpy/tests/enhancement_tests/test_ahi.py
```diff
@@ -0,0 +1,52 @@
...
+import unittest
+import numpy as np
+import xarray as xr
+import dask.array as da
+
+
+class TestAHIEnhancement(unittest.TestCase):
+    """Test the AHI enhancement functions."""
+
+    def setUp(self):
```
mraspaud, line 29: Moving towards pytest, this should be equivalent

```suggestion
class TestAHIEnhancement():
    """Test the AHI enhancement functions."""

    def setup(self):
```

mraspaud (approved):
Looks good to me! I have two minor stylistic comment, otherwise I think it's good to merge!

satpy/etc/composites/ahi.yaml
mherbertson: good idea!

mraspaud:
pre-commit.ci autofix

simonrp84 (approved):
LGTM!

PR #1677, opened 2021-05-17
Fix SATPY_CONFIG_PATH being unusable when imported in a subprocess

### LINKED ISSUE #1676: New config feature does not support a subprocess call to another script which uses satpy too.
**Describe the bug**
The introduction of the new satpy config feature broke the back compatibility to support a subprocess call from a script uses satpy to another script which uses satpy too. Because at the caller level of satpy config ([L76 of _config.py](https://github.com/pytroll/satpy/blob/main/satpy/_config.py#L76)) it **modified** the environment variable SATPY_CONFIG_PATH as a string representation of a python list,  however, the satpy config module at the callee level in the subprocess call, it just parse the string representation of the list as a string,  which is not correct and case the config path variable is invalid in the downstream.

**To Reproduce**
```
export SATPY_CONFIG_PATH=/myconfig1:/myconfig2
python ./test_subprocess.py
``` 
where test_subprocess.py is as below
```python
import satpy
import subprocess

subprocess.run('''python -c "import satpy; print(satpy.config.get('config_path'))"''', shell=True)
```

**Expected behavior**
It should be a list contains two entries as:
['/myconfig1', '/myconfig2']

**Actual results**

The config_path become a list single entry with a not valid path as:
["**['/myconfig1', '/myconfig2']**"]

**Environment Info:**
 - OS: OSX and Linux Ubuntu 18.04
 - Satpy Version: 0.28.0

yufeizhu600:
This PR try to add back the feature to support to call another script which also uses satpy in a python subprocess call. See Issue #1676 for detail of the bug. 

 - [x] Closes #1676 
 - [x] Tests added

### DIFF
--- a/satpy/tests/test_config.py
+++ b/satpy/tests/test_config.py
@@ -166,6 +166,7 @@ class TestConfigObject:
 
     def test_config_path_multiple_load(self):
         """Test that config paths from subprocesses load properly.
+
         Satpy modifies the config path environment variable when it is imported.
         If Satpy is imported again from a subprocess then it should be able to parse this
         modified variable.
--- a/satpy/tests/test_config.py
+++ b/satpy/tests/test_config.py
@@ -165,8 +165,9 @@ class TestConfigObject:
                                                        '/my/configs3']
 
     def test_config_path_multiple_load(self):
-        """Test that config paths are parsed properly after multiple load of satpy configs,
-           for example a subprocess call from a python script."""
+        """Test that config paths are parsed properly after multiple load of
+           satpy configs, for example a subprocess call from a python script.
+        """
         from importlib import reload
         import satpy
         old_vars = {
--- a/satpy/tests/test_config.py
+++ b/satpy/tests/test_config.py
@@ -165,8 +165,11 @@ class TestConfigObject:
                                                        '/my/configs3']
 
     def test_config_path_multiple_load(self):
-        """Test that config paths are parsed properly after multiple load of
-           satpy configs, for example a subprocess call from a python script.
+        """Test that config paths from subprocesses load properly.
+        
+        Satpy modifies the config path environment variable when it is imported.
+        If Satpy is imported again from a subprocess then it should be able to parse this
+        modified variable.
         """
         from importlib import reload
         import satpy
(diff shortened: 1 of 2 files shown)


djhoese (approved):
Looks good. I kept trying to come up with ways so that `ast` wouldn't have to be used, but I see that because it is used later on with `_CONFIG_PATHS` we have to parse it completely before hand.

I see this is a draft. What is left to do?

satpy/tests/test_config.py
```diff
@@ -164,6 +164,28 @@ def test_config_path_multiple(self):
                                                        '/my/configs2',
                                                        '/my/configs3']
 
+    def test_config_path_multiple_load(self):
+        """Test that config paths are parsed properly after multiple load of
+           satpy configs, for example a subprocess call from a python script.
+        """
```
djhoese, line 170:
```suggestion
        """Test that config paths from subprocesses load properly.
        
        Satpy modifies the config path environment variable when it is imported.
        If Satpy is imported again from a subprocess then it should be able to parse this
        modified variable.
        """
```

satpy/tests/test_config.py
```diff
@@ -164,6 +164,31 @@ def test_config_path_multiple(self):
                                                        '/my/configs2',
                                                        '/my/configs3']
 
+    def test_config_path_multiple_load(self):
+        """Test that config paths from subprocesses load properly.
+
```
djhoese, line 169:
```suggestion

```

satpy/tests/test_config.py
```diff
@@ -164,6 +164,30 @@ def test_config_path_multiple(self):
                                                        '/my/configs2',
                                                        '/my/configs3']
 
+    def test_config_path_multiple_load(self):
+        """Test that config paths from subprocesses load properly.
+        Satpy modifies the config path environment variable when it is imported.
+        If Satpy is imported again from a subprocess then it should be able to parse this
+        modified variable.
```
djhoese, line 171:
```suggestion
        """Test that config paths from subprocesses load properly.

        Satpy modifies the config path environment variable when it is imported.
        If Satpy is imported again from a subprocess then it should be able to parse this
        modified variable.
```

yufeizhu600:
Thanks for fixing the lint issue. Yes it is good to go.

PR #1680, opened 2021-05-19
Implement reading of index map and auxiliary data in FCI L1c reader

ameraner:
This PR add the functionality of reading the index map and the related auxiliary geometric/acquisition parameters in the FCI L1c reader. 

 - [x] Tests added


ameraner:
A quick overview of the index map and auxiliary datasets loaded with this PR on the latest released test data:

image

adybbroe:
@ameraner Can you explain for a dummy, what an `index map` is in this context and perhaps explain why it is needed/important?

satpy/tests/reader_tests/test_fci_l1c_fdhsi.py
```diff
@@ -400,6 +422,52 @@ def test_load_bt(self, reader_configs, caplog):
...
+        res = reader.load(
+            [name + '_index_map' for name in
+             self._chans["solar"] + self._chans["terran"]], pad_data=False)
+        assert 16 == len(res)
+        for ch in self._chans["solar"] + self._chans["terran"]:
+            assert res[ch + '_index_map'].shape == (200, 11136)
+            numpy.testing.assert_array_equal(res[ch + '_index_map'][1, 1], 5138)
+
+    def test_load_aux_data(self, reader_configs):
+        """Test loading of index_map."""
```
mraspaud, line 447:
```suggestion
        """Test loading of auxiliary data."""
```

satpy/tests/reader_tests/test_fci_l1c_fdhsi.py
```diff
@@ -400,6 +422,52 @@ def test_load_bt(self, reader_configs, caplog):
...
+
+        filenames = [
+            "W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--"
+            "CHK-BODY--L2P-NC4E_C_EUMT_20170410114434_GTT_DEV_"
+            "20170410113925_20170410113934_N__C_0070_0067.nc"
+        ]
+
+        reader = load_reader(reader_configs)
+        loadables = reader.select_files_from_pathnames(filenames)
+        reader.create_filehandlers(loadables)
```
mraspaud, line 459: these lines could make a small function and be factorized across the tests.

mraspaud (commented):
Thanks for all the hard work with adding all the datasets! I have some small questions and suggestions inline, otherwise it looks good!

ameraner:
@adybbroe The index map is basically an array assigning integer indices to each image pixel. Each index represents the number of fixed time intervals passed since the start of the repeat cycle. The time interval is variable, currently it's set to 0.1s. So, in principle, the index map alone is a proxy of the acquisition time for each pixel, with a 0.1s precision.

Now, inside the files there are several 1-d arrays containing geometric and acquisition parameters, valid at each time interval within the repeat cycle. With a time interval of 0.1s and a repeat cycle of 10 min, there are 600s/0.1s = 6000 values in each 1-d array. By using the indices inside the index map (that go from 1 to 6000), you can use the 1-d arrays as a sort of look-up-tables to retrieve the geometric/acquisition parameters valid for each image pixel. By looking up the value for each pixel, you obtain the arrays I'm plotting above.

TLDR: In other words, the index map is a mechanism to allow the extraction of per-pixel geometric/acquisition parameters, with a limited memory size, at cost of having "discretised" time steps (limited temporal resolution).

This is useful for users wanting to precisely analyse the scenes, do accurate calibrations, etc. I suppose most users will only be interested in the acquisition time array.

satpy/tests/reader_tests/test_fci_l1c_fdhsi.py
ameraner: true, thanks!

satpy/readers/fci_l1c_fdhsi.py
```diff
@@ -220,55 +263,78 @@ def _get_dataset_measurand(self, key, info=None):
...
+    @staticmethod
+    def _getitem(block, lut):
+        return lut[block.astype('uint16')]
+
+    def _get_dataset_aux_data(self, dsname):
+        """Get the auxiliary data arrays using the index map."""
+        # get index map
+        index_map = self._get_dataset_index_map(_get_channel_name_from_dsname(dsname))
+        # subtract one as index map indexing starts from 1
+        index_map += -1
```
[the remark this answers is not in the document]
ameraner: Done, thanks!

satpy/tests/reader_tests/test_fci_l1c_fdhsi.py
ameraner: Done, thanks!

satpy/readers/fci_l1c_fdhsi.py
```diff
@@ -220,55 +268,75 @@ def _get_dataset_measurand(self, key, info=None):
...
+    @staticmethod
+    def _getitem(block, lut):
+        return lut[block.astype('uint16')]
+
+    def _get_dataset_aux_data(self, dsname):
+        """Get the auxiliary data arrays using the index map."""
+        # get index map
+        index_map = self._get_dataset_index_map(_get_channel_name_from_dsname(dsname))
+        # index map indexing starts from 1
+        index_map -= 1
```
mraspaud, line 296: I think this belongs to the `_get_dataset_index_map` method.

satpy/readers/fci_l1c_fdhsi.py
```diff
@@ -220,55 +268,75 @@ def _get_dataset_measurand(self, key, info=None):
...
+
+        # get lut values from 1-d vector
+        lut = self[AUX_DATA[_get_aux_data_name_from_dsname(dsname)]]
+
+        fv = default_fillvals.get(lut.dtype.str[1:], np.nan)
+        lut = lut.where(lut != fv)
+
```
mraspaud, line 303: Could this be delegated to its own method?

mraspaud (commented):
LGTM, I just have a couple of style comments inline.

satpy/readers/fci_l1c_fdhsi.py
ameraner: I would actually like to keep it here: this way, a user that is only interested in reading the index map for his own use, gets the values as they're stored in the file and described in the docs. 
The `-= 1` is only relevant when using the index map in combination with the lut (so in `_get_dataset_aux_data`), as the (python) indexing needs to start from 0.

satpy/readers/fci_l1c_fdhsi.py
ameraner: Done, thanks!

PR #1691, opened 2021-05-26
Fix reference to sector_id global key in 'awips_tiled' writer YAML

joleenf:
This edits one line in awips_tiled.py so that the key "add_sector_id_global" is used for applying the sector_id to the global attributes.  This key is chosen since it matches the key defined in the awips_tiled.yaml file 
https://github.com/pytroll/satpy/blob/88938ab8e0eb2b2f6ae89665733b7e2281720fce/satpy/etc/writers/awips_tiled.yaml#L58

### DIFF
--- a/satpy/etc/writers/awips_tiled.yaml
+++ b/satpy/etc/writers/awips_tiled.yaml
@@ -118,7 +118,7 @@ templates:
         var_name: data
         attributes:
           physical_element:
-            raw_value: 'CLAVR-x {name}'
+            value: 'CLAVR-x {name}'
       clavrx_cloud_type:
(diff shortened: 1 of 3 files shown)


djhoese:
Do you think you could add a check for the sector_id in one of the tests, maybe this one:

https://github.com/pytroll/satpy/blob/88938ab8e0eb2b2f6ae89665733b7e2281720fce/satpy/tests/writer_tests/test_awips_tiled.py#L159

and make sure that it fails without your change and passes with it?

djhoese (approved):
LGTM. Thanks.

PR #1692, opened 2021-05-26
Add raw 'counts' calibration to 'abi_l1b' reader

tommyjasmin:
Adds ability to return the raw counts as a calibration type. This is sometimes used by McIDAS clients, for example when copying images from an ADDE server.

 - [x] Tests added 
 - [x] Fully documented 


djhoese:
Failures are caused by a new version of xarray and bad usage of creating Dataset objects in one of Satpy's readers. Fixes are in https://github.com/pytroll/satpy/pull/1777 and will be merged. I don't see any failures of your tests so far.

satpy/etc/readers/abi_l1b.yaml
```diff
@@ -102,6 +102,9 @@ datasets:
       reflectance:
         standard_name: toa_bidirectional_reflectance
         units: "%"
+      counts:
+        standard_name: raw_counts
```
djhoese, line 106: Where is the name "raw_counts" from? The existing readers that have counts use `standard_name: counts`.

satpy/readers/abi_l1b.py
```diff
@@ -38,21 +38,27 @@ class NC_ABI_L1B(NC_ABI_BASE):
...
-        if key['calibration'] == 'reflectance':
-            logger.debug("Calibrating to reflectances")
-            res = self._vis_calibrate(radiances)
-        elif key['calibration'] == 'brightness_temperature':
-            logger.debug("Calibrating to brightness temperatures")
-            res = self._ir_calibrate(radiances)
-        elif key['calibration'] != 'radiance':
-            raise ValueError("Unknown calibration '{}'".format(key['calibration']))
+        # For raw cal, don't apply scale and offset, use true file counts
+        if (key['calibration'] == 'counts'):
```
djhoese, line 42:
```suggestion
        if key['calibration'] == 'counts':
```

satpy/readers/abi_l1b.py
```diff
@@ -38,21 +38,27 @@ class NC_ABI_L1B(NC_ABI_BASE):
...
+        }
+
+        try:
+            res = cal_dictionary[key['calibration']](radiances)
+        except KeyError:
+            raise ValueError("Unknown calibration '{}'".format(key['calibration']))
 
         # convert to satpy standard units
-        if res.attrs['units'] == '1':
+        if (res.attrs['units'] == '1') and (key['calibration'] != 'counts'):
```
djhoese, line 61:
```suggestion
        if res.attrs['units'] == '1' and key['calibration'] != 'counts':
```

satpy/etc/readers/abi_l1b.yaml
djhoese: Ah the standard_name from the YAML isn't actually used as far as I can tell. The calibrate method sets its own information. This should probably be updated to match that or removed.

satpy/readers/abi_l1b.py
```diff
@@ -38,21 +38,27 @@ class NC_ABI_L1B(NC_ABI_BASE):
...
-            logger.debug("Calibrating to reflectances")
-            res = self._vis_calibrate(radiances)
-        elif key['calibration'] == 'brightness_temperature':
-            logger.debug("Calibrating to brightness temperatures")
-            res = self._ir_calibrate(radiances)
-        elif key['calibration'] != 'radiance':
-            raise ValueError("Unknown calibration '{}'".format(key['calibration']))
+        # For raw cal, don't apply scale and offset, use true file counts
+        if (key['calibration'] == 'counts'):
+            radiances = self.nc['Rad']
```
djhoese, line 43: I'm not a huge fan of something being called radiances and meaning two different things depending on the calibration. Others may disagree with me but what about a shortcut here like:

```
if key['calibration'] == 'counts':
    return self._raw_calibrate(self.nc['Rad'])
```

Then you can remove the `'counts': self._raw_calibrate` entry in the dict below.

Open for discussion, just a feeling.

satpy/tests/reader_tests/test_abi_l1b.py
```diff
@@ -269,6 +269,42 @@ def test_vis_calibrate(self):
...
+        super(Test_NC_ABI_L1B_raw_cal, self).setUp(rad=rad)
+
+    def test_raw_calibrate(self):
+        """Test RAW calibration."""
+        res = self.reader.get_dataset(
+            make_dataid(name='C05', calibration='counts'), {})
+
+        # We expect the raw data to be unchanged
+        expected = res.data
+        self.assertTrue(np.allclose(res.data, expected, equal_nan=True))
```
djhoese, line 298: Could you add an assertion here that `res.data.dtype == np.int16`? Basically to make sure the data isn't cast to another data type. I think `==` is the preferred way of comparing dtypes now. Or maybe it should be `res.data.dtype.type`.

djhoese (changes requested):

satpy/etc/readers/abi_l1b.yaml
tommyjasmin: Agreed, we made this change Dave.

satpy/readers/abi_l1b.py
tommyjasmin: Agreed, change has been made Dave.

satpy/readers/abi_l1b.py
tommyjasmin: Agreed, change has been made Dave.

satpy/tests/reader_tests/test_abi_l1b.py
```diff
@@ -269,6 +269,43 @@ def test_vis_calibrate(self):
...
+        self.assertTrue(res.data.dtype == np.int16, "int16 data type expected")
```
djhoese, line 299:
```suggestion
        self.assertEqual(res.data.dtype, np.int16, "int16 data type expected")
```

djhoese (approved):
I made one small change to the test assert that you added. Once the tests pass I think we can merge this!

PR #1710, opened 2021-06-01
Fix NUCAPS reader having incorrect _FillValue attribute

### LINKED ISSUE #1104: NUCAPS reader uses incorrect _FillValue
**Describe the bug**
Tommy and I noticed while looking at data from the NUCAPS reader that floating point variables with a `_FillValue` attribute set to `-9999.0` were replacing fill values with `NaN` but still had the attribute `_FillValue` set to `-9999.0`. The reader should be updated to remove `_FillValue` from the metadata (`get_metadata` method I think).

CC @tommyjasmin

djhoese:
The `_FillValue` attribute was being returned in NUCAPS DataArrays even after they had been masked out and replaced with NaNs. This PR makes sure the attribute is removed.

CC @tommyjasmin and @joleenf who were the most recent people to modify this reader.

 - [x] Closes #1104 
 - [x] Tests added

### DIFF
--- a/satpy/readers/nucaps.py
+++ b/satpy/readers/nucaps.py
@@ -221,6 +221,9 @@ class NUCAPSFileHandler(NetCDF4FileHandler):
             data = data.where((data <= valid_max))  # | (data >= valid_min))
         if fill_value is not None:
             data = data.where(data != fill_value)
+            # this _FillValue is no longer valid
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM, feel free to merge

PR #1711, opened 2021-06-01
Rewrite installation instructions to make conda use clearer

### LINKED ISSUE #1060: Reorder installation instructions to put conda before PyPI
We often get users who have difficulty installing Satpy and its dependencies using PyPI. I noticed in our installation instructions that PyPI is listed first. What do we think about putting conda before PyPI and emphasizing that conda is highly recommended due to the number of complex C dependencies we have?

djhoese:
We generally tell new users to use conda for installing Satpy. This PR rearranges the installation instructions so conda comes before pip and also adds information about miniconda.

 - [x] Closes #1060 
 - [x] Fully documented

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -23,20 +23,41 @@ to always use conda-forge by running:
 
     $ conda config --add channels conda-forge
 
-We recommend creating a separate environment for your work with Satpy. If
-you haven't created and activated one already, you can by running:
+In a new conda environment
+--------------------------
+
+We recommend creating a separate environment for your work with Satpy. To
+create a new environment and install Satpy all in one command you can
+run:
+
 
 .. code-block:: bash
 
-    $ conda create -c conda-forge -n my_satpy_env python
+    $ conda create -c conda-forge -n my_satpy_env python satpy
+
+You must then activate the environment so any future python or
+conda commands will use this environment.
+
+.. code-block::
+
     $ conda activate my_satpy_env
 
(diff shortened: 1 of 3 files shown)


mraspaud (approved):
LGTM

doc/source/install.rst
```diff
@@ -47,7 +28,7 @@ you haven't created and activated one already, you can by running:
 
 .. code-block:: bash
 
-    $ conda create -n my_satpy_env python
+    $ conda create -c conda-forge -n my_satpy_env python
```
pnuu, line 31: Would it be better to advise to install `satpy` directly at env creation to make the installation faster? Or add a short comment like "It is also possible to define all the wanted packages in the environment creation command, which makes the installation faster and might solve some dependency issues between package versions."

pnuu (approved):
LGTM. Just one comment.

doc/source/install.rst
djhoese: Yeah good point. I'll see if I can restructure this.

djhoese:
Not sure if the two `note` sections look bad in the "existing" section.

image

mraspaud:
Could the second note link to the `check_satpy` utility?

mraspaud:
And the notes are fine.

djhoese:
> Could the second note link to the check_satpy utility?

Done.

mraspaud (approved):
LGTM

PR #1716, opened 2021-06-03
Update Clavrx reader for netcdf files

### LINKED ISSUE #1661: Support for clavrx netcdf files
## Feature Request

The clavrx team is now producing netcdf files in real-time, for example see files in the AHI directory:  ftp://ftp@ftp.ssec.wisc.edu/pub/clavrx/real_time/level2/ahi/

Currently, the clavrx reader only supports hdf files.

**Solution**
If possible, update the current reader to support both the old hdf and new netcdf files.  If that is not the best solution, create a reader to support the new netcdf files.

joleenf:
The PR updates the current CLAVRX HDF reader to include a netcdf reader for geostationary files currently being produced. 

 - [ ] Closes #1661
 - [ ] Tests for netcdf still needed

### DIFF
--- a/satpy/readers/clavrx.py
+++ b/satpy/readers/clavrx.py
@@ -19,10 +19,14 @@
 
 import os
 import logging
+
 import numpy as np
 import netCDF4
+import xarray as xr
 from glob import glob
 from satpy.readers.hdf4_utils import HDF4FileHandler, SDS
+from satpy.readers.file_handlers import BaseFileHandler
+from satpy import CHUNK_SIZE
 from pyresample import geometry
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


joleenf:
The PR updates the current CLAVRX HDF reader to include a netcdf reader for geostationary files currently being produced.

Closes #1661
Tests for netcdf reader have been added.

satpy/readers/clavrx.py
```diff
@@ -32,183 +37,103 @@
...
-                    new_info['coordinates'] = coordinates
-                yield True, new_info
-            elif is_avail is None:
-                # if we didn't know how to handle this dataset and no one else did
-                # then we should keep it going down the chain
-                yield is_avail, ds_info
+    if factor is not None and offset is not None:
+        def scale_inplace(data):
+            data *= factor
+            data += offset
```
djhoese, line 107: I'm not sure I feel strongly enough about this to request a change, but we have to be careful with inplace operations on dask-backed DataArrays. I've had problems in some code where multiplying an integer array by a float (factor) can lead to an error because the result of `int * float` should be a float but the inplace array is `int`. If this doesn't need to be in-place, could you make it not in-place?

satpy/readers/clavrx.py
```diff
@@ -32,183 +37,103 @@
...
-        data = data.where(data != fill)
-        scale_inplace(data)
-        if valid_range is not None:
-            valid_min, valid_max = scale_inplace(valid_range[0]), scale_inplace(valid_range[1])
-            data = data.where((data >= valid_min) & (data <= valid_max))
-            data.attrs['valid_min'], data.attrs['valid_max'] = valid_min, valid_max
-
-        return data
+class _CLAVRxHelper:
+    """A base class for the CLAVR File Handlers."""
```
djhoese, line 136:
```suggestion
    """A base class for the CLAVRx File Handlers."""
```

satpy/readers/clavrx.py
```diff
@@ -305,16 +237,257 @@ def _read_axi_fixed_grid(self, l1b_attr):
...
+        return _CLAVRxHelper._read_axi_fixed_grid(self.filename, l1b_att)
+
+    def get_metadata(self, attrs, ds_info):
+        """Get metadata."""
+        i = {}
+        i.update(attrs)
+        i.update(ds_info)
+
+        flag_meanings = i.get('flag_meanings', None)
+        if not i.get('SCALED', 1) and not flag_meanings:
```
djhoese, line 463: The NetCDF files use `SCALED`? Not a more CF standard `scale_factor`? :cry:

satpy/tests/reader_tests/test_clavrx_nc.py
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 Satpy developers
```
djhoese, line 3:
```suggestion
# Copyright (c) 2021 Satpy developers
```

djhoese (changes requested):
A couple small changes, otherwise this looks pretty good to me. You mentioned you were worried about using the utilities classes correctly. Was there something specific you wanted me to look at?

satpy/readers/clavrx.py
joleenf: Yes, I can change that.  I tried to change as little as possible from the previous code, but I am not attached to the scale in place operations.

satpy/readers/clavrx.py
djhoese: I would assume that as clavrx development continues that things become more and more CF-y so if `scale_factor` and `add_offset` are there, let's use those.

satpy/readers/clavrx.py
djhoese: It may not actually be a problem, but with dask there also shouldn't be a point to doing it...so let's go not in place.

satpy/readers/clavrx.py
joleenf: Actually, I am not sure I should change this.  This isolates the variables that are flags well without pulling in extra variables which are not.  If I would switch to scale_factor, there is one an extra variable which would be picked up that probably shouldn't be assigned a flag value unknown attribute.

djhoese: I think I'm misunderstanding what the original purpose of this code is. What do you mean by "extra variables"? Aren't SCALED and scale_factor just attributes in the dictionary?

I realize now that this if statement isn't controlling if scale/offset are applied which is what I originally thought it was. I'm now fine with however you have to make this work. I assumed SCALED and scale_factor were the same thing and that you were then going to use SCALED to scale the data later. So...maybe just ignore that I said anything?

joleenf: I know this is no excuse, but that was in the code before I started working on it.  As far as I can see, this does not do anything to the data scaling, that is handled with scale_factor and add_offset in get_dataset.  This adds some attribute information to the data variable: if these are flags, and for some reason, they do not have "flag_meaning" metadata to describe them, add a "meaning unknown" attribute to the metdata.  All the flag variables + 1 "duplicate" variable have a scale_factor of 1.0, but the duplicate variable is SCALED as 1b versus the flags which are 0b.

djhoese: Sounds good.

djhoese (approved):

PR #1726, opened 2021-06-11
Point out get_area_def in resample documentation

gerritholl:
In the main resample documentation, mention the nice satpy utility
function get_area_def.

 - [x] Closes #1726
 - [x] Fully documented


djhoese (approved):
My excuse for this is that this function wasn't supposed to be public originally, but it seems it has become more and more useful and needed. Thanks for this update.

PR #1733, opened 2021-06-21
Fix geotiff writer ignoring dtype argument

### LINKED ISSUE #1730: geotiff writer ignores dtype argument, always writes float if enhance=False
**Describe the bug**

When writing a geotiff with `enhance=False`, satpy always writes floating points, even if `dtype="uint8"` is passed.  For single channels, this leads to an unwanted dtype in the image.  For RGBs, this leads to ValueError.

**To Reproduce**

```python
import os
from glob import glob
from satpy import Scene
sc = Scene(filenames=seviri_files, reader=["seviri_l1b_hrit"])
sc.load(["IR_108", "airmass"])
ls = sc.resample("eurol")
ls.save_dataset("IR_108", enhance=False, dtype="uint8")
ls.save_dataset("airmass", enhance=False, dtype="uint8")
```

**Expected behavior**

I expect that two unenhanced images with dtype=uint8 are written.

**Actual results**

For the single channel image, `gdalinfo` reveals:

```
Band 1 Block=2560x1 Type=Float32, ColorInterp=Gray
```

For the RGB, Python throws a ValueError:

```
  File "/home/gholl/checkouts/protocode/mwe/rgb-unenhanced.py", line 10, in <module>
    ls.save_dataset("airmass", enhance=False, dtype="uint8")
  File "/home/gholl/checkouts/satpy/satpy/scene.py", line 986, in save_dataset
  File "/home/gholl/checkouts/satpy/satpy/writers/__init__.py", line 811, in save_dataset
    return self.save_image(img, filename=filename, compute=compute, fill_value=fill_value, **kwargs)
  File "/home/gholl/checkouts/satpy/satpy/writers/geotiff.py", line 208, in save_image
    raise ValueError("Image must be in 'L' mode for floating "
ValueError: Image must be in 'L' mode for floating point geotiff saving
```

**Environment Info:**
 - Satpy Version: v0.29.0-51-g7b8eb8d4 (latest main at the time of writing)
 - PyResample Version: v1.12.0
 - trollimage version: v1.15.0-1-g6a663c1

**Additional context**

`git bisect` reveals this bug was introduced in #1499.

BENR0:
This fixes #1730

 - [x] Closes #1730 
 - [x] Tests added

### DIFF
--- a/satpy/tests/writer_tests/test_geotiff.py
+++ b/satpy/tests/writer_tests/test_geotiff.py
@@ -111,7 +111,7 @@ class TestGeoTIFFWriter(unittest.TestCase):
         w.save_datasets(datasets)
 
     def test_dtype_for_enhance_false(self):
-        """Test that dtype of dataset is used if enhance=False."""
+        """Test that dtype of dataset is used if parameters enhance=False and dtype=None."""
         from satpy.writers.geotiff import GeoTIFFWriter
         datasets = self._get_test_datasets()
         w = GeoTIFFWriter(base_dir=self.base_dir, enhance=False)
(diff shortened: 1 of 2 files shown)


djhoese:
> While this definitely is a bug it seems to me that saving an unenhanced image with "unit8" is in most cases not really wanted because there will be clipping of the data. Maybe a warning should be added for the case when clipping possibly might occur?

I'm inclined to avoid the "annoying" warning and assume that the user understands what they are saying when they do this. But I don't feel that strongly. I'm also not sure how common this "most cases" is. For reflectance data this wouldn't be that bad (major loss of precision, but still). For BT data, yeah not good. For RGBs, yeah not great either probably.

If you want to add the warning go for it. If @mraspaud or @pnuu or @gerritholl have feelings about this, speak now.

BENR0:
> I'm inclined to avoid the "annoying" warning and assume that the user understands what they are saying when they do this

I totally agree. I just suggested it because some users might not be aware and would wonder why their image looks so weird :-D. So if anyone want's me to add it I will, otherwise I am ok with how it is.

gerritholl:
In DWD we've been generating images with `stretch="no"` but I don't know how we've prevented problems with values not between 0 and 256.

djhoese (approved):

djhoese:
Seems I lost track of this. Thanks again @BENR0.

PR #1734, opened 2021-06-22
Remove valid_range from attributes in VIRR L1b reader

joleenf:
The valid_range is removed from the attributes after it is a applied to the data.  The format as a numpy array was not working with another writer.  This was chosen over the alternative approach to make the valid_range a list.  The valid_range was in the attributes but not scaled to match the final data output and so it didn't seem to have much relevant meaning.  

Test has been added to make sure that valid_range is not in the final dataset attributes.


djhoese (approved):
LGTM

PR #1737, opened 2021-06-23
Add additional GLM accumulated product types and overlays

deeplycloudy:
- [ ] Tests added 
 - [ ] Fully documented 


deeplycloudy:
As part of a recent project for EUMETSAT, I produced a variety of GLM overlays on ABI data. This PR includes all of the "operational" US products, as well as a new product type produced by `glmtools` that mimics the construction of a future MTG LI accumulated product. I also changed the overlay channel to ABI Ch13 (10µm or so) to match the MTG FCI channel that was the best thermal IR match to what is available on ABI.

I'm opening this PR for discussion about whether these overlays belong in the main SatPy repo, or if these would be better kept private / particular to a specific project. I assume it's somehow possible to use these `.yaml`s independently … just haven't looked up that documentation.

djhoese:
> I'm opening this PR for discussion about whether these overlays belong in the main SatPy repo, or if these would be better kept private / particular to a specific project

I think these are good for the Satpy package. We don't have a lot of GLM composites and these being "operational" (whether in quotes or not) is a good argument for getting them out to the Satpy community. I think the only time I would turn down a contribution like this is if it required really complicated python code and/or was only useful to a very small group. These are great.

I'd like to hear from some of the other maintainers, but I know @adybbroe and @mraspaud are on holiday. I think @gerritholl has done some stuff with GLM and ABI so he might be interested in these. Bottom line, I'm hoping for at least one other review, but I think these are great.

Thanks @deeplycloudy

satpy/etc/enhancements/glm.yaml
```diff
@@ -17,6 +82,13 @@ enhancements:
           stretch: crude
           min_stretch: [0, 0, 0]
           max_stretch: [1, 1, 1]
+  # c14_total_energy:
+  #   standard_name: c14_total_energy
+  #   operations:
+  #     - name: gamma
+  #       method: !!python/name:satpy.enhancements.gamma
+  #       kwargs:
+  #         gamma: [1.4, 1, 1]
```
pnuu, line 91: Can this be removed?

pnuu (commented):
These are very interesting while waiting for LI to be available! 

One thing I think is required is crude stretch from 0 to 1 for the `C14_*` composites. Without those, every result will be stretched using the default `linear` stretch, and the results will vary depending on the actual data in the image.

And the one commented enhancement should be removed if not inteded for use.

satpy/etc/enhancements/glm.yaml
deeplycloudy: Yes, will do.

deeplycloudy:
> One thing I think is required is crude stretch from 0 to 1 for the `C14_*` composites. Without those, every result will be stretched using the default `linear` stretch, and the results will vary depending on the actual data in the image.

I'm happy to add the stretch as you describe, though I don't know how to add stretch to `composites/glm.yaml`. There are no examples of that in the ABI or SEVIRI composites. Do you mean I should add stretch to `enhancements/glm.yaml` like in the existing `c14_flash_extent_density`? 

```
  c14_flash_extent_density:
    standard_name: c14_flash_extent_density
    operations:
      - name: stretch
        method: !!python/name:satpy.enhancements.stretch
        kwargs:
          stretch: crude
          min_stretch: [0, 0, 0]
          max_stretch: [1, 1, 1]
```

So if I understand, what I need to do is replicate each of the composites with a corresponding enhancements? I'm happy to do so, but wanted to be sure I understood.

djhoese:
> Do you mean I should add stretch to enhancements/glm.yaml like in the existing c14_flash_extent_density?

Yes.

> So if I understand, what I need to do is replicate each of the composites with a corresponding enhancements?

Yes.

Note in your enhancement you could specify the `standard_name:` to match the standard_name you defined in your composite definition or you could specify `name:` to match the name of your composite. You *could* make the `standard_name:` for all of your composites the same (ex. `standard_name: glm_abi_overlay`) and then you'd only have to specify a single enhancement (ex. `standard_name: glm_abi_overlay`). What is best up to you. I don't have a preference. Your current standard names are more specific, but the `glm_abi_overlay` route has more future flexibility and less enhancement configuration. You don't have to use `glm_abi_overlay` either, that's just a descriptive example. I could also see this backfiring if any overlay composites you want to make aren't RGB (grayscale ABI data + grayscale GLM data) as that would require a different enhancement configuration.

...and I'm sure I've confused you now.

deeplycloudy:
@djhoese thanks for the advice, as I figured I should somehow be able to avoid repeating the enhancement definitions at the overlay step. @pnuu your requested change is in, too. 

I've rerun a few of my basic overlay plots and nothing broke. Do I need to add any other docs or unit tests somewhere, per the checklist? I'm happy to do so but I would need to learn how SatPy handles adding test datasets and documentation of available overlays.

I'm also tagging @gerritholl who might find this work interesting given his recent issues opened re: blending GLM data into other scenes.

simonrp84:
I'm quite interested in the GLM data, so will have a play with these new composites in the next few days. Can't promise I can provide much useful feedback but will do what I can, thanks for adding these :)

djhoese:
We generally don't have any tests specifically for new composites that don't involve new Python code (we test the python code, not necessarily the configuration in the YAML using it beyond basic syntax). The new reader datasets would be nice to add to:

https://github.com/pytroll/satpy/blob/main/satpy/tests/reader_tests/test_glm_l2.py

But for unit tests we typically make fake data as you can see near the top of the module. This isn't a "real" test of reading the data, but for now it is what we have.

djhoese:
> The new reader datasets would be nice to add to:
>> https://github.com/pytroll/satpy/blob/main/satpy/tests/reader_tests/test_glm_l2.py
>
> But for unit tests we typically make fake data as you can see near the top of the module. This isn't a "real" test of reading the data, but for now it is what we have.

@deeplycloudy Sorry for losing track of this. We are preparing for a new Satpy release and this PR is basically ready to go. Do you have time to look at adding some of the new reader datasets to the glm reader tests?

PR #1751, opened 2021-07-02
Improve documentation for MultiScene.blend

gerritholl:
In the `MultiScene.blend` docstring, explain what it actually does and
what the interface of the blend function should be.

 - [x] Closes #1750 
 - [x] Fully documented

Issue #1747, opened 2021-06-30
Load composites without file handlers.


gerritholl:
## Feature Request

**Is your feature request related to a problem? Please describe.**

Composite loading depends on file handles.  If I manually assign datasets to a scene object, I cannot use the scene class to calculate a composite from those datasets.  Toy example:

```
sc = Scene()
sc["flash_extent_density"] = xarray.DataArray(np.arange(100).reshape(10, 10), dims=("y", "x"))
sc["C14"] = xarray.DataArray(np.arange(100).reshape(10, 10), dims=("y", "x"))
sc.load(["C14_flash_extent_density"])
```

Although all dependencies for `C14_flash_extent_density` are available and calculating this composite would not require handling any files, this scene object is not aware of any file handlers and can therefore not calculate this composite.

Real world situations in which this might occur:

- calculating a composite from the result of `MultiScene.blend`, for example, when aggregating `flash_extent_density` from GLM; or
- unit tests in which a developer has assigned synthetic data to a scene and wants to test if a composite is calculated correctly.

**Describe the solution you'd like**

I would like that composite loading does not need a file handler if dependencies are all available.  I'm not sure how exactly it would know what composite to load, but in principle it should be possible for the Scene class to figure out any composites with name `x` for which the dependencies are met.

**Describe any changes to existing user workflow**

Any implementation would need to be careful not to break existing user workflow.

**Additional context**

I don't know if this can be cleanly implemented without refactoring the entire Scene class in the sense of #1469.

The term `load` in this case is somewhat misleading/confusing, because it suggests data are read from storage.  When we "load" a composite, we're not reading anything from storage, we're calculating the composite based on other datasets.

djhoese:
Regarding the `.load` term, under the hood this method is actually calling a "generate" (or maybe its "generate_composites") method which is a little more accurate.

The technical side of this feature request is that the compositor loading code is created during `Scene.__init__`:

https://github.com/pytroll/satpy/blob/f736315178ebeb26a15a48302251cf39534ed143/satpy/scene.py#L113-L116

This comp/mods information is given to the DependencyTree which is used during `.load` to figure out what we have access to and where we can get it from (readers or already in the Scene). Additionally, the comps/mods information depends on what sensors the instantiated Readers represent (ex. `abi_l1b` is `abi`). This let's Satpy find the sensor-specific YAML configs so that the DependencyTree can know about those sensor-specific composites.

[...]

PR #1752, opened 2021-07-02
Blend into extant scene

### LINKED ISSUE #1749: Load from blended scene
## Feature Request

I'm blending `flash_extent_density` from a `MultiScene` to get a scene in which I have summed the values over multiple scenes.  I cannot load composites from the scene I get back from `MultiScene.blend`.

```python
import satpy
import xarray
filenames = ["/media/nas/x21308/scratch/prob-glm-abi/" + f for f in [
    "OR_ABI-L1b-RadF-M6C14_G16_s19000010000000_e19000010005000_c20403662359590.nc",
    "OR_GLM-L2-GLMF-M3_G16_s19000010000000_e19000010001000_c20403662359590.nc",
    "OR_GLM-L2-GLMF-M3_G16_s19000010001000_e19000010002000_c20403662359590.nc",
    "OR_GLM-L2-GLMF-M3_G16_s19000010002000_e19000010003000_c20403662359590.nc",
    "OR_GLM-L2-GLMF-M3_G16_s19000010003000_e19000010004000_c20403662359590.nc",
    "OR_GLM-L2-GLMF-M3_G16_s19000010004000_e19000010005000_c20403662359590.nc"]]
ms = satpy.MultiScene.from_files(
        filenames,
        reader=["abi_l1b", "glm_l2"],
        group_keys=["start_time"],
        time_threshold=35,
        missing="pass")
ms.load(["flash_extent_density"])
def myblend(ds):
    return xarray.concat(ds, dim="z").sum("z")
sc = ms.blend(blend_function=myblend)
sc.load(["C14_flash_extent_density"])
```

I would like to load composites from the blended scene.

Perhaps I could pass an existing scene to `blend()`, that would be populated, rather than always creating a new `Scene()`.  This way, I could first create a `Scene` that is at least aware of the relevant `FileHandler`s, and then maybe it can load the composite when the `Scene` has been populated by `blend` with the necessary dependencies, as far as they aren't loaded directly from disk?

Not directly, but as with #1748 and #1747, doing this properly may rely on #1469 which may break some user worflow.

**Additional context**

I'm not sure if loading from a blended scene would be easy to implement cleanly or even solve my problem.  I've wanted to load from resampled scenes in the past without luck...

gerritholl:
When using `MultiScene.blend`, allow the user to bring their own scene into which the new datasets will be written.  With the added flexibility this allows, the user may be able to load from a blended scene.
 
 - [x] Closes #1749 with workaround 
 - [x] Tests added 
 - [ ] Fully documented


gerritholl:
If #1751 is merged first, I can commit the documentation of the new feature here after.

djhoese:
Is this a workaround for the #1747? Is it needed otherwise? I'd prefer not to merge this if not. If something like this is needed I think I'd prefer something like a `Scene.update` method to combine to Scenes together.

gerritholl:
> Is this a workaround for the #1747?

Partly.  More directly, it's a workaround for #1749.  I suppose that if #1747 was fixed properly, then #1749 would be closed as well.

> Is it needed otherwise?

I don't know.

> I'd prefer not to merge this if not. If something like this is needed I think I'd prefer something like a Scene.update method to combine to Scenes together.

How would that work?  The object I get back from `MultiScene.blend` doesn't have any file handlers associated.  Could you sketch how I could load from a blended scene with a `Scene.update`?

PR #1766, opened 2021-07-16
Optimize "crefl" rayleigh correction

djhoese:
I recently added VIIRS SDR benchmarks (#1754) to prepare for this work. I knew crefl would be slow due to poor understanding of dask on my part (while helping a student update the algorithm to be dask friendly), but the benchmarks show it is just terrible. This PR is my attempt at fixing that. The main goals are to move all individual dask calls into a single `map_blocks` function. After that I'd like to play with numba or even cython locally, but that may not make it into this PR due to the complications it would cause in the build process.

 - [ ] Closes #xxxx 
 - [ ] Tests added


djhoese:
So I was able to run this PR with asv by doing:

```bash

```

I had to add my branch to the asv config so that publish could find the commits. One gotcha when viewing the individual plots is you have to choose your branch in the left pane otherwise the plots are blank.

here are the peakmem results:

image

On the left (main), we have 22.3GB, on the very right we have 6.7GB. So a pretty decent improvement.

Here's time:

image


In main it's about 12s, in this PR 10.7s. So not as great of an improvement overall, but on a low-memory system or when processing other things I think the memory usage could be huge. Also note I wasn't configuring/limiting dask at all with this. ... @mraspaud maybe we should make sure to set `OMP_NUM_THREADS=1` in the EWC instance. Maybe in the benchmark setups themselves.

Anyway, this PR is ready for review and merge.

mraspaud (approved):
LGTM, thanks for refactoring along the way!

Issue #1756, opened 2021-07-09
Physical Element in awips_tiled writer


joleenf:
**Describe the bug**
Physical element is being written as an attribute of the data variable rather than a global value. AWIPS looks for the physical element in the global attributes
**To Reproduce**
ncdump an output file from the tiled writer and view the location of the physical_element

**Expected behavior**
AWIPS depends on the physical element being in the global attributes for ingest.

joleenf:
This might be specific to awips tiled generated for the clavrx products.

djhoese:
This may have been a mixup on my end when I added support for the "templates" in the awips_tiled writer for GLM support. In those cases there is no physical element (I think) specified, but it required handling multiple data variables per tile file which likely caused confusion on this. The tests definitely need to be updated to check this (surprised they didn't do this already).

FYI in your Polar2Grid issue you made (https://github.com/ssec/polar2grid/issues/360), the vmin/vmax difference is expected I think and I don't think AWIPS uses it. It was just helpful to me. The x/y signed/unsigned difference is something I had to play with and I think is the way I want it.

PR #1770, opened 2021-07-21
Fix CLAVR-x reader and 'awips_tiled' writer to produce AWIPS-compatible output

joleenf:
Changes are made to the awips_tiled yaml so that the physical_element is made into a global attribute addressing #1756
for the clavrx products.
Makes valid_range a list/adds test in test_clavrx reader test.
Adds units to the variables in the awips_tiled yaml for clavrx products so that units are in the output data.
Does not scale integer/categorical data so that the reader presents integer arrays to the writers for integer data.


satpy/etc/writers/awips_tiled.yaml
```diff
@@ -66,9 +66,12 @@ templates:
 #        value: "${ORGANIZATION}"
       awips_id: {}
 #        value: "{awips_id}"  # special variable created by awips_tiled.py
+      physical_element: {}
+#        value: "{physical_element}"   #special variable created by awips_tiled.py
```
djhoese, line 70: Why does this need to be handled by that fancy method? In the case of the polar template we only ever send one data array at a time to be rendered (if this isn't true then it is a bug) by the template class.

Oh...you need each variable to have a custom physical_element so that information needs to come from the variable element...interesting. I guess this makes sense then because the only other ways I could see this being done are:

1. Don't allow custom ones and create a clavrx template that makes the global physical element `"CLAVR-x {name}"`.
2. Create a separate template for each clavrx variable...very verbose.

Could you please verify that by adding that method in the python code that the GLM output doesn't have a global physical element? You could print out the attrs from one of the GLM test cases just to double check. Thanks.

djhoese (changes requested):
I reviewed what I could. I trust you on most of the clavrx stuff since I release "expert-hood" of that reader to you.

satpy/etc/readers/clavrx.yaml
```diff
@@ -21,6 +21,7 @@ file_types:
     file_patterns:
       - 'clavrx_OR_{sensor}-L1b-Rad{sector}-{mode}C{channel_number}_{platform_shortname}_s{start_time:%Y%j%H%M%S%f}.level2.nc'
       - 'clavrx_{platform_shortname}_{start_time:%Y%m%d_%H%M}_B{channel_number}_{sector}_R.level2.nc'
+      - 'clavrx_{platform_shortname}_{start_time:%Y%m%d_%H%M}_B{channel_number}_{sector}_DK_R10_S0110.DAT.level2.nc'
```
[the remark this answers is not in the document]
joleenf: I am discussing this with clavrx group, I think the naming convention might revert back to the old name and then I can remove this line.

satpy/etc/writers/awips_tiled.yaml
joleenf: I thought about creating a separate template for each variable and thought that might be a pain in the long term.  It is true, that in the polar template only one variable is sent at a time (as far as I can tell).  You are correct, I need a custom physical_element for every variable element and that physical_element has to be global, AWIPS does not ingest the file without the global physical_element.  That must be unique because that is how AWIPS assigns the file path when ingesting (as I understand it).  

In the first solution, you are saying that instead of using the "polar" template, I create a separate clavrx template?  How will the awips_tiled writer know to use the clavrx template?  In this case, I remove the fancy method from the awips_tiled writer.

djhoese: AWIPS uses the physical element for the display name usually in the product browser. It is configured for "Polar" files here:

https://github.com/ssec/polar2grid/blob/f19c604d03c0f4fc1452a7ec48a1b484f2349cf1/awips_scmi/Polar/PolarCommonDescription.xml#L17-L19

>  I create a separate clavrx template? How will the awips_tiled writer know to use the clavrx template? In this case, I remove the fancy method from the awips_tiled writer.

Yes. You can provide the template name as a keyword argument to the writer. It is available in Polar2Grid here:

https://github.com/ssec/polar2grid/blob/f19c604d03c0f4fc1452a7ec48a1b484f2349cf1/polar2grid/writers/awips_tiled.py#L165-L170

We use it for the CSPP Gridded GLM project to select the GLM-specific templates. There could be one for CLAVR-x if that makes the most sense and is the easiest to maintain. However, I am not opposed to the solution you went with here. It is unfortunate, but I don't see a way around it while keeping the rest of the code the same.

If you keep it as you have it here could you provide a docstring to the method to explain what it is doing and why?

joleenf: I checked the glm and in the code does add a physical element to the global physical element, so I will have to try a different path.

joleenf: Correction, if I use template="glm_l2_radc" as a keyword in save_datasets, no global physical element in written.  However, the awips_tiled writer default polar template at least writes a file with glm data In that case, the global physical element is written.

djhoese: Ah ok, that is expected. Sounds good.

joleenf: The change to the awips_tiled.yaml and code to add the global physical attribute is okay then?

djhoese: Yes, I think so. I'm not sure there is a cleaner way to do it but that attribute is definitely needed for that template. So unless the CLAVR-x project is ok with generated `{name}` physical_element attributes then I'd say leave it as-is. It defaults to `{name}` right?

joleenf: I am no sure what you are asking when you write 
> unless the CLAVR-x project is ok with generated `{name}` physical_element attributes

Are you asking if the clavrx group would generate the physical element attribute in the original netcdf file?  I am not sure that would work well and then there would still need to be code which moves the physical element attribute from the variable to the global.

On the second question:
> It defaults to {name} right?

Yes.

djhoese: I meant the Polar template could be updated to have the global physical_element set to `{name}` which would mean the satpy product name would always be the physical_element in AWIPS. This would mean CLAVR-x files wouldn't have any "CLAVR-x {name}" prefix. If `{name}` was ok (granted it is not very flexible) then we wouldn't need the custom global attribute method you added.

Forget I said anything. Leave it as is.

joleenf: Let me check how that attribute is being used to ingest.  I have a feeling that product names will collide without the "CLAVR-x" descriptor before the product name.  However, AWIPS ingest is new to me and I need to check.

joleenf: I think we should leave it as is.  The clavrx products include product names that can get very general and then the AWIPS style rules might be applied incorrectly.  The other way to work around this would be to change the reader to alter the satpy name assigned to each datsaset_id?  Is that a better approach?  It seems a little odd.

djhoese: Nah, let's leave it.

djhoese: Ah I just realized, so this results in a physical_element global and variable attribute, right? I can live with that if so.

joleenf: Yes.  This is true.

djhoese (approved):
This looks good to me. I'm going to merge it so we can get my fixes from #1776 and your fixes.

satpy/tests/writer_tests/test_awips_tiled.py
```diff
@@ -414,32 +373,11 @@ def test_lettered_tiles_no_fit(self):
...
-            attrs=dict(
-                name='test_ds',
-                platform_name='PLAT',
-                sensor='SENSOR',
-                units='1',
-                area=area_def,
-                start_time=now,
-                end_time=now + timedelta(minutes=20))
-        )
+        data = da.full((2000, 1000), np.nan, chunks=500, dtype=np.float32),
```
djhoese, line 377:
```suggestion
        data = da.full((2000, 1000), np.nan, chunks=500, dtype=np.float32)
```

PR #1772, opened 2021-07-23
Add access point to global_attrs to netCDF4FileHandler

tsukada-cs:
In the `NetCDF4FileHandler`, it can't get all global_attributes in NetCDF with a single command.
I think `NetCDF4FileHandler` should have a method to get all global_attributes at once without the attribute’s name.

In this PR, we can access all global_attributes with 

`wrapper['/attr']` or `wrapper['/attrs']`

 - [x] Tests added
 - [x] Fully documented

### DIFF
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -196,7 +196,7 @@ class NetCDF4FileHandler(BaseFileHandler):
         val = self.file_content[key]
         if isinstance(val, netCDF4.Variable):
             return self._get_variable(key, val)
-        elif isinstance(val, netCDF4.Group):
+        if isinstance(val, netCDF4.Group):
             return self._get_group(key, val)
         return val
 
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -136,24 +136,25 @@ class NetCDF4FileHandler(BaseFileHandler):
         """Collect all the global attributes for the provided file object."""
         global_attrs = {}
         for key in obj.ncattrs():
-            value = getattr(obj, key)
             fc_key = f"/attr/{key}"
-            try:
-                value = np2str(value)
-            except ValueError:
-                pass
+            value = self._get_attr_value(obj, key)
             self.file_content[fc_key] = global_attrs[key] = value
         self.file_content["/attrs"] = global_attrs
 
     def _collect_attrs(self, name, obj):
         """Collect all the attributes for the provided file object."""
         for key in obj.ncattrs():
-            value = getattr(obj, key)
             fc_key = f"{name}/attr/{key}"
-            try:
-                self.file_content[fc_key] = np2str(value)
-            except ValueError:
-                self.file_content[fc_key] = value
+            value = self._get_attr_value(obj, key)
+            self.file_content[fc_key] = value
+
+    def _get_attr_value(self, obj, key):
+        value = getattr(obj, key)
+        try:
+            value = np2str(value)
+        except ValueError:
+            pass
+        return value
 
     def collect_metadata(self, name, obj):
         """Collect all file variables and attributes for the provided file object.
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -227,6 +227,7 @@ class NetCDF4FileHandler(BaseFileHandler):
         return val
 
     def _get_group(self, key, val):
+        """"Get a group from the netcdf file."""
         # Full groups are conveniently read with xr even if file_handle is available
         with xr.open_dataset(self.filename, group=key,
                              **self._xarray_kwargs) as nc:
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -201,6 +201,7 @@ class NetCDF4FileHandler(BaseFileHandler):
         return val
 
     def _get_global_attrs(self):
+        """Get the global attributes of the netcfd file."""
         global_attrs = {}
         for _key, _val in self.file_content.items():
             if _key.startswith('/attr/'):
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -187,6 +187,12 @@ class NetCDF4FileHandler(BaseFileHandler):
 
     def __getitem__(self, key):
         """Get item for given key."""
+        if key in ('/attr', '/attrs'):
+            global_attrs = {}
+            for _key, _val in self.file_content.items():
+                if _key.startswith('/attr/'):
+                    global_attrs[_key] = _val
+            return global_attrs
         val = self.file_content[key]
         if isinstance(val, netCDF4.Variable):
             if key in self.cached_file_content:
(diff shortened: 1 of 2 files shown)


satpy/tests/reader_tests/test_netcdf_utils.py
```diff
@@ -145,6 +145,16 @@ def test_all_basic(self):
...
+        global_attrs = {
+            '/attr/test_attr_str': 'test_string',
+            '/attr/test_attr_str_arr': 'test_string2',
+            '/attr/test_attr_int': 0,
+            '/attr/test_attr_float': 1.2
+            }
+        self.assertEqual(file_handler['/attr'], global_attrs)
+        self.assertEqual(file_handler['/attrs'], global_attrs)
+
+
```
mraspaud, line 157: Too many blank lines :)

```suggestion
```

mraspaud (commented):
This looks good to me, just a linting error to fix so far :)

satpy/readers/netcdf_utils.py
```diff
@@ -187,6 +191,12 @@ def collect_cache_vars(self, cache_vars, obj):
 
     def __getitem__(self, key):
         """Get item for given key."""
+        if key in ('/attr', '/attrs'):
```
mraspaud, line 194: As you can see in the checks, Codebeat is complaining that this method has become too nested: https://codebeat.co/projects/github-com-pytroll-satpy-main/pull_requests/487392

If you are up to it, you could try refactoring a bit, maybe to something like this:
```python
def __getitem__(self, key):
    """Get item for given key."""
    if key in ('/attr', '/attrs'):
        return self._get_global_attrs()
    val = self.file_content[key]
    if isinstance(val, netCDF4.Variable):
        return self._get_variable(key, val)
    elif isinstance(val, netCDF4.Group):
        return self._get_group(key, val)
```
and create the new methods `_get_global_attrs`, `_get_var`, `_get_group` by moving the corresponding code in there.

tsukada-cs: Thanks for the comment. I've moved those!

satpy/readers/netcdf_utils.py
```diff
@@ -187,27 +191,44 @@ def collect_cache_vars(self, cache_vars, obj):
...
+            return self._get_variable(key, val)
         elif isinstance(val, netCDF4.Group):
-            # Full groups are conveniently read with xr even if file_handle is available
-            with xr.open_dataset(self.filename, group=key,
-                                 **self._xarray_kwargs) as nc:
-                val = nc
+            return self._get_group(key, val)
+        return val
+
+    def _get_global_attrs(self):
```
mraspaud, line 203:
```suggestion
    def _get_global_attrs(self):
        """Get the global attributes of the netcfd file."""
```

satpy/readers/netcdf_utils.py
```diff
@@ -187,27 +191,44 @@ def collect_cache_vars(self, cache_vars, obj):
...
+        global_attrs = {}
+        for _key, _val in self.file_content.items():
+            if _key.startswith('/attr/'):
+                global_attrs[_key] = _val
+        return global_attrs
+
+    def _get_variable(self, key, val):
```
mraspaud, line 210:
```suggestion
    def _get_variable(self, key, val):
        """Get a variable from the netcdf file."""
```

satpy/readers/netcdf_utils.py
```diff
@@ -187,27 +191,44 @@ def collect_cache_vars(self, cache_vars, obj):
...
+            group, key = parts
+        else:
+            group = None
+        if self.file_handle is not None:
+            val = self._get_var_from_filehandle(group, key)
+        else:
+            val = self._get_var_from_xr(group, key)
+        return val
+
+    def _get_group(self, key, val):
```
mraspaud, line 227:
```suggestion
    def _get_group(self, key, val):
        """"Get a group from the netcdf file."""
```

satpy/readers/netcdf_utils.py
```diff
@@ -187,27 +191,44 @@ def collect_cache_vars(self, cache_vars, obj):
...
-            if len(parts) == 2:
-                group, key = parts
-            else:
-                group = None
-            if self.file_handle is not None:
-                val = self._get_var_from_filehandle(group, key)
-            else:
-                val = self._get_var_from_xr(group, key)
+            return self._get_variable(key, val)
         elif isinstance(val, netCDF4.Group):
```
mraspaud, line 199: Just to make codefactor.io happy

```suggestion
        if isinstance(val, netCDF4.Group):
```

mraspaud (commented):
Ok, I think these are my last comments, after that it should be good to merge :)

tsukada-cs:
Thanks a lot!

satpy/readers/netcdf_utils.py
```diff
@@ -187,27 +191,47 @@ def collect_cache_vars(self, cache_vars, obj):
...
+        else:
+            group = None
+        if self.file_handle is not None:
+            val = self._get_var_from_filehandle(group, key)
+        else:
+            val = self._get_var_from_xr(group, key)
+        return val
+
+    def _get_group(self, key, val):
+        """"Get a group from the netcdf file."""
```
mraspaud, line 230: Sorry about that one!

```suggestion
        """Get a group from the netcdf file."""
```

tsukada-cs: No problem ;)

mraspaud (approved):
LGTM

djhoese (approved):
Nice work! One more consideration: what do you think about storing the global attributes in a separate dictionary (ex. `self._global_attrs = {}`) or as a separate key (ex. `self.file_content['/attrs'] = {}`) and have that filled in in the `_collect_attrs` method when the file is read. We're already storing all the global attributes in `self.file_content` so having another copy wouldn't be that bad. You could have a special `_collect_global_attrs` method to do it maybe. Just a thought.

tsukada-cs:
That's also nice idea! But I'd like to see what's going on with the current implementation. Thank you :)

djhoese:
> But I'd like to see what's going on with the current implementation.

I'm not sure I understand what you mean. Are you saying you like the current implementation and don't want to change it?

tsukada-cs:
@djhoese  Sorry for the confusion. I don't really have time for this. If someone wants to change the implementation, that's fine.

djhoese:
@mraspaud I'm working on refactoring this and I'd like to remove the ability to use `/attr` **and** `/attrs` and instead make it only `/attrs`. Are you and @tsukada-cs ok with that?

tsukada-cs:
@djhoese

> Are you and @tsukada-cs ok with that?

Thank you. There is no problem with that.

I prefer
```
self["/attrs"]["some_key"]
```
, but I have no strong preference or reason for it.

mraspaud:
Sounds good. I agree that `self["/attrs"]["some_key"]` looks better.

djhoese:
Ok this is ready for re-review. Global attributes are available as `handle["/attrs"]` and the keys in that dictionary are just the name of the attribute. I also refactored some of the other code to hopefully clean it up a bit (smaller methods, less redundant code).

mraspaud (approved):
LGTM, it's nice with the simplification of the interface

PR #1776, opened 2021-07-27
Fix 'awips_tiled' writer producing an invalid y coordinate

djhoese:
@joleenf pointed out that something was wrong in the coordinates with the 'awips_tiled' writer. After spending a while tracking it down I discovered that the `add_offset` was being set to the `.min()` value of the Y coordinate data. The Y data is always in descending order (negative scale_factor) so using the min as the add_offset results in the scale in-file values being `-5119, -5118, ..., 0`. For the existing templates, these values are int16 with `_Unsigned = 'true'` which means these negative values are incorrect (and -1 is the fill value). It is all just bad overall and resulted in invalid Y coordinates when unscaled from the file.

@joleenf do you think you'd be able to test this with some of the other stuff you've been doing before I merge it?

 - [x] Tests added

### DIFF
--- a/satpy/tests/writer_tests/test_awips_tiled.py
+++ b/satpy/tests/writer_tests/test_awips_tiled.py
@@ -46,6 +46,7 @@ def check_required_common_attributes(ds):
     assert x_attrs.get('standard_name') == 'projection_x_coordinate'
     assert x_attrs.get('units') == 'meters'
     assert 'scale_factor' in x_attrs
+    assert x_attrs['scale_factor'] > 0
     assert 'add_offset' in x_attrs
 
     assert 'y' in ds.coords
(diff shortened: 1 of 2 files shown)


djhoese:
@mraspaud @sfinkens Looks like the fiduceo reader is failing tests now. Not sure how it didn't fail in the unstable environment first (or maybe it would have but we didn't have enough recent pull requests).

djhoese:
Looks like it was made into a PR here: https://github.com/pydata/xarray/pull/5630 4 days ago, merged, and then released immediately.

djhoese:
See #1777 for fixes.

joleenf:
@djhoese As noted in slack, the area_extent and range of y coordinates are consistent.  The y coordinate is increasing from a negative value to a positive value.  I only checked the fixed grid that was broken before the bug fix.  I will check the grid that was okay before the bug fix and make sure it is still okay and get back to you.

joleenf:
Navigation and tiles looking good for both 211e and us_1km grids with clavrx data.

PR #1777, opened 2021-07-27
Fix mviri l1b fiduceo reader compatibility with newer xarray

djhoese:
Discovered in #1776, the new version of xarray now doesn't allow you to assign a Variable object with a DataArray. This can accidentally happen when you create a Dataset with `('y', 'x', some_data)` where `some_data` is a DataArray. The dimensions are redundant and being a tuple tells xarray that you are trying to make a Variable from scratch.

This PR is my best attempt at fixing this.


mraspaud (approved):
LGTM. We will let @sfinkens have the last word on this when he's back, but I'll just merge for now.

sfinkens:
LGTM, thanks for fixing this!

PR #1784, opened 2021-08-05
Fix ABI readers not assigning 'platform_name' for GOES-18/19

djhoese:
Simple fix so that the ABI-based readers know about GOES-18/19. This came up in SIFT development where the data and most of the metadata was perfectly fine, but `platform_name` was missing (was set to `None`).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


mraspaud (approved):
Lgtm

PR #1792, opened 2021-08-13
Maintain categorical clavrx data as integer arrays

joleenf:
Fixes bug that turns categorical data into float arrays after the valid_range is applied.  This is done by using the fill value to fill the data outside of the valid_range for integer.  This PR also adds tests to make verify that int arrays have _FillValue and remain integer, while float arrays have no _FillValue in the output attributes and remain float.


djhoese (approved):
Looks good to me. Is there anything else you wanted to do on this?

joleenf:
No.

PR #1795, opened 2021-08-18
Make raw metadata optional in SEVIRI HRIT/Native

### LINKED ISSUE #1718: Raw metadata handling impacts performance
**Describe the bug**
As reported by @BENR0 on slack, the handling of raw SEVIRI HRIT metadata seems to impact performance (factor of ~2 slower).

**Expected behavior**
Raw metadata handling should not significantly affect performance.

sfinkens:
Make the provision of raw metadata optional in SEVIRI HRIT and Native readers. This avoids the associated performance penalty for the majority of use cases where those metadata are not needed. For this purpose a new reader kwarg `include_raw_metadata` is added. The option for limiting the array size in the metadata remains, because it still allows users to speed up the loading if they don't need those large arrays. Example:

```python
scn = satpy.Scene(
    filenames,
    reader='seviri_l1b_hrit',
    reader_kwargs={'include_raw_metadata': True, 'mda_max_array_size': 10}
)
```

The performance penalty in the current version is ~10% at max (see table below). Its origin is the comparison of large nested dictionaries from multiple segments (HRIT) or files (Native). Array comparison is responsible for approx. half of that penalty. The other half is spent with dictionary iteration and comparison of scalar objects. The latter amount can only be reduced if raw metadata are excluded.

Here are some timings for a single IR channel in HRIT format. For higher resolution channels the relative penalty is smaller, because the actual loading takes longer.

| Variant  | Time | Relative to current version | 
| ------------- | ------------- | ------------- |
| Without raw metadata (the default in this PR) | 0.74s | 89% |
| Max metadata array size = 1  | 0.79s | 95% |
| Max metadata array size = 10  | 0.81s | 98% |
| Max metadata array size = 100 (current version)  | 0.83s | 100% |
| Max metadata array size = unlimited  | 0.85s | 102% |

 - [X] Closes #1718  
 - [X] Tests added 
 - [X] Fully documented

### DIFF
--- a/satpy/readers/seviri_base.py
+++ b/satpy/readers/seviri_base.py
@@ -114,16 +114,6 @@ The SEVIRI L1.5 readers provide the following metadata:
 * The ``orbital_parameters`` attribute provides the nominal and actual satellite
   position, as well as the projection centre. See the `Metadata` section in
   the :doc:`../readers` chapter for more information.
-* The ``raw_metadata`` attribute provides raw metadata from the file header
-  (HRIT and Native format). By default, arrays with more than 100 elements are
-  excluded to limit memory usage. This threshold can be adjusted using the
-  ``mda_max_array_size`` reader keyword argument:
-
-  .. code-block:: python
-
-       scene = satpy.Scene(filenames,
-                          reader='seviri_l1b_hrit/native',
-                          reader_kwargs={'mda_max_array_size': 1000})
 
 * The ``acq_time`` coordinate provides the mean acquisition time for each
   scanline. Use a ``MultiIndex`` to enable selection by acquisition time:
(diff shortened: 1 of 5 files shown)


sfinkens:
@loreclem Would that solution be ok for you?

loreclem:
Thank you @sfinkens, I'll do some internal checks and report back here asap.

simonrp84:
Hi @sfinkens, just to confirm: The raw metadata is disabled by default in this PR, right? And the user will need to add the `kwarg` if they want to include the metadata?

sfinkens:
@simonrp84 Yes, exactly!

simonrp84:
Great, thanks for the confirmation!

PR #1800, opened 2021-08-23
Fix for missing attributes when requesting 'counts' calibration from ABI L1B reader.

jon4than:
As part of Pull Request #1692, we added an early return to abi_l1b.py's get_dataset method when the requested calibration is "counts".

The problem is that the early return results in a number of missing attributes in the object returned from get_dataset.

I also took the liberty of updating the unit tests to check for some of the missing attributes, as well as including a test for invalid calibration.

As an aside, I agree with the logic for _why_ it was decided to return early; the `self.nc['Rad']` values aren't radiances. That said, I couldn't think of a more straightforward way to write this code.

 - [x] Tests added 
 - [x] Fully documented 

### DIFF
--- a/satpy/readers/abi_l1b.py
+++ b/satpy/readers/abi_l1b.py
@@ -40,14 +40,16 @@ class NC_ABI_L1B(NC_ABI_BASE):
         logger.debug('Reading in get_dataset %s.', key['name'])
         # For raw cal, don't apply scale and offset, return raw file counts
         if key['calibration'] == 'counts':
-            return self._raw_calibrate(self.nc['Rad'])
-        radiances = self['Rad']
+            radiances = self.nc['Rad']
+        else:
+            radiances = self['Rad']
 
         # mapping of calibration types to calibration functions
         cal_dictionary = {
             'reflectance': self._vis_calibrate,
             'brightness_temperature': self._ir_calibrate,
-            'radiance': self._rad_calibrate
+            'radiance': self._rad_calibrate,
+            'counts': self._raw_calibrate,
         }
 
         try:
--- a/satpy/readers/abi_l1b.py
+++ b/satpy/readers/abi_l1b.py
@@ -40,7 +40,7 @@ class NC_ABI_L1B(NC_ABI_BASE):
         logger.debug('Reading in get_dataset %s.', key['name'])
         # For raw cal, don't apply scale and offset, return raw file counts
         if key['calibration'] == 'counts':
-            radiances = self.nc['Rad']
+            radiances = self.nc['Rad'].copy()
         else:
             radiances = self['Rad']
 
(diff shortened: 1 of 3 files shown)


satpy/readers/abi_l1b.py
```diff
@@ -40,14 +40,16 @@ def get_dataset(self, key, info):
         logger.debug('Reading in get_dataset %s.', key['name'])
         # For raw cal, don't apply scale and offset, return raw file counts
         if key['calibration'] == 'counts':
-            return self._raw_calibrate(self.nc['Rad'])
-        radiances = self['Rad']
+            radiances = self.nc['Rad']
+        else:
+            radiances = self['Rad']
```
djhoese, line 45: Your PR made me realize something, I don't think `self.nc['Rad']` makes a copy of the `DataArray` underneath. The rest of this all looks good, but I'm wondering if we're introducing another bug here. Could you try loading a channel with counts and then loading it with the default calibration? Something like:

```python
scn = Scene(reader='abi_l1b')
scn.load(['C01'], calibration='counts')
scn.load(['C01'], calibration='reflectance')
```

I'm wondering if you print out the `.attrs` in the file handlers `.get_dataset` method (right before the DataArray is returned) if the reflectance data will be correct. Otherwise, a `.copy()` might be needed at the end of `self.nc['Rad'].copy()`.

jon4than: That's a good point. I went ahead and added a logger.debug(...) to the end of abi_l1b.py's `get_dataset`, and gave your test a go...and the output looks sensible _to me_, but I trust your judgement on this a lot more than mine :wink:, so here's what I saw:

https://gist.github.com/jon4than/0a16ddd756efa5217cc28ddb479a0b1a

Not sure if that's the best way to include the output, and let me know if I misinterpreted how I should be testing things.

At any rate, the explicit `copy()` seems like it might be a good idea regardless of how the `.attrs` appear.

jon4than: Though FWIW it looks as though the `self.nc[item]` in abi_base.py's `__getitem__`, `_adjust_data`, and `_adjust_coords` methods is not explicitly copied either.

djhoese: So I think we're getting lucky here. So the calibration methods do `res.attrs = attrs` and inside xarray it looks like this actually makes a shallow copy of the dictionary. The abi_base.py `__getitem__` also does this attrs assignment.

I wouldn't mind the `.copy()` being added if you have the time.

jon4than: Done. Thanks for pointing this stuff out, as it can be pretty hard to spot!

djhoese (approved):

djhoese:
Thanks @jon4than!

PR #1801, opened 2021-08-23
added link to the GOES-2-go package in the docs as a download source.

blaylockbk:
I just saw that your documents link to my rclone tutorial for downloading GOES data. Thanks for sharing that! I also have a python package called "GOES-2-go" that downloads GOES data from AWS to your local machine. This link could be helpful to others who use satpy, and I hoped you consider including this link in your docs.


mraspaud (approved):
This looks good to me, thanks for providing the PR.

PR #1812, opened 2021-08-30
Fix composites failing when inputs are different chunk sizes

djhoese:
While working on using the CREFL rayleigh correction on MODIS data I noticed a couple issues. This PR fixes:

1. The variable name was incorrect for the DEM file typically used with MODIS data (tbase.hdf)
2. The tbase.hdf file doesn't have a fill value specified so it was causing the code to fail when reading it.
3. Any composites that use `map_blocks` or `blockwise` will fail if they have different chunk sizes for their inputs. This is happening for MODIS data due to the angle datasets being returned as 64-bit floats when they only need to be 32-bit floats (that'll be another PR). Anyway, I've added a `unify_chunks` method that is automatically run as part of `match_data_arrays` that all Composites typically follow.

 - [x] Tests added

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -19,6 +19,7 @@
 
 import logging
 import os
+import string
 import warnings
 
 import dask.array as da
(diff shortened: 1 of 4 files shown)


djhoese:
:man_facepalming: I already made #1808 which includes these commits. Let me redo this.

PR #1822, opened 2021-09-14
Add SEVIRI-specific Rayleigh correction and Natural Enhanced night composite

simonrp84:
Previously, the `rayleigh` corrector would not function as expected with SEVIRI, as it would choose the `HRV` channel rather than `VIS006` for the required red band. This PR adds a SEVIRI specific definition of the Rayleigh corrector that includes the correct channel name.
In addition, this PR also adds a night-lights composite using `natural_enh` to complement the existing composite using `natural_color`. This creates more realistic day/night images that are a bit closer to those produced from other sats.

### DIFF
--- a/satpy/etc/composites/seviri.yaml
+++ b/satpy/etc/composites/seviri.yaml
@@ -11,6 +11,19 @@ modifiers:
     - IR_108
     - IR_134
 
+  rayleigh_corrected:
+    modifier: !!python/name:satpy.modifiers.PSPRayleighReflectance
+    atmosphere: us-standard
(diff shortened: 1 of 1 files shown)


djhoese:
Could this mean that technically the rayleigh correction is not as "good" as it could be if someone was using HRV?

simonrp84:
Problem is, jrv doesn't cover the full disk - so by using that cha nel for Rayleigh you end up with big missing data gaps in the output image.

mraspaud (approved):
LGTM!

PR #1824, opened 2021-09-16
Add additional ACSPO reader file patterns

djhoese:
New versions of the ACSPO files can have new filenames where "SSTskin" was replaced with "SSTsubskin". This PR also changes the unittest-based test class to a pytest one so that I could do parametrize tests.

 - [x] Tests added


mraspaud (approved):
LGTM

PR #1846, opened 2021-10-04
Replace 'libtiff' python dependency with 'pylibtiff'

djhoese:
The python libtiff packages was renamed a long time ago to pylibtiff. This PR replaces that PyPI (setup.py) dependency name with the new name.


djhoese:
Gotta love when CI starts failing for unrelated reasons...

mraspaud (approved):
LGTM

PR #1849, opened 2021-10-11
Update AGRI reader to ensure that angles can be correctly loaded.

simonrp84:
The AGRI L1 reader could not correctly load angles, as it was trying to find a `calibration` that does not exist, and does not make sense, for angles datasets.
This PR checks if angles are being loaded and, if so, doesn't attempt calibration. Tests have also been added.

Furthermore, this PR also correctly sets the fill value in the data for angles datasets.

 - [x ] Tests added

### DIFF
--- a/satpy/readers/agri_l1.py
+++ b/satpy/readers/agri_l1.py
@@ -45,6 +45,48 @@ PLATFORM_NAMES = {'FY4A': 'FY-4A',
                   'FY4C': 'FY-4C'}
 
 
+def scale(dn, slope, offset):
+    """Convert digital number (DN) to calibrated quantity through scaling.
+
+    Args:
+        dn: Raw detector digital number
+        slope: Slope
+        offset: Offset
+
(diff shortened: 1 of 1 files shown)


yukaribbba:
@simonrp84 Thanks for this! I have another issue not that related: Do you think the atmospheric correction is some kind of "too strong" for AGRI? I mean it works perfect for ABI/AHI/AMI family, but for AGRI, not that satisfying.

simonrp84:
Aah, ok. The AGRI sensor has much wider bands than ABI, which leads to problems in generating the simulated green channel required for `true_color` composites. Here's spectral response function plots of GOES-16 and Fengyun-4, you can see how wide the 0.6 and 0.8 micron AGRI channels are!
FY4A
GOES
!

yukaribbba:
Ok got it. This has troubled me for a long time. I have to turn to other ways like Photoshop to make it look better.

PR #1860, opened 2021-10-21
Catch ImportError on UnsafeLoader in composites/config_loader

### LINKED ISSUE #1859: Yaml UnsafeLoader ImportErrror on colab.google
Importing Satpy is increasingly difficult on google colab.  This is mostly due to colab still using pyyaml 3.13 by default.  Force updating libyaml and pyyaml used to work as a workaround but that's problematic (pinning pyyaml==5.4.1 actually fixes the workaround).  While this is a colab thing, there's an easy and I think intended work around missing from satpy since 0.24.0

6 of the 7 UnsafeLoader imports in satpy are wrapped in a try:except.   
```python
try:
    from yaml import UnsafeLoader
except ImportError:
    from yaml import Loader as UnsafeLoader
```

But composites/config_loader.py introduced an unprotected import back in 0.24.0.   This seems to have just been an oversight while moving code around in https://github.com/pytroll/satpy/pull/1360.  

I see there is a bigger issue/request #789 around yaml loaders but is there anything against a 3 line PR to try:except this one import. 
https://github.com/pytroll/satpy/blob/e92b6b2992e2adc731ff3621f17caf331525faec/satpy/composites/config_loader.py#L24

**To Reproduce**
colab.research.google.com
```python
!pip install satpy

import satpy

```

**Expected behavior**
Loader is imported as UnsafeLoader for pyyaml versions without Unsafeloader

**Environment Info:**
 - colab.research.google.com
 - Satpy==0.23.0 == success | Satpy == 0.24.0+ errors

abrammer:
Add Try/Except around UnsafeLoader import to match the other 6 imports and support old versions of pyyaml.  

The default pyyaml on colab.google is still 3.13.  This individual import prevents satpy from being used without also manually installing an updated pyyaml.  6 of the 7 other UnsafeLoader imports already have this logic so just cover this one remaining import with it.  

 - [ ] Closes #1859 
 - [ ] Tests ? | Would require testing against various pyyaml versions.

### DIFF
--- a/satpy/composites/config_loader.py
+++ b/satpy/composites/config_loader.py
@@ -21,7 +21,11 @@ import logging
 import warnings
 
 import yaml
-from yaml import UnsafeLoader
+
+try:
+    from yaml import UnsafeLoader
+except ImportError:
+    from yaml import Loader as UnsafeLoader
 
 from satpy import DatasetDict, DataQuery, DataID
 from satpy._config import (get_entry_points_config_dirs, config_search_paths,


djhoese:
CI jobs had a weird HTTP hangup. I've restarted them.

PR #1868, opened 2021-10-30
Fix MiRS reader not working with new versions of dask

djhoese:
While testing the mirs reader with Polar2Grid I noticed an error about dtype inference:

```
ValueError: `dtype` inference failed in `map_blocks`.
```

This PR fixes this and adds more complex tests. The tests now:

1. Include all 22 ATMS channels
2. Make sure that requested datasets in the tests are actually loaded

I also renamed some variables in the coefficient reading for easier debugging in the future (same overall performance).

I've also included a small update to the AMSR2 L1b reader to make the sensor lowercase to match other readers in Satpy.

CC @joleenf

 - [x] Tests added 
 - [ ] Fully documented


djhoese:
No complaints on slack. I'm going to merge this given the lack of complaints and the small amount of changes.

PR #1876, opened 2021-11-08
Add Python 3.9 to CI tests

gerritholl:
Add GitHub CI tests for newer versions of Python.

 - [x] Closes #1876 
 - [x] Tests added

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -63,7 +63,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ["windows-latest", "ubuntu-latest", "macos-latest"]
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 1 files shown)


gerritholl:
Not sure if this is enough; there are some other references to 3.8; I don't know if those would need updating too?

djhoese:
I don't think conda-forge has all packages up for Python 3.10, so these tests will probably fail. What other references to Python 3.8 are there? In the other CI environments? I'd be OK with those being bumped to 3.9 or left as-is.

gerritholl:
I mean that within the same file there are two references to `python-version: "3.8"`, but I don't know if that affects the actual tests.

djhoese:
If you follow the indentation up, the python-versions map to Python 3.9 for the lint and style checks, 3.8 for the website building, 3.8 for the "experimental/unstable" tests. I think all of those are reasonable for now.

PR #1882, opened 2021-11-12
Update GDAL_OPTIONS with driver= and COG-specific options

howff:
See issue https://github.com/pytroll/trollimage/issues/93 (Add support for Cloud Optimized GeoTIFF)

This PR allows the `driver` option to be specified, so you can save a dataset with `driver='COG'` to get a cloud-optimized GeoTIFF. It also add additional options which are specific to the COG format. (Whether or not they are actually supported might depend on the rasterio code).

The PR only makes sense after merging PR https://github.com/pytroll/trollimage/pull/94 (Use COG driver to write cloud optimized GeoTIFF)


satpy/writers/geotiff.py
```diff
@@ -76,7 +76,27 @@ class GeoTIFFWriter(ImageWriter):
                     "profile",
                     "bigtiff",
                     "pixeltype",
-                    "copy_src_overviews",)
+                    "copy_src_overviews",
+                    # Not a GDAL option, but allows driver='COG'
+                    "driver",
```
djhoese, line 81: I would make this an explicit option on the save method and remove it from this list.

Plantain:
I've been testing this over the last few days, it seems to work well.
My only observation is I seem to get a "ERROR 4: filename.tiff No such file or directory error" from GDAL(?) on successful runs.

howff:
> I've been testing this over the last few days, it seems to work well. My only observation is I seem to get a "ERROR 4: filename.tiff No such file or directory error" from GDAL(?) on successful runs.

That's right, I noticed that too, although the output file always validated ok. I think it's related to this https://github.com/rasterio/rasterio/issues/2339

djhoese:
This makes me think we need to switch to rioxarray sooner rather than later. I've seen a lot of issues on the rioxarray repository lately about optimizing "windowed writes" and other fixes for creating geotiffs from xarray/dask.

pnuu:
I never saw the `ERROR 4` with my (now closed) #1928. There I just added `"driver"` to the `GDAL_OPTIONS` tuple.

pnuu:
I don't see the `ERROR 4` even with this PR and `rasterio=1.2.10` and `gdal=3.3.3`.

djhoese:
I tried this out on my local machine with dask's default scheduler and didn't see any errors. I also tried not using this and specifying `overviews=` in different ways but the validate script from GDAL always complains in one way or another about how the overviews are ordered in the file. It seems like a bug with rasterio's method for building overviews. Let's merge this and see how we feel about it. I didn't see any major performance difference.

pnuu:
I on the other hand get valid results from `validate_cloud_optimized_geotiff.py`. I used `glbl.save_datasets(base_dir='/tmp/', tiled=True, blockxsize=512, blockysize=512, driver='COG', overviews=[])` when saving the images.

djhoese:
To clarify, using `driver='COG'` should always give a valid COG. I was trying to create one by using the basic "gtiff" driver and specify overviews in different ways. No matter what I did the validate script complained for one or more reasons.

PR #1884, opened 2021-11-15
Fix nucaps reader failing when given multiple input files

djhoese:
@scottlindstrom pointed out while working with Polar2Grid that he wasn't able to give the nucaps reader multiple files as input. This PR should fix the issues he was running into.

 - [x] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #1886, opened 2021-11-16
Update quickstart documentation so that HRV channel is not loaded

simonrp84:
The quickstart guide currently says:
`global_scene.load([0.6, 0.8, 10.8])`
But, for SEVIRI, this will load the `HRV` channel and *not* `VIS006`. This PR corrects the quickstart so that `VIS006` is loaded as expected.

 - [ ] Fully documented


gerritholl:
The real problem here is Satpys logic to implicitly load HRV when asking for 0.6 µm, which may or may not be what users want.  Thereby Satpy doesn't even show a warning that the request is ambiguous.  Better would be a (global) config where priority con be configured between different behaviours in case of ambiguity: guess without warning, guess with warning, or error.  When guessing, it could be the channel with the narrowest band, channel with the highest spatial resolution, or something else yet.  See the dedicated issue https://github.com/pytroll/satpy/issues/1721.

djhoese:
Sure but that doesn't change that this PR requires more changes than it has.

djhoese (approved):

doc/source/quickstart.rst
```diff
@@ -130,23 +130,23 @@ Calculating measurement values and navigation coordinates
...
 ================
 
-To visualize loaded data in a pop-up window:
-
-    >>> global_scene.show(0.6)
-
+To visualize loaded data in a pop-up window:                                                        
+                                                                                                    
+    >>> global_scene.show(0.8)
+
```
djhoese, line 149:
```suggestion
To visualize loaded data in a pop-up window:

    >>> global_scene.show(0.8)

```

djhoese:
pre-commit.ci autofix

PR #1887, opened 2021-11-17
Remove confusing and unused cache_dir default of '.'

### LINKED ISSUE #1456: Default cache directory should respect XDG Base Directory Specification.
## Feature Request

As a default cache directory, satpy uses `'.'`:

https://github.com/pytroll/satpy/blob/a18b6e2fae74a71beea01a26a3a0ffcaabd91016/satpy/resample.py#L426-L432

This is potentially problematic: the current directory may be unwriteable or otherwise a poor choice to create files.  

I would like for satpy to respect the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and use the `$XDG_CACHE_HOME` directory (probably `$XDG_CACHE_HOME/satpy`) as a cache directory if none is specified.

Users who have `$XDG_CACHE_HOME` set will find cache files where they previously didn't.

Simply taking `$XDG_CACHE_HOME` will not introduce additional dependencies.  Satpy could consider to use `appdirs.user_cache_dir("satpy")` using the [`appdirs` package](https://pypi.org/project/appdirs/).  This solution would add `appdirs` as a satpy dependency.

**Additional context**

I can pass `cache_dir=appdirs.user_cache_dir("satpy")` explicitly to `.resample(...)`

djhoese:
As described in #1456, the source code for resampling suggests that if resampling `cache_dir` is not provided it will default to the current directory. This is never used as this portion of the code is only used *if* `cache_dir` is specified.

 - [x] Closes #1456

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,9 @@
 name: CI
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency
+# https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pullrequestevent
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.type }}
+  cancel-in-progress: true
 
 on: [push, pull_request]
 
--- a/satpy/resample.py
+++ b/satpy/resample.py
@@ -536,6 +536,8 @@ class KDTreeResampler(BaseResampler):
     def _check_numpy_cache(self, cache_dir, mask=None,
                            **kwargs):
         """Check if there's Numpy cache file and convert it to zarr."""
+        if cache_dir is None:
+            return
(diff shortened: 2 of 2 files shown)


djhoese:
I noticed the CI runs weren't cancelling themselves with new commits so I added a concurrency group to the config which should make them cancel in the future if new ones are started.

gerritholl:
I'm indeed confused :)  Does this change existing behaviour or not?

djhoese:
@gerritholl No change in functionality except for unintended issues. The only unintended thing that I can figure out could happen is if you didn't specify `cache_dir` but you were currently in the cache directory then it would find those cache files. I don't remember if it would actually use them, but still. Bottom line, no functionality change.

mraspaud (approved):
LGTM

PR #1901, opened 2021-11-25
Update Scene.sensor_names to include sensors from readers and contained data

### LINKED ISSUE #1900: Load composites mixed from files or provided data
## Feature Request

I'm loading ABI but calculating GLM FED from external sources (actually summed and coming from a `MultiScene.blend`, but it could be anywhere).  I then want to load a composite that should load one dependency from a file (ABI) and the other from a dataset provided manually (GLM).  This doesn't work, because the scene "forgets" the sensor names derived from filenames provided on scene creation, when it gets instead the sensor names from data provided "manually".  In pseudocode:

```python
sc_both = satpy.Scene(filenames=...)
sc_glm = some_calculation(...)  # calculate sum of FED from multiple files
sc_both["flash_extent_density"] = sc_glm["flash_extent_density"]
sc_both.load(["C14_flash_extent_density"])
```

The last step fails `KeyError` due to missing dependencies.

The problem is more succinctly illustrated by:

```python
from satpy import Scene, MultiScene

glm_file = "/media/nas/x21308/GLM-processed/C/1min/2021/06/02/00/OR_GLM-L2-GLMC-M3_G16_s20211530000200_e20211530001200_c20211932137440.nc"
abi_file = "/media/nas/x21308/abi/G16_ABI_CONUS_2021153/00/OR_ABI-L1b-RadC-M6C14_G16_s20211530001159_e20211530003532_c20211530003595.nc"

sc_abi = Scene(filenames=[abi_file], reader="abi_l1b")
sc_abi.load(["C14"])
sc_both = Scene(filenames={"abi_l1b": [abi_file], "glm_l2": [glm_file]})
print(sc_both.sensor_names)  # both
sc_both["C14"] = sc_abi["C14"]
print(sc_both.sensor_names)  # only ABI
```

which gives

```
{'glm', 'abi'}
{'abi'}
```

(this assumes the fix in #1899 is already merged)

I would like that `Scene.sensor_names` can merge those sensor names derived from the filenames provided on Scene creation and those derived from datasets assigned directly.

Tricky.  I _think_ it shouldn't matter, but I can't promise there aren't corner cases that may be affected.  Let's hope our unit tests are good enough to catch those.

**Additional context**

As a workaround, I could load something from both datasets before doing the direct assignment, or assign datasets from both sensors.  Neither appear particularly clean or maintainable, so I would prefer the approach I suggested above.

djhoese:
This PR updates Scene.sensor_names so it includes sensor information from both the loaded readers (if any) and the contained data (if any). Previously it would only include the reader information if there were no contained datasets with sensor information. The case I was hoping to avoid (reader's including more sensors than we care about) turns out to be more prohibitive than this new solution.

This should be tested by @gerritholl before being merged just to be sure. If there is another test case that you think is worth adding let me know.

 - [x] Closes #1900 
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -135,16 +135,15 @@ class Scene:
         Sensor information is collected from data contained in the Scene
         whether loaded from a reader or generated as a composite with
         :meth:`load` or added manually using ``scn["name"] = data_arr``).
-        If no data is currently contained in the Scene or no sensor
-        information is found in the data metadata, loaded readers will be
-        consulted for sensor information.
+        Sensor information is also collected from any loaded readers.
+        In some rare cases this may mean that the reader includes sensor
+        information for data that isn't actually loaded or even available.
 
         """
-        sensor_names = self._contained_sensor_names()
-        if not sensor_names:
-            sensor_names = set([sensor for reader_instance in self._readers.values()
-                                for sensor in reader_instance.sensor_names])
-        return sensor_names
+        contained_sensor_names = self._contained_sensor_names()
+        reader_sensor_names = set([sensor for reader_instance in self._readers.values()
+                                   for sensor in reader_instance.sensor_names])
+        return contained_sensor_names | reader_sensor_names
 
     def _contained_sensor_names(self) -> set[str]:
         sensor_names = set()
--- a/.bandit
+++ b/.bandit
@@ -1,2 +1,3 @@
-skips: ["B506"]
-exclude_dirs: ["satpy/tests"]
+[bandit]
+skips: B506
+exclude: satpy/tests
--- a/.bandit
+++ b/.bandit
@@ -1,2 +1,2 @@
 skips: ["B506"]
-exclude: ["satpy/tests"]
+exclude_dirs: ["satpy/tests"]
--- a/.bandit
+++ b/.bandit
@@ -1,1 +1,2 @@
 skips: ["B506"]
+exclude: ["satpy/tests"]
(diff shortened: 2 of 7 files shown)


djhoese:
Codefactor WHY?!?!?!?!?

I just can't seem to get it to properly use the `.bandit` config. When I run `bandit --ini .bandit -r satpy/` locally then it works.

gerritholl:
I tested this PR for my GLM/ABI mixed use case, and everything seems to work as desired and expected.

satpy/scene.py
```diff
@@ -135,16 +135,15 @@ def sensor_names(self) -> set[str]:
         Sensor information is collected from data contained in the Scene
         whether loaded from a reader or generated as a composite with
         :meth:`load` or added manually using ``scn["name"] = data_arr``).
-        If no data is currently contained in the Scene or no sensor
-        information is found in the data metadata, loaded readers will be
-        consulted for sensor information.
+        Sensor information is also collected from any loaded readers.
+        In some rare cases this may mean that the reader includes sensor
+        information for data that isn't actually loaded or even available.
```
gerritholl, line 140: How can this happen?

gerritholl (approved):
Looks good to me.  As already stated, I tested it for my real use case, and it works for me.

satpy/scene.py
djhoese: Most likely bad configuration by the reader developer. Like if the file handler doesn't look at what file it is and just blindly says all possible sensors. Or if the file handler have a sensor names property, I think the reader defaults to the sensors listed in the YAML.

mraspaud (approved):
LGTM! Thanks for solving the .bandit issue with codefactor

PR #1905, opened 2021-11-29
Updated mpef product header to include new fields

ColinDuff:
- [x] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented 


satpy/tests/reader_tests/test_eum_base.py
```diff
@@ -86,6 +86,14 @@ def test_fun(self):
             ('PlannedRepeatCycleEnd', time_cds_expanded)
         ])
 
+        test_mph = [
```
sfinkens, line 89: Can you please split that in two test methods, for example `test_timestamps` and `test_mpef_header`?

satpy/readers/seviri_base.py
```diff
@@ -449,7 +449,12 @@ def get(self):
             ('TerminationSpace', 'S1'),
             ('EncodingVersion', np.uint16),
             ('Channel', np.uint8),
-            ('Filler', 'S20'),
+            ('ImageLocation', 'S3'),
```
sfinkens, line 452: Is there a way to test these changes directly? At the moment they are only tested indirectly (there's a simplified version of that data type in the test).

sfinkens (commented):
Thanks for adding this! I have a couple of questions.

satpy/readers/eum_base.py
```diff
@@ -76,9 +76,9 @@ def recarray2dict(arr):
                     try:
                         data = data.decode()
                     except ValueError:
-                        pass
+                        continue
```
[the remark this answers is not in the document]
ColinDuff: hi, thanks for the review. When i was trying to read in an MSG Bufr product with the new mpef product header format it raised a ValueError when trying to decode one of the header strings but as it just called pass it then tried to use the data value and it crashed. I would have thought continue would be better here as then would skip trying to use the data variable and continue without crashing. thoughts?

satpy/tests/reader_tests/test_eum_base.py
ColinDuff: yes , will do

satpy/readers/eum_base.py
```diff
@@ -76,9 +76,9 @@ def recarray2dict(arr):
...
                     data = data.split(':')[0].strip()
-                res[key] = data
+                    res[key] = data
```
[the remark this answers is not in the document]
ColinDuff: let me check this

satpy/tests/reader_tests/test_eum_base.py
```diff
@@ -101,6 +102,27 @@ def test_fun(self):
 
         self.assertEqual(recarray2dict(pat), expected)
 
+    def test_mpef_product_header(self):
+        """Test function for TestRecarray2Dict and mpef product header."""
+        mpef_header = np.dtype([
```
sfinkens, line 107: I'm still a bit concerned that the datatype used for testing is a copy of the original datatype (`mpef_product_header`). I guess testing the conversion of the entire header would be the best solution, but that's probably beyond the scope of this PR :slightly_smiling_face:  Maybe subsetting the original is a good compromise?

```python
from satpy.readers.seviri_base import mpef_product_header
names = ['ImageLocation', 'GsicsCalMode', 'GsicsCalValidity', 'Padding', 'OffsetToData' 'Padding2']
mpef_header = np.dtype([(name, mpef_product_header.fields[name][0]) for name in names])
mph_string = np.array(..., dtype=mpef_header)
```

satpy/readers/eum_base.py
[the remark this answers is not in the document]
sfinkens: Thanks for the clarification! I see now that the original exception was about Python 2/3 compatibility. This seems to be a different problem. In that case it doesn't make sense to proceed. But if the code just continues with the next key, the problematic key will be missing in the result. Is it possible to set `res[key] = None` in this case?

satpy/tests/reader_tests/test_eum_base.py
sfinkens: Thanks!

sfinkens (approved):
Nice work, looks good to me now!

djhoese:
pre-commit.ci autofix

satpy/readers/eum_base.py
```diff
@@ -76,7 +76,8 @@ def recarray2dict(arr):
                     try:
                         data = data.decode()
                     except ValueError:
-                        pass
+                        res[key] = None
+                        continue
                     data = data.split(':')[0].strip()
```
djhoese, line 81:
```suggestion
                        data = None
                    else:
                        data = data.split(':')[0].strip()
```

Thoughts?

ColinDuff:
hi, if i understand your proposed change correctly the  
    data = data.split(':')[0].strip()
line has to be within the 
    if ntype[:2] == '|S':
condition as we also pass non string types in here so the data.split would fail for these types

djhoese:
I was attempting to make a `try/except/else` block (meaning, not part of the other `if/else`s). If I missed some logic that makes this not work then ignore the suggestion.

ColinDuff:
understand what you mean now David, and code updated as per your proposal

PR #1908, opened 2021-11-30
Update AHI HSD calibration coefficients

simonrp84:
Currently, the AHI HSD reader uses the default calibration coefficients for VIS channels: `gain_count2rad_conversion` and `offset_count2rad_conversion` in Block 5 of the header.
However, updated calibration coefficients are available in Block 6 of the header: `cali_gain_count2rad_conversion` and `cali_offset_count2rad_conversion`.

This PR switches from using the old coefficients by default to using the new ones by default.

In addition, this PR removes the clipping of negative radiances. In the current `main` branch any radiances below zero are automatically set to zero. This is undesirable as negative radiances are useful for analysing sensor noise, amongst other things, so I remove this unnecessary clipping.
There is also clipping for the reflectance values, but I've left that in for further discussion.

### DIFF
--- a/satpy/readers/ahi_hsd.py
+++ b/satpy/readers/ahi_hsd.py
@@ -307,7 +307,7 @@ class AHIHSDFileHandler(BaseFileHandler):
     """
 
     def __init__(self, filename, filename_info, filetype_info,
-                 mask_space=True, calib_mode='nominal',
+                 mask_space=True, calib_mode='update',
                  user_calibration=None):
         """Initialize the reader."""
         super(AHIHSDFileHandler, self).__init__(filename, filename_info,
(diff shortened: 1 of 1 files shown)


mraspaud (commented):
Tests are failing, otherwise lgtm

sfinkens (commented):
Thanks, makes sense to use the latest coefficients! Can you please also update in the docs that this is the new default? 

Regarding radiance clipping: Although I agree that clipping should at least be optional, maybe we should wait for a decision in #1911?

simonrp84:
Ok, it seems that the tests are complaining about coverage in an unaffected file. Odd. Anyway, AHI coverage remains the same.

Thanks for the pointer about documentation @sfinkens - I've updated the docs now.

mraspaud (approved):
LGTM

djhoese (approved):

PR #1909, opened 2021-12-01
Refactor 'ahi_hsd' reader and make overview/overview_raw consistent with ABI

djhoese:
This is a mini-PR on the path toward improving the performance of the ahi_hsd/abi_l1b true_color generation as discussed in #1902. This is mostly just cleanup with a small modification to the space masking code to pass the chunk size of the data that the mask will be used on. This is just to avoid future issues where fancy chunking may not match with the `PYTROLL_CHUNK_SIZE`.

I also updated the `overview` composite to add an `overview_raw` so they match what the ABI composites do. I also made the composites based on channel name instead of wavelength.

About pre-commit: bandit complained about the use of PIPE in the utils module, but I'm not sure there is an easy way around that. Mypy also complained about an import in `_compat.py` but I couldn't seem to make it happy (the import only exists in Python 3.8+ and it must be running in Python 3.7). I turned off hooks so I could do the last commit.

 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/etc/composites/ahi.yaml
+++ b/satpy/etc/composites/ahi.yaml
@@ -155,13 +155,22 @@ composites:
     standard_name: fire_temperature
     name: fire_temperature_39refl
 
-
   overview:
     compositor: !!python/name:satpy.composites.GenericCompositor
     prerequisites:
-    - 0.65
-    - 0.85
-    - 10.4
+    - name: B03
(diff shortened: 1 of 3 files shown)


sfinkens (approved):
Nice work, LGTM!

PR #1917, opened 2021-12-02
Add support to read and visualize NOAA GOESR L2+ cloud mask products

yufeizhu600:
- [x] Fully documented 

NOAA has added two new cloud mask products into GOES-R L2+ Cloud Mask products since Nov. 29th 2021. source:  [NOAA OSPO satellite messages](https://www.ospo.noaa.gov/data/messages/2021/MSG333194402.html). These two products are the `4 Level Cloud Mask` product with variable name `ACM` and the `Cloud Probability` product with variable name `Cloud_Probabilities`.   
This PR add support to read those two new variables in the abi_l2_nc reader, and also set the colorization enhancement as default for `BCM`, `ACM` and `Cloud_Probabilities`. The colormap definitions can be found in the enhancement config file. 
Here are a set samples of the enhancements:

1. Colorized binary cloud mask `BCM`

binary_cloud_mask_2021336181021

2. Colorized four level cloud mask `ACM`

four_level_cloud_mask_2021336181021-1

3. Colorized cloud probability  `Cloud_Probabilities`

cloud_probability_2021336181021

Here is code sample to generate the enhanced images:
```
#!/usr/bin/env python
import fsspec
from satpy import Scene

import satpy

debug_on()

#https://noaa-goes16.s3.amazonaws.com/ABI-L2-ACMF/2021/336/12/OR_ABI-L2-ACMF-M6_G16_s20213361200213_e20213361209521_c20213361211176.nc
start_time = '20213361200213'

filename = 'noaa-goes16/ABI-L2-ACMF/2021/336/12/OR_ABI-L2-ACMF-M6_G16_s' + start_time + '*.nc'

s3_files = fsspec.open_files("simplecache::s3://" + filename, s3={'anon': True})

fs_files = [FSFile(open_file) for open_file in s3_files]

scn = Scene(filenames=fs_files, reader='abi_l2_nc')

variables = ['BCM', 'ACM', 'Cloud_Probabilities']
scn.load(variables)

for var in variables:
    scn.save_dataset(var, './' + var + '_' + start_time + '.tif')
```

### DIFF
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -363,6 +363,7 @@ composites:
       - name: C04
       - name: C02
       - name: C05
+
   ash:
     description: >
       Ash RGB, for GOESR: NASA, NOAA
@@ -680,3 +681,22 @@ composites:
     prerequisites:
       - name: C14
     standard_name: highlighted_toa_brightness_temperature
+
+  # NOAA GOES-R Level-2 ABI Cloud Mask product  
+  binary_cloud_mask:
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+      - BCM
+    standard_name: binary_cloud_mask
+
+  four_level_cloud_mask:
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+      - ACM
+    standard_name: four_level_cloud_mask
+
+  cloud_probability:
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+      - Cloud_Probabilities
+    standard_name: cloud_probability
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -682,7 +682,7 @@ composites:
       - name: C14
     standard_name: highlighted_toa_brightness_temperature
 
-  # NOAA GOES-R Level-2 ABI Cloud Mask product  
+  # NOAA GOES-R Level-2 ABI Cloud Mask product
   binary_cloud_mask:
     compositor: !!python/name:satpy.composites.GenericCompositor
     prerequisites:
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -619,12 +619,10 @@ composites:
 
   cloud_phase:
     description: >
-      Cloud Phase RGB, for EUMETSAT
-      Day Cloud Phase RGB, for EUMETSAT (https://www.eumetsat.int/website/home/Images/ImageLibrary/DAT_2861499.html)
-      "When we use the NIR2.3 instead of the VIS0.8 on the green beam, we can devise a new RGB product (let us call it 'Day Cloud Phase RGB') that has similar cloud colours than the Natural Colour RGB, but with improved separation of ice and water clouds."
+      EUMETSAT Cloud Phase RGB product
     references:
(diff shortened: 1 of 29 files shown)


satpy/etc/composites/abi.yaml
```diff
@@ -679,3 +679,22 @@ composites:
...
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+      - ACM
+    standard_name: four_level_cloud_mask
+
+  cloud_probability:
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+      - Cloud_Probabilities
+    standard_name: cloud_probability
```
djhoese, line 700: Do we need separate composites for this? Could you make the enhancements use `name: BCM` (for example) and have this work the same?

djhoese: Or does the enhancement change the end result so much that you would prefer them to be separate products?

yufeizhu600: These composites added customized color map/color palette to the visualization of these variables. In my experience these color maps are quite useful especially for the 4-level cloud mask and the cloud probability; because without the color map the default image will be in grey scale, which is hard to differentiate the different mask levels/cloud probabilities. 
The binary cloud mask (BCM) works OK in a black/white image, however, by assign same colors scheme as the new 4-level-cloud-mask, it is easy to see the difference/improvement of the new products.

djhoese: Sure, but the composite isn't doing anything here. Why not remove the composite definition and update the enhancements to use the product from the reader directly? That way someone doing `.load(['BCM'])` would get the nice colorized version by default.

yufeizhu600: OK, I see what you mean now, I will try to make it as default in the enhancement config file.

yufeizhu600: I have given a second though and think the composites are still necessary, because the cloud masks are not only use for visualization, but also can be used for masking other products. In the later case, user need to load the values in the data file instead of the colors by default enhancement; for example, to generate an ABI cloud product similar to the [one for SEVIRI](https://github.com/pytroll/satpy/blob/72dbcd25696e39423a56c078c3d7038e19bfdc7b/satpy/etc/composites/seviri.yaml#L29). Is that make sense?

djhoese: The enhancements are **only** applied when saving an image (PNG, geotiff, etc). They will not effect someone using the dataset inside a composite unless the composite explicitly pre-enhances the dataset. For example, the DayNightCompositor could put cloud mask on the day side and night microphysics on the night side of an image, but it must enhance both of them before combining them.

If someone wants to use a cloud mask outside of Satpy then they shouldn't be saving it to an image format and should instead either be using NetCDF or saving as a geotiff with `enhance=False` to make sure the enhancement is not used.

Let me know if I'm missing something here.

djhoese: As for your reference of the seviri cloud product, doesn't that `ct` that is used in that composite come directly from the reader?

yufeizhu600: Yes, you are right. I realized that the enhancement is applied at the stage save dataset to image not at the time loading the dataset. I confused the functionalities which applied to the 'raw' dataset during the loading with enhancement at certain point. Thanks for clarification.

djhoese (approved):
Looks good to me. Thanks for modifying it for me.

PR #1920, opened 2021-12-03
Bugfixes seviri l2 grib reader

strandgren:
This purpose of this PR is to standardize the area definitions of the SEVIRI L2 grib products. During implementation three bugs were identified and resolved:

1. When looping over multiple grib files, information from a previously opened grib file would sometimes be passed on to the next call. This has been resolved by always lopping over all messages in the grib file, such that no additional information can be passed on. An issue for this unexpected behavior has been created in the python-ecCodes github (https://github.com/ecmwf/eccodes-python/issues/58).

2. The computation of the area extent did not provide correct results for segmented products. The implementation used in the seviri_l1b_native reader is now used (the method has been moved to seviri_base to allow both reader to access it).

3. In the aerosol over sea product the Earth's minor axis is provided in the wrong unit (off by a factor 1e8). A work-around has been temporarily implemented that will make sure that all earth axis dimensions are of the correct magnitude. This work-around can be used until the issue has been resolved by EUMETSAT. 

 - [x] Tests added

### DIFF
--- a/satpy/etc/areas.yaml
+++ b/satpy/etc/areas.yaml
@@ -114,6 +114,58 @@ msg_seviri_iodc_1km:
     lower_left_xy: [-5571248.412732527, -5566247.740968115]
     upper_right_xy: [5566247.740968115,  5571248.412732527]
 
+# Full disk - segmented products
+msg_seviri_fes_9km:
+  description:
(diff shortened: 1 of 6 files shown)


mraspaud (approved):
Nice job, LGTM!

PR #1922, opened 2021-12-06
Fix ABI cloud_phase composite recipe and enhancement

yufeizhu600:
The current Eumetsat cloud phase composite in the abi.yaml config file seems not correct according to the recipe found on the  [Eumetsat website](http://eumetrain.org/RGBguide/recipes/RGB_recipes.pdf). 
eum_cloud_phase
Also the link to the reference documentation in the cloud phase composite description is dead. 
This PR has fixed the recipe in the enhancement and composite config files. Below is a sample of the cloud phase composite.
cloud_phase_2021340175021-1
For the interpretation of the colours in the image, please refer to this [Cloud Phase RGB Quick Guide](http://www.eumetrain.org/rgb_quick_guides/quick_guides/CloudPhaseRGB.pdf)

For comparison, here is a sample from the current implementation of the cloud phase composite recipe.
cloud_phase_nogood

### DIFF
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -363,6 +363,7 @@ composites:
       - name: C04
       - name: C02
       - name: C05
+
   ash:
     description: >
       Ash RGB, for GOESR: NASA, NOAA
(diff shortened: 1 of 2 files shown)


yufeizhu600:
For comparison, here is a sample from the current implementation of the cloud phase composite recipe.
cloud_phase_nogood

djhoese (approved):
Wow, nice catch. Looks like the enhancement was never added at all. This looks good to me. The codebeat issues seem to be unrelated and probably due to you starting from an old `main` branch. I'll give others a chance to review this too, but otherwise I can merge this tomorrow if no one has anything else to say (remind me to merge this on slack if I forget).

PR #1926, opened 2021-12-07
Update seadas_l2 chlor_a enhancement to use new log10 stretch

djhoese:
Depends on https://github.com/pytroll/trollimage/pull/97 but since this won't fail until it is actually used this could be merged before that trollimage PR is merged and released.

This enhancement now matches what the Polar2Grid user is expecting (as far as we were told).

 - [ ] Closes #xxxx 
 - [ ] Tests added


djhoese:
This is such a small PR and effects no one else so I'm merging without review with the assumption that no one has a problem with this.

PR #1928, opened 2021-12-08
Enable writing to cloud-optimized geotiff (COG)

pnuu:
This PR enables the writing of cloud-optimized geotiffs added to Trollimage in https://github.com/pytroll/trollimage/pull/94

At the bare minimum, this enables the writing using the `'COG'` GDAL driver:
```python
scn.save_datasets(driver='COG')
```

The downside is that this driver seems to be a lot slower. For my test data (Himawari-8/AHI in HRIT format) and images (18 composites) it takes 112 seconds compared to 93 s for `scn.save_datasets(base_dir='/tmp/', tiled=True, blockxsize=512, blockysize=512, overviews=[32, 16, 8, 4, 2])`.

 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/tests/test_writers.py
+++ b/satpy/tests/test_writers.py
@@ -530,6 +530,10 @@ class TestYAMLFiles(unittest.TestCase):
 class TestComputeWriterResults(unittest.TestCase):
     """Test compute_writer_results()."""
 
+    @pytest.fixture(autouse=True)
+    def inject_fixtures(self, caplog):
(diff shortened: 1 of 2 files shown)


pnuu:
Ok, setting `GDAL_NUM_THREADS=4` makes also the writing multithreaded and the processing time dropped to 70 seconds.

pnuu:
Nevermind, there already is #1882.

PR #1934, opened 2021-12-12
Factorize area computation in goes_imager_hrit

### LINKED ISSUE #1022: Factorize area def computation in goes_imager_hrit
## Feature Request

`HRITMSGFileHandler` now has two methods `_get_area_extent` and `get_area_extent` to compute the area extent. The latter is inherited from `HRITFileHandler` and yields incorrect results for SEVIRI.

The only reader using `HRITFileHandler.get_area_extent` is `HRITGOESFileHandler`. So we could remove that method and use the new helper methods from `satpy.readers._geos_area` in `HRITGOESFileHandler` instead.

None

**Additional context**
None

sfinkens:
Use the helper methods from `satpy.readers._geos_area`, which have been around for a while now. Added the test first to make sure the extents remain unchanged.

 - [X] Closes #1022 
 - [X] Tests added

### DIFF
--- a/satpy/readers/goes_imager_hrit.py
+++ b/satpy/readers/goes_imager_hrit.py
@@ -30,8 +30,8 @@ from datetime import datetime, timedelta
 import dask.array as da
 import numpy as np
 import xarray as xr
-from pyresample import geometry
 
+from satpy.readers._geos_area import get_area_definition, get_area_extent, get_geos_area_naming
 from satpy.readers.eum_base import recarray2dict, time_cds_short
 from satpy.readers.hrit_base import (
     HRITFileHandler,
@@ -352,6 +352,8 @@ SPACECRAFTS = {
     14: "GOES-14",
     15: "GOES-15"}
 
+SENSOR_NAME = 'goes_imager'
+
 
 class HRITGOESFileHandler(HRITFileHandler):
     """GOES HRIT format reader."""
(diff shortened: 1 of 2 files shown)


sfinkens:
I think that this was the last reader to make use of `satpy.readers.hrit_base.HRITFileHander.get_area_def`. At least I can remove that method and all `*hrit*` reader tests still pass. Should we remove it then?

sfinkens:
I also noticed that many GEO reader tests are very similar when it comes to verifying the area def. For example

```
        a, b = proj4_radius_parameters(area.proj_dict)
        assert a == EQUATOR_RADIUS
        assert b == POLE_RADIUS
        assert area.proj_dict['h'] == ALTITUDE
        assert area.proj_dict['lon_0'] == 100.1640625
        assert area.proj_dict['proj'] == 'geos'
        assert area.proj_dict['units'] == 'm'
        assert area.width == 2816
        assert area.height == 464
        assert area.area_id == 'goes-15_goes_imager_fd_3km'
        area_extent_exp = (-5639254.900260435, 1925159.4881528523,
                           5643261.475678028, 3784210.48191544)
        np.testing.assert_allclose(area.area_extent, area_extent_exp)
```

Should we add a helper method for that to `satpy.tests.utils`?

djhoese:
> Should we add a helper method for that to satpy.tests.utils?

If you can make it reusable and generic enough then I'd say go for it.

mraspaud (approved):
LGTM!

PR #1935, opened 2021-12-13
Add additional pre-commit checks for yaml syntax

djhoese:
I found an issue in my own software where I had some bad YAML where a dictionary key was used twice. I noticed xarray uses a yaml checker in their pre-commits so copied what they were using over here. I also updated the flake8 hook to use a newer version.


djhoese:
pre-commit.ci autofix

sfinkens (approved):
LGTM

djhoese:
One option to check for duplicate keys in the YAML files loaded/used by Satpy would be to add a custom constructor/loader as described in this gist:

https://gist.github.com/pypt/94d747fe5180851196eb

This wouldn't help the pre-commit in this PR, but at least the pre-commit is checking syntax.

djhoese:
Looks like pre-commit.ci choked so I ran the hooks for all files locally.

sfinkens:
A custom constructor sounds good. I guess this wouldn't hit performance?

djhoese:
> A custom constructor sounds good. I guess this wouldn't hit performance?

It depends how it is implemented in pyyaml. If it is all done in C in pyyaml then yeah big performance hit. If it is done in python then there is still a performance hit because we're adding extra conditions/checks. Since most of our configs are dictionaries/mappings this would be executed a lot. The custom constructor could be used only on unit tests so that we verify that the YAML files we have are what we expect, but I think the better long term solution is to rely on outside YAML linters/checkers. I'm not really liking the custom constructor idea the more I think about it.

mraspaud (approved):
Lgtm

PR #1942, opened 2021-12-15
Fix dynamic_dnb composite converting NaNs to 0s

djhoese:
Noticed this in a small edge case with Polar2Grid where on the edge of a bad granule where half was all NaNs and the other half was good data, there was a layer of "gray" pixels between. This was caused by resampling blending the valid data with the 0s of the bad data, but the bad data were meant to be NaNs and excluded. This all comes down to handling NaNs properly in the compositor: `data < 0` is not the same as `~(data >= 0)`.

 - [x] Tests added


mraspaud (approved):
LGTM, nice catch!

PR #1952, opened 2021-12-21
Keep original varnames of tropomi_l2

### LINKED ISSUE #1949: keep `delta_time` instead of renaming to `offset_time`
**Describe the bug**

In the official TROPOMI products, it's `delta_time` but renamed to `offset_time`.
https://github.com/pytroll/satpy/blob/f9013963a632873731dee3d8f5c2112392495854/satpy/etc/readers/tropomi_l2.yaml#L45-L49

```
int delta_time(time=1, scanline=2906);
  :long_name = "offset of start time of measurement relative to time_reference";
  :units = "milliseconds since 2019-07-02 00:00:00";
  :_FillValue = -2147483647; // int
  :_ChunkSizes = 1U, 2906U; // uint
```

Is it better to keep it as the original name? I'm producing the data for some KNMI researchers. They usually use `delta_time`.

zxdawn:
Change `offset_time` and `ref_time` to `delta_time` and `time`, respectively, as same as the original file.

 - [x] Closes #1949


satpy/etc/readers/tropomi_l2.yaml
```diff
@@ -42,16 +42,16 @@ datasets:
...
-    offset_time:
-        name: 'offset_time'
+    delta_time:
+        name: 'delta_time'
         file_type: tropomi_l2
         file_key: 'PRODUCT/delta_time'
-        standard_name: offset_time
-    ref_time:
-        name: 'ref_time'
+        standard_name: delta_time
```
djhoese, line 49: Just curious, do the input files have a standard_name for this? Is it also `delta_time`? Could we (maybe in a future PR) just copy the standard_name from the file's attributes?

satpy/etc/readers/tropomi_l2.yaml
```diff
@@ -42,16 +42,16 @@ datasets:
...
+    time:
+        name: 'time'
         file_type: tropomi_l2
         file_key: 'PRODUCT/time'
-        standard_name: ref_time
+        standard_name: time
```
djhoese, line 54: Isn't there a more CF standard name for time? What are the units of this dataset?

satpy/etc/readers/tropomi_l2.yaml
zxdawn: It's a good idea. The `standard_name` is `time`:
```
int time(time=1);
  :units = "seconds since 2010-01-01 00:00:00";
  :standard_name = "time";
  :axis = "T";
  :long_name = "reference time for the measurements";
  :comment = "The time in this variable corresponds to the time in the time_reference global attribute";
  :_FillValue = -2147483647; // int
  :_ChunkSizes = 1U; // uint
```

satpy/etc/readers/tropomi_l2.yaml
zxdawn: Units is `  :units = "seconds since 2010-01-01 00:00:00";`. The standard name is `time`.
```
int time(time=1);
  :units = "seconds since 2010-01-01 00:00:00";
  :standard_name = "time";
  :axis = "T";
  :long_name = "reference time for the measurements";
  :comment = "The time in this variable corresponds to the time in the time_reference global attribute";
  :_FillValue = -2147483647; // int
  :_ChunkSizes = 1U; // uint
```

gerritholl:
@zxdawn Any update on this?

tommyjasmin:
Hey all - me and Bob took a peek at this today and I think your change is fine Xin.  I was concerned about backward compatibility but I have sample data going back to 2018 which have the same naming convention.   I'm not sure why I would have used the names `offset_time` and `ref_time` in the first place, other than possibly because the variable long name attributes have those terms:

```     
 int time(time=1);
        ...
        :long_name = "reference time for the measurements";
```

```
int delta_time(time=1, scanline=373);
        ...
        :long_name = "offset of start time of measurement relative to time_reference";
```

In any case, I agree preserving the terms from the source data files makes sense.

zxdawn:
@tommyjasmin Thanks for the check Tommy.

I merged the main branch now and got the lint error:
```
satpy/tests/writer_tests/test_ninjotiff.py:99:15: E275 missing whitespace after keyword
Error: Process completed with exit code 1.
```
Em ... It's not caused by my PR.

djhoese:
Strange, that should have been caught a lot earlier. Would you mind fixing it? If it seems like more than just a space or more than one error then I'll try to find the time to do it.

djhoese:
@zxdawn Could you try re-merging `main`? I just merged a PR that fixes all the linting issues.

zxdawn:
@djhoese Sorry for the late reply. Merged now.

PR #1956, opened 2021-12-27
Fix broken links in composite references

gerritholl:
In the references included in the composite definition YAML files, fix broken links to documents on the CIRA and EUMETSAT websites.

There is one change I am not sure about. @jhbravo, could you please confirm: "Cloud Type recipe and typical colours" previously linked to https://www.eumetsat.int/website/wcm/idc/idcplg?IdcService=GET_FILE&dDocName=PDF_IL_18_05_13_A&RevisionSelectionMethod=LatestReleased&Rendition=Web, which is now Error 404.  The closest equivalent I could find is https://www.eumetsat.int/fr/media/45659.  Is that the correct new link?

 - [x] Fully documented

### DIFF
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -321,7 +321,7 @@ composites:
       appear red. With the increasing intensity and temperature the fires will also be detected
       by the 2.2 μm and 1.6 μm bands resulting very intense fires in white.
     references:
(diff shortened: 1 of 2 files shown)


djhoese (approved):

gerritholl:
NB: @jhbravo has confirmed in slack that the new link is correct.

PR #1969, opened 2022-01-07
Improve modifier documentation

gerritholl:
Improve documentation on modifiers.  Added a section to the
documentation dedicated to modifiers.

 - [ ] Closes #xxxx 
 - [x] Fully documented


mraspaud (approved):
LGTM

doc/source/composites.rst
```diff
@@ -347,6 +347,12 @@ Here we see two changes:
 The modifier above is a built-in that normalizes the Solar zenith
 angle to Sun being directly at the zenith.
 
+More examples can be found in Satpy source code directory
+``satpy/etc/composite.yaml``.
```
djhoese, line 351: Could/should this maybe be a link to the actual directory? As is `composite.yaml` isn't a file or directory in the repository.

doc/source/index.rst
```diff
@@ -59,6 +59,7 @@ the base Satpy installation.
     composites
     resample
     enhancements
+    modifiers
```
djhoese, line 62: Tempted to say this should go under composites. Thoughts?

djhoese (changes requested):
Had a few questions/suggestions but otherwise looks pretty good. The only other thing is that I noticed you use two spaces after every period. Speaking for myself, I have always used a single space in satpy documentation. We should probably be consistent. Maybe others can way in on single versus double spaces between sentences?

doc/source/composites.rst
gerritholl: Oops, of course this should be `satpy/etc/composites` and not `composite.yaml`.  As for link to the actual directory, you mean on GitHub?

doc/source/index.rst
gerritholl: I don't mind either way.  I can move it if you prefer.

doc/source/composites.rst
djhoese: Yeah, link to the directory on GitHub.

gerritholl:
> you use two spaces after every period

Yes, I do.  I know most style guides speak for one space after a period.  I find two spaces after a period easier to read, and old habits die hard.  I can change it to one space if you prefer.

doc/source/index.rst
djhoese: Logically composites and modifiers are the most similar and are used at the same time in Satpy data generation/loading. However, `modifiers` are also not a "primary" topic that users need to worry about initially. I guess if you're making other modifications I'd prefer if you moved this.

gerritholl: Moved to be a subsection of composites.

doc/source/composites.rst
gerritholl: Fixed.

gerritholl:
What about the build failure?

```
intersphinx inventory 'https://docs.scipy.org/doc/numpy/objects.inv' not readable due to ValueError: unknown or unsupported inventory version: ValueError('invalid inventory header: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">')
```

does not sound like something I've done, does it?

PR #1978, opened 2022-01-14
Fix reflectances for VII l1b reader V1

ninahakansson:
Fix the scaling of reflectances for VII l1b reader.
There should be a multiplication with 100 in the conversion to reflectances as reflectances have unit %.
The band averaged solar irradiance is already calculated per µm and the wavelengths
should not need to be used for scaling. However there seem to be a scale factor
missing (1e3) for band averaged solar irradiance. Also channel cw thermal has a missing scale factor 1e-3.
This might be fixed in later versions of the testdata or in the actual data.
With this update the resulting reflectances are more similar to the metop reflectances used to simulate the testdata.

 - [ ] Closes #xxxx 
 - [ ] Tests added


ninahakansson:
This is no longer needed for the new version of data.

PR #1988, opened 2022-01-21
Update geometry.py docstring from compositor to modifier

mherbertson:
This changes the docstring example from compositor to modifier.


pnuu (approved):
LGTM

PR #1991, opened 2022-01-24
Update reference to dask distributed setup page

mherbertson:
Updated the intersphinx reference to the dask deploying-python.rst

 - [x] Fully documented


mraspaud (approved):
LGTM! The sphinx build passes now, so merging. Thanks @mherbertson !

PR #1993, opened 2022-01-24
Add 'l2_flags' quality filtering to 'seadas_l2' reader

djhoese:
In an attempt to improve the quality of the chlorophyll product, @kathys and I are looking at adding the ability to use the file's `l2_flags` to remove low-quality pixels. This PR is an initial attempt at this work while we try different flag combinations.

 - [x] Tests added 
 - [ ] Fully documented


djhoese:
Confirmed with @kathys that these settings are what we want. Merging.

PR #1994, opened 2022-01-24
Refactor 'native' resampler to produce less dask tasks

djhoese:
This is a small quality of debugger life change. While debugging task graphs it because confusing to determine when the native resampler was being used in the overall graph. Additionally, my new general goal is to remove unnecessary dask tasks being created in satpy. Although the changes in this PR don't improve performance in a measurable way, it does overall clean up the dask graph and code when debugging.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #2002, opened 2022-01-31
Update tests to be more flexible to CRS and enhancement changes

djhoese:
This is an update needed for https://github.com/pytroll/pyresample/pull/415 so that satpy tests still pass with those pyresample changes. One lesson learned is that `+init=EPSG:XXXX` does not respect the axis order of the EPSG definition. Using `EPSG:XXXX` does.

This PR also includes refactoring of the enhancements tests because they were causing failures when combined with my trollimage PR here: https://github.com/pytroll/trollimage/pull/99

 - [x] Tests added


pnuu (approved):
LGTM

PR #2011, opened 2022-02-07
Fix memory leak in cached_property backport

### LINKED ISSUE #2008: abi_l1b reader leaks memory in Python-3.7
**Describe the bug**
The `abi_l1b` reader leaks memory in Python-3.7, see the screenshot below. I suspect this is due to the custom `cached_property` decorator that is used to cache the file object: https://github.com/pytroll/satpy/blob/34bd429235293a87de99062e6b99a98eb17df8ce/satpy/readers/abi_base.py#L57

At least if I replace that with a normal `@property` the memory footprint is constant. But I don't know why.

**To Reproduce**

```python
import satpy
import glob

filenames = glob.glob('*C13*')  # one day of data
for filename in filenames:
    scn = satpy.Scene([filename], reader='abi_l1b')
    scn.load(['C13'])
    scn['C13'].compute()
```

**Screenshots**
memory_leak_abi_l1b
Blue: Python-3.7
Black: Python-3.8

**Environment Info:**
 - OS: Linux
 - Satpy Version: 0.33.1

sfinkens:
Closes #2008 by replacing the existing backport with the `cached_property` implementation from the Python-3.8 repository.

 - [X] Closes #2008 
 - [X] Tests added

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -264,7 +264,7 @@ intersphinx_mapping = {
     'dask': ('https://docs.dask.org/en/latest', None),
     'geoviews': ('http://geoviews.org', None),
     'jobqueue': ('https://jobqueue.dask.org/en/latest', None),
-    'numpy': ('https://docs.scipy.org/doc/numpy', None),
+    'numpy': ('https://numpy.org/doc/stable', None),
     'pydecorate': ('https://pydecorate.readthedocs.io/en/stable', None),
     'pyorbital': ('https://pyorbital.readthedocs.io/en/stable', None),
     'pyproj': ('https://pyproj4.github.io/pyproj/dev', None),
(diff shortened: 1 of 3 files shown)


sfinkens:
Example with one day of ABI data.

memory_leak_abi_l1b(2)

Black: Current implementation with Python-3.7
Blue: Fix with Python-3.7

djhoese:
Where does this solution come from? Stackoverflow? Official CPython docs somewhere?

djhoese:
If you update line 267 in `doc/source/conf.py` to say:

```
    'numpy': ('https://numpy.org/doc/stable', None),
```

it should fix the build_website failure. Numpy updated their documentation structure.

sfinkens:
> 'numpy': ('https://numpy.org/doc/stable', None),

Done, thanks!

djhoese (approved):
This looks good to me. It is a much more involved solution than I thought it was going to be. I think given that we are removing Python 3.7 support I think we should merge this, make a release, then immediately remove this as part of a PR to drop Python 3.7 support. Thoughts @sfinkens @mraspaud @pnuu ?

mraspaud (approved):
LGTM as a patch for the last release supporting python 3.7 :)

PR #2015, opened 2022-02-09
Fix various dask array bugs in CREFL modifier

djhoese:
@ralphk11 pointed out to me on slack some issues he was having with processing ABI data with the CREFL modifier. Turned out to be two separate bugs. One was that there was a `map_blocks` call with no `dtype` or `chunks` defined so dask was computing the result immediately. The second issue is that crefl was passing DataArrays to `da.map_blocks`. This apparently passes the entire DataArray array as a single chunk (it isn't recognized as containing a dask array). This wasn't caught in the tests because the tests were using dask arrays of one giant chunk (one giant chunk == entire array).

 - [x] Tests added


mraspaud (approved):
LGTM

PR #2025, opened 2022-02-16
Allow skipping unit conversion in NinJoTIFF

### LINKED ISSUE #2024: Allow to skip unit conversion in ninjotiff writer
## Feature Request

We are using infrared `SingleBandCompositor` composites with a crude stretch enhancement to write files with multiple writers, including the ninjotiff writer.  The ninjotiff and ninjogeotiff files are described in °C units (user requirement), whereas the input data are in K.  In this case, the ninjotiff writer converts pixel values from K to °C before enhancing the data:

https://github.com/pytroll/satpy/blob/82112a60f843ad0d3fa32c6d49f224bcb00f28d7/satpy/writers/ninjotiff.py#L118-L122

Other writers do not perform any unit conversion or do so in a different way (see #2021).  The consequence is that data values passed to the crude stretch enhancement for ninjotiff production are offset by 273.15 compared to those passed for production with other writers.  For example, if the geotiff or ninjogeotiff writer get

```yaml
kwargs: {stretch: crude, min_stretch: 313.5, max_stretch: 186}
```

then producing the same images with the ninjotiff writer would need

```yaml
kwargs: {stretch: crude, min_stretch: 40, max_stretch: -87}
```

which is a problem.

The good news is that unit conversion isn't actually needed with the ninjotiff writer.  So far, we accidentally didn't use any, by using a `GenericCompositor` rather than a `SingleBandCompositor`, which means our units attribute was lost and conversion did not occur.  Instead, we pass `ch_min_measurement_unit` and `ch_max_measurement_unit` (in °C), which the ninjotiff writer uses to calculate the scale (slope) and offset (axisintercept), independently of what's happening in the crude stretch (it's up to the user to ensure consistency).  By skipping unit conversion altogether and setting temperature limits both in the enhancement definition and in the writer parameters, we get images that look as intended and have the correct values in °C encoded.  This is not ideal either, but corresponds to the status quo.

I would like that the ninjotiff writer gets a way to skip unit conversion.  For example, a parameter `convert_temperature_units`, which defaults to True.  If set to False, no unit conversion is performed, and the same crude stretch enhancement parameters can be used when producing ninjotiff or other formats.

None.  The status quo would remain the default behaviour.

**Additional context**

Other solutions could be:

- Define special composites for ninjotiff production, with their own enhancements, where parameters are defined in °C rather than K.  This would work, but it would require duplicating all IR composites and enhancements and make trollflow2 production more complicated.  Products would be defined in two different places, which is not ideal for maintenance purposes.
- Change the behaviour of the ninjogeotiff writer to be similar to the ninjotiff writer (see #2021).  This would mean that the same enhancement configuration (in °C) can be used for both writers.  This is not ideal, because other writers (such as regular geotiff) would still need different parameters.
- Produce images in units of K and leave it to the client to convert units if users wish to display other units.  In this case the client would be NinJo.  There are good arguments for this approach and this might even be implemented eventually, but is not under my control and I need a solution sooner than such an implementation can be expected.
- Something else I'm not thinking of.

See also #2018 and #2021.

gerritholl:
Add an option to the NinJoTIFF writer to skip unit conversion.  For a
detailed rationale, see https://github.com/pytroll/satpy/issues/2024.

 - [x] Closes #2024
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/tests/writer_tests/test_ninjotiff.py
+++ b/satpy/tests/writer_tests/test_ninjotiff.py
@@ -67,6 +67,19 @@ class TestNinjoTIFFWriter(unittest.TestCase):
             uconv.assert_called_once_with(dataset, 'K', 'CELSIUS')
         self.assertEqual(iwsd.call_count, 1)
 
+    @mock.patch('satpy.writers.ninjotiff.ImageWriter.save_dataset')
+    @mock.patch('satpy.writers.ninjotiff.nt', pyninjotiff_mock.ninjotiff)
+    def test_dataset_skip_unit_conversion(self, iwsd):
+        """Test saving a dataset without unit conversion."""
+        from satpy.writers.ninjotiff import NinjoTIFFWriter
+        ntw = NinjoTIFFWriter()
+        dataset = xr.DataArray([1, 2, 3], attrs={'units': 'K'})
+        with mock.patch('satpy.writers.ninjotiff.convert_units') as uconv:
+            ntw.save_dataset(dataset, physic_unit='CELSIUS',
+                             convert_temperature_units=False)
+            uconv.assert_not_called()
+        self.assertEqual(iwsd.call_count, 1)
+
     @mock.patch('satpy.writers.ninjotiff.NinjoTIFFWriter.save_dataset')
     @mock.patch('satpy.writers.ninjotiff.ImageWriter.save_image')
     @mock.patch('satpy.writers.ninjotiff.nt', pyninjotiff_mock.ninjotiff)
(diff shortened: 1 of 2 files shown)


djhoese (approved):
I definitely shouldn't be the final word on this, but this seems like a reasonable solution. I think the general confusion that does/will arise for anyone "out of the know" trying to compare enhancements and data values will be happy with this and with stretch limits being specified in the units of the input data. I think there are other options (don't remember if you mentioned this in the related issue) that would require some duplicate code to reimplement `save_datasets` or maybe even a custom Enhancer object for the writer so that the unit conversion can happen after the initial Kelvin-based stretch.

PR #2029, opened 2022-02-20
Remove unused caching in sunz_corrected and rayleigh modifiers

djhoese:
As discussed on slack, it seems through my testing that the WeakValueDictionary being used isn't actually retaining any of the items after the modifier is called. My best guess is that we are using the dask array that is contained in a DataArray, but caching the DataArray. Since the dask array is the only thing that has a pointer to it after the modifier is called the DataArray gets garbage collected and removed from the WVD.

Although I could make this caching work I don't think it is buying us anything in our "modern" world of dask arrays. Also, I think this method of caching opens the possibility of false hits if our cache keys don't match the actual arguments of the function being called.

 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/modifiers/atmosphere.py
+++ b/satpy/modifiers/atmosphere.py
@@ -18,7 +18,6 @@
 """Modifiers related to atmospheric corrections or adjustments."""
 
 import logging
-from weakref import WeakValueDictionary
 
 import dask.array as da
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM

PR #2031, opened 2022-02-20
Update AHI HSD reader with observation/scheduled times and nominal satellite position

### LINKED ISSUE #2012: Define time metadata options and usage

djhoese:
This is the proof of concept for the time discussion had in #2012 and towards using #2030 to optimize AHI HSD processing. This also includes changes to satellite position metadata in the reader as was discussed on slack. I'll try to describe this more below.

This PR:

1. Adds `observation_start/end_time` to the DataArray metadata (see 3 below) using data from the file.
2. Adds `nominal_start/end_time` using previously implemented "rounding" of the time based on the sector of the data (full disk versus regional).
3. Put the above new times in a `time_parameters` sub-dictionary in `.attrs`.
4. Changes `start_time` and `end_time` to be `scheduled` times instead of `observation` times. This is inline with the discussion in #2012.
5. Adds a `round_actual_position` keyword argument to the reader that will round `satellite_actual_longitude/latitude/altitude` in a way to make them consistent across bands of a single time step. Without this satellite position differed between bands and even between segments of a single band in a single time step. This meant that sensor angle generation had to be redone for every input.
6. Removes legacy satellite position values. I will remove them for other readers in a future PR.

Questions/concerns:

1. ~I added checks in the tests for the new time information, but it isn't a really good test as scheduled start/end are equal for the existing test data.~ Tests have been rewritten to use more realistic numbers.
2. We may need to update either the base reader or the `combine_metadata` utility function otherwise these observation/nominal times will be averaged instead of min/max'd. I think @gerritholl updated this code last in `combine_metadata`.
3. Should I include information on other times we discussed in #2012 in the `readers.rst` docs even if I'm not using them?
4. ~I got rid of the legacy satellite position metadata for the AHI HSD reader. I recall @sfinkens doing a big PR a while ago to update all readers to use the newer `orbital_parameters` but the legacy values were kept. Do these need to be kept around? Should I remove them in other reader?~ I will remove them in a future PR.

 - [x] Closes #2012 
 - [x] Tests added 
 - [x] Fully documented


ameraner:
One point confused me a bit: reading the documentation on the `satellite_nominal_*` position, I understood that this position shall be the satellite target position (center of the station keeping box), or "the position we're keeping the satellite close to", or "where the satellite is supposed to be if the orbit was perfect and unperturbed". So I would have expected this position to be hardcode-able, like for AHI `satellite_nominal_latitude=0` and `satellite_nominal_longitude=140.7`.  
But from the code, the description above, and the Slack discussion, I see that you're defining as the nominal position the actual position rounded to the third decimal, which is (potentially) something different to the above.  
I don't know AHI so well, is it perhaps a special situation here? Or maybe shall the documentation be updated to specify that this doesn't need to be a fixed value, but can also be an approximation of the actual position, depending on the behavior of the specific sat (e.g. when big oscillations from the nominal position are possible?)?

Otherwise it looks good to me, thank you for leading the discussions and providing an example for the timing metadata!

djhoese:
Very good questions @ameraner! I will admit I "stole" the use of the "nominal" position for my rounded position use case since it wasn't being used for AHI before this PR. I think @sfinkens originally started the orbital parameters discussion so probably has a lot more use cases in mind for why using "nominal" for what I did is a bad idea. My main goal though is to get a satellite position that is close to the actual position and accurate enough to not produce visual differences in composites, but also be consistent between bands/segments of a single time step/repeat cycle. If the satellite moves and the next satellite position is way different than the previous one, that's fine and expected.

I don't think for what I'm doing we can hardcode these values since the satellite could move from its intended position and we'd want to get those position values from the file data. From what I was told by @simonrp84 on slack, the values you suggested above (140.7, 0) would be the projection coordinates and would not be accurate enough (would show differences in imagery).

I think we either:

1. Update the nominal position documentation to support what I'm going for here with the rounding (depends on input from @sfinkens).
2. Add a new possible orbital parameter for satellite position to hold this rounded position.

I guess an alternative would be to add rounding on-the-fly in the angle generation code based on the actual position and just assume it works for all satellites, but I'm not a big fan of this.

simonrp84:
To be clear, for Himawari the difference between actual and nominal satellite position is likely to be small enough that it doesn't manifest in the imagery. But, I think taking such an approach for Himawari would be a slippery slope as this *does* make a difference for other sats (Met-8 and Met-9 in particular, also some other older ones) and for these sats it wouldn't be possible to use the nominal position.
I don't much like the idea of using nominal for one sat but not others, so rounding the actual satellite position to ensure it's consistent across all bands / segments in one timeslot seems like the most sensible compromise to me.

ameraner:
Looking for `satellite_nominal_longitude` in the codebase, there are already readers fetching this value from similarly named variables in the files (Electrol `NominalLongitude`, ABI+GLM `nominal_satellite_subpoint_lon`, AGRI `NOMCenterLon`, SEVIRI `NominalLongitude`) , so maybe we would need a new variable describing what we want here (a rounded actual position to give consistent values across channels/datasets). Maybe this could still be done on the fly, but with a configurable rounding, to give the interested user the possibility to increase the precision?

To add to the list of sats where nominal != projection, reading from the [ABI L1b  PUG ](https://www.goes-r.gov/users/docs/PUG-L1b-vol3.pdf) page 20

> Note: GOES-East nominal satellite subpoint longitude is -75.2 degrees east. However, the image product data has
been resampled to be centered at -75.0 degrees east longitude.

To add another reference value, Met-11 is currently drifting in longitude at 0.02719 deg/day (see [here](https://www.eumetsat.int/meteosat-orbital-parameters)) from the nominal (0,0) position.

djhoese:
I'm less concerned about nominal equalling projection and more concerned with nominal being close to "actual". For these various readers/sensors that have a nominal position, do we know if it is consistent between bands? If we can make that assumption, then adding to the documentation of nominal something like "these values may be generated by the reader by rounding the 'actual' position to produce consistent position coordinates between bands" and use the nominal position in the angle generation.

ameraner:
As those nominal position values in the files are (semi)fixed and satellite platform-related, they should definitely be consistent across bands (they are for sure for SEVIRI).  
I guess the question is if the nominal position from the files is not "too fixed" I.e. if we are ok with generating the angles using the nominal position defined as the station-keeping-box-center, when this position can differ a degree or more from the actual position. From the discussions, it sounds like we need something in-between the nominal (never changing apart from a few times per satellite lifetime, e.g. to change service) and the actual (that varies continuously, causing measureable changes already across channels and/or image segments).

djhoese:
Yeah, it probably is better to be safe than sorry. Too bad we would have yet another orbital parameter. :thinking: Why can't things just be consistent :cry:

I'll wait for more feedback before changing anything. @ameraner @simonrp84 any ideas on a name for this in-between "actual" and "nominal" position? "approximate"?

ameraner:
I feel you!  
I think "approximate" might be confusing ("why is there an approximate position if we know precisely the actual?"/"Is approximate some sort of estimated position, rather than the actual?").  
Why not a simple and clear "rounded_actual"? It tells you what it is 😄

djhoese:
Just hoping for a single word that has the same/similar meaning :wink:

simonrp84:
Yes, I think `rounded_actual_satellite_position` is the most clear neame. there's a few single word possibilites but I fear they might all lead to further confusion in the future.

FWIW, this is only an issue for AHI isn't it? I mean, all the other satellties report an `actual` position and a `nominal` position per timeslot rather than per channel/segment to my knowledge.

djhoese:
> FWIW, this is only an issue for AHI isn't it? I mean, all the other satellties report an actual position and a nominal position per timeslot rather than per channel/segment to my knowledge.

That's a good question. I really only have experience with ABI and AHI as far as geostationary satellite sensors go. If other segmented geostationary readers show that times are consistent between bands then maybe we special case AHI for its nominal position in the metadata (since it doesn't have one previously)? Then the documentation for the reader can be updated to reference this and the documentation for the generic reader metadata (readers.rst) can be updated to mention this possibility that a reader may include a rounded position for "nominal"...I don't know. I'm definitely leaning towards a design that works for me and may not be the best long term decision.

pnuu:
How about `average_satellite_position_for_imaging_period`?

doc/source/readers.rst
```diff
@@ -141,46 +141,83 @@ time etc. The following attributes are standardized across all readers:
...
+  has stopped being recorded.
+* ``scheduled_start_time``: The "human friendly" time describing the start of
+  the data observation interval or repeat cycle. This time is often on a round
+  minute (seconds=0). Along with the scheduled end time, these times define the
+  regular interval of the data collection. For example, GOES-16 ABI full disk
+  images are collected every 10 minutes (in the common configuration) so
+  ``scheduled_start_time`` and ``scheduled_end_time`` would be 10 minutes apart
+  regardless of when the instrument recorded data inside that interval.
+* ``scheduled_end_time``: Same as ``scheduled_start_time``, but the end of the
+  interval.
```
pnuu, line 179: Like @ameraner said in [here](https://github.com/pytroll/satpy/issues/2012#issuecomment-1033077346), these should be `repeat_cycle_{start,end}_time`, or similar.

djhoese: Are you expressing your opinion or are you telling me I did it wrong? In that discussion Andrea said that "scheduled" would be an acceptable term and based on other discussions (including in that thread) "scheduled" seemed like the more accepted term.

djhoese: Also it was pointed out that the "planned time" isn't actually available in the SEVIRI files so there wasn't a real need to have a planned versus scheduled distinction.

pnuu: That's my opinion, and also what I read @ameraner meant. In the end he was okay with these terms if the explanations didn't make sense, but the differences between "scheduled" and "repeat cycle"/"slot" will still be there. But yeah, just my opinion on the terminology :-)

sfinkens:
To my knowledge, AHI HSD is the only format where the actual position varies across bands/segments of a single image. And I guess in most cases users wouldn't want this. So what if the actual position in the AHI reader is rounded by default and we add a reader kwarg to disable this? Then we could avoid adding another orbital parameter.

@djhoese Regarding the legacy attributes: I think we can remove them now, the deprecation note in the docs has been around for two years now.

satpy/readers/ahi_hsd.py
```diff
@@ -364,25 +367,44 @@ def __del__(self):
...
+    def scheduled_start_time(self):
         """Time this band was scheduled to be recorded."""
+        return self._modify_obs_time_for_schedule(self.observation_start_time)
+
+    @property
+    def scheduled_end_time(self):
+        """Get the scheduled end time."""
+        return self._modify_obs_time_for_schedule(self.observation_end_time)
+
+    def _modify_obs_time_for_schedule(self, obs_time):
```
sfinkens, line 398: Can you please explain in the docstring what the scheduled time looks like? It's hard for me to grasp that from the `obs_time.replace()` call

sfinkens:
And I think the documentation updates are very clear!

satpy/readers/ahi_hsd.py
djhoese: Done.

djhoese:
Ok I think this PR is ready for re-review. I got a little annoyed trying to add a test for the new keyword argument and just rewrote most of the tests to use pytest and created some fake data structures so the tests didn't need so much mocking. There is still a lot of room for improvement, but I think overall it is better.

So the current summary of this PR is:

1. Update documentation regarding new metadata: `scheduled_start/end_time` and `observation_start/end_time`. I never got word from @simonrp84 about what he meant in the related issue about not liking "observation". Otherwise I think everyone being "OK" with "scheduled" is better than some people not liking it or being confused by it.
2. Updated AHI HSD reader to produce new time metadata.
3. Updated AHI HSD reader to have a `round_actual_position` keyword argument that is True by default and rounds actual lon/lat/alt to be consistent between the various bands.
4. Update AHI HSD tests.

Next step: although the times should be more consistent, they aren't accurate since the `combine_metadata` function averages them instead of min/maxes them. Anybody have a problem with me fixing that in `combine_metadata`?

djhoese:
Oh! The problem I'm having with the coordinates in my real world case is that there is also the `nadir` position coordinates which in current `get_satpos` are preferred over the `actual`. I would need my #2030 to get around this.

djhoese:
3 minutes if I comment out nadir lon/lat so the rounded actual is used:

image

satpy/readers/ahi_hsd.py
```diff
@@ -33,6 +33,27 @@
 ``observation_end_time`` keys as well as the ``start_time`` and ``end_time``
 keys.
 
+Satellite Position
+******************
+
+As discussed in the :ref:`orbital_parameters` documentation, a satellite
+position can be described by a specific "actual" position, a "nominal"
+position, a "projection" position, or sometimes a "nadir" position. Not all
+readers are able to produce all of these times. In the case of AHI HSD data
```
sfinkens, line 42: Typo: ... are able to produce all of these _positions_

sfinkens (approved):
Nice work, thanks a lot! Also for improving the tests.

satpy/readers/ahi_hsd.py
djhoese: Done.

simonrp84:
I'm unable to run with AHI data today and actually test what you've done, but the code changes you made LGTM.

doc/source/dev_guide/custom_reader.rst
```diff
@@ -500,6 +500,11 @@ needs to implement a few methods:
 
 On top of that, two attributes need to be defined: ``start_time`` and
 ``end_time``, that define the start and end times of the sensing.
+See the :ref:`time_metadata` section for a description of the different
+times that Satpy readers typically use and what times should be used
+for the ``start_time`` and ``end_time``. If these times are not part of
+the DataArray metadata returned by ``get_dataset`` it will be added from
+these attributes.
```
ameraner, line 507: Maybe it's worth specifying here that `start_time` and `end_time` are not only being added to the dataset `attrs` if missing, but they will also be overwritten if already present. See https://github.com/pytroll/satpy/blob/8aa4280be8ea2b27565fd077a4f67f05c883036b/satpy/readers/yaml_reader.py#L726-L727
CC @strandgren as we found this out lately together.

ameraner (approved):
Looks good to me, thank you again for the discussion and the implementations!
Just a minor optional suggestion below.

doc/source/dev_guide/custom_reader.rst
djhoese: Good point!

djhoese: Ok I've just pushed a clarification for this. Thanks. And if you or @strandgren want to make a PR to have start/end_time set only if they don't already exist I think I'd be OK with that. I'm not sure I can think of a use case where we *need* to overwrite these values. Overall, as you've noticed, it is surprising to the reader developer.

satpy/readers/ahi_hsd.py
```diff
@@ -204,20 +228,37 @@
     ("numof_correction_info_data", "<u2"),
 ])
 
+# Navigation correction sub-info
+_NAVCORR_SUBINFO_TYPE = np.dtype([
```
mraspaud, line 232:
```suggestion
_NAVIGATION_CORRECTION_SUBINFO_TYPE = np.dtype([
```

satpy/readers/ahi_hsd.py
```diff
@@ -364,25 +406,56 @@ def __del__(self):
...
+    def scheduled_start_time(self):
         """Time this band was scheduled to be recorded."""
+        return self._modify_obs_time_for_schedule(self.observation_start_time)
+
+    @property
+    def scheduled_end_time(self):
+        """Get the scheduled end time."""
+        return self._modify_obs_time_for_schedule(self.observation_end_time)
+
+    def _modify_obs_time_for_schedule(self, obs_time):
```
mraspaud, line 437:
```suggestion
    def _modify_observation_time_for_schedule(self, observation_time):
```

satpy/readers/ahi_hsd.py
```diff
@@ -364,25 +406,56 @@ def __del__(self):
...
+        """
         timeline = "{:04d}".format(self.basic_info['observation_timeline'][0])
         if self.observation_area == 'FLDK':
             dt = 0
         else:
-            observation_freq = {'JP': 150, 'R3': 150, 'R4': 30, 'R5': 30}[self.observation_area[:2]]
-            dt = observation_freq * (int(self.observation_area[2:]) - 1)
-        return self.start_time.replace(hour=int(timeline[:2]), minute=int(timeline[2:4]) + dt//60,
-                                       second=dt % 60, microsecond=0)
+            observation_freq_minutes = {'JP': 150, 'R3': 150, 'R4': 30, 'R5': 30}[self.observation_area[:2]]
```
mraspaud, line 454: isn't that seconds?

mraspaud, line 454:
```suggestion
            observation_frequency_minutes = {'JP': 150, 'R3': 150, 'R4': 30, 'R5': 30}[self.observation_area[:2]]
```

satpy/readers/ahi_hsd.py
```diff
@@ -204,20 +228,37 @@
...
+])
+
 # 9 Observation time information block
 _OBS_TIME_INFO_TYPE = np.dtype([
     ("hblock_number", "u1"),
     ("blocklength", "<u2"),
     ("number_of_observation_times", "<u2"),
 ])
 
+_OBS_LINE_TIME_INFO_TYPE = np.dtype([
```
mraspaud, line 245:
```suggestion
_OBSERVATION_LINE_TIME_INFO_TYPE = np.dtype([
```

satpy/readers/ahi_hsd.py
```diff
@@ -572,21 +636,25 @@ def read_band(self, key, info):
                               b=float(self.proj_info['earth_polar_radius'] * 1000))
         actual_alt = float(self.nav_info['distance_earth_center_to_satellite']) * 1000 - re
 
+        if self._round_actual_position:
+            actual_lon = round(actual_lon, 3)
+            actual_lat = round(actual_lat, 2)
```
mraspaud, line 641: The number of digits provided here: are they taken from documentation or have been estimated empirically?

satpy/readers/ahi_hsd.py
```diff
@@ -572,21 +636,25 @@ def read_band(self, key, info):
...
+            standard_name=ds_info['standard_name'],
+            wavelength=ds_info['wavelength'],
             resolution='resolution',
             id=key,
             name=key['name'],
-            scheduled_time=self.scheduled_time,
+            scheduled_start_time=self.scheduled_start_time,
+            scheduled_end_time=self.scheduled_end_time,
+            observation_start_time=self.observation_start_time,
+            observation_end_time=self.observation_end_time,
```
mraspaud, line 655: Similarly to the orbital parameters, could we have a nested dict for times here?

mraspaud (commented):
Thanks for taking the time to reorganise the header types! Some comments inline

doc/source/readers.rst
```diff
@@ -141,46 +141,85 @@ time etc. The following attributes are standardized across all readers:
...
+-------------
+
+In addition to the generic ``start_time`` and ``end_time`` pieces of metadata
+there are other time fields that may be provided if the reader supports them.
+These items include:
+
+* ``observation_start_time``: The point in time when a sensor began recording
+  for the current data.
+* ``observation_end_time``: Same as ``observation_start_time``, but when data
+  has stopped being recorded.
```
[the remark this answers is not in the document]
djhoese: It is mentioned in the `Coordinates` section above.

doc/source/readers.rst
```diff
@@ -141,46 +141,85 @@ time etc. The following attributes are standardized across all readers:
...
+  images are collected every 10 minutes (in the common configuration) so
+  ``scheduled_start_time`` and ``scheduled_end_time`` would be 10 minutes apart
+  regardless of when the instrument recorded data inside that interval.
+* ``scheduled_end_time``: Same as ``scheduled_start_time``, but the end of the
+  interval.
+
+In general, ``start_time`` and ``end_time`` will be set to the "scheduled"
+time by the reader. This ensures that other Satpy components get a
+consistent time for calculations (ex. generation of solar zenith angles)
+and can be reused between bands.
```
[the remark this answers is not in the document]
djhoese: There is no setting to configure this. `start/end_time` are determined by the reader, but it should be considered "best practice" to set them to the scheduled time. If a user or other satpy component needs to choose scheduled versus observation (or some other) time then they can find those values in the metadata.

doc/source/readers.rst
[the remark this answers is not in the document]
mraspaud: Yes, but this is relevant here too.

doc/source/readers.rst
```diff
@@ -141,46 +141,85 @@ time etc. The following attributes are standardized across all readers:
...
+  * ``nadir_longitude/latitude``: Intersection of the instrument's Nadir with the surface of the
+    earth. May differ from the actual satellite position, if the instrument is pointing slightly
+    off the axis (satellite, earth-center). If available, this should be used to compute viewing
+    angles etc. Otherwise, use the actual satellite position.
+  * ``projection_longitude/latitude/altitude``: Projection center of the re-projected data. This
+    should be used to compute lat/lon coordinates. Note that the projection center can differ
+    considerably from the actual satellite position. For example MSG-1 was at times positioned
+    at 3.4 degrees west, while the image data was re-projected to 0 degrees.
+  * [DEPRECATED] ``satellite_longitude/latitude/altitude``: Current position of the satellite at
+    the time of observation in geodetic coordinates.
```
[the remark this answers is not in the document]
djhoese: I think this is dependent on the reader. For example, I see some readers used the `satellite_` with the projection coordinates. Others I think were using the actual position. I think that's why @sfinkens did all the work on `orbital_parameters` (because of the inconsistencies).

satpy/readers/ahi_hsd.py
djhoese: I'll do this, but `obs_time` wasn't obvious enough?

satpy/readers/ahi_hsd.py
djhoese: :man_facepalming:

satpy/readers/ahi_hsd.py
djhoese: Estimated by running with real data and discussions with @simonrp84. Other ideas?

satpy/readers/ahi_hsd.py
djhoese: Yes, that could work. Do you have an idea for the name?

satpy/readers/ahi_hsd.py
mraspaud: I like to have things spelled out for me

satpy/readers/ahi_hsd.py
mraspaud: No, just curious 😁

satpy/readers/ahi_hsd.py
mraspaud: Time_information?

doc/source/readers.rst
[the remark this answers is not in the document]
djhoese: Keep in mind this documentation was here before, I just moved it. Also, I plan on making another PR removing these deprecated keys anyway so there is no need to document what they changed to. The orbital_parameters for these have existed for a long time.

doc/source/readers.rst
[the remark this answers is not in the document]
djhoese: Done.

satpy/readers/ahi_hsd.py
djhoese: This PR or a separate one? I suppose it would have to be done in this one as we are defining the metadata. Shoot.

`time_parameters`?

doc/source/readers.rst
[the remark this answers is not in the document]
mraspaud: Ok

satpy/readers/ahi_hsd.py
mraspaud: Yes, that works

djhoese:
As discussed on slack I through together some updates for this PR:

1. Use `nominal_start/end_time` instead of `scheduled` to avoid confusion with scheduled versus planned time, but also avoid my dislike of "repeat_cycle".
2. Put these new time fields in a sub-dictionary called `time_parameters` to mimic `orbital_parameters`. I'm open to suggestions for a different name for this. Let me know.

CC @ameraner @pnuu @mraspaud @simonrp84 @sfinkens

djhoese:
I just pushed another commit to this which updates the base file handlers support for these new time attributes. It does a min/max on the start/end times in the sub-dictionary just like it already does for start_time/end_time.

Besides the naming of the `time_parameters` sub-dictionary this PR is ready for re-review and merge.

mraspaud (approved):
This LGTM. I'm good with calling the new attribute time_parameters.

djhoese:
So I got github approval from Martin, slack approval from Panu and Simon, and previous approval from Andrea (but not on nominal). I'm going to merge this since I need it for a lot of other work. If @ameraner or anyone else has a problem with something in this PR or the general design of this *stuff* speak now (or as soon as possible) and I will update things as quickly as I can. You can comment here, the related issue, or a new issue.

Merging now...

doc/source/readers.rst
```diff
@@ -141,46 +141,85 @@ time etc. The following attributes are standardized across all readers:
...
+
+Orbital parameters describe the position of the satellite. As such they
+typically come in a few "flavors" for the common types of orbits a satellite
+may have.
+
+For *geostationary* satellites it is described using the following scalar attributes:
+
+  * ``satellite_actual_longitude/latitude/altitude``: Current position of the satellite at the
+    time of observation in geodetic coordinates (i.e. altitude is relative and normal to the
+    surface of the ellipsoid).
```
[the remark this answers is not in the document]
sfinkens: @djhoese As far as I know the ellipsoid is mission-specific, for example Meteosat First Generation was using a different ellipsoid than Meteosat Second Generation.

sfinkens:
Thanks a lot for the great effort! Looks very good.

PR #2032, opened 2022-02-21
Fix various metadata bugs in 'awips_tiled' writer

djhoese:
Closes https://github.com/ssec/polar2grid/issues/417

Kathy (@kathys) has been doing some testing and it looks like there were a couple metadata related issues with the AWIPS tiles being produced by this writer. Most of these are from past refactors and incomplete tests not catching them.

 - [ ] Closes #xxxx 
 - [x] Tests added

### DIFF
--- a/satpy/etc/writers/awips_tiled.yaml
+++ b/satpy/etc/writers/awips_tiled.yaml
@@ -111,6 +111,7 @@ templates:
         attributes:
           physical_element:
             raw_value: ACSPO SST
+          units: {}
 
       # CLAVR-x Products
       default_clavrx:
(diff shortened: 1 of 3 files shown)


djhoese:
Just committed a blank commit to see if codefactor was happy after a bug fix on their side but as I pushed it I realized it was already passing. Oops.

djhoese:
CC @joleenf 

We have confirmation from Lee Cronce at SSEC that the files produced with these changes now load successfully and look as expected. I'll merge this in a little bit.

djhoese:
Given the size of these changes and that I'm the only one of the core developers familiar with this writer, I'm just going to merge it. If it is decided that this should have had a review then let's define that somewhere.

PR #2033, opened 2022-02-24
Remove use of legacy satellite position attributes

djhoese:
As mentioned in #2031 and #2030, some readers are still producing the legacy satellite position parameters. Some include the newer orbital_parameter sub-dictionaries and some don't. This PR removes all the deprecated usages outside of the AHI HSD reader which was updated in #2031.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented


sfinkens (approved):
Nice work, thanks for updating all the readers!

simonrp84:
I've only looked at the `electrol_hrit.py` file but at least in that one I think the parameters are the wrong way around.
On lines 291-293:
```
            'satellite_nominal_longitude': self.mda['orbital_parameters']['satellite_nominal_longitude'],
            'satellite_nominal_latitude': 0.,
            'projection_longitude': self.mda['projection_parameters']['SSP_longitude'],
```
`projection_longitude` is not the same as `SSP_longitude`. The latter is the *actual* position of the satellite while the former is the position onto which the data is reprojected. In general, the `projection_longitude` is the same as the `satellite_nominal_longitude`.

With the changes in nomenclature I've lost track - what should the `SSP_longitude` should be stored as? It's the one that's needed to generate the viewing angles for Rayleigh calculations etc.

djhoese:
@simonrp84 In general the nomenclature shouldn't have changed from changes in my PRs. Actual is still actual, nominal is still nominal. The main nomenclature change/decision was with times. The code you referenced was not changed by me, but I can fix it in this PR if we can figure out what it should be.

You pointed out the SSP_longitude is the "actual" longitude, but based on the code in the `__init__` this is how it is set:

```
        sublon = self.epilogue['GeometricProcessing']['TGeomNormInfo']['SubLon']
        sublon = sublon[self.chid]
        self.mda['projection_parameters']['SSP_longitude'] = np.rad2deg(sublon)
        self.mda['orbital_parameters']['satellite_nominal_longitude'] = np.rad2deg(self.prologue['SatelliteStatus']['NominalLongitude'])
```

Does this change your mind about whether or not these are correct? I mean if SubLon/SSP_longitude is the "actual" location then maybe it should be stored as "actual" and we shouldn't have any "projection" position in the orbital parameters?

mraspaud (approved):
LGTM

PR #2039, opened 2022-02-28
Fix angle caching using a large amount of memory

djhoese:
While generating some profiling plots to show the effects of the recent optimizations in satpy/pyresample and what caching angles and lon/lat arrays did, I discovered that generating the arrays used a lot of memory. This PR implements a simple fix to work around this by using a trick I developed for my profiling scripts. It essentially tells dask to compute chunks and give the resulting numpy array to an object that will do the writing. Instead of writing or storing anything it just throws the result away. Let's look at some results generating GOES-16 ABI `true_color`...

Before and After Caching and other optimizations:

resource_profiling_figure

The plot that isn't in the above is what it looks like when you first fill the cache (12-13GB max memory):

image

After this PR, this looks like:

image

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/modifiers/angles.py
+++ b/satpy/modifiers/angles.py
@@ -178,7 +178,7 @@ class ZarrCacheHelper:
                 new_sub_res = sub_res.to_zarr(zarr_path, compute=False)
             new_res.append(new_sub_res)
         # actually compute the storage to zarr
-        da.compute(new_sub_res)
+        da.compute(new_res)
 
 
 def cache_to_zarr_if(
--- a/satpy/modifiers/angles.py
+++ b/satpy/modifiers/angles.py
@@ -180,7 +180,25 @@ class ZarrCacheHelper:
                                               compute=False)
             new_res.append(new_sub_res)
         # actually compute the storage to zarr
-        da.compute(new_res)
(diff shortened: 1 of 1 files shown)


mraspaud:
I see `to_zarr` has a `chunk_store` argument, could this be useful?

mraspaud:
Ah sorry, I didn't realize that dask arrays were used.

djhoese:
Newest commit profiling results for saving an ABI L1b true_color to geotiff *and* caching to zarr:

image

The change is that I stopped using `return_stored=True`. That was returning a dask array because in the original design my plan was to pass this zarr-based dask array to the rest of the computation. Somewhere along the development path I realized I should cache the data immediately instead I'm not sure how, but somehow it is even faster using this method.

djhoese:
And when using that cache:

image

The CPU/memory is almost identical which is expected since my change should only be during the writes to the cache.

djhoese:
Oh! it is faster because I'm doing it wrong!

djhoese:
Darn. With the fix it takes the same amount of time as before...expected but I got excited.

mraspaud (approved):
LGTM

mraspaud:
Feel free to merge even if the macos builds are hanging

PR #2045, opened 2022-03-07
Add type annotations and cleanup code for geotiff writer

djhoese:
While working on a feature that ended up not being needed, I cleaned up the geotiff writer code.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented


mraspaud (approved):
LGTM, thanks for taking the time to make a PR

PR #2052, opened 2022-03-11
Add resolution dependent chunk sizing to 'modis_l1b' reader

djhoese:
This pull request is my attempt to do something similar to #1980 with using dask's auto chunk sizing in a reader. However, this PR takes a different approach by choosing chunk sizes based on what auto-chunking would be for the 250m version of a dataset. The idea being that a user is probably going to upsample/interpolate the data to a higher resolution at some point, especially for 5km data. I've been doing a lot of other optimizations with my polar2grid tool and python-geotiepoints so I don't have hard numbers to whether or not this feature makes a difference over hardcoding the chunk size for all variables. I'll try taking a look at that in the future if needed.

@mraspaud I wanted to make this to get your feedback about how much you hate, like, or want changes on what I'm doing.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented


djhoese:
The tests are failing because they depend on https://github.com/pytroll/python-geotiepoints/pull/35

satpy/readers/hdfeos_base.py
```diff
@@ -216,14 +217,48 @@ def load_dataset(self, dataset_name, is_category=False):
...
+        num_columns_to_multiplier = {
+            271: 20,  # 5km
+            1354: 4,  # 1km
+            2708: 2,  # 500m
+            5416: 1,  # 250m
+        }
+        for max_columns, res_multiplier in num_columns_to_multiplier.items():
+            if var_shape[-1] <= max_columns:
+                return res_multiplier
+        return 1
```
mraspaud, line 260: We already have a mechanism to find out the resolution of a file here: https://github.com/pytroll/satpy/blob/main/satpy/readers/modis_l1b.py#L61-L71

Could we factorise these two?

satpy/readers/modis_l1b.py
```diff
@@ -122,7 +121,9 @@ def get_dataset(self, key, info):
 
             array = array.where(array >= np.float32(valid_range[0]))
             array = array.where(array <= np.float32(valid_range[1]))
-            array = array.where(from_sds(uncertainty, chunks=CHUNK_SIZE)[index, :, :] < 15)
+            uncertainty = self.sd.select(dataset + "_Uncert_Indexes")
+            uncertainty_chunks = self._chunks_for_variable(uncertainty)
+            array = array.where(from_sds(uncertainty, chunks=uncertainty_chunks)[index, :, :] < 15)
```
mraspaud, line 126: This looks like it could be it's own function

satpy/readers/hdfeos_base.py
```diff
@@ -216,14 +217,48 @@ def load_dataset(self, dataset_name, is_category=False):
...
 
         return data
 
+    def _chunks_for_variable(self, hdf_dataset):
+        scan_length_250m = 40
+        var_shape = hdf_dataset.info()[2]
+        res_multiplier = self._get_res_multiplier(var_shape)
+        non_yx_chunks = tuple()
+        if len(var_shape) == 3:
+            # assume (band, y, x)
```
mraspaud, line 235: Is this documented somewhere? If it is, this comment should be rephrased to remove "assume"

satpy/readers/hdfeos_base.py
```diff
@@ -216,14 +217,48 @@ def load_dataset(self, dataset_name, is_category=False):
...
+            non_yx_chunks = ((1,) * var_shape[0],)
+            var_shape = var_shape[-2:]
+        elif len(var_shape) != 2:
```
mraspaud, line 238: When is len(var_shape) not 2 or 3? For that matter, when is len(var_shape) 2?

mraspaud (commented):
Thanks for this work! I'm a bit worried that CHUNK_SIZE disappears completely, as it still is the standard way to steer the chunk size in satpy.
Other than that, I have a bunch of questions inline

satpy/readers/hdfeos_base.py
djhoese: Depends what you mean by "documented". I suppose there could be some MODIS L2 files out there somewhere that have a 3D variable that doesn't have a "band" dimension. Due to some of the magic of the MODIS readers reading additional variables it wasn't easy to use other logic from things like `get_dataset` where we know the dimensions.

Other than the odd L2 product that I'm not aware of, I think all L1b variables are 2D or 3D...wait:

```
        byte DC\ Restore\ Change\ for\ Thermal\ Bands(number\ of\ scans, number\ of\ emissive\ bands, detectors\ per\ 1km\ band) ;
        byte DC\ Restore\ Change\ for\ Reflective\ 250m\ Bands(number\ of\ scans, number\ of\ 250m\ bands, detectors\ per\ 250m\ band) ;
        byte DC\ Restore\ Change\ for\ Reflective\ 500m\ Bands(number\ of\ scans, number\ of\ 500m\ bands, detectors\ per\ 500m\ band) ;
        byte DC\ Restore\ Change\ for\ Reflective\ 1km\ Bands(number\ of\ scans, number\ of\ 1km\ reflective\ bands, detectors\ per\ 1km\ band) ;
```

These are in a IMAPP L1B file I have. Not sure the right approach here.

satpy/readers/hdfeos_base.py
djhoese: There are a band number variables:

```
        float Band_250M(Band_250M) ;
                Band_250M:long_name = "250M Band Numbers for Subsetting" ;
        float Band_500M(Band_500M) ;
                Band_500M:long_name = "500M Band Numbers for Subsetting" ;
        float Band_1KM_RefSB(Band_1KM_RefSB) ;
                Band_1KM_RefSB:long_name = "1KM Reflective Solar Band Numbers for Subsetting" ;
        float Band_1KM_Emissive(Band_1KM_Emissive) ;
                Band_1KM_Emissive:long_name = "1KM Emissive Band Numbers for Subsetting" ;
```

but we don't currently use them as far as I know. Overall I just didn't like leaving a case that wasn't handled and rather than fail loudly I thought I'd just do auto chunking.

satpy/readers/hdfeos_base.py
djhoese: That code gets the resolution for the overall file, but as you know files can contain multiple resolutions of data (geo at one resolution, band data at another, etc). This is another case where if the MODIS readers were completely rewritten and didn't do magic "load by name" functions then we might know this information before loading the variable data.

satpy/readers/hdfeos_base.py
```diff
@@ -216,14 +217,48 @@ def load_dataset(self, dataset_name, is_category=False):
...
+            # don't guess
+            return dask.array.core.normalize_chunks("auto", shape=var_shape, dtype=np.float32)
+        shape_for_250m = tuple(dim_size * res_multiplier for dim_size in var_shape)
+        chunks_for_250m = dask.array.core.normalize_chunks(("auto", -1), shape=shape_for_250m, dtype=np.float32)
+        row_chunks_for_250m = chunks_for_250m[0][0]
+        scanbased_row_chunks_for_250m = np.round(row_chunks_for_250m / scan_length_250m) * scan_length_250m
+        var_row_chunks = scanbased_row_chunks_for_250m / res_multiplier
+        var_row_chunks = max(var_row_chunks, scan_length_250m / res_multiplier)  # avoid getting 0 chunk size
+        return non_yx_chunks + (var_row_chunks, -1)
```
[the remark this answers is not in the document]
djhoese: I can refactor this after the other discussions are settled.

satpy/readers/hdfeos_base.py
mraspaud: can we extract the dimension names from the hdf file?

djhoese: If pyhdf puts the dimensions on the variable object then I suppose we could, but since this code is used for L1b and L2 I'm not sure there is a good/stable way of doing dimension name -> meaning.

satpy/readers/hdfeos_base.py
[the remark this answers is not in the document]
djhoese: This has been refactored into a first draft of a shared utility function.

djhoese:
I'm not sure I can track down all variations of my profiling of these changes, but the bottom line is it maybe goes from 32s to 34s in execution time (so worse, but depends) but drops 2-4GB in memory usage. It is hard to compare because EWA resampling only depends on dask's `array.chunk-size` while this PR makes MODIS change from needing PYTROLL_CHUNK_SIZE to dask's config. So the chunking is more consistent on the reader side of things, but now changing the reader's chunk size (via dask's config option) also changes the chunk size of the output area. And as mentioned on slack the high memory usage here is not because of the reader at all, it is the EWA resampling doing way too much fancy stuff with chunking. If we could reliably take a SwathDefinition and say "give me the chunk that overlaps this sub-area of the target area" (similar to gradient search) then we'd be able to do things in a much smarter way I think.

satpy/utils.py
```diff
@@ -631,6 +633,112 @@ def _get_pytroll_chunk_size():
...
+
+    """
+    if len(input_shape) not in (2, 3):
+        # we're not sure about this shape so don't guess
+        return dask.array.core.normalize_chunks("auto", shape=input_shape, dtype=input_dtype)
+
+    pre_non_yx_chunks, yx_shape, post_non_yx_chunks = _split_non_yx_chunks(input_shape)
+    high_res_shape = tuple(dim_size * low_res_multiplier for dim_size in yx_shape)
+    col_chunks = -1 if whole_scan_width else "auto"
+    chunks_for_high_res = dask.array.core.normalize_chunks(
```
mraspaud, line 699: I think we should make use of the `previous_chunks` argument in order to keep the new chunks aligned with the underlying chunked storage if possible. Might not apply for modis files, although the scan size might play a role here.

satpy/utils.py
```diff
@@ -631,6 +633,112 @@ def _get_pytroll_chunk_size():
...
+            a time. For example, polar-orbiter scan geolocation being
+            interpolated from low resolution to high resolution.
+
+    Returns:
+        A tuple where each element is the chunk size for that axis/dimension.
+
+    """
+    if len(input_shape) not in (2, 3):
+        # we're not sure about this shape so don't guess
+        return dask.array.core.normalize_chunks("auto", shape=input_shape, dtype=input_dtype)
```
mraspaud, line 694: Codecov says this is not tested?

satpy/utils.py
```diff
@@ -631,6 +633,112 @@ def _get_pytroll_chunk_size():
         return None
 
 
+def chunks_by_resolution(
+        input_shape: tuple[int, ...],
```
mraspaud, line 637: Not sure if this is possible, but would it make sense to work on xarray DataArrays in order to be able to pick just the x and y dimensions?

satpy/tests/test_utils.py
```diff
@@ -44,182 +45,100 @@
 # - caplog
 
 
-class TestUtils(unittest.TestCase):
-    """Testing utils."""
+class TestGeoUtils:
```
mraspaud, line 48: Thanks for the refactor, looks much cleaner now!

satpy/readers/hdfeos_base.py
```diff
@@ -215,14 +214,40 @@ def load_dataset(self, dataset_name, is_category=False):
         from satpy.readers.hdf4_utils import from_sds
 
         dataset = self._read_dataset_in_file(dataset_name)
-        dask_arr = from_sds(dataset, chunks=CHUNK_SIZE)
+        chunks = self._chunks_for_variable(dataset)
+        dask_arr = from_sds(dataset, chunks=chunks)
```
mraspaud, line 218: I forgot again, can we get the storage chunking from pyhdf?

mraspaud (commented):
Thanks a lot for tackling the issue of multi-resolution chunk compatibility! Hopefully that will make satpy run better :)
Looks good code-wise, just a few comments and questions.

satpy/utils.py
djhoese: That's covered by `num_high_res_elements`. It is up to the caller to provide a number there that makes sense for the scanning pattern of the instrument and the on-disk storage.

satpy/utils.py
djhoese: No because we'd have to have a chunked dask array to give to the DataArray to be able to use the DataArray's dimensions (chicken/egg problem). The caller could only pass the y and x dimensions as its `input_shape` and then use the resulting chunks accordingly. I could update MODIS to do this now that I think of it. Right now it just works though too.

satpy/readers/hdfeos_base.py
djhoese: No. Not with pyhdf. It doesn't have any wrappers around the C API for chunk related stuff.

satpy/utils.py
djhoese: Codecov is wrong. It is covered when run locally. I'm not sure why coveralls isn't being reported in the PR. I had to make a small change to fix the docs build so let's see if it recognizes the coverage with this new run.

satpy/utils.py
```diff
@@ -631,6 +633,112 @@ def _get_pytroll_chunk_size():
         return None
 
 
+def chunks_by_resolution(
```
mraspaud, line 636: Ok, I did quite some thinking about this, and I believe we can make this simpler and more flexible.
Basically the idea is to follow the signature of dask's `normalize_chunks` more closely. So that for example the signature of this function could be:
```python
def chunks_by_resolution(chunks, shape, low_res_multipliers, dtype, previous_chunks, high_res_shape):
```
where `chunks` would be eg `(-1, "auto")` for one polar data (-1 meaning full width chunk), `shape` is the shape of the current array, `low_res_multipliers` would be eg `(4, 4)` for 1km data, `previous_chunks` would be optional and related to on disk storage, and `high_res_shape` would be optional (good for the case where hi res data shape is not the low res data shape times the multiplier)
I believe this is all data that the reader has when calling this utility function, and using this signature, we allow having n-dimensional arrays without having to guess anything.
So with this, we can just pass "auto" and -1 from the chunks spec straight to normalize chunks. And I guess that would be the most used cases for the x and y dimensions right?
If we do that, maybe this function should be renamed to eg `normalize_low_res_chunks`?

mraspaud (commented):
I have a suggestion for making the utility function more flexible/generic, otherwise looks good.

satpy/readers/hdfeos_base.py
```diff
@@ -215,14 +214,40 @@ def load_dataset(self, dataset_name, is_category=False):
...
+            np.float32,
+            scan_length_250m,
+            res_multiplier,
+            whole_scan_width=True
+        )
+
+    @staticmethod
+    def _get_res_multiplier(var_shape):
+        num_columns_to_multiplier = {
+            271: 20,  # 5km
```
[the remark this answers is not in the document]
mraspaud: I realise this might not be an issue if the full dataset is in 5km res anyway...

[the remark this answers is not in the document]
djhoese: The 270 case is handled by the logic below it which does a `<=`.

satpy/utils.py
djhoese: I'm OK with this I think. I'll have to see how it looks when I implement it. The main idea of allowing parameters for each dimension is worth it enough. I'm not sure I'm sold on the `high_res_shape` needing to be a parameter as I'd think that rounding any fractional results would produce the same thing.

djhoese: Ok. Major refactor done. Let me know what you think. I didn't add high res shape yet as I'm not sure I need it. Willing to be told otherwise.

djhoese: Oh I should have mentioned, I didn't pass previous chunks to dask because it gave me different results in my unit tests. I'll check on that in a bit and see if that's expected.

djhoese: Ok, added previous chunks now. This actually fixed something I hadn't noticed which was that chunks were being returned that weren't actually aligned to the disk chunking (previous chunks). At least not at the coarser resolutions.

I'm not sure I 100% like this logic, but I think it is good enough:

```python
low_res_chunks.append(int(max(hr_chunks[0] / lr_mult, prev_chunks / lr_mult)))
```

My main concern is that `prev_chunks / lr_mult` could produced a smaller chunk size that the on-disk chunking since `prev_chunks` could be coming from on-disk and it would be for this particular resolution. So if we took a on-disk chunk of 100 and got 100 / 4, then we'd be getting 4 dask chunks per on-disk chunk which has been shown to hurt performance pretty bad (ex. AGRI files). It could just be set to `prev_chunks` as a lower-limit, but then we could end up negating the resolution stuff...actually yeah I'm going to change this.

djhoese:
@mraspaud I'm running into a logical issue with `previous_chunks`. If I pass it to dask's `normalize_chunks` to get the chunking for the highest resolution band, then I'm assuming that `previous_chunks` is referring to the high resolution data storage. For most cases this isn't true, I'm using the on-disk chunks for the current resolution of data.

In the case of ABI at least, on-disk chunks are always the same between resolutions which makes sense since the chunking is for compression and disk bandwidth purposes, not resolution-relative processing. In the case of MODIS though, I don't know the on-disk chunking so I'm providing the minimum number of rows per high resolution band (40 rows). For the 1km bands this means I take the chunk size returned by dask for the 250m band, divide by 4 (1km -> 250m), and get my new chunk size. If I do something similar for ABI and use on-disk chunk sizes I'll end up with misaligned chunks. In one of my tests modeled after ABI I get:

* chunks for 500m => 2034
* chunks for 1km => 2034 / 2 => 1017
* 1017 / 226 on-disk chunk size => 4.5 (not aligned)

Hhhmmm maybe I can pass dask `previous_chunks * low_res_multiplier` and it would behave sort of as I expect...maybe. This does assume that on-disk chunks do not differ widely between resolutions though.

djhoese:
Ugh, no, this helps with MODIS (as far as I can tell), but breaks ABI cases because dask is producing a larger chunk size to accommodate the `previous_chunks * low_res_multiplier` previous chunk size. I'm not sure I can use previous_chunks in the dask call. I think that was one thing you were hoping for.

mraspaud:
I was hoping ondisk chunks would be consistent across resolutions, if that's not the case, we are kind of screwed.

djhoese:
About the CodeScene failures: for one of the failures it is taking the docstring on the new utility function as the number of lines in that function. That's not fair.

The other complaints are valid, but I'm not really sure what to do about them. Too many arguments in the parametrized test function. Yeah, but any other form would make the tests harder to follow/configure.

I'll have to think about the number of arguments on the utility function.

And lastly, it is right about `satpy/utils.py`. It does too much and should be separated into specific modules (ex. `satpy/_chunking.py`).

mraspaud:
Splitting utils sounds like a pr on its own...

mraspaud (approved):
LGTM, thanks for starting this work on multi resolution chunking!

PR #2054, opened 2022-03-11
Fix DifferenceCompositor not using metadata from YAML

djhoese:
Found out the `standard_name:` I was assigning in the YAML wasn't being assigned.

 - [ ] Closes #xxxx 
 - [x] Tests added

### DIFF
--- a/satpy/tests/test_composites.py
+++ b/satpy/tests/test_composites.py
@@ -269,8 +269,8 @@ class TestDifferenceCompositor(unittest.TestCase):
     def test_basic_diff(self):
         """Test that a basic difference composite works."""
         from satpy.composites import DifferenceCompositor
-        comp = DifferenceCompositor(name='diff')
-        res = comp((self.ds1, self.ds2), standard_name='temperature_difference')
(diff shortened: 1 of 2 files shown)


satpy/tests/test_composites.py
```diff
@@ -269,8 +269,8 @@ def setUp(self):
     def test_basic_diff(self):
         """Test that a basic difference composite works."""
         from satpy.composites import DifferenceCompositor
-        comp = DifferenceCompositor(name='diff')
-        res = comp((self.ds1, self.ds2), standard_name='temperature_difference')
+        comp = DifferenceCompositor(name='diff', standard_name='temperature_difference')
+        res = comp((self.ds1, self.ds2))
```
djhoese, line 273: This test was incorrect. Passing `standard_name` to the `__call__` method of a compositor will never happen. Only the Scene calls `__call__` and it passed the DataID properties computed during dependency decisions. The `__init__` method is when metadata from the YAML are passed.

mraspaud (approved):
LGTM

PR #2056, opened 2022-03-11
Update SLSTR calibration coefficients

simonrp84:
The SLSTR calibration coefficients supplied in the reader are old and incorrect, this PR updates the coefficients to the latest ones supplied by EUM/ESA/Copernicus.

 - [x] Tests passed

### DIFF
--- a/satpy/readers/slstr_l1b.py
+++ b/satpy/readers/slstr_l1b.py
@@ -36,23 +36,23 @@ PLATFORM_NAMES = {'S3A': 'Sentinel-3A',
                   'S3B': 'Sentinel-3B'}
 
 # These are the default channel adjustment factors.
-# Defined in the product notice: S3.PN-SLSTR-L1.06
-# https://www4-int.eumetsat.int/media/42788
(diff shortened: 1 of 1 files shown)


djhoese:
Just because I know people...should we make it possible to switch between them? I think you did this for other readers, right?

simonrp84:
No, it should not be possible to switch between the old and new ones: The old ones were incorrect.

Also, the option already exists in the code (via `user_calib` kwarg) for the user to supply their own calibration values.

djhoese:
Ah your description says "outdated" so I assumed they were corrected/updated, but not necessarily invalid before. Were these like pre-launch coefficients or something?

simonrp84:
Heh, true. Have updated the description. Long story regarding why these values are needed: but the old values were from a previous iteration of the calibration efforts. New values are much more robust as they're derived from a) A longer time series and b) More validation sites.

djhoese (approved):
LGTM

mraspaud (approved):
LGTM

PR #2065, opened 2022-03-17
Add filename to YAML for NASA NRT VIIRS files with creation date/time.

simonrp84:
The `viirs_l1b` format currently doesn't support files from the NASA / earthdata NRT data feed as these files contain the creation date in the filename.
This PR adds support for the NRT files by updating the YAML appropriately.


djhoese (approved):
Another filenaming scheme from the same vendor? Ok.

simonrp84:
I have no idea if the existing NRT file format in the YAML is still applicable (the one without the creation date) but all the VIIRS L1b NRT files I've seen recently do contain the date. Will keep an eye on it, maybe we can delete the old one in the future.

PR #2070, opened 2022-03-22
Fix linting errors

sfinkens:
This PR fixes a couple of linting errors recently reported by Flake8 Bugbear (see https://github.com/pytroll/satpy/runs/5641087592?check_suite_focus=true for example)

#### B019 Use of `functools.lru_cache` or `functools.cache` on class methods can lead to memory leaks. The cache may retain instance references, preventing garbage collection.

- [x] satpy/readers/aapp_l1b.py:237:6
- [x] satpy/readers/aapp_l1b.py:289:6
- [x] satpy/readers/mviri_l1b_fiduceo_nc.py:608:6
- [x] satpy/readers/mviri_l1b_fiduceo_nc.py:692:6
- [X] satpy/readers/eps_l1b.py:192:6
- [X] satpy/readers/eps_l1b.py:242:6
- [X] satpy/readers/sar_c_safe.py:137:6
- [X] satpy/readers/sar_c_safe.py:159:6
- [X] satpy/readers/sar_c_safe.py:189:6
- [X] satpy/readers/sar_c_safe.py:619:6
- [X] satpy/readers/nwcsaf_nc.py:248:6

I used the following workaround (from https://www.youtube.com/watch?v=sVjtp6tGo0g):

```python
class FileHandler:
    def __init__(self):
        self.compute = lru_cache(maxsize=123)(self._compute_uncached)
    
    def _compute_uncached(self, x, y):
        # expensive computation
        pass
```
#### B020 Found for loop that reassigns the iterable it is iterating with each iterable value.

- [x] satpy/tests/reader_tests/test_nwcsaf_msg.py:463:32
- [x] satpy/tests/reader_tests/test_viirs_compact.py:2432:32

This was basically the same problem in two locations, so I factorized that method.

### DIFF
--- a/satpy/readers/mviri_l1b_fiduceo_nc.py
+++ b/satpy/readers/mviri_l1b_fiduceo_nc.py
@@ -153,8 +153,8 @@ References
 """
 
 import abc
+import functools
 import warnings
-from functools import lru_cache
 
 import dask.array as da
 import numpy as np
@@ -565,6 +565,9 @@ class FiduceoMviriBase(BaseFileHandler):
         self.projection_longitude = float(filename_info['projection_longitude'])
         self.calib_coefs = self._get_calib_coefs()
 
+        self._get_angles = functools.cache(self._get_angles_uncached)
+        self._get_acq_time = functools.cache(self._get_acq_time_uncached)
+
     def get_dataset(self, dataset_id, dataset_info):
         """Get the dataset."""
         name = dataset_id['name']
@@ -605,8 +608,7 @@ class FiduceoMviriBase(BaseFileHandler):
         ds['acq_time'] = self._get_acq_time(resolution)
         return ds
 
-    @lru_cache(maxsize=8)  # 4 angle datasets with two resolutions each
-    def _get_angles(self, name, resolution):
+    def _get_angles_uncached(self, name, resolution):
         """Get angle dataset.
 
         Files provide angles (solar/satellite zenith & azimuth) at a coarser
@@ -689,8 +691,7 @@ class FiduceoMviriBase(BaseFileHandler):
 
         return coefs
 
-    @lru_cache(maxsize=3)  # Three channels
-    def _get_acq_time(self, resolution):
+    def _get_acq_time_uncached(self, resolution):
         """Get scanline acquisition time for the given resolution.
 
         Note that the acquisition time does not increase monotonically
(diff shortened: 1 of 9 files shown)


satpy/tests/reader_tests/utils.py
```diff
@@ -0,0 +1,45 @@
...
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
+"""Utilities for reader tests."""
+
+
+def default_attr_processor(root, attr):
+    """Do not change the attribute."""
+    return attr
+
+
+def fill_h5(root, contents, attr_processor=default_attr_processor):
```
djhoese, line 26: Was this just cleanup of tests or was this actually required to make things work?

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -565,6 +565,9 @@ def __init__(self, filename, filename_info, filetype_info,
         self.projection_longitude = float(filename_info['projection_longitude'])
         self.calib_coefs = self._get_calib_coefs()
 
+        self._get_angles = functools.cache(self._get_angles_uncached)
+        self._get_acq_time = functools.cache(self._get_acq_time_uncached)
```
djhoese: As you probably know, you have to use `lru_cache` here as `cache` isn't available until Python 3.9.

djhoese (approved):
Once the `functools.cache` is switched to `functools.lru_cache` I'm good with this. Thanks for the quick work on updating all of these usages.

satpy/readers/mviri_l1b_fiduceo_nc.py
sfinkens: Thanks for the hint, that was a testing leftover

satpy/tests/reader_tests/utils.py
sfinkens: No that was just cleanup. The actual problem was re-using the `var` variable in the inner loop, so I renamed it to `attr_val`.

djhoese:
I think given the urgency of this set of fixes we'll have to ignore the codebeat issues introduced. They are all still "info" level but are basically complaining about the `__init__` methods doing too much.

sfinkens:
Yeah that was me adding new init methods with the same signature as the base class.

djhoese:
Let's merge this now and if anyone else has problems we can change it later. Thanks again @sfinkens!

PR #2071, opened 2022-03-24
Add example to the documentation using multiple readers

adybbroe:
Just giving an example of providing a dataset that requires multiple readers.

It was quite clear from the documentation already, but thought it was nice giving an explicit example, as at least I have been in doubt a couple of times how to do it correct.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Fully documented 

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -72,7 +72,7 @@ class Scene:
         """Initialize Scene with Reader and Compositor objects.
 
         Notice (see parameters list below) that it is possible to load a
-        combinations of files or sets of files each requiring their specific
+        combination of files or sets of files each requiring their specific
         reader. For that the `filenames` needs to be a ``dict``, like e.g.::
 
             scn = Scene(filenames={'nwcsaf-pps_nc': glob('/path/to/nwc/saf/pps/files/*'),
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2010-2017, 2022 Satpy developers
+# Copyright (c) 2010-2022 Satpy developers
 #
 # This file is part of satpy.
 #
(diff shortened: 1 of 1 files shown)


adybbroe:
I thought RTD pages should be build for the PR, so I could check it? Wasn't that the case previously? Or am I thinking of the Pyspectral repo?

satpy/scene.py
```diff
@@ -71,6 +71,13 @@ def __init__(self, filenames=None, reader=None, filter_parameters=None,
                  reader_kwargs=None):
         """Initialize Scene with Reader and Compositor objects.
 
+        Notice (see parameters list below) )that it is possible to load a
```
gerritholl, line 74:
```suggestion
        Notice (see parameters list below) that it is possible to load a
```

satpy/scene.py
```diff
@@ -71,6 +71,13 @@ def __init__(self, filenames=None, reader=None, filter_parameters=None,
...
+        combinations of files or sets of files each requiring their specific
```
gerritholl, line 75:
```suggestion
        combination of files or sets of files each requiring their specific
```

satpy/scene.py
```diff
@@ -71,6 +71,13 @@ def __init__(self, filenames=None, reader=None, filter_parameters=None,
...
+        reader. For that the `filenames` needs to be a ``dict``, like e.g.::
```
gerritholl, line 76:
```suggestion
        reader. For that ``filenames`` needs to be a `dict`, like e.g.::
```

gerritholl (changes requested):
Just a few typographic corrections.

adybbroe:
> Just a few typographic corrections.

Thanks @gerritholl ! That was sloppy of me, good you saw it.

adybbroe:
Anything you are okay with @djhoese and @mraspaud and you think we can merge?
Not that it is in a hurry of course - it is just a tiny PR, so can also be forgotten easily...

satpy/scene.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2010-2017 Satpy developers
+# Copyright (c) 2010-2017, 2022 Satpy developers
```
djhoese, line 3: Why not add to the range? `2010-2022`? We could also discuss just removing the years...or the copyright and license header altogether (as I think dask and other projects have started doing).

satpy/scene.py
```diff
@@ -71,6 +71,13 @@ def __init__(self, filenames=None, reader=None, filter_parameters=None,
                  reader_kwargs=None):
         """Initialize Scene with Reader and Compositor objects.
 
+        Notice (see parameters list below) that it is possible to load a
+        combination of files or sets of files each requiring their specific
+        reader. For that ``filenames`` needs to be a `dict`, like e.g.::
+
+            scn = Scene(filenames={'nwcsaf-pps_nc': glob('/path/to/nwc/saf/pps/files/*'),
+                                   'modis_l1b': glob('/path/to/modis/lvl1/files/*')})
+
```
djhoese, line 80: This, being the less likely use case (especially by new users), should come after the below example(s). Preferably in the long run (would appreciate it in this PR, but I would understand if you don't have the time), this docstring would include a small example of creating a Scene with `reader` and `filenames` (a list), an empty Scene (as it does already), and with multiple readers (as you're adding here). I don't care about the order of the last two, but the reader+filenames case should definitely be the first thing users see.

We should also make sure that the generated sphinx API docs contain proper syntax highlighting for these code blocks. If they don't we should use `.. code-block:: python` instead of the generic `::`.

djhoese (changes requested):

satpy/scene.py
adybbroe: Yes, this is just handled automatically in my IDE (emacs :-) )
I let you decide. Now I have added the full range. Think it requires a more over reaching discussion and decision if we should remove the headers alltogether

satpy/scene.py
adybbroe: Good suggestions, better now?

gerritholl (approved):

satpy/scene.py
```diff
@@ -71,14 +71,30 @@ def __init__(self, filenames=None, reader=None, filter_parameters=None,
...
+        created with no Readers available requiring Datasets to be added
+        manually::
 
             scn = Scene()
             scn['my_dataset'] = Dataset(my_data_array, **my_info)
 
+        Further, notice that it is also possible to load a combination of files
+        or sets of files each requiring their specific reader. For that
+        ``filenames`` needs to be a `dict` (see parameters list below), like
+        e.g.::
```
djhoese, line 92: The `like` might be redundant or unnecessary here. It reads a little funny.

djhoese (approved):
LGTM thanks for making the changes.

satpy/scene.py
djhoese: Looks good. Could you try running the website generation locally and make sure these code fragments look good?

satpy/scene.py
adybbroe: Agree. Done.

satpy/scene.py
adybbroe: Sure, I actually did that, and normally does that. But it would also be nice to have RTD built for PRs so that others can see. I think it looks okay. I was unsure to how much we should use the marking of what is a keyword. I kept only the first mentions of `reader` and `filenames` in italic, the following mentions are highlighted as we do for key word arguments.
scene_reader_doc_example

PR #2074, opened 2022-03-30
Added test to confirm string IDs are accepted in ninjogeotiff writer

gerritholl:
On the side of NinJo there exists an interest to move from numerical IDs
to string IDs for the satellite and channel IDs in the ninjogeotiff
headers.  The ninjogeotiff writer has no objection to this.  Confirm
this no-objection with a unit test.

Offline testing shows the ninjotiff writer does _not_ take string values.  There are no plans to change the ninjotiff writer to change this.

 - [x] Tests added

### DIFF
--- a/satpy/tests/writer_tests/test_ninjogeotiff.py
+++ b/satpy/tests/writer_tests/test_ninjogeotiff.py
@@ -904,3 +904,17 @@ def test_create_unknown_tags(test_image_small_arctic_P):
             PhysicValue="N/A",
             SatelliteNameID=6500014,
             Locatie="Hozomeen")
+
+
+def test_str_ids(test_image_small_arctic_P):
+    """Test that channel and satellit IDs can be str."""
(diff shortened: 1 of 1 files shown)


djhoese (approved):

djhoese:
I'm marking this as a "code cleanup" since it doesn't fix or add any functionality. If you have an idea for a more accurate label feel free to change it or tell me to.

It makes sense to me that given the low-level nature of the ninjotiff stuff that it wouldn't support different data types since you have to declare the TIFF tag data type.

gerritholl:
> I'm marking this as a "code cleanup" since it doesn't fix or add any functionality. If you have an idea for a more accurate label feel free to change it or tell me to.

That's fine.

When I started adding the test, I expected it to fail, but was confused with the ninjotiff writer, where it does fail.

mraspaud (approved):
LGTM

PR #2084, opened 2022-04-09
Fix CREFL using incorrect coefficients for MODIS

djhoese:
While working on some Polar2Grid 3.0 validation and comparison, I noticed that the MODIS true_color composites with CREFL rayleigh correction looked different than in past versions. The past versions use the official C version of the CREFL/CVIIRS code. After some debugging it was clear the Satpy CREFL implementation was different. What was used in Satpy was only the VIIRS and ABI coefficients worked on by @ralphk11. The other coefficients needed for MODIS were hidden behind a global `if` statement.

I think this PR fixes everything I was concerned about. It is a little uglier than I was hoping for now that I have to pass `sensor_name` further into the atmosphere code, but the alternative was some duplicated code. I'm satisfied that the MODIS tests are the only ones that needed modification.

CC @wroberts4 who originally tried to merge all of the crefl algorithms together.

Edit: To clarify, the MODIS processing was incorrectly using VIIRS coefficients and code paths rather than the official MODIS ones.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented


djhoese:
Old:

terra_modis_true_color_20220401_153116_wgs84_fit old

New:

terra_modis_true_color_20220401_153116_wgs84_fit new

mraspaud (changes requested):
Good that the coeffs are not mixed up anymore. However, we should have a nicer way to handle the coefficients for the different sensors. Hardcoded index offsets ?!?

simonrp84:
Not directly related to this PR itself, but is there some documentation / a paper / something that describes CREFL that we can add to the docstring? Would be good to have something more insightful than "Ralph did it" if possible ;-)

satpy/modifiers/_crefl_utils.py
```diff
@@ -337,19 +335,21 @@ def _atm_variables_finder(mus, muv, phi, height, tau, tO3, tH2O, taustep4sphalb,
     return sphalb, rhoray, TtotraytH2O, tOG
 
 
-def get_atm_variables(mus, muv, phi, height, ah2o, bh2o, ao3, tau):
+def get_atm_variables(mus, muv, phi, height, sensor_name, ah2o, bh2o, ao3, tau):
```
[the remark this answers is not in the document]
djhoese: Interesting idea. I'll see what I can do.

djhoese:
@simonrp84 The CREFL software is the NASA MODIS CREFL package:
https://directreadout.sci.gsfc.nasa.gov/?id=dspContent&cid=91

And the NASA VIIRS CVIIRS SPA software (link requires login so not copying here).

The stuff by Ralph is nothing official as far as I know. @ralphk11 can comment. I'm not sure if it has been published. I can, should, and will document more about this being from the NASA algorithms.

As far as you question on slack:

> Why does ABI use a different set of atmospheric variables to MODIS/VIIRS?

Ralph said this about the changes as a whole:

> The main changes that I made with Min Oo was that we calculated new coefficients for AHI based upon a radiative transfer model that didn’t assume the atmosphere was infinite plane-parallel. It takes into account the sphericity of the earth, which is important for the geostationary sensors at medium view angles. For VIIRS and MODIS this isn’t necessary and a plane-parallel approximation is fine.

satpy/modifiers/_crefl_utils.py
```diff
@@ -68,49 +69,46 @@ def _csalbr(tau):
...
-    # /*const float taur0[Nbands] = { 0.0507,  0.0164,  0.1915,  0.0948,  0.0036,  0.0012,  0.0004,
-    # 0.3109, 0.2375, 0.1596, 0.1131, 0.0994, 0.0446, 0.0416, 0.0286, 0.0155};*/
-    taur0 = np.array([0.04350, 0.01582, 0.16176, 0.09740, 0.00369, 0.00132,
-                      0.00033, 0.05373, 0.01561, 0.00129, 0.1131, 0.0994,
-                      0.0446, 0.0416, 0.0286, 0.0155,
-                      184.7200e-003, 52.3490e-003, 15.8450e-003, 1.3074e-003, 311.2900e-006])
-    # add last 5 from bH2O to aO2
-    aO2 = 0
+# Values from crefl 1.7.1
+rg_fudge = .55  # This number is what Ralph says "looks good" for ABI/AHI
```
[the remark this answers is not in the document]
djhoese: Yeah...this has all been here for a while though so not really much we can do.

djhoese:
@mraspaud I've refactored the parts of the code I could. I wanted to avoid passing instance methods to `map_blocks` because I don't think they serialize properly on a distributed scheduler. Whether or not my refactor is an overall improvement, I'm not sure. I think the code felt cleaner the more I worked with it, but that could be because I was getting used to it.

@simonrp84 @ralphk11 I've added more information to the docstring including linking to the NASA DRL pages for CREFL and CVIIRS. This whole module has an `_` prefix so it is not included in the Satpy API docs by default so no one will see this on the Satpy documentation site.

ralphk11:
> @simonrp84 The CREFL software is the NASA MODIS CREFL package: https://directreadout.sci.gsfc.nasa.gov/?id=dspContent&cid=91
> 
> And the NASA VIIRS CVIIRS SPA software (link requires login so not copying here).
> 
> The stuff by Ralph is nothing official as far as I know. @ralphk11 can comment. I'm not sure if it has been published. I can, should, and will document more about this being from the NASA algorithms.
> 
> As far as you question on slack:
> 
> > Why does ABI use a different set of atmospheric variables to MODIS/VIIRS?
> 
> Ralph said this about the changes as a whole:
> 
> > The main changes that I made with Min Oo was that we calculated new coefficients for AHI based upon a radiative transfer model that didn’t assume the atmosphere was infinite plane-parallel. It takes into account the sphericity of the earth, which is important for the geostationary sensors at medium view angles. For VIIRS and MODIS this isn’t necessary and a plane-parallel approximation is fine.

For reference the original CREFL code is similar to what is described in appendix A1 (page 74) of the ATBD for the MODIS MOD04/MYD04 data product. Note that for some reason the linked filename says MOD02(??) https://modis.gsfc.nasa.gov/data/atbd/atbd_mod02.pdf.

The AHI/ABI implementation is based on the MODIS collection 6 algorithm, where a spherical-shell atmosphere was assumed rather than a plane-parallel. See appendix A in: "The Collection 6 MODIS aerosol products over land and ocean" Atmos. Meas. Tech., 6, 2989–3034, 2013 www.atmos-meas-tech.net/6/2989/2013/ doi:10.5194/amt-6-2989-2013

djhoese:
Added. Thanks @ralphk11

djhoese:
Unstable environment failures are from this PR in xarray which should be fixed shortly (hopefully): https://github.com/pydata/xarray/pull/6471

satpy/modifiers/_crefl_utils.py
```diff
@@ -49,6 +88,434 @@
...
+    def _run_crefl(self, mus, muv, phi, solar_zenith, sensor_zenith, height, coeffs):
+        LOG.debug("Using VIIRS CREFL algorithm")
+        return da.map_blocks(_run_crefl, self._refl.data, mus.data, muv.data, phi.data,
+                             height, self._refl.attrs.get("sensor"), *coeffs,
+                             meta=np.ndarray((), dtype=self._refl.dtype),
+                             chunks=self._refl.chunks, dtype=self._refl.dtype,
+                             )
+
+
+class _MODISCREFLRunner(_VIIRSCREFLRunner):
```
mraspaud, line 354: There should be a common base class for the viirs and modis variants, at least because the debug message would be wrong in the modis crefl runner.

mraspaud (commented):
Thanks for the refactoring, it looks definitely better. We could surely go further with it, but that could be in another future PR if you don't feel like going further on this right now. I have one comment inline which I think needs addressing, and then there are quite a few places where coverage is allegedly missing...

satpy/modifiers/_crefl_utils.py
djhoese: Good catch. I see a couple options, which one below do you think is cleanest? This assumes that the problem you have is that the log message is wrong not MODIS and VIIRS should share a base class versus MODIS subclassing VIIRS.

1. I make the runner base class have a "name" class attribute that is used in the debug log message to say which runner is running.
2. I remove the log message.
3. Create a separate class for MODIS (based off of the _CREFLRunner base class) that duplicates the `map_blocks` call but with the correct log message.
4. Have a VIIRS/MODIS base class with a `_run_crefl` method that just calls another method to run map_blocks. The VIIRS and MODIS subclasses then override this method adding a single debug message.
5. Move the log message to the `__init__`.

mraspaud: I was thinking about 4.

djhoese: Ended up finding another way using `super()`. Let me know what you think.

satpy/modifiers/_crefl_utils.py
```diff
@@ -49,6 +88,443 @@
...
+            raise ValueError("Unrecognized data resolution: {}", resolution)
+        # Find the best wavelength of coefficients
+        if isinstance(wavelength_range, str):
+            # wavelength range is actually a band name
+            return index_map[wavelength_range]
+        for k, v in index_map.items():
+            if isinstance(k, str):
+                # we are analyzing wavelengths and ignoring dataset names
+                continue
+            if k[0] <= wavelength_range[1] <= k[2]:
```
mraspaud, line 136: We have a class that does this... https://github.com/pytroll/satpy/blob/main/satpy/dataset/dataid.py#L100

satpy/modifiers/_crefl_utils.py
```diff
@@ -49,6 +88,443 @@
...
+            (0.8455, 0.865, 0.8845): 3,  # C03
+            "C03": 3,
+            # (1.3705, 1.378, 1.3855): None,  # C04
+            # "C04": None,
+            (1.580, 1.610, 1.640): 4,  # C05
+            "C05": 4,
+            (2.225, 2.250, 2.275): 5,  # C06
+            "C06": 5
+        },
+    }
```
mraspaud, line 169: I don't see a value of 1, is this deliberate? in which case maybe the second element of the abi coeffs could be deleted?

mraspaud (commented):
Looks good, just a few comments.

satpy/modifiers/_crefl_utils.py
djhoese: This looks like it is a bug. There are only 5 values in the LUT right now but comparing to @ralphk11's code on SSEC gitlab it looks like the value for C02 is missing. @ralphk11 any ideas what's going on. For context, the ABI/AHI `taur0` values are:

```
np.array([184.7200e-003, 52.3490e-003, 15.8450e-003, 1.3074e-003, 311.2900e-006]),
```

We have in the code above commented out C04 from being used, but your code shows the following values for `taur0`:

```
taur0 = np.array([184.7200e-003, 132.1000e-003, 52.3490e-003, 15.8450e-003, 1.3074e-003, 311.2900e-006])
```

The 132.1e-3 is missing from our values.

djhoese: Ok so I talked to @ralphk11 on slack. The code I was looking at was for AHI which has the "green" 0.51 band. ABI has C04/1.37um. The coefficients in the crefl code were updated in Satpy to not include the 0.51 band but the index values were never updated. I've now added tests for the other support channels and have fixed the coefficient indexes.

satpy/modifiers/_crefl_utils.py
djhoese: WavelengthRange is now being used.

djhoese:
Surprisingly the precision differences between my system and CI seems decent (1e-6). I've updated the tests to be a little friendly.

djhoese:
I just pushed a small commit that removed the `computing_meta` keyword argument from dask. This is unnecessary now that I pass `meta=` to the map_blocks call.

mraspaud (approved):
LGTM!

PR #2096, opened 2022-04-27
Convert remote files to FSFile objects automatically

pnuu:
To make it possible to read files directly from a remote location using `fsspec` and `FSFile`, I'm adding a feature that automatically converts file paths that contain the transfer protocol to `FSFile` objects.

With this, all the readers supporting reading for example from S3 object storage the files can be given simply like this:
```python
filenames = [
    's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3__-MSG3_RSS____-WV_073___-000006___-202204260855-__',
    's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3__-MSG3_RSS____-WV_073___-000007___-202204260855-__',
    's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3__-MSG3_RSS____-WV_073___-000008___-202204260855-__',
    's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3__-MSG3_RSS____-_________-EPI______-202204260855-__',
    's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3__-MSG3_RSS____-_________-PRO______-202204260855-__',
]
scn = Scene(reader='seviri_l1b_hrit', filenames=filenames)
scn.load(['WV_073'])
```

The credentials etc. are automatically read from the `fsspec` configuration file, which needs to be described somewhere in the documentation. Alternately the credentials etc. can be given via Scene `reader_kwargs`.

 - [x] Tests added 
 - [x] Fully documented


pnuu:
I'll be working more on this (to fix the failing tests, documentation, etc.), but waiting for possible comments until tomorrow.

djhoese:
This is something I was hoping for when FSFile was first created, or rather before it was created. I wanted something similar for some of the fancier NetCDF URLs that can end with `#mode=bytes` in URLs. Although @mraspaud wanted to keep this functionality separate I think for UX we need to have the convenience of something like this.

djhoese:
I should add...I was a little surprised it was added in the Scene. Is there maybe a spot lower in the code that this could work?

pnuu:
`Scene` was the first place I found handling the filenames, so thought it would be the most logical place to make this work for all supported readers.

pnuu:
Documentation added to `quickstart.html`.

Any thoughts on:

1. what would be a better place for this instead of `Scene`?
2. should there be a way to pass the configuration in the code via `Scene` kwargs?

Already existing alternatives to 2.:
- `fsspec` has an environment variable `FSSPEC_CONFIG_DIR` to point to the relevant config directory, this could be set by the user via `os.environ`
- the `fsspec.config.conf` dictionary can be edited by hand

djhoese:
We should maybe consider putting the documentation somewhere outside of quickstart. Our quickstart is starting to be a "learn everything about satpy"-start.

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,55 @@
...
+        's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3__-MSG3_RSS____-WV_073___-000008___-202204260855-__',
+        's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3__-MSG3_RSS____-_________-EPI______-202204260855-__',
+        's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3__-MSG3_RSS____-_________-PRO______-202204260855-__',
+    ]
+    scn = Scene(reader='seviri_l1b_hrit', filenames=filenames)
+    scn.load(['WV_073'])
+
+As this is a private resource, credentials and server end-point need to be configured::
+
+    {
```
simonrp84, line 49: I think this needs to be a bit clearer - where should the user configure these credentials? In the python file? In a config file elsewhere?

pnuu: It is said above that the configuration should be put in `fsspec` configuration directory, which is by default `~/.config/fsspec/` in Linux.

pnuu: Although now that I read it, I've mixed the texts around so it's not that clear. I'll cleanup it a bit.

simonrp84: Thanks!

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,81 @@
+====================
+Reading remote files
```
mraspaud, line 2: For consistency, should this page contain the FSFile examples too?

pnuu: Yes, I'll add them.

mraspaud:
So, as mentioned live yesterday, I think we need to add a option to provide storage option in the call. So either a `storage_options` kwarg to the scene constructor or use the files/readers dict? Just thinking out loud here...

pnuu:
I started with the `storage_options` route to see what you think. The currently supported `Scene(filenames={'reader1': [...], 'reader2': [...]`)` structure isn't directly usable. I think the `reader_kwargs` could be used for reader specific settings so that the `storage_options` would be `.pop()`'d from it before passing the dictionary to the reader.

pnuu:
I removed the `storage_options` kwarg from `Scene` and added the handling to `reader_kwargs`. This way the options can be given for each reader separately.

mraspaud:
Reviewing this now, I see codefactor and codescene have some comments?

pnuu:
Yep. CodeFactor complains about a function I didn't touch (fixed one similar already), CodeScene didn't like the complexity there was before, so I fixed that and got a new complaint that now there are too many functions in the module (76/75). I'll see if I can do something about those.

pnuu:
At least they are now happy...

satpy/utils.py
```diff
@@ -538,3 +539,86 @@ def get_chunk_pixel_size():
...
+    if filenames:
+        fsspec_files = fsspec.open_files(filenames, **storage_options)
+        return [FSFile(f) for f in fsspec_files]
+    return []
+
+
+def _get_storage_options_from_reader_kwargs(reader_kwargs):
+    if reader_kwargs is None:
+        return None, None
+    storage_options = reader_kwargs.pop('storage_options', None)
```
mraspaud, line 602: `pop` is changing the dict in place, right? so why return `reader_kwargs`?

mraspaud (commented):
Thanks for adding a very nice feature!

some comments.

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,118 @@
...
+    filenames = [
+        's3://satellite-data-eumetcast-seviri-rss/H-000-MSG3*202204260855*',
+    ]
+    storage_options = {
+        "client_kwargs": {"endpoint_url": "https://PLACE-YOUR-SERVER-URL-HERE"},
+        "secret": "VERYBIGSECRET",
+        "key": "ACCESSKEY"
+    }
+    scn = Scene(reader='seviri_l1b_hrit', filenames=filenames, reader_kwargs={'storage_options': storage_options})
+    scn.load(['WV_073'])
```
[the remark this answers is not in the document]
pnuu: Are the doctests run by default on all files? I don't see any skips on https://satpy.readthedocs.io/en/stable/quickstart.html which definitely wouldn't work without having the data on the defined path.

satpy/scene.py
```diff
@@ -106,8 +107,15 @@ def __init__(self, filenames=None, reader=None, filter_parameters=None,
...
                 reader instances.
 
+                Keyword arguments for remote file access are also given in this dictionary.
+                See `documentation <https://satpy.readthedocs.io/en/stable/remote_reading.html>`_
+                for usage examples.
+
         """
         self.attrs = dict()
+
+        storage_options, reader_kwargs = _get_storage_options_from_reader_kwargs(reader_kwargs)
```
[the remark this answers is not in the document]
pnuu: Done.

satpy/scene.py
```diff
@@ -118,6 +126,9 @@ def __init__(self, filenames=None, reader=None, filter_parameters=None,
         if filenames and isinstance(filenames, str):
             raise ValueError("'filenames' must be a list of files: Scene(filenames=[filename])")
 
+        if filenames:
+            filenames = check_file_protocols(filenames, storage_options)
```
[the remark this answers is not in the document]
pnuu: I renamed `check_file_protocols` to `convert_remote_files_to_fsspec`, but kept the storage option collection and `reader_kwargs` cleanup separate. I'll think about this a bit.

satpy/scene.py
```diff
@@ -1501,3 +1512,31 @@ def _get_prereq_datasets(self, comp_id, prereq_nodes, keepables, skip=False):
...
+
+
+def _get_storage_options_from_reader_kwargs(reader_kwargs):
+    if reader_kwargs is None:
+        return None, None
+    storage_options = reader_kwargs.pop('storage_options', None)
+    storage_opt_dict = _get_storage_dictionary_options(reader_kwargs)
+    storage_options = _merge_storage_options(storage_options, storage_opt_dict)
+
+    return storage_options, reader_kwargs
```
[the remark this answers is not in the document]
pnuu: I thought it were clearer that way they are modified, but will remove the return value.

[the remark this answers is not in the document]
pnuu: Actually now that I've renamed the return value in `Scene`, I'll keep the return.

satpy/utils.py
pnuu: It is, but as discussed above it's more clear that the dict has been changed if a new version is returned and collected to a new variable.

doc/source/remote_reading.rst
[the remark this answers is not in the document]
mraspaud: oh, ok, I thought we were doctesting everything. I got confused by the doctest setup at the top of the file. Should we have a clear comment on the top of the files that are not tested? Should this file actually be tested?

pnuu:
1. the `simplecache` is not automatically cleaned, so I'm really hesitant to add that as a default as it can cause unrelated issues on the processing environment by sneakily filling up the disk.
2. I think `reader_kwargs={'simple_cache': {'cache_storage': '/path/to/cache', 'expiry_time': 12345}}` should work
3. That is something for `fsspec` to handle, right?

doc/source/remote_reading.rst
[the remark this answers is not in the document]
pnuu: I didn't know what that meant so just copied it from `quickstart.rst` where the documentation was originally.

I'm not sure these can be tested apart from the GOES case, but I'm also not sure whether having the tests to access AWS/S3 and interact with the data is reasonable...

mraspaud:
1. I agree with @pnuu , but we could make an explicit comment in the documentation to encourage the user to experiment with caching for speedups.
3. We don't know in advance what the user needs, it could be that the user passes the files for all the channels but in the ends just reading a couple. By not preemptively downloading the data, we can save a lot of bandwidth :)

pnuu:
Added a short section on caching the remote files to documentation.

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,172 @@
...
+As an example, reading ABI data from public AWS S3 storage can be done in the following way::
+
+    from satpy import Scene
+
+    storage_options = {'anon': True}
+    filenames = ['s3://noaa-goes16/ABI-L1b-RadC/2019/001/17/*_G16_s20190011702186*']
+    scn = Scene(reader='abi_l1b', filenames=filenames, reader_kwargs={'storage_options': storage_options})
+    scn.load(['true_color_raw'])
+
+In addition to `fsspec` the `s3fs` library needs to be installed.
```
djhoese, line 27:
```suggestion
Reading from S3 as above requires the `s3fs` library to be installed in addition to `fsspec`.
```

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,172 @@
+====================
+Reading remote files
+====================
+
+.. testsetup:: *
+    >>> import sys
+    >>> reload(sys)
+    >>> sys.setdefaultencoding('utf8')
```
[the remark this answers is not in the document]
pnuu: Just copied from `quickstart.rst` where the documentation was earlier. I didn't know what it was for, so just played safe. Removing it.

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,166 @@
...
+    {
+        "s3": {
+            "anon": "true"
+        }
+    }
+
+.. note::
+
+    Using `reader_kwargs` doesn't completely override the configuration file, so in case of problems
+    remove the configuration file to see if that solves the issue.
```
djhoese, line 37: Does this mean the options specified in `reader_kwargs` only "supplement" or add on to the options specified in the config?

doc/source/remote_reading.rst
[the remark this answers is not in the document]
djhoese: I don't think I want regular unit tests running S3 downloads, but doctests may be "OK". I don't think we run doctests as part of CI as most of our examples use fake non-existent paths.

mraspaud (approved):
LGTM

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,166 @@
...
+    }
+    scn = Scene(filenames=filenames, reader_kwargs=reader_kwargs)
+
+
+Caching the remote files
+========================
+
+Caching the remote file locally can speedup the overall processing time significantly, especially if the data are re-used
+for example when testing. The caching can be done by taking advantage of the `fsspec caching mechanism
+<https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally>`_::
```
djhoese, line 94: The fsspec docs say:

>  Only “simplecache” is guaranteed thread/process-safe.

Should we maybe only mention `simplecache` here in that case?

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,166 @@
...
+      - Notes
+    * - No caching
+      - 650 s
+      - remove `reader_kwargs` and `filecache::` from the code
+    * - File cache
+      - 66 s
+      - Initial run
+    * - File cache
+      - 13 s
+      - After the inital run
```
djhoese, line 131:
```suggestion
      - Second run
```

doc/source/remote_reading.rst
```diff
@@ -0,0 +1,166 @@
...
+`fsspec documentation <https://filesystem-spec.readthedocs.io/en/latest/index.html>`_ for more details on connection options
+and detailes.
+
+
+Supported readers
+=================
+
+.. _reader_table:
+
+.. list-table:: Satpy Readers capable of reading remote files using `fsspec`
```
djhoese, line 151: In a future PR I'll really need to include this information in a big table of all the readers.

satpy/scene.py
```diff
@@ -34,6 +34,7 @@
 from satpy.node import CompositorNode, MissingDependencies, ReaderNode
 from satpy.readers import load_readers
 from satpy.resample import get_area_def, prepare_resampler, resample_dataset
+from satpy.utils import _get_storage_options_from_reader_kwargs, convert_remote_files_to_fsspec
```
djhoese, line 37: I think a couple of us talked about this a little while ago. An underscore function should only be used inside its own module. An underscore module can be used inside the package, but its functions should still be public (no underscore) if they are used outside the module.

satpy/utils.py
```diff
@@ -538,3 +539,86 @@ def get_chunk_pixel_size():
...
+    if remote_files:
+        return local_files + fs_files + _filenames_to_fsfile(remote_files, storage_options)
+
+    return local_files + fs_files
+
+
+def _sort_files_to_local_remote_and_fsfiles(filenames):
+    from urllib.parse import urlparse
+
+    from satpy.readers import FSFile
```
djhoese, line 573: Both of these imports are in satpy or standard library so these can be moved to the top of the module right?

djhoese (changes requested):
A couple comments and requests, but otherwise looks really good. Codecov says a lot of the stuff in `satpy/utils.py` is not covered. Are those old messages or do more tests need to be added?

satpy/utils.py
pnuu: Oh, indeed. I was going to argue `fsspec` would then be a hard requirement, but `FSFile` doesn't import it. Which reminds me I need to add `fsspec` as optional dependency to `setup.py`...

doc/source/remote_reading.rst
pnuu: Forgot: @BENR0 started on that front in https://github.com/pytroll/satpy/pull/1547 but the PR has been at draft stage since Feb 2021.

doc/source/remote_reading.rst
pnuu: I tested `filecache` and `blockcache` with the default threaded scheduler using 4 Dask workers and it worked great. I think that's because the cache is filled with full files when the `Scene` is created and that's a single threaded operation. Would the distributed scheduler change that somehow?

doc/source/remote_reading.rst
djhoese: Nah, because then you couldn't run it with the secret keys in the config and the anon access stuff in the kwargs. As-is is probably fine.

djhoese: Or...how does xarray do it?

doc/source/remote_reading.rst
djhoese: I could see that depending on the reader. The fact that it says it explicitly and we're talking about concurrency makes me worried that it may be hard to produce an error.

pnuu: Roger, I'll switch to `simplecache`. The cache expiry on `filecache` isn't really something that is a huge consern on satellite data processing.

doc/source/remote_reading.rst
pnuu: Oh yeah, having the secrets in a config is better than having them in code.

pnuu: Clarified the note a bit.

doc/source/remote_reading.rst
pnuu: Done.

satpy/scene.py
pnuu: Changed `get_storage_options_from_reader_kwargs()` to public function.

satpy/utils.py
pnuu: Done and done.

pnuu: Needed to move `FSFile` import back to the function due to circular import.

pnuu:
> A couple comments and requests, but otherwise looks really good. Codecov says a lot of the stuff in `satpy/utils.py` is not covered. Are those old messages or do more tests need to be added?

Forgot to answer this. All that code is covered, it's just `codecov` being confused for some reason.

mraspaud:
So, I tested this with filenames generated using https://github.com/pytroll/trollmoves/pull/114 using the following snippet:
```python
from posttroll.message import Message
from getpass import getpass
from satpy import Scene

with open("message.txt") as fd:
    msg = Message(rawstr=fd.read())
filenames = [item["uri"] for item in msg.data["dataset"]]

password = getpass()

scn = Scene(
    filenames=filenames,
    reader="olci_l1b",
    reader_kwargs={
        "storage_options": {"ssh": {"password": password}},
        "engine": "h5netcdf",
    },
)
scn.load(["true_color"])
scn.save_dataset("true_color")
```
and it works fine. 

The message looks like this:
```
pytroll://segment/1b/safe-olci/s3b/ dataset myuser@myserver.smhi.se 2022-05-06T07:20:12.753644 v1.01 application/json {"dataset": [{"filesystem": {"cls": "fsspec.implementations.tar.TarFileSystem", "protocol": "tar", "args": [], "fo": "/local_disk/tellicast/received/TER-1/T01-S3B-01/S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3.tar", "target_protocol": "ssh", "target_options": {"host": "myserver.smhi.se", "protocol": "ssh"}}, "uri": "tar://S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3/Oa01_radiance.nc::ssh://myserver.smhi.se/local_disk/tellicast/received/TER-1/T01-S3B-01/S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3.tar", "uid": "tar://S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3/Oa01_radiance.nc"}, {"filesystem": {"cls": "fsspec.implementations.tar.TarFileSystem", "protocol": "tar", "args": [], "fo": "/local_disk/tellicast/received/TER-1/T01-S3B-01/S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3.tar", "target_protocol": "ssh", "target_options": {"host": "myserver.smhi.se", "protocol": "ssh"}}, "uri": "tar://S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3/Oa02_radiance.nc::ssh://myserver.smhi.se/local_disk/tellicast/received/TER-1/T01-S3B-01/S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3.tar", "uid": "tar://S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3/Oa02_radiance.nc"}, {"filesystem": {"cls": "fsspec.implementations.tar.TarFileSystem", "protocol": "tar", "args": [], "fo": "/local_disk/tellicast/received/TER-1/T01-S3B-01/S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3.tar", "target_protocol": "ssh", "target_options": {"host": "myserver.smhi.se", "protocol": "ssh"}}, "uri": "tar://S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3/Oa03_radiance.nc::ssh://myserver.smhi.se/local_disk/tellicast/received/TER-1/T01-S3B-01/S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3.tar", "uid": "tar://S3B_OL_1_EFR____20220506T050140_20220506T050440_20220506T065548_0179_065_304_3780_MAR_O_NR_002.SEN3/Oa03_radiance.nc"}
...
```

For the fun of it, here is the resulting image :)
true_color_20220506_111537

pnuu:
Sweet, thanks for the test!

djhoese (approved):

PR #2098, opened 2022-05-02
Fix Scene.coarsest/finest_area not returning consistent results

djhoese:
In rare cases some readers return multiple AreaDefinitions or SwathDefinitions of the same resolution, but with different coordinates. The case I ran into is the AMSR2 sensor with its A and B views. They have the same resolution (shape), but are slightly shifted from one another. Having Scene.coarsest_area and Scene.finest_area return either one of these is fine, but you run into issues if they are not consistent. In my case I was using the result to freeze a `DynamicAreaDefinition` and then compare my final geotiffs against expected results. Currently the Scene's choice between the A and B geometries is based on the order that the DataArrays is sorted which is essentially random. My tests would sometimes match and sometimes not match depending on what order was ultimately used.

This PR fixes this to include other properties of the geometry objects to decide which one should be returned. It isn't perfect, especially when it comes to SwathDefinitions, but it is a start.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/viirs_sdr.py
+++ b/satpy/readers/viirs_sdr.py
@@ -604,7 +604,7 @@ class VIIRSSDRReader(FileYAMLReader):
             LOG.warning("Required file type '%s' not found or loaded for "
                         "'%s'", ds_info['file_type'], dsid['name'])
         else:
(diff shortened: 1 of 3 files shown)


mraspaud (approved):
LGTM, thanks for reorganising the tests too!

PR #2099, opened 2022-05-03
Fix MultiScene.group in case of differing identifier properties

### LINKED ISSUE #2089: MultiScene.group doesn't work for differing identifier properties
**Describe the bug**
This issue was described on the mailing list. A user created a MultiScene with GOES-16, GOES-17, H8, and Meteosat data in each of the Scenes. They then did `mscn.group(...)`, resampled, blended, and did save_datasets on the resulting blended Scene. This blending didn't actually do anything because the DataIDs of the individual Scenes had different `resolution` parameters so the DataIDs were something like `DataID(name="ir_group", resolution=3000.123)`. Blending can only work on DataArrays where the DataIDs are equal.

I think the proper fix might be to take the DataID of the DataArray being put in the group and remove every property from `.attrs` that appears in the DataArray (ex. `resolution`, `calibration`, and even `_satpy_id`). I don't think it is supported and may be mentioned in another issue, but allowing DataQuery objects for selecting DataArrays for a group might be nice.

**To Reproduce**

```python
from satpy import Scene
from glob import glob
from satpy import MultiScene, DataQuery

debug_on()
h8_scene = Scene(filenames=glob('./*.DAT'), reader='ahi_hsd')
h8_scene.load(['b13_inv'])

g16_scene = Scene(filenames=glob('./*G16*'), reader='abi_l1b')
g16_scene.load(['c13_inv'])

g17_scene = Scene(filenames=glob('./*G17*'), reader='abi_l1b')
g17_scene.load(['c13_inv'])

m08_scene = Scene(filenames=glob('./MSG1*'), reader='seviri_l1b_native')
m08_scene.load(['i18_inv'], upper_right_corner='NE')

m11_scene = Scene(filenames=glob('./MSG4*'), reader='seviri_l1b_native')
m11_scene.load(['i18_inv'], upper_right_corner='NE')

mscn = MultiScene([g17_scene, h8_scene, g16_scene, m08_scene, m11_scene])
groups = {DataQuery(name='ir_group', wavelength=(10, 11, 12)): ['c13_inv', 'b13_inv', 'i18_inv']}
mscn.group(groups)

resampled = mscn.resample('worldeqc3km70', reduce_data=False, radius_of_influence=10000)
blended = resampled.blend()
blended.save_dataset('ir_group')


```

**Expected behavior**
Able to blend a grouped MultiScene.

**Actual results**
No results (empty blended Scene).

**Screenshots**

**Environment Info:**
 - Satpy Version: [e.g. 0.9.0]
 - PyResample Version:

**Additional context**

I think @zxdawn originally adding the grouping functionality?

sfinkens:
Fix #2089 by dropping identifier attributes before adding the group alias.

 - [X] Tests added

### DIFF
--- a/satpy/multiscene.py
+++ b/satpy/multiscene.py
@@ -69,59 +69,72 @@ def timeseries(datasets):
     return res
 
 
-def add_group_aliases_to_scenes(scenes, groups):
-    """Add group aliases to multiple scenes."""
+def group_datasets_in_scenes(scenes, groups):
+    """Group different datasets in multiple scenes by adding aliases.
+
+    Args:
+        scenes (iterable): Scenes to be processed.
+        groups (dict): Groups of datasets that shall be treated equally by
+            MultiScene. Keys specify the groups, values specify the dataset
+            names to be grouped. For example::
+
+            from satpy import DataQuery
+            groups = {DataQuery(name='odd'): ['ds1', 'ds3'],
+                      DataQuery(name='even'): ['ds2', 'ds4']}
+    """
     for scene in scenes:
-        yield _duplicate_datasets_to_be_grouped(scene, groups)
+        grp = GroupAliasGenerator(scene, groups)
+        yield grp.duplicate_datasets_with_group_alias()
 
 
-def _duplicate_datasets_to_be_grouped(scene, groups):
-    scene = scene.copy()
-    for group_id, group_members in groups.items():
-        _duplicate_dataset_to_be_grouped(group_id, group_members, scene)
-    return scene
+class GroupAliasGenerator:
+    """Add group aliases to a scene."""
 
+    def __init__(self, scene, groups):
+        """Initialize the alias generator."""
+        self.scene = scene.copy()
+        self.groups = groups
 
-def _duplicate_dataset_to_be_grouped(group_id, group_members, scene):
-    member_ids = _get_dataset_id_of_group_members_in_scene(group_members, scene)
-    if len(member_ids) == 1:
-        _duplicate_dataset_with_different_id(
-            dataset_id=member_ids[0],
-            alias_id=group_id,
-            scene=scene
-        )
-    elif len(member_ids) > 1:
-        raise ValueError('Cannot add multiple datasets from a scene '
-                         'to the same group')
+    def duplicate_datasets_with_group_alias(self):
+        """Duplicate datasets to be grouped with a group alias."""
+        for group_id, group_members in self.groups.items():
+            self._duplicate_dataset_with_group_alias(group_id, group_members)
+        return self.scene
 
+    def _duplicate_dataset_with_group_alias(self, group_id, group_members):
+        member_ids = self._get_dataset_id_of_group_members_in_scene(group_members)
+        if len(member_ids) == 1:
+            self._duplicate_dataset_with_different_id(
+                dataset_id=member_ids[0],
+                alias_id=group_id,
+            )
+        elif len(member_ids) > 1:
+            raise ValueError('Cannot add multiple datasets from a scene '
+                             'to the same group')
 
-def _get_dataset_id_of_group_members_in_scene(group_members, scene):
-    return [
-        scene[member].attrs['_satpy_id']
-        for member in group_members if member in scene
-    ]
+    def _get_dataset_id_of_group_members_in_scene(self, group_members):
+        return [
+            self.scene[member].attrs['_satpy_id']
+            for member in group_members if member in self.scene
+        ]
 
+    def _duplicate_dataset_with_different_id(self, dataset_id, alias_id):
+        dataset = self.scene[dataset_id].copy()
+        self._prepare_dataset_for_duplication(dataset, alias_id)
+        self.scene[alias_id] = dataset
 
-def _duplicate_dataset_with_different_id(dataset_id, alias_id, scene):
-    dataset = scene[dataset_id].copy()
-    _prepare_dataset_for_being_duplicated(dataset, alias_id)
-    scene[alias_id] = dataset
+    def _prepare_dataset_for_duplication(self, dataset, alias_id):
+        # Drop all identifier attributes from the original dataset. Otherwise
+        # they might invalidate the dataset ID of the alias.
+        self._drop_id_attrs(dataset)
+        dataset.attrs.update(alias_id.to_dict())
 
+    def _drop_id_attrs(self, dataset):
+        for drop_key in self._get_id_attrs(dataset):
+            dataset.attrs.pop(drop_key)
 
-def _prepare_dataset_for_being_duplicated(dataset, alias_id):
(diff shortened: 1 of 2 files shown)


satpy/multiscene.py
```diff
@@ -69,29 +69,72 @@ def timeseries(datasets):
...
+    def _duplicate_dataset_with_group_alias(self, group_id, group_members):
+        member_ids = self._get_dataset_id_of_group_members_in_scene(group_members)
+        if len(member_ids) == 1:
+            self._duplicate_dataset_with_different_id(
+                dataset_id=member_ids[0],
+                alias_id=group_id,
+            )
+        elif len(member_ids) > 1:
+            raise ValueError('Cannot add multiple datasets from a scene '
+                             'to the same group')
```
djhoese, line 113: Is there a case where `member_ids` is 0 or negative? If not, maybe the check for `> 1` could go first in this method (raising the exception) and the `== 1` case would be all that's left (less indentation).

satpy/multiscene.py
```diff
@@ -69,29 +69,72 @@ def timeseries(datasets):
...
+
+    def _get_dataset_id_of_group_members_in_scene(self, group_members):
+        return [
+            self.scene[member].attrs['_satpy_id']
+            for member in group_members if member in self.scene
+        ]
+
+    def _duplicate_dataset_with_different_id(self, dataset_id, alias_id):
+        dataset = self.scene[dataset_id].copy()
```
djhoese, line 122: Just to triple check, this copies `.attrs` but probably doesn't deep copy, right? That's probably OK.

satpy/tests/test_multiscene.py
```diff
@@ -211,80 +213,96 @@ def test_from_files(self):
...
-        scene2['ds4'] = ds4
-
-        multi_scene = MultiScene([scene1, scene2])
-        groups = {make_dataid(name='odd', wavelength=(1, 2, 3)): ['ds1', 'ds3'],
-                  make_dataid(name='even', wavelength=(2, 3, 4)): ['ds2', 'ds4']}
-        multi_scene.group(groups)
+class TestMultiSceneGrouping:
+    """Test dataset grouping in MultiScene."""
+
+    @pytest.fixture
```
djhoese, line 220: Interesting. I never thought about creating the fixtures inside the class. :+1:

satpy/tests/test_multiscene.py
```diff
@@ -211,80 +213,96 @@ def test_from_files(self):
...
-
-        # Adding an alias for multiple datasets in one scene should fail
-        gen = add_group_aliases([scene3], {g1: ['ds3', 'ds31']})
-        self.assertRaises(ValueError, list, gen)
+    def test_fails_to_add_multiple_datasets_from_the_same_scene_to_a_group(self, multi_scene):
+        """Test that multiple datasets from the same scene in one group fails."""
+        groups = {DataQuery(name='mygroup'): ['ds1', 'ds2']}
+        multi_scene.group(groups)
+        with pytest.raises(ValueError):
+            next(multi_scene.scenes)
```
djhoese, line 305: I'm not sure how I feel about this way of testing. I'm curious what @mraspaud and @pnuu think. From what I understand of it you've made a test for each *check* but they use a single test case essentially via the `grouped_multi_scene` fixture. I think my preference would be to have all the checks in a single test method, but I can be persuaded to be OK with this.

djhoese (commented):
Looks really good, nice job. I had a couple comments and one concern about the tests.

satpy/multiscene.py
sfinkens: Yes, that's also my understanding

satpy/tests/test_multiscene.py
djhoese: > I tried to follow the recommendation of making only one assertion per test.

Is that a "Clean Code" recommendation?

sfinkens: I think so. Dave Farley mentioned it in a Continuous Delivery Video.

sfinkens: Something along the lines "A test should fail for only one reason"

djhoese: `<pulls "Clean Code" from bookshelf next to desk>`

The testing examples in Clean Code by Robert C Martin have multiple asserts in a single test.

djhoese: If you can find the Dave Farley video that'd be good. Actually...I have Dave Farley's book downstairs, let me see if he says anything in there. To me failing for one reason could mean something different than "there is only one spot in the test that can fail"

djhoese: Found it!

https://youtu.be/UWtEVKVPBQ0?t=584

He does seem to be saying exactly what you've done here, but I'm not sure I agree. We may want to discuss this with others. To me there is a fine line about what he is saying and *maybe* what he actually means. I get what he's saying about having multiple asserts but I would like to think the main takeaway is that you are only checking *one* behavior not necessarily checking only one assertion. For example,

```
def test_data_copy():
    orig_data = ...
    copied_data = orig_data.copy()
    copied_data["new_key"] = "new_value"
    assert copied_data["new_key"] == "new_value"
    assert "new_key" not in copied_data
```

This test contains two assertions, but it is really only testing one overall behavior. ...but now that I'm writing this I see how the first assertion is testing the "behavior" that setting data worked and the second assertion is testing that the copy worked. But I suppose the setting would be yet another test while the setting here is just one necessity of testing the copy. :man_shrugging:

sfinkens: I agree that the definition of behaviour is a bit vague. I just tried it out to see how it feels. And for me the fact that I had to come up with a very explicit name for those tiny tests made the biggest difference in readability.

sfinkens: So I simplified the tests as we discussed in the meeting, but now I feel an urgent need to explain the intent of these assertions with comments...

satpy/tests/test_multiscene.py
```diff
@@ -211,80 +213,80 @@ def test_from_files(self):
...
+            DataQuery(name='even'): ['ds2', 'ds4']
+        }
+
+    def test_multi_scene_grouping(self, multi_scene, groups, scene1):
+        """Test grouping a MultiScene."""
+        multi_scene.group(groups)
+        shared_ids_exp = {make_dataid(name="odd"), make_dataid(name="even")}
+        assert multi_scene.shared_dataset_ids == shared_ids_exp
+        assert DataQuery(name='odd') not in scene1
+        xr.testing.assert_allclose(multi_scene.scenes[0]["ds1"], scene1["ds1"])
```
djhoese, line 282: Nice.

mraspaud (approved):
LGTM, thanks for cleaning up the tests!

djhoese (approved):

PR #2100, opened 2022-05-04
Fix handling of non-existent reflectance bands in 'viirs_l1b' reader

djhoese:
The VIIRS L1b format was changed a couple years ago to save space by only including reflectance band when there is day data in the scene. The variables in these cases do not exist in the files. This resulted in two problems:

1. Trying to load a reflectance band would produce an unexpected error related to trying to treat an integer shape (default return value to handle scalar variables) as a tuple.
2. Listing the reflectance bands as available in `Scene.available_dataset_ids/names`.

This PR fixes both those cases.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM, just a small improvement suggestion inline.

PR #2102, opened 2022-05-05
Add more products to fci_l2_nc reader

strandgren:
This PR adds some more products to the `fci_l2_nc` reader. It also makes the `file_patterns` in the yaml-file less stringent in order to be compatible with L2 files from different processing environments.

The PR is mostly about adding new product entries entries in the `yaml`-file, but also include some minor changes in the `py`-file in order to allow both the CRM and OCA products to use the `_slice_dataset` method, initially developed for the multi-dimensional segmented ASR product.

 - [x] Make `file_patterns` less stringent.
 - [x] Add FIR product
 - [x] Add OLR product
 - [x] Add CRM product


strandgren:
Coverage tests are most likely failing because some lines of tested code have been removed.

mraspaud (approved):
LGTM!

PR #2107, opened 2022-05-09
Fix 'ahi_hsd' reader crashing when 'observation_timeline' was invalid

simonrp84:
If an AHI data file contains fill value for the `'observation_timeline'` property then the `ahi_hsd` reader crashes when it tried to round the observation times across segments.
This PR checks whether the timeline is valid and only performs rounding if it is. If invalid, no rounding is done and a warning is shown to the user.

### DIFF
--- a/satpy/readers/ahi_hsd.py
+++ b/satpy/readers/ahi_hsd.py
@@ -436,6 +436,13 @@ class AHIHSDFileHandler(BaseFileHandler):
         """Get the nominal end time."""
         return self._modify_observation_time_for_nominal(self.observation_end_time)
 
+    @staticmethod
+    def _is_valid_timeline(timeline):
+        """Check that the `observation_timeline` value is not a fill value."""
+        if int(timeline[:2]) > 23:
+            return False
+        return True
+
     def _modify_observation_time_for_nominal(self, observation_time):
         """Round observation time to a nominal time based on known observation frequency.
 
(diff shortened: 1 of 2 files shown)


satpy/readers/ahi_hsd.py
```diff
@@ -455,9 +455,12 @@ def _modify_observation_time_for_nominal(self, observation_time):
         else:
             observation_frequency_seconds = {'JP': 150, 'R3': 150, 'R4': 30, 'R5': 30}[self.observation_area[:2]]
             dt = observation_frequency_seconds * (int(self.observation_area[2:]) - 1)
-        return observation_time.replace(
-            hour=int(timeline[:2]), minute=int(timeline[2:4]) + dt//60,
-            second=dt % 60, microsecond=0)
+        if int(timeline[:2]) < 24:
```
sfinkens, line 458: Could you please move the comparison (hour < 24) to a self-describing method, so that the if statement becomes (for example) `if _is_valid_timeline(timeline)`?

sfinkens (commented):
Thanks for the fix! Just one comment regarding readability. Would you mind adding a test for this?

satpy/readers/ahi_hsd.py
simonrp84: Ok, test added and I've also added the function you suggest.

satpy/tests/reader_tests/test_ahi_hsd.py
```diff
@@ -359,6 +359,22 @@ def test_blocklen_error(self, *mocks):
                 fh._check_fpos(fp_, fpos, 0, 'header 1')
                 assert len(w) > 0
 
+    def test_time_rounding(self):
+        """Test rounding of the nominal time."""
+
+        assert AHIHSDFileHandler._is_valid_timeline(FAKE_BASIC_INFO['observation_timeline']) is True
+        assert AHIHSDFileHandler._is_valid_timeline('65526') is False
```
sfinkens, line 365: These two assertions are independent of the rest and could be moved into a separate test method.

sfinkens (commented):
Thanks a lot for adding the tests! Just one tiny comment :)

satpy/tests/reader_tests/test_ahi_hsd.py
```diff
@@ -37,11 +37,11 @@
...
-    'satellite': np.array(['Himawari-8']),
-    'observation_area': np.array(['FLDK']),
-    'observation_start_time': np.array([58413.12523839]),
-    'observation_end_time': np.array([58413.12562439]),
-    'observation_timeline': np.array([300]),
+    'satellite': 'Himawari-8',
+    'observation_area': 'FLDK',
+    'observation_start_time': 58413.12523839,
+    'observation_end_time': 58413.12562439,
+    'observation_timeline': '0300',
```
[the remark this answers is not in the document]
simonrp84: Yes, it does: While debugging I had some `print` statements inside the `ahi_hsd.py` file and they showed that there's no numpy arrays for these values, they're all simple strings or floats.

satpy/tests/reader_tests/test_ahi_hsd.py
simonrp84: Done!

simonrp84:
Tests are failing but I *think* the failures are not to do with the changes here.

djhoese:
The MacOS job seems to be a hiccup, the other one is the unstable environment which is failing because of changes in dask that are incompatible with xarray. We can track the dask/xarray issue at https://github.com/pydata/xarray/issues/6578

I've restarted the MacOS job.

simonrp84:
Thanks, seems to have worked now

djhoese (approved):

PR #2114, opened 2022-05-27
Fix ACSPO 'sensor' attribute not being lowercase

djhoese:
Sensor names were coming from file attributes, but weren't lowercase which is semi-standard in Satpy. This was causing issues in Polar2Grid where the sensor wasn't matching custom config entries.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,7 @@ repos:
           - types-pkg-resources
           - types-PyYAML
           - types-requests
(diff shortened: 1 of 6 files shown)


djhoese:
@pytroll/satpy-core FYI this PR includes the final changes necessary to drop Python 3.7 properly. Pretty simple, but wanted to tag you here just to point it out.

PR #2119, opened 2022-06-02
Add initial support for Fengyun-4B AGRI data.

simonrp84:
This PR adds support for the AGRI instrument aboard Fengyun-4B.
The format has changed from the original AGRI format used for -4A, and the reader should now be applicable for both satellite. Tests have not yet been added!
 - [ ] Tests added

### DIFF
--- a/satpy/readers/agri_l1.py
+++ b/satpy/readers/agri_l1.py
@@ -42,6 +42,11 @@ _CFAC_list = [81865099.0, 40932549.0, 20466274.0, 10233137.0]
 _LOFF_list = [10991.5, 5495.5, 2747.5, 1373.5]
 _LFAC_list = [81865099.0, 40932549.0, 20466274.0, 10233137.0]
 
+CHANS_ID = 'NOMChannel'
+SAT_ID = 'NOMSatellite'
+SUN_ID = 'NOMSun'
+
+
 PLATFORM_NAMES = {'FY4A': 'FY-4A',
                   'FY4B': 'FY-4B',
                   'FY4C': 'FY-4C'}
@@ -101,7 +106,16 @@ class HDF_AGRI_L1(HDF5FileHandler):
         ds_name = dataset_id['name']
         logger.debug('Reading in get_dataset %s.', ds_name)
         file_key = ds_info.get('file_key', ds_name)
+        if self.filename_info['platform_id'] == 'FY4B':
+            if CHANS_ID in file_key:
+                file_key = f'Data/{file_key}'
+            elif SUN_ID in file_key or SAT_ID in file_key:
+                file_key = f'Navigation/{file_key}'
+
         data = self.get(file_key)
+
+        print(file_key)
+
         if data.ndim >= 2:
             data = data.rename({data.dims[-2]: 'y', data.dims[-1]: 'x'})
 
@@ -154,7 +168,10 @@ class HDF_AGRI_L1(HDF5FileHandler):
         """Calibrate to reflectance [%]."""
         logger.debug("Calibrating to reflectances")
         # using the corresponding SCALE and OFFSET
-        cal_coef = 'CALIBRATION_COEF(SCALE+OFFSET)'
+        if self.filename_info['platform_id'] == 'FY4B':
+            cal_coef = 'Calibration/CALIBRATION_COEF(SCALE+OFFSET)'
+        else:
+            cal_coef = 'CALIBRATION_COEF(SCALE+OFFSET)'
         num_channel = self.get(cal_coef).shape[0]
         if num_channel == 1:
             # only channel_2, resolution = 500 m


satpy/readers/agri_l1.py
```diff
@@ -101,7 +106,16 @@ def get_dataset(self, dataset_id, ds_info):
...
         file_key = ds_info.get('file_key', ds_name)
+        if self.filename_info['platform_id'] == 'FY4B':
+            if CHANS_ID in file_key:
+                file_key = f'Data/{file_key}'
+            elif SUN_ID in file_key or SAT_ID in file_key:
+                file_key = f'Navigation/{file_key}'
+
         data = self.get(file_key)
+
+        print(file_key)
```
zxdawn, line 117: This should be deleted.

satpy/readers/agri_l1.py
```diff
@@ -154,7 +168,10 @@ def calibrate_to_reflectance(self, data, channel_index, ds_info):
         """Calibrate to reflectance [%]."""
         logger.debug("Calibrating to reflectances")
         # using the corresponding SCALE and OFFSET
-        cal_coef = 'CALIBRATION_COEF(SCALE+OFFSET)'
+        if self.filename_info['platform_id'] == 'FY4B':
+            cal_coef = 'Calibration/CALIBRATION_COEF(SCALE+OFFSET)'
```
zxdawn, line 172: Maybe, it's better to define `CAL_ID` at the beginning like the `CHANS_ID`?

zxdawn (changes requested):

satpy/readers/agri_l1.py
```diff
@@ -101,7 +106,16 @@ def get_dataset(self, dataset_id, ds_info):
...
+        if self.filename_info['platform_id'] == 'FY4B':
+            if CHANS_ID in file_key:
+                file_key = f'Data/{file_key}'
+            elif SUN_ID in file_key or SAT_ID in file_key:
+                file_key = f'Navigation/{file_key}'
+
         data = self.get(file_key)
+
+        print(file_key)
+
```
mraspaud, line 118:
```suggestion
```

satpy/readers/agri_l1.py
simonrp84: In this case I didn't do so, as there's no confusion about what we want to load (it's always in `Calibration/` while for the other datasets it coul dbe `Data/` or `Navigation/`).

satpy/readers/agri_l1.py
```diff
@@ -42,6 +42,11 @@
 _LOFF_list = [10991.5, 5495.5, 2747.5, 1373.5]
 _LFAC_list = [81865099.0, 40932549.0, 20466274.0, 10233137.0]
 
+CHANS_ID = 'NOMChannel'
+SAT_ID = 'NOMSatellite'
+SUN_ID = 'NOMSun'
```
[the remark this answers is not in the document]
zxdawn: Actually, I never tried to load these variables ;) Agree with you, that should be another PR.

MengXin001:
In addition, FengYun4B AGRI has a new 6.95 µm channel, and the original Band10(C10) central wavelength has been updated to 7.92 µm

MengXin001:
In addition, FengYun4B AGRI has a new 6.95 µm channel, and the original Band10(C10) central wavelength has been updated to 7.92 µm

MengXin001:
In addition, FengYun4B AGRI has a new 6.95 µm channel, and the original Band10(C10) central wavelength has been updated to 7.92 µm

mraspaud:
@MengXin001 Thanks a lot for the heads up!

simonrp84:
Yes, I was exchanging emails with a CMA colleague about this today. Will update the YAML later once I've added the tests.

simonrp84:
OK, unfortunately I can't figure out how the AGRI tests work as they're based on the HDF4 code that's always remained opaque to me. Hopefully someone else can point to how the tests should be updated!

I have, however, added the new pseudo red channel used for the `true_color` composite. Here's how it looks, I'll add tests for this new composite next week.
New:
new
Original:
orig


Could still be better, we need the FY-4B spectral response functions to do a proper Rayleigh correction.

satpy/readers/agri_l1.py
```diff
@@ -154,7 +166,10 @@ def calibrate_to_reflectance(self, data, channel_index, ds_info):
         """Calibrate to reflectance [%]."""
         logger.debug("Calibrating to reflectances")
         # using the corresponding SCALE and OFFSET
-        cal_coef = 'CALIBRATION_COEF(SCALE+OFFSET)'
+        if self.filename_info['platform_id'] == 'FY4B':
+            cal_coef = 'Calibration/CALIBRATION_COEF(SCALE+OFFSET)'
+        else:
+            cal_coef = 'CALIBRATION_COEF(SCALE+OFFSET)'
         num_channel = self.get(cal_coef).shape[0]
```
djhoese, line 173: As discussed on slack, there are some optimizations that could be made to this reader. The main one I noticed was this `.get` of the coefficients variable in this method. It does 3 `.get` calls which is creating a DataArray 3 times and on two of them it does `.values` which means they are computed twice. Additionally, the coefficients are used for every channel that is loaded (even if it is just a sub-index). I refactored this into a cached_property locally that I think you could apply here. Here is my diff:

```diff
diff --git a/satpy/readers/agri_l1.py b/satpy/readers/agri_l1.py
--- a/satpy/readers/agri_l1.py	(revision 6a0e6da340023971f4a54c480bfe037b702c9219)
+++ b/satpy/readers/agri_l1.py	(date 1656293104023)
@@ -32,6 +32,7 @@
 
 from satpy.readers._geos_area import get_area_definition, get_area_extent
 from satpy.readers.hdf5_utils import HDF5FileHandler
+from satpy._compat import cached_property
 
 logger = logging.getLogger(__name__)
 
@@ -96,6 +97,12 @@
         """Init filehandler."""
         super(HDF_AGRI_L1, self).__init__(filename, filename_info, filetype_info)
 
+    @cached_property
+    def reflectance_coeffs(self):
+        # using the corresponding SCALE and OFFSET
+        cal_coef = 'CALIBRATION_COEF(SCALE+OFFSET)'
+        return self.get(cal_coef).values
+
     def get_dataset(self, dataset_id, ds_info):
         """Load a dataset."""
         ds_name = dataset_id['name']
@@ -153,14 +160,13 @@
     def calibrate_to_reflectance(self, data, channel_index, ds_info):
         """Calibrate to reflectance [%]."""
         logger.debug("Calibrating to reflectances")
-        # using the corresponding SCALE and OFFSET
-        cal_coef = 'CALIBRATION_COEF(SCALE+OFFSET)'
-        num_channel = self.get(cal_coef).shape[0]
+        coeffs = self.reflectance_coeffs
+        num_channel = coeffs.shape[0]
         if num_channel == 1:
             # only channel_2, resolution = 500 m
             channel_index = 0
-        data.attrs['scale_factor'] = self.get(cal_coef)[channel_index, 0].values.item()
-        data.attrs['add_offset'] = self.get(cal_coef)[channel_index, 1].values.item()
+        data.attrs['scale_factor'] = coeffs[channel_index, 0].item()
+        data.attrs['add_offset'] = coeffs[channel_index, 1].item()
         data = scale(data, data.attrs['scale_factor'], data.attrs['add_offset'])
         data *= 100
         ds_info['valid_range'] = (data.attrs['valid_range'] * data.attrs['scale_factor'] + data.attrs['add_offset'])
```

PR #2123, opened 2022-06-10
Fix negative channel 3A slope calibration coefficient

ninahakansson:
Handle negative slopes for 3A in calibration the same way as for channel 1

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/aapp_l1b.py
+++ b/satpy/readers/aapp_l1b.py
@@ -593,15 +593,14 @@ def _vis_calibrate(data,
         intercept2 = da.from_array(data["calvis"][:, chn, coeff_idx, 3],
                                    chunks=line_chunks) * 1e-7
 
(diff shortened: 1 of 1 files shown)


djhoese:
So the original fix was for Channel 2 (at index 1) and this is for 3a (at index 2). Given that it took so long to discover the issue with Channel 2 and that it took an email from Nigel Atkinson, do you @ninahakansson have any idea if any other channels (channel 1?) suffer from this?

adybbroe:
> I think is could in principle also happen to channel 1. So maybe we should just apply it to all visible channels?

Yes, please do that, also just agreed via mail by Nigel Atkinson (NWPSAF)

gerritholl:
Is this ready to merge?  It seems the codecov failure is due to the total number of lines of code being reduced?

PR #2131, opened 2022-06-22
Refactor the ratio-sharpened RGB compositor to use dask map_blocks

djhoese:
While working on some performance benchmarking I got annoyed at the number of dask tasks produced for ratio-sharpening of an RGB. This PR refactors the code to run in a dask `map_blocks` function so it is done in a single task. This has limited performance improvements given the limited tasks involved...but it is obvious now that this sharpening takes a long time. That's a separate PR though :wink: 

 - [ ] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #2133, opened 2022-06-25
Rewrite 'apply_enhancement' as individual decorators to allow for easier dask map_blocks usage

djhoese:
Edit: This PR was originally just a new keyword argument to `apply_enhancement` but @mraspaud has rewritten everything as decorators. This makes it much more clear what each enhancement function is doing without the indirection of passing callables/functions between various pieces of code with a lot of keyword arguments. The original description of this PR is below but does not necessarily still apply.

### Original Description:

I've wanted to do this for a while. Some enhancements require multiple numpy/dask function calls which each produce their own dask task. This is unnecessary in most cases and makes debugging more difficult since you don't know if all the tasks are related. It also results in a small overhead from dask itself due to processing each task.

This PR adds a `use_map_blocks` keyword argument to the `apply_enhancement` function used by a lot of builtin satpy enhancements so that they can use dask's `map_blocks` to make one dask task instead of many little tasks. I didn't add tests when doing this as it started out as just a refactor (require a switch to using map_blocks) but then I realized someone may need to not use map_blocks so I added the kwarg.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/enhancements/__init__.py
+++ b/satpy/enhancements/__init__.py
@@ -19,7 +19,8 @@
 import logging
 import os
 import warnings
-from functools import partial
+from collections import namedtuple
+from functools import wraps
 from numbers import Number
 
 import dask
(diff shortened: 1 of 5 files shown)


gerritholl:
What is the status of this PR?  What would still need to be done?

djhoese:
@gerritholl this mostly needs a review from the various people that might care. There is enough refactoring that a "clean code" review would be good, but otherwise this shouldn't change the end result of things. This does change the default processing path for enhancements using this helper method and passing dask arrays. If their functions don't expect per-chunk arrays then they might need to change things...but for the tests we have for the builtin enhancements it seems we're good.

mraspaud (changes requested):
I'm fine with functionality and logic. However, I think the number of arguments to apply_enhancement is starting to bug me. From what I see, the past last keywords (separate, pass_dask, use_mapblock) are actually predefined per enhancement, so my bet is that the code would be much cleaner if we turned them into decorators to use on each individual enhancement function. This will make each function accept a full dask array once decorated.
I'm going to try implementing is and making a PR to your PR, see if you like it.

mraspaud (approved):
LGTM

PR #2140, opened 2022-07-08
Add Natural Color / IR composites

simonrp84:
This PR adds new composites to blend Natural Color daytime imagery with night time IR imagery, similar to the existing 'night lights' composites but without the background image.


mraspaud (commented):
Thanks for adding these. The night_ir composite added here is very similar to the existing cloudtop one, only the linear stretching cutoffs change. I was wondering if we could factorise these two?

mraspaud (approved):
LGTM

PR #1979, opened 2022-01-14
Update VII reader for test data v2

pepephillips:
- [ ] Closes #1842  and should also address #1462 and #1460 in that the issues are solved for the new test data V2 (V1 is now obsolete).
 - [ ]  
 - [ ] Tests added

PR #2146, opened 2022-07-19
Update Metimage reader for L2 test data

pepephillips:
- [ ] This is essentially the L2 part to PR #1979 concerning the new METimage product format V4A (lat,lon instead of lon,lat)
 - [ ] vii_l2_nc.yaml has been updated so L2 product readers accept the new lat/lon format and have the correct file pattern.  
 - [ ] test_vii_l2_nc.py has also been updated accordingly.


mraspaud (approved):
LGTM, thanks for fixing the tests too!

PR #2151, opened 2022-07-28
Change default type of AWIPS tiled writer to int16 + _Unsigned

djhoese:
Apparently AWIPS can't handle signed 16-bit integers.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM, thanks for cleaning up old comments.

PR #2160, opened 2022-08-02
Add reader for ESA's Ocean Color CCI data

simonrp84:
This PR adds support for the data products by ESA's [Ocean Color CCI](https://climate.esa.int/en/projects/ocean-colour/about/). Currently it supports the gridded lat/lon data but not the sinusoidal grid. All products and composite periods are supported.

 - [x] Tests added
 - [x] Fully documented


doc/source/readers.rst
```diff
@@ -368,3 +368,13 @@ Arctica-M N1 HDF5 format reader
...
     :noindex:
+
+
+netcdf based readers
+--------------------
+
+ESA Ocean Color CCI L3S reader
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. automodule:: satpy.readers.oceancolorcci_l3_nc
```
djhoese, line 379: I think this list of readers will be going away shortly and if necessary will be automatically generated. Please remove this new section. At least, that's my opinion.

simonrp84:
pre-commit.ci autofix

simonrp84:
This should be ready to go now.

satpy/etc/readers/oceancolorcci_l3_nc.yaml
```diff
@@ -0,0 +1,826 @@
+reader:
+  name: oceancolorcci_l3_nc
+  short_name: OCCCI Level 3
+  long_name: Ocean color CCI Level 3S data reader
+  description: NetCDF Reader for ESA Oceancolor CCI data
+  status: Nominal
+  supports_fsspec: false
+  default_channels: []
+  sensors: [merged]
```
djhoese, line 9: What are all the possible sensors that this could include?

simonrp84: This product uses data from multiple sensors and the exact sensors used depends on the temporal coverage of the file being loaded (f.ex: A file covering Jan 2002 won't include VIIRS). However, the actual data is always a merged dataset of all sensors being used. It's not a single-sensor product.

djhoese: Yes, I understand that, but we don't typically put a "catch all" sensor name in Satpy. So is there a finite list of sensors that could go here? In the file handler you can have a `sensor_names` property that reads information from the file and provide a set of sensors that are included in that specific file. This will/would make composite loading more accurate to the data being loaded. If this doesn't seem reasonable then I guess "merged" if fine for now, but we should maybe come up with a standard for this.

simonrp84: Sorry, I don't follow this bit: 
> This will/would make composite loading more accurate to the data being loaded.

There aren't any composites for this data and knowing which sensors are involved doesn't really achieve anything - as the product is always a fusion of multiple sensors.

djhoese: Yeah it is kind of a weird case for me to make this point I guess. But yeah, the sensor-specific composite configs that are loaded are based on what sensors the reader says were loaded. If it is a made up sensor like `merged` then no composites will be available (I suppose visir.yaml *might* be), but unless someone added a `merged.yaml` then no composites specific to this reader would be loaded. Maybe we just leave this as you have it and cross this bridge when we come to it.

simonrp84: Ok, I see what you mean. Would it help if I renamed the sensor to something more specific like `oc_cci_merged`? Then if anyone comes up with a composite in the future they can add it to a nicely named YAML rather than `merged.yaml`, which is probably a bit too generic.

djhoese: Merged is probably fine for now. As an example, for other readers like those that take the output of algorithm software and support multiple sensors we might have `sensors: [viirs, modis, avhrr]`. In other readers we might have `sensor_names` in the file handler or `.attrs["sensor"]` in the DataArray equal to some global attribute that the input files have that say which sensors the merged data came from. `.attrs['sensor']` can be a set of strings, not only a single sensor string.

djhoese: Another question that hasn't be explicitly asked: do the input data files contain the information about what sensor data went in to making them?

simonrp84: I have some files that do, and some files that don't contain that info. I think all new versions of the data contain an attribute with the list of sensors used, but older versions do not.

djhoese:
This looks good to me. I think in the future if/when you add another reader with this many datasets you should maybe consider learning how to implement the `available_datasets` method. It would probably save you a lot of time typing everything into the YAML.

As for the sensor, we'll deal with any complications of that when they cause problems.

PR #2166, opened 2022-08-05
Correct the sun azimuth angle range within satpy.

simonrp84:
At present, the azimuth angles returned for sun and satellite are on different ranges.
`satpy.modifiers.angles._get_sun_azimuth_ndarray` returns values in the range `-180. -> 180.` degrees.
`satpy.modifiers.angles._get_sensor_angles_ndarray` returns values in the range `0. -> 360.` degrees.

For things like Rayleigh correction this is not a problem as the angles are normalised on lines 56 and 57 of `atmosphere.py`. However, it would be better to make this normalisation directly on the solar angles before they're passed to any other functions.
My reasoning for this is:
1) Users that call either of the azimuth angle calculation functions do not currently know that the angles are on different ranges - this is not documented. This is obviously unwanted and potentially problematic if they don't check the angles themselves before use.
2) Although only the azimuth difference is used in satpy at the moment this may not always be the case (for more complex atmospheric correction, f.ex). We should therefore be prepared for future changes by ensuring azimuths are correct.

This PR applies a very simple fix (taken from `atmosphere.py`) to correct the solar azimuth range. Both azimuths will now be in the `0 -> 360` range as expected.


djhoese:
@adybbroe or @mraspaud any memory of why the azimuth angle is in a different range in pyorbital?

@simonrp84 I'm tempted to request that the rayleigh correction be updated to only do the `% 360` when data is loaded from provided "optional prerequisites" but avoids the extra calculation if the `get_angles` functions were used. Thoughts?

Also, tests please.

simonrp84:
I've added some tests to this now. What I did was rearrange the tests as was done in #2175, I moved a bunch of tests from `test_modifiers.py` into `modifier_tests\test_angles.py`. I then added a `test_solazi_correction` function to the `test_angles.py` file. Hopefully that enables either #2175 or this PR to be merged without causing issues when merging the other PR later.

satpy/tests/modifier_tests/test_angles.py
```diff
@@ -0,0 +1,340 @@
...
+        from datetime import datetime
+
+        from satpy.modifiers.angles import _get_sun_azimuth_ndarray
+
+        lats = np.array([-80, 40, 0, 40, 80])
+        lons = np.array([-80, 40, 0, 40, 80])
+
+        dt = datetime(2022, 1, 5, 12, 50, 0)
+
+        azi = _get_sun_azimuth_ndarray(lats, lons, dt)
```
djhoese, line 338: Can you please right the test against the higher level function(s) instead of this "private" one?

djhoese (changes requested):

satpy/tests/modifier_tests/test_angles.py
simonrp84: the only higher level one that isn't private is `get_angles`. It's hard to get against that as it needs area definitions that would cover a bit enough region to have the whole range of angles.
So is it really necessary to test against the non-private function, especially considering most of the other tests in there are already testing private functions?
(the function above is using `_get_sensor_angles_from_sat_pos` for example)

djhoese: Fine.

djhoese (approved):

PR #2181, opened 2022-08-16
Add true color RGB for FCI

gerritholl:
Add a true color RGB for FCI.

 - [ ] Closes #xxxx
 - [ ] Tests added 
 - [ ] Fully documented 

Marking as draft because I don't know if it looks right...

`true_color_raw` (was already supported):

201709201220-MTG-I1-fci-mtg_fci_fdss_1km-true_color_raw

The `true_color` with this PR looks like:

201709201220-MTG-I1-fci-mtg_fci_fdss_1km-true_color

I find it quite dark, compared to, for example, MODIS true color images such as [visible in EOSDIS](https://worldview.earthdata.nasa.gov/?v=-100.47848594224924,-33.7358947568389,93.19932560790274,65.8266052431611&t=2022-08-16-T14%3A35%3A09Z), but I wouldn't know what to do differently on the Satpy side of things, so maybe this is due to the test data or perhaps this is expected.

This is with pyspectral-0.11.1.dev3+g6948a7a (latest main).

### DIFF
--- a/satpy/etc/composites/fci.yaml
+++ b/satpy/etc/composites/fci.yaml
@@ -11,3 +11,14 @@ composites:
       - name: 'cloud_state'
     lut: [.nan, 0, 1, 1, 1, 1, 1, 1, 0, .nan]
     standard_name: binary_cloud_mask
+
+  true_color:
+    compositor: !!python/name:satpy.composites.SelfSharpenedRGB
+    prerequisites:
+      - name: vis_06
+        modifiers: [sunz_corrected, rayleigh_corrected]
+      - name: vis_05
+        modifiers: [sunz_corrected, rayleigh_corrected]
+      - name: vis_04
+        modifiers: [sunz_corrected, rayleigh_corrected]
+    standard_name: true_color


gerritholl:
Apparently, the brown Congo Basin is not because everything burnt down, but because the 0.51 µm channel at FCI deliberately misses the chlorophyl peak.  There is a similar problem for AHI.  When I add a corrected green composite consisting of a fraction of 0.93 VIS 0.5 and 0.07 NIR 0.8, the forest looks greener.  Those ratios were taken from @pnuu who posted them in Slack at https://pytroll.slack.com/archives/C011NR3LE20/p1653310377583929.

satpy/etc/composites/fci.yaml
```diff
@@ -11,3 +26,17 @@ composites:
       - name: 'cloud_state'
     lut: [.nan, 0, 1, 1, 1, 1, 1, 1, 0, .nan]
     standard_name: binary_cloud_mask
+
+  true_color:
+    compositor: !!python/name:satpy.composites.SelfSharpenedRGB
```
ameraner, line 31: what is the advantage here to use the SelfSharpenedRGB compositor? Aren't the channels already all in the same resolution?

ameraner (commented):
Thank you @gerritholl for this work! The start of many discussions and tunings for FCI :)

FYI, we have recently noticed an anomaly in the test dataset, namely due to a mistake in the radiance simulation framework the vis_08 channel is too bright, so that a factor of 0.8 should be applied to get correct radiances/reflectances. As the vis_08 is used in the green correction now , this 0.8 factor is probably influencing the fraction factor you chose for the correction. With real data we'll likely need to re-tune anyway, but this is one more thing to consider.

satpy/etc/composites/fci.yaml
ameraner: Or are you already thinking of the HRFI case? Would the sharpening work automatically if the vis_06 is provided in a higher resolution?

gerritholl: I'm thinking of the HR case.  I believe the sharpening should work if `vis_06` is provided at 500 metre.

ameraner: Nice!

gerritholl:
> the vis_08 channel is too bright, so that a factor of 0.8 should be applied to get correct radiances/reflectances

Although I don't think tuning based on buggy test data is the most productive, I've tried to replace the `0.07` by `0.056` all else being equal, I get:

201709201220-MTG-I1-fci-mtg_fci_fdss_1km-true_color

which is very slightly darker and not an improvement IMHO, so I think further tuning should be done when we have real data.

simonrp84:
LGTM!

satpy/composites/ahi.py
```diff
@@ -15,30 +13,8 @@
 #
 # You should have received a copy of the GNU General Public License along with
 # satpy.  If not, see <http://www.gnu.org/licenses/>.
-"""Composite classes for the AHI instrument."""
+"""Composite classes for AHI adjustments."""
```
djhoese, line 16: Adjustments? This module is for AHI specific composites, even if it only has the green correction it is still a composites module. Not a huge deal, but it stood out to me.

satpy/composites/ahi.py
```diff
@@ -15,30 +13,8 @@
...
-    def __call__(self, projectables, optional_datasets=None, **attrs):
-        """Boost vegetation effect thanks to NIR (0.8µm) band."""
-        LOG.info('Boosting vegetation on green band')
-
-        projectables = self.match_data_arrays(projectables)
-        new_green = sum(fraction * value for fraction, value in zip(self.fractions, projectables))
-        new_green.attrs = combine_metadata(*projectables)
-        return super(GreenCorrector, self).__call__((new_green,), **attrs)
+# The green corrector used to be defined here, but was moved to spectral.py
+# in Satpy 0.38 because it also applies to FCI.
```
djhoese, line 19: We should maybe do a grep. I think there was another sensor (AGRI?) that used this and could be updated.

satpy/composites/spectral.py
```diff
@@ -0,0 +1,72 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+class GreenCorrector(GenericCompositor):
+    """Corrector of the FCI or AHI green band.
+
+    The green band in FCI and AHI deliberately misses the chlorophyll peak
+    in order to focus on aerosol and ash rather than on vegetation.  This
+    makes true colour RGBs look like all the forests have burnt down, which
```
djhoese, line 33: While silly, I'm not sure this should be included.

satpy/composites/spectral.py
```diff
@@ -0,0 +1,72 @@
...
+    they haven't yet.  To make them look a bit greener, this corrector allows
+    to simulate the green band as a fraction of two or more other channels.
+
+    To be used, the composite takes two re more input channels and a parameter
```
djhoese, line 37: Typo? re

djhoese (changes requested):
Thanks for the cleanup. Just a couple small changes.

satpy/composites/ahi.py
gerritholl: Fixed in d053f22e.

satpy/composites/spectral.py
gerritholl: Changed formulation in d053f22e.

satpy/composites/spectral.py
gerritholl: Fixed typo in d053f22e (now reads "or" rather than "re").

gerritholl:
This fails only for the experimental run now, which, as I understand it, is allowed.  So the PR is ready as far as I'm concerned.

satpy/composites/spectral.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
mraspaud, line 2: These aren't needed, right?

mraspaud:
```suggestion
```

mraspaud (commented):
Just a minor cleanup suggestion, otherwise LGTM

satpy/composites/spectral.py
gerritholl: Removed in 718a27a1.

mraspaud (approved):
LGTM

PR #2190, opened 2022-08-22
Fix some enhancements producing dask arrays wrapped in dask arrays

djhoese:
I noticed while counting the number of dask computations that some enhancements were causing more computations than they used to. This was caused by some of the dask arrays producing dask arrays so the dask arrays were being passed to rasterio during geotiff saving which caused yet another compute.

The issue was that due to the refactoring in #2133, the functions were accidentally doing dask "stuff" (delayed functions or `da.where`) inside a function that was already map_blocks'd. This PR fixes all of that and adds checks to the tests to make sure it hopefully doesn't happen again in the future.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #2191, opened 2022-08-23
Fix automatic reader table not listing readers with missing dependencies

### LINKED ISSUE #2189: Supported readers page does not include many readers
**Describe the bug**
The table with the supported readers at the RTD pages does not seem to update correctly. 
I was looking for the EPSSG MWI/ICI which was added by a recent PR, but can't find them?

**To Reproduce**

```python
# Your code here

```

**Expected behavior**

**Actual results**

**Screenshots**

**Environment Info:**
 - Satpy Version: [e.g. 0.9.0]
 - PyResample Version:

**Additional context**

BENR0:
This makes a minor modification to the `available_readers` function to be able to pass a yaml loader type.
Based on this for reader table generation when `BasicLoader` is passed reader classes are not instantiated and thus readers with missing dependencies in the RTD build environment are now included in the reader table.

 - [x] Closes #2189

### DIFF
--- a/doc/source/reader_table.py
+++ b/doc/source/reader_table.py
@@ -20,6 +20,11 @@
 
 from satpy.readers import available_readers
 
+try:
+    from yaml import BaseLoader
+except ImportError:
+    from yaml import BaseLoader  # type: ignore
+
 
 def rst_table_row(columns=None):
     """Create one row for a rst table.
@@ -72,7 +77,7 @@ def generate_reader_table():
     table = [rst_table_header("Satpy Readers", header=["Description", "Reader name", "Status", "fsspec support"],
                               widths=[45, 25, 30, 30])]
 
-    reader_configs = available_readers(as_dict=True)
+    reader_configs = available_readers(as_dict=True, yaml_loader=BaseLoader)
     for rc in reader_configs:
         table.append(rst_table_row([rc.get("long_name", "").rstrip("\n"), rc.get("name", ""),
                                     rc.get("status", ""), rc.get("supports_fsspec", "false")]))
--- a/doc/source/reader_table.py
+++ b/doc/source/reader_table.py
@@ -18,12 +18,9 @@
 # along with satpy.  If not, see <http://www.gnu.org/licenses/>.
 """Module for autogenerating reader table from config files."""
 
-from satpy.readers import available_readers
+from yaml import BaseLoader
 
-try:
-    from yaml import BaseLoader
-except ImportError:
-    from yaml import BaseLoader  # type: ignore
+from satpy.readers import available_readers
 
 
 def rst_table_row(columns=None):
--- a/satpy/readers/__init__.py
+++ b/satpy/readers/__init__.py
@@ -26,11 +26,7 @@ from datetime import datetime, timedelta
 from functools import total_ordering
 
 import yaml
-
-try:
-    from yaml import UnsafeLoader
-except ImportError:
-    from yaml import Loader as UnsafeLoader  # type: ignore
+from yaml import UnsafeLoader
 
 from satpy._config import config_search_paths, get_entry_points_config_dirs, glob_config
 
--- a/satpy/readers/__init__.py
+++ b/satpy/readers/__init__.py
@@ -370,7 +370,7 @@ def get_valid_reader_names(reader):
     return new_readers
 
 
-def available_readers(as_dict=False):
+def available_readers(as_dict=False, yaml_loader=UnsafeLoader):
     """Available readers based on current configuration.
 
     Args:
@@ -385,7 +385,7 @@ def available_readers(as_dict=False):
     readers = []
     for reader_configs in configs_for_reader():
         try:
-            reader_info = read_reader_config(reader_configs)
+            reader_info = read_reader_config(reader_configs, loader=yaml_loader)
         except (KeyError, IOError, yaml.YAMLError):
             LOG.debug("Could not import reader config from: %s", reader_configs)
             LOG.debug("Error loading YAML", exc_info=True)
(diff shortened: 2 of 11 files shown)


mraspaud:
@BENR0 don't worry about the linting error, I have fixed it in my PR, which I will merge as soon as the tests pass

doc/source/reader_table.py
```diff
@@ -20,6 +20,11 @@
 
 from satpy.readers import available_readers
 
+try:
+    from yaml import BaseLoader
+except ImportError:
+    from yaml import BaseLoader  # type: ignore
```
mraspaud, line 26: I'm not sure how this can work? this is importing twice the same thing, right?

satpy/readers/__init__.py
```diff
@@ -370,7 +370,7 @@ def get_valid_reader_names(reader):
     return new_readers
 
 
-def available_readers(as_dict=False):
+def available_readers(as_dict=False, yaml_loader=UnsafeLoader):
```
mraspaud, line 373: Is there a reason to use the unsafe loader here or would it just work with the BaseLoader?

mraspaud (commented):
Thanks for the quick fix, any chance of adding a test?

doc/source/reader_table.py
BENR0: Indeed. I removed to much from the second import I guess. Apart from that this is something I forgot to add to the PR. I think this try/except was introduced in Satpy during the transition phase of pyyaml. The first import works since pyyaml version 5.1 which is already 2 years old. My suggestion would be to update the Satpy requirements to need pyyaml>= 5.1 then we could do away with the try/except.

satpy/readers/__init__.py
BENR0: Well I was just setting the default here to be "backwards" compatible meaning leave the old default functionality unchanged as much as possible. I don't know what the intentions are. If this is set to BaseLoader as a default this will always list all readers with yaml config files even if they are missing dependencies. So if a user uses this to find out which readers are available to him this might be irritating if he then tries one that can't work due to missing dependencies. If this is changed it, at least, needs to be documented somewhere. if this has other implications I didn't check.

doc/source/reader_table.py
BENR0: I can add a test for this if it is decided that this indeed should be the solution (@djhoese mentioned something about running the table generating function inside the mocking but that seemed a little more complicated to me and I don't know if that is even possible).

mraspaud: yes, I'm good with setting a limit on pyyaml version

satpy/readers/__init__.py
mraspaud: Good point.

djhoese: I thought the BaseLoader raised an error when it encountered python imports and other "dynamic" YAML features. This makes sense to me for backwards compatibility and to keep the dependency checking the default.

doc/source/reader_table.py
djhoese: I actually have a special loader in this test code that ignores all tags so that imports can be skipped over:

https://github.com/pytroll/satpy/blob/267e779cea71cb5286d5edc1b4da7478eb62fc62/satpy/tests/test_readers.py#L632-L635

If that isn't needed and the BaseLoader can be used that'd be pretty nice.

BENR0: It seems the BaseLoader works. I will add a test tomorrow.

mraspaud: Yes, the base loader ignore anything python-specific iirc

djhoese:
So what's left on this? I think @mraspaud was hoping for some type of test?

BENR0:
@mraspaud @djhoese I added the test.

satpy/tests/test_readers.py
```diff
@@ -662,6 +671,28 @@ def test_available_readers(self):
...
+            return real_import(name, globals=globals, locals=locals, fromlist=fromlist, level=level)
+
+        self.monkeypatch.delitem(sys.modules, 'netcdf4', raising=False)
+        self.monkeypatch.setattr(builtins, '__import__', patched_import_error)
+
+        with pytest.raises(ImportError):
+            import netcdf4  # noqa: F401
+
+        reader_names = available_readers(yaml_loader=yaml.BaseLoader)
+        self.assertIn('abi_l1b', reader_names)  # needs netcdf4
```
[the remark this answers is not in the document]
BENR0: Well to be honest I got this monkey patching idea from the web. From my understanding the monkey patch should patch all imports of netcdf4 even in xarray and throw an ImportError.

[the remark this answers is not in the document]
djhoese: So...I'm not questioning the monkey patching, I'm saying that `abi_l1b` is one reader that won't be affected by this failed import, at least not at import time.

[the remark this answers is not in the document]
BENR0: ah ok, missunderstanding then. I added the `viirs_l1b`. Missed that one because I copied the line from above but didn't change it. Will push the change in a minute.

satpy/readers/__init__.py
```diff
@@ -370,7 +366,7 @@ def get_valid_reader_names(reader):
...
```
djhoese, line 369: Can you add `yaml_loader` to the docstring please.

djhoese (changes requested):
One small request

satpy/readers/__init__.py
BENR0: @djhoese done.

djhoese (approved):

PR #2200, opened 2022-09-06
Remove mention of fallback to GDAL in geotiff writer

djhoese:
This bad information came up in the monthly meeting we had today. The geotiff writer does not use GDAL's python bindings directly at all anymore. This PR removes the mention of this and states that `rasterio` is required.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


djhoese:
Got approval on slack.

PR #2203, opened 2022-09-18
Fix experimental dependency stdlibc++ issues in CI

djhoese:
See https://github.com/rasterio/rasterio/issues/2591 for more detailed debugging.

The shortest answer I can give is that unstable wheels built for PyPI are typically built with a manylinux container using the system glibcxx (stdlibc++.so) which is relatively old. Conda-forge packages are built with a conda-forge-specific glibcxx which is much newer. If the wheels are installed and imported before a conda-forge package that uses stdlibc++.so then it will load the system .so and when the conda-forge package is imported it will reuse the already loaded but *OLDER* .so.

This is my attempt at forcing our CI to load the conda-forge libstdc++ no matter what.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -86,6 +86,8 @@ jobs:
       - name: Install unstable dependencies
         if: matrix.experimental == true
         shell: bash -l {0}
+        # We must get LD_PRELOAD for stdlibc++ or else the manylinux wheels
+        # may break the conda-forge libraries trying to use newer glibc versions
         run: |
           python -m pip install \
           --index-url https://pypi.anaconda.org/scipy-wheels-nightly/simple/ \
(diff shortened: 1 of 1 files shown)


djhoese:
Note the latest comment by the scipy maintainer here: https://github.com/rasterio/rasterio/issues/2591#issuecomment-1251042465

He reminded me that this wasn't an issue with old scipy builds, but their new process is not including a "symbol hiding" step they used to do. These symbols that are being included seem to be leading the dynamic library loader to behave differently and cause this issue.

I think the changes in this PR are still useful in the long run since we are "crossing the streams" as far as manylinux wheels and conda-forge packages, but it is probably avoidable by some of these higher-compiled libraries.

mraspaud (approved):
LGTM, thanks for the hard detective work :detective: !

PR #2217, opened 2022-09-30
Fix failing get_satpos precision test and other CI failures

djhoese:
A unit test has started failing. This PR adds an `rtol` to allow it to pass now.


djhoese:
WTF I think it just seg faulted.

mraspaud (approved):
LGTM, thanks for fixing this so fast

PR #2120, opened 2022-06-02
Add reader for MWS onboard EPS-SG-A

adybbroe:
This is a first draft reader for the EPS-SG MWS level-1b netCDF file format.
 

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 
 - [x] Add your name to `AUTHORS.md` if not there already

This is an early draft so far, and several things needs to be addressed:
 - [x] Rearrange the `FrquencyRange` and related classes defined in the AAPP-MHS reader so they can be reused in several readers
 - [x] Add unittests for the reader
 - [x] Decide how the channel comparisons should behave!
 - [x] Add handling of auxiliary data sets, see FCI and VII readers

PR #2221, opened 2022-10-05
Fix ici after mws merge

adybbroe:
When supporting MWS in #2120 the use of the `FrequencyDoubleSideBand` type was moved to another module. I/we forgot to update the yaml file for ICI, so the code stopped working as is for ICI.
This PR fixes this.

In addition I had not tested that the code is not able to handle the ancillary datasets like surface type and terrain yet. So I removed this from the yaml file. Reading such datasets is considered of less importance at the moment. We could add an issue on that, but I would rather take that when (and if) a need appears.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/readers/ici_l1b_nc.yaml
+++ b/satpy/etc/readers/ici_l1b_nc.yaml
@@ -12,7 +12,7 @@ reader:
     name:
       required: true
     frequency_double_sideband:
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM

PR #2229, opened 2022-10-10
Writer for Zarr format

BENR0:
Implements a Zarr writer.

This draft basically just reused all of the logic of the netcdf writer and uses xarray `to_zarr` instead of `to_netcdf`.
Except for groups this seems to work at first sight. There are no tests yet so this was just visually "tested" with and without using a `LocalCluster`.  

Todos:
- write tests
- refactor code (duplication with netcdf writer)
- "performance" analysis
- work on groups
- write reader for round trip use (separate PR)


BENR0:
closed in favour of #2259

PR #2235, opened 2022-10-13
Fix Scene.load modifiers keyword argument having no effect

### LINKED ISSUE #2234: Scene `modifiers` keyword argument has no effect
**Describe the bug**
Originally described on Satpy, doing the following doesn't actually change the loaded bands. You must use a DataQuery to get it to work.

**To Reproduce**

```python
from satpy import Scene; from glob import glob

scn = Scene(reader='modis_l1b', filenames=glob("/data/satellite/modis/input_modis_test1/*.hdf"))
scn.load(["1", "2"])

scn["1"].data.mean().compute()
# 9.824119

scn = Scene(reader='modis_l1b', filenames=glob("/data/satellite/modis/input_modis_test1/*.hdf"))
scn.load(["1", "2"], modifiers=("sunz_corrected",))

scn["1"].data.mean().compute()
# 9.824119

scn["1"].attrs["modifiers"]
# ()

```

**Expected behavior**
A modifier is applied to all bands loaded.

**Actual results**
See above.

**Additional context**

After seeing this I thought maybe the data needed to be resampled to work, but that also has no effect.

djhoese:
As described in the related issue, passing `modifiers=('some_modifier',)` to `Scene.load` has no effect, but it should unless I'm forgetting some conversation where we decided no to do this.

 - [x] Closes #2234  
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/dataset/dataid.py
+++ b/satpy/dataset/dataid.py
@@ -508,7 +508,7 @@ class DataQuery:
     """The data query object.
 
     A DataQuery can be used in Satpy to query for a Dataset. This way
-    a fully qualified DataID can be found even if some of the DataID
+    a fully qualified DataID can be found even if some DataID
     elements are unknown. In this case a `*` signifies something that is
     unknown or not applicable to the requested Dataset.
     """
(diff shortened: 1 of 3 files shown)


mraspaud (approved):
LGTM, thanks for taking the time to reformat docstring along the way!

PR #2239, opened 2022-10-19
Fix incorrect rows_per_scan in 'acspo' reader

djhoese:
In #2114 I updated the ACSPO reader so the `sensor_name` attribute was lowercase, but I didn't update the reader's internal dictionary mapping sensor name to rows per scan. This meant that the file handler was always returning "0" rows per scan. This made the EWA resampling results look bad. This PR fixes the case of this internal dictionary.

 - [x] Closes https://github.com/ssec/polar2grid/issues/498 
 - [x] Tests added

### DIFF
--- a/satpy/readers/acspo.py
+++ b/satpy/readers/acspo.py
@@ -33,9 +33,9 @@ LOG = logging.getLogger(__name__)
 
 
 ROWS_PER_SCAN = {
-    'MODIS': 10,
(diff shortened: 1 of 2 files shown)


djhoese:
Only failure is the unstable environment and given the simplicity of these changes I'm merging without review. Talk to me on slack if you have concerns.

PR #2240, opened 2022-10-21
Refactor RGB ratio sharpening again for better performance

djhoese:
This reverts most of the changes in #2131 due to regressions seen in the benchmarks. The changes here show about 1GB less memory usage on my machine and ~25s saved in execution time. This is not a complete reversion of #2131 as it still uses map_blocks to get the ratio used in the ratio sharpening. This seems to me the best of both worlds. It means I can use numpy inplace operations (ex. specifying the `out` parameter in a numpy ufunc) while not having to copy/create the large RGB stacked array in numpy-land. I think this creation of the RGB array chunks was the main reason for the increased memory usage seen in the benchmarks.

This PR also includes the removal of the inter-band masking for NaN values. This isn't needed and was redundant as the GenericCompositor base class already does this.

 - [ ] Closes #xxxx 
 - [ ] Tests added

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -924,11 +924,11 @@ class RatioSharpenedRGB(GenericCompositor):
 
     def __init__(self, *args, **kwargs):
         """Instanciate the ration sharpener."""
-        self.high_resolution_color_name = kwargs.pop("high_resolution_band", "red")
-        if self.high_resolution_color_name not in ['red', 'green', 'blue', None]:
+        self.high_resolution_color = kwargs.pop("high_resolution_band", "red")
+        if self.high_resolution_color not in ['red', 'green', 'blue', None]:
             raise ValueError("RatioSharpenedRGB.high_resolution_band must "
                              "be one of ['red', 'green', 'blue', None]. Not "
-                             "'{}'".format(self.high_resolution_color_name))
+                             "'{}'".format(self.high_resolution_color))
         kwargs.setdefault('common_channel_mask', False)
         super(RatioSharpenedRGB, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


djhoese:
Oh I should also add that I added some `np.errstate` context managers around operations that we *know* will produce numpy warnings in regular cases containing space pixels.

djhoese:
Ran `asv` on commits from May to present skipping every 2 weeks or so. Here are the results including the changes in this PR as the last data point:

image

This is AHI HSD true_color memory which is what of the biggest regressions on the public benchmark results (mraspaud's github.io). This brings us back to a reasonable amount of memory usage. I should note that execution time (as I think was seen in the public benchmarks) was not effected by these changes (in the original PR or this one).

satpy/composites/__init__.py
```diff
@@ -945,79 +945,72 @@ def __call__(self, datasets, optional_datasets=None, **info):
...
+            )
+            old_red = red
+            old_green = green
+            old_blue = blue
+            red = high_res if low_resolution_index == 0 else red * ratio
+            green = high_res if low_resolution_index == 1 else green * ratio
+            blue = high_res if low_resolution_index == 2 else blue * ratio
+            red.attrs = old_red.attrs
+            green.attrs = old_green.attrs
+            blue.attrs = old_blue.attrs
```
mraspaud, line 992: would `with xr.set_options(keep_attrs=True):` be an alternative here?

satpy/resample.py
```diff
@@ -1415,6 +1415,7 @@ def resample_dataset(dataset, destination_area, **kwargs):
         return dataset
 
     fill_value = kwargs.pop('fill_value', get_fill_value(dataset))
+    print(fill_value)
```
mraspaud, line 1418:
```suggestion
```

satpy/composites/__init__.py
```diff
@@ -945,79 +945,72 @@ def __call__(self, datasets, optional_datasets=None, **info):
...
+        blue = datasets[2]
         if optional_datasets and self.high_resolution_band is not None:
             LOG.debug("Sharpening image with high resolution {} band".format(self.high_resolution_band))
             high_res = datasets[3]
             if 'rows_per_scan' in high_res.attrs:
                 new_attrs.setdefault('rows_per_scan', high_res.attrs['rows_per_scan'])
             new_attrs.setdefault('resolution', high_res.attrs['resolution'])
             colors = ['red', 'green', 'blue']
-            low_res_idx = colors.index(self.high_resolution_band)
+            low_resolution_index = colors.index(self.high_resolution_band)
```
mraspaud, line 968: I'm not sure I understand here: how can the position of the high resolution band name become the low resolution index?

oh, ok, got it now. To avoid confusion in he future, could the `red`, `green`, `blue` variables be renamed to `low_res_red`, `low_res_green`, `low_res_blue` and `colors` be renamed to `low_res_colors`?

satpy/resample.py
pnuu, line 1418: Left-over debug print
```suggestion
```

satpy/composites/__init__.py
mraspaud: And maybe rename `self.high_resolution_band` to `self.high_resolution_band_name`?

mraspaud (changes requested):
Thanks for fixing the performance of the ration sharpener. Do I understand correctly that the performance gain is only when there is no high res band provided?
Otherwise I have some styling comments.

PR #2164, opened 2022-08-03
Add low level moisture composite

gerritholl:
Add the ESSL composite to visualise low level moisture using the ratio between 0.91 µm and 0.86 µm.  Experimental.

Greyscale version:

MTI1-low_level_moisture-nq0008km-20170920_1120-1129-devel

Test image produced from FCI test data for 2017-09-20 11:20, stretched between 0.35 and 0.86.

In the greyscale version, clouds are white.  Areas with high low level moisture are dark.

The fine structure will be more visible in a colorised version.

This visualisation was developed by Hans-Peter Rösli, Pieter Groenemeijer, and the European Severe Storms Laboratory..

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented 
 - [ ] Add your name to `AUTHORS.md` if not there already

PR #2246, opened 2022-10-27
Fix MODIS reader tests failing with new geotiepoints

djhoese:
The new version of python-geotiepoints implements MODIS interpolation in Cython. In addition to being faster, it adds some more strict-ness/rigidity to what data types can be passed to a function. To simplify the code in geotiepoints, when the user passes more than one array they must all be the same data type (ex. all 32-bit floats or all 64-bit floats). In the real world MODIS cases I've never seen an issue with mixing (mismatching) array data types, but the tests in Satpy do introduce this.

In this PR I fix the small issue in the reader which enforces casting to 32-bit floats. In the tests, I wasn't able to force the HDF4 files produced to match the real MODIS data exactly where this casting wouldn't be needed.

Note: I could update geotiepoints to be more flexible...but this seems so much simpler.

CC @gerritholl for #2164 CI failures

 - [ ] Closes #xxxx 
 - [ ] Tests added

### DIFF
--- a/satpy/readers/hdfeos_base.py
+++ b/satpy/readers/hdfeos_base.py
@@ -243,7 +243,7 @@ class HDFEOSBaseFileReader(BaseFileHandler):
         # we still need to convert integers to floats
         if scale_factor is not None and not is_category:
             if add_offset is not None and add_offset != 0:
-                data = data - add_offset
+                data = data - np.float32(add_offset)
             data = data * np.float32(scale_factor)
 
         if good_mask is not None:


gerritholl (approved):
Looks good to me, even if I don't understand what's going on :)

djhoese:
Basically you subtract a python `float` from a 32-bit floating numpy array it turns it into a 64-bit floating numpy array. Now it doesn't because I'm forcing the scalar to a 32-bit float.

PR #2247, opened 2022-10-27
Document behaviour on default enhancement

gerritholl:
Document Satpys behaviour on applying a default enhancement if no enhancement is defined, and document on how to avoid this enhancement from being applied.  Also add a recommendation to always define an enhancement even if using the default.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Fully documented

### DIFF
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -536,6 +536,23 @@ the file) as::
           kwargs:
             gamma: [1.7, 1.7, 1.7]
 
+.. warning::
+   If you define a composite with no matching enhancement, Satpy will by
+   default apply a crude stretch enhancement.  If you want no enhancement
+   at all (maybe you are enhancing a composite based on
+   :class:`DayNightCompositor` where the components have their own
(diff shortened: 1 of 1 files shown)


doc/source/composites.rst
```diff
@@ -536,6 +536,23 @@ the file) as::
           kwargs:
             gamma: [1.7, 1.7, 1.7]
 
+.. warning::
+   If you define a composite with no matching enhancement, Satpy will by
+   default apply a crude stretch enhancement.  If you want no enhancement
```
djhoese, line 541: I think technically it isn't a crude stretch but the "linear" stretch which defaults to a 5% cutoff/quantile on either end.

https://github.com/pytroll/trollimage/blob/93e547a45a1c3a13768da4a5169a3f5bf5ebc75a/trollimage/xrimage.py#L1014

gerritholl: Fixed in 8dba3d9.

djhoese (approved):

PR #2262, opened 2022-11-04
Fix ratio sharpening not sharing invalid mask between bands

djhoese:
Fixes a bug introduced in #2240 where I refactored the duplicate masking behavior because it was already handled in the parent class, GenericCompositor. However, I didn't change the default for the ratio sharpening class to enable this behavior in the parent class so it wasn't being done at all. The tests were not complex enough to pick up this mistake but have been updated in this PR.

The behavior I'm talking about is where a mask is created of the NaN locations in each band, merged, and then applied to all the bands. Without this, you get strange colors where one band has missing data but another doesn't. This is especially necessary on the edges of resampled data where different resolution may "stretch" further geographically than other bands.

 - [x] Closes https://github.com/ssec/polar2grid/issues/556 
 - [x] Tests added

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -929,7 +929,6 @@ class RatioSharpenedRGB(GenericCompositor):
             raise ValueError("RatioSharpenedRGB.high_resolution_band must "
                              "be one of ['red', 'green', 'blue', None]. Not "
                              "'{}'".format(self.high_resolution_color))
-        kwargs.setdefault('common_channel_mask', False)
         super(RatioSharpenedRGB, self).__init__(*args, **kwargs)
 
     def __call__(self, datasets, optional_datasets=None, **info):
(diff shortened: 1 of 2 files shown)


djhoese:
When I run asv on my own machine (while I'm doing other things) for the last 4 commits where this PR is the last entry, I get this for peakmem on the AHI true_color load test:

image

So not much of a change. Time:

image

It may look like it goes up, but those other commits weren't even related to these changes so this is just system variance from my point of view. I think this is safe to merge in that regard.

I'm still tracking down another issue that @kathys identified in https://github.com/ssec/polar2grid/issues/558 regarding quality/sharpness of the images, but I'm not sure it is related to these refactorings. I should know more in the next hour or so.

djhoese:
CI is having a rough day today. First it was a mamba hiccup, now it is a coveralls connection hiccup. :crossed_fingers:

djhoese:
Merging given the simplicity of the fix, the fact that I broke it originally, and the issues it is causing for me downstream. Slap my wrist if you must. :straight_ruler:

PR #2271, opened 2022-11-11
Add LI L2 reader

### LINKED ISSUE #2213: Update `li_l2`-reader to read MTG LI L2 test data
## Feature Request

The current `li_l2`-reader does not support reading of the EUMETSAT released LI L2 test data.
https://www.eumetsat.int/new-li-test-data-released-2021

Update the reader to support the latest test data.

None

seenno:
Add LI L2 reader and remove the old code as it's obsolete. 
The reader handles both, LI L2 point products (LI-LE, LI-LEF, LI-LGR, LI-LFL) and LI L2 accumulated products (LI-AF, LI-AFR, LI-AFA). For the accumulated products, the reader supports the output as a gridded product. This can be activated by adding the reader kwarg `with_area_definition=True`.

 - [x] Closes #2213  
 - [x] Tests added 
 - [x] Fully documented 


ameraner:
The tests are failing with 
```self = <[RecursionError('maximum recursion depth exceeded') raised in repr()] DataArray object at 0x7feb90f7dc40>
data = dask.array<broadcast_to, shape=(123,), dtype=datetime64[ns], chunksize=(123,), chunktype=numpy.ndarray>
coords = Coordinates:
    *empty*, dims = ('scalar',), name = None, attrs = {}
indexes = Indexes:
    *empty*, fastpath = False

    def __init__(
        self,
        data: Any = dtypes.NA,
        coords: Sequence[Sequence[Any] | pd.Index | DataArray]
        | Mapping[Any, Any]
        | None = None,
        dims: Hashable | Sequence[Hashable] | None = None,
        name: Hashable = None,
        attrs: Mapping = None,
        # internal parameters
        indexes: dict[Hashable, Index] = None,
        fastpath: bool = False,
    ) -> None:
        if fastpath:
            variable = data
            assert dims is None
            assert attrs is None
            assert indexes is not None
        else:
            # TODO: (benbovy - explicit indexes) remove
            # once it becomes part of the public interface
            if indexes is not None:
>               raise ValueError("Providing explicit indexes is not supported yet")
E               ValueError: Providing explicit indexes is not supported yet

/usr/share/miniconda3/envs/test-environment/lib/python3.9/site-packages/xarray/core/dataarray.py:395: ValueError
```

at line https://github.com/pytroll/satpy/pull/2271/files#diff-3b2bff08b4001ec6f72cca67791cc4322b38e0db97e68f2109791093e56e6052R595
We actually need to define the indexes for the reader to work as intended...

It seems a regression of xarray, since the tests are passing e.g. with xarray-2022.3.0, but failing with 2022.11.0 and the current xarray master. 
Any idea? @djhoese is this related to the RecursionError issues we saw for the ancillary_variables recently?

djhoese:
No this isn't related to my stuff at all. Did this work in the past? Why do you need to provide the indexes from the base DataArray?

mraspaud (commented):
Some preliminary comments. Also, CodeScene makes good points

satpy/readers/li_base_nc.py
```diff
@@ -0,0 +1,754 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
```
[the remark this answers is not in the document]
seenno: Removed in all files.

satpy/readers/li_base_nc.py
```diff
@@ -0,0 +1,754 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class LINCFileHandler(NetCDF4FileHandler):
+    """Base class used as parent for the concrete LI reader classes."""
+
+    def __init__(self, filename, filename_info, filetype_info, cache_handle=True):
+        """Initialize LINCFileHandler."""
+        super(LINCFileHandler, self).__init__(filename, filename_info, filetype_info,
```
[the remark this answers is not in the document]
seenno: Changes as suggested.

satpy/readers/li_base_nc.py
```diff
@@ -0,0 +1,754 @@
...
+    def register_coords_from_scan_angles(self):
+        """Register lat lon datasets in this reader."""
+        lat_name, lon_name = self.get_latlon_names()
+        self.register_dataset(lon_name)
+        self.register_dataset(lat_name)
+
+    def get_area_def(self, _dsid):
+        """Compute area definition for a dataset, not supported in this reader."""
+        raise NotImplementedError
+
```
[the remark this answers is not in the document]
seenno: Removed `get_area_def()` and `get_bounding_box()`.

ameraner:
Sorry @djhoese, I just noticed that the code formatting in my comment swallowed the first line (edited now), which somehow points to a `RecursionError`.. that's why I was asking about the `RecursionError`s we saw lately.  

But the issue seems actually to be in the definition of the `indexes`. The reader actually works without it, so we just removed it (note that this code has been written by external contractors, so we are not fully aware of all the code details).

More info on the `indexes` if you're interested, from my investigation:
In old-ish versions of xarray (<=2022.03), `xr.DataArray` was accepting `indexes` in the init, but I think it was not actually using it unless when passing `fastpath=True` (see https://github.com/pydata/xarray/blob/v2022.03.0/xarray/core/dataarray.py#L404-L416, the docstring doesn't mention this..). From 2022.06.0rc0 on, they implemented the `ValueError` check referenced above, I suppose to avoid this silent ignoring.. However, it seems that xarray devs are picking up this again in the last weeks, see https://github.com/pydata/xarray/pull/7214.

satpy/readers/li_base_nc.py
```diff
@@ -0,0 +1,754 @@
...
+        elevation = self.get_measured_variable(self.swath_coordinates['elevation'])
+        elevation = self.apply_use_rescaling(elevation)
+
+        # Daskify inverse projection computation:
+        lon, lat = da.map_blocks(self.inverse_projection, azimuth, elevation, proj_dict,
+                                 chunks=(2, azimuth.shape[0]),
+                                 meta=np.array((), dtype=azimuth.dtype),
+                                 dtype=azimuth.dtype,
+                                 )
```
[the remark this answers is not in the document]
ameraner: The code was inspired by this code snippet here https://github.com/pytroll/pyresample/blob/952cec6ee4bce55bb70f2c1237401356645bc3f6/pyresample/geometry.py#L2434-L2441, that is encapsulated in `get_lonlats` of `AreaDefinition`. As in the end it's only a simple `map_blocks` call, I would propose to leave it like this, if you agree.

seenno:
All comments have been addressed and most of CodeScene complaints solved. We are not sure if it makes sense to solve the remaining ones...

ameraner:
Re-adding Sauli and Dave to the reviewers as Github automatically removed them when requesting a new review from Martin.

sjoro (commented):
i have difficulties following the `_li_test_utils.py` structure. it has these `li_xx_schema`-methods that build up a correct test-dictionary for each product.
do we need this flexibility of passing a `settings`-dictionary to these tests? is it actually used? if it is NOT passed, the method builds up the required dictionary anyways...
a clearer way would be to define these dictionaries outside of the `li_xx_schema`-methods with fixed values, as is done with SEVIRI native-reader. with that it would be clear what is being tested... also, this would get rid of the CodeScene complaints with too long functions.
my two cents, i'll leave further comments on this with @mraspaud and @djhoese 
otherwise the reader looks good to me.

sjoro (changes requested):
see comments on `product_schemas` and `products` and refactor.

satpy/tests/reader_tests/test_li_l2_nc.py
```diff
@@ -0,0 +1,632 @@
...
+
+        for sname in sectors:
+            for vname, desc in sector_vars.items():
+                # variable should be in list of dataset:
+                assert vname in all_vars
+                self._test_dataset_single_sector_variable([sname, vname], desc, settings, handler)
+
+    def test_dataset_loading(self, filetype_infos):
+        """Test loading of all datasets from all products."""
+        products = [
```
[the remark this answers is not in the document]
seenno: Fixed as suggested, now only has one definition dict in `_li_test_utils`, imported in `_li_test_utils`.

djhoese:
Looks like pre-commit isn't happy but I think @mraspaud said he fixed these already and merged them in main. Could this branch be updated with the current state of pytroll's main branch?

satpy/etc/readers/li_l2_nc.yaml
```diff
@@ -0,0 +1,193 @@
...
+      product_type: '2-AFA'
+      variables: ['mtg_geos_projection',
+                   'accumulation_start_times',
+                   'accumulation_offsets',
+                   'x',
+                   'y',
+                   'accumulated_flash_area',
+                   'l1b_geolocation_warning',
+                   'l1b_radiometric_warning',
+                   'average_flash_qa']
```
pnuu, line 185: Intendation is a bit off
```suggestion
      variables: ['mtg_geos_projection',
                  'accumulation_start_times',
                  'accumulation_offsets',
                  'x',
                  'y',
                  'accumulated_flash_area',
                  'l1b_geolocation_warning',
                  'l1b_radiometric_warning',
                  'average_flash_qa']
```

satpy/etc/readers/li_l2_nc.yaml
```diff
@@ -0,0 +1,193 @@
...
+                        'filter_values',
+                        'l1b_filter_qa',
+                        'l2_group_filter_qa',
+                        'l2_flash_filter_qa']
+      variables: []
+      variable_transforms:
+        l1b_chunk_offsets:
+          accumulate_index_offset: "{sector_name}/l1b_window"
+        epoch_time:
+          seconds_to_datetime: "2000-01-01T00:00:00.000000"
```
pnuu, line 33: Are these epoch times such that they might change separately for different products, or could it be defined in the code as a single constant?

satpy/readers/li_base_nc.py
```diff
@@ -0,0 +1,769 @@
...
+        """Apply the accumulate_index_offset transform on a given array."""
+        # retrieve the __index_offset here, or create it if missing:
+        # And keep track of the shared ds_info dict to reset it later in combine_info()
+        self.current_ds_info = ds_info
+        offset = ds_info.setdefault('__index_offset', 0)
+
+        ref_var = self.get_transform_reference('accumulate_index_offset', ds_info)
+
+        # Apply the current index_offset already reached on the indices we have in the current dataset:
+        data_array += offset
```
pnuu, line 624: In-place operations for `xr.DataArray`s or Dask arrays isn't generally a good idea.
```suggestion
        data_array = data_array + offset
```

satpy/readers/li_l2_nc.py
```diff
@@ -0,0 +1,107 @@
...
+            data_array = self.get_array_on_fci_grid(data_array)
+        return data_array
+
+    def get_area_def(self, dsid):
+        """Compute area definition for a dataset, only supported for accumulated products."""
+        is_var_with_swath_coord = self.is_var_with_swath_coord(dsid)
+        if is_var_with_swath_coord and self.with_area_def:
+            return get_area_def('mtg_fci_fdss_2km')
+
+        raise NotImplementedError
```
pnuu, line 68: A message here could be helpful.

satpy/readers/li_l2_nc.py
```diff
@@ -0,0 +1,107 @@
...
+        # Integer values without the application of scale_factor and add_offset
+        # hence no projection/index calculation.
+        # Note that x and y have origin in the south-west corner of the image
+        # and start with index 1.
+
+        rows = self.get_measured_variable('y')
+        cols = self.get_measured_variable('x')
+        attrs = data_array.attrs
+
+        rows, cols, data = da.compute(rows, cols, data_array)
```
pnuu, line 91: Could we avoid computing these and keep everything lazy?

satpy/readers/li_l2_nc.py
```diff
@@ -0,0 +1,107 @@
...
+        # origin is in the south-west corner, so we flip the rows (applying
+        # offset of 1 implicitly)
+        # And we manually offset the columns by 1 too:
+        rows = (5568 - rows.astype(int))
+        cols = cols.astype(int) - 1
+
+        data_2d = np.empty((5568, 5568), dtype=data_array.dtype)
+        data_2d[:] = np.nan
+
+        data_2d[rows, cols] = data
```
pnuu, line 102: I think this is also doable laziliy with `da.where()`.

pnuu (changes requested):
Few comments, suggestions and requests. I think the `get_array_on_fci_grid()` could be done in a lazy way, but I'd need to play with it a while to see how.

satpy/etc/readers/li_l2_nc.yaml
seenno: Now defined as a single constant in the code as suggested.

satpy/readers/li_l2_nc.py
seenno: OK, please check and suggest a solution, if there is a better one.

satpy/readers/li_l2_nc.py
seenno: OK, please check and suggest a solution, if there is a better one.

pnuu: So, this might work:
```python
import dask.array as da
import numpy as np

GRID_COLS = 5568
GRID_ROWS = 5568
NUM_POINTS = 123

rows = da.random.randint(low=0, high=GRID_ROWS - 1, size=NUM_POINTS)
cols = da.random.randint(low=0, high=GRID_COLS - 1, size=NUM_POINTS)
data = da.random.random(NUM_POINTS)

# Only the rows and columns need to be computed
rows, cols = da.compute(rows, cols)

# Create an empyt 1-D array for the results
flattened_result = np.nan * da.zeros((GRID_ROWS * GRID_COLS), dtype=data.dtype)
# Insert the data. Dask doesn't support this for more than one dimension at a time, so ...
flattened_resultresult[rows * GRID_COLS + cols] = data
# ... reshape to final 2D grid
data_2d = da.reshape(flattened_result, (GRID_ROWS, GRID_COLS))
```

seenno: @pnuu Now implemented.

pnuu (approved):
LGTM

satpy/readers/li_l2_nc.py
```diff
@@ -0,0 +1,108 @@
...
+        rows = self.get_measured_variable('y')
+        cols = self.get_measured_variable('x')
+        attrs = data_array.attrs
+
+        rows, cols = da.compute(rows, cols)
+
+        # origin is in the south-west corner, so we flip the rows (applying
+        # offset of 1 implicitly)
+        # And we manually offset the columns by 1 too:
+        rows = (5568 - rows.astype(int))
```
sjoro, line 96: i'd use the above defined constant here and in the next lines to avoid hard coding `5568`

satpy/tests/reader_tests/test_li_l2_nc.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def test_get_area_def(self, filetype_infos):
+        """Test retrieval of area def for accumulated products."""
+        handler = LIL2NCFileHandler('filename', {}, extract_filetype_info(filetype_infos, 'li_l2_af_nc'),
+                                    with_area_definition=True)
+
+        dsid = make_dataid(name="flash_accumulation")
+
+        area_def = handler.get_area_def(dsid)
+        assert area_def.shape == (5568, 5568)
```
sjoro, line 618: and test against it here

sjoro (changes requested):
a small request to maybe define a new constant in order to avoid hard coding `5568` in various places in the code.

satpy/readers/li_l2_nc.py
```diff
@@ -0,0 +1,108 @@
...
+import numpy as np
+import xarray as xr
+
+from satpy import CHUNK_SIZE
+from satpy.readers.li_base_nc import LINCFileHandler
+from satpy.resample import get_area_def
+
+logger = logging.getLogger(__name__)
+
+
```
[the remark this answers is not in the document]
seenno: Implemented.

satpy/readers/li_l2_nc.py
seenno: Implemented.

satpy/tests/reader_tests/test_li_l2_nc.py
```diff
@@ -0,0 +1,709 @@
...
+from unittest import mock
+
+import numpy as np
+import pytest
+import xarray as xr
+from pyproj import Proj
+
+from satpy._config import config_search_paths
+from satpy.readers.li_base_nc import LINCFileHandler
+from satpy.readers.li_l2_nc import LIL2NCFileHandler
```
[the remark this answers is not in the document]
seenno: Implemented.

satpy/tests/reader_tests/test_li_l2_nc.py
seenno: Implemented.

satpy/readers/li_l2_nc.py
```diff
@@ -0,0 +1,109 @@
...
+        rows, cols = da.compute(rows, cols)
+
+        # origin is in the south-west corner, so we flip the rows (applying
+        # offset of 1 implicitly)
+        # And we manually offset the columns by 1 too:
+        rows = (LI_GRID_SHAPE[0] - rows.astype(int))
+        cols = cols.astype(int) - 1
+
+        # Create an empyt 1-D array for the results
+        flattened_result = np.nan * da.zeros((LI_GRID_SHAPE[0] * LI_GRID_SHAPE[0]), dtype=data_array.dtype)
```
sjoro, line 101: this probably needs to be `(LI_GRID_SHAPE[0] * LI_GRID_SHAPE[1]`, right?

sjoro (changes requested):
only one comment to clarify, otherwise looks good to me! thanks for your efforts on this reader @seenno and @ameraner !

satpy/readers/li_l2_nc.py
seenno: Well spotted, fixed now.

ameraner:
So we added some new tests to improve coverage (now we're at 100% for both the reader files 🙂). Also most CodeScene comments were addressed, the few remaining ones complain about the tests setup file, that contains the mocked files structures... I don't think it makes sense to improve this further.  
So from our side this is ready now.

sjoro (approved):
LGTM!

seenno:
Thank you everyone who contributed, either as a reviewer or helping me to understand Satpy and its coding style :)

PR #2272, opened 2022-11-13
Add ability to export Scene's dependency tree to a graphviz dot format

djhoese:
This is a proof of concept for something that might be useful. The idea was to create a way to semi-easily convert a series of Satpy dependency relationships into a graphviz graph. Not sure how useful it is, but I wanted to try it. Here's a basic graph for the ABI `true_color`.

test_graph3

I think this could be cleaned up to treat modifiers differently and shorten the size of each bubble.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/dependency_tree.py
+++ b/satpy/dependency_tree.py
@@ -145,6 +145,21 @@ class Tree:
         """Render the dependency tree as a string."""
         return self._root.display()
 
+    def to_graphviz(self, *args, node_kwargs=None, **kwargs):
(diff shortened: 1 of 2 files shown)


djhoese:
@mraspaud I noticed this PR has been sitting around doing nothing. It might be a nice little utility, but if I remember right your comment on slack when I first made this was something along the lines of "what's the point?". Maybe I should just make it a gist? I admit it doesn't need to be a method on the node classes but might still be interesting to have. Thoughts?

PR #2274, opened 2022-11-14
Documentation correction: Not all built-in compositors derive from GenericCompositor

gerritholl:
Fix an incorrect statement in the documentation alleging that all compositors are derived from the GenericCompositor.  This isn't true. Also add a paragraph on why one might want to use SingleBandCompositor.

 - [x] Fully documented


pnuu (approved):
LGTM

PR #2275, opened 2022-11-14
Multiscene blend with weights

adybbroe:
This PR aims adding a way to blend several scenes (polar and geo cloud parameters is what we have in mind) using pixel weights. Weights could for instance simply be the satellite zenith angle (or rather `1/satz`) so that the pixel with the smallest satellite zenith angle is chosen from all scenes being blended. Weights could also be a comlex mixture of several parameters, like time, satellite zenith, retrieval uncertainty (if available) and so on.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented 

Example/pseudo code creating a multi scene of one geo (SEVIRI) and two polar (here AVHRR/3 from N18 and N19) remapping to a pre-defined area, and blending the NWCSAF cloudtype datasets by stacking with weighs. Here the weights are the inverse of the satellite-zenith angles:
```
    mscn = MultiScene([geo, polar1, polar2])
    groups = {DataQuery(name='CTY_group'): ['ct']}
    mscn.group(groups)

    resampled = mscn.resample(areaid, reduce_data=False)

    weights = [1./geo_satz_angles, 1./polar1_satz_angles, 1./polar2_satz_angles]
    blended = resampled.blend(blend_function=stack_weighted, weights=weights)

    # blended.show('CTY_group')
    blended.save_dataset('CTY_group', filename='./blended_stack_weighted_geo_polar_2polars.nc')
```


adybbroe:
@djhoese Just a question. The stack function retains the attributes from the first dataset, but wouldn't it more intuitive to store the attributes from the last one in the stack, the on on top? I realize that can be debated, so if you had a particular reason for keeping the first I am absolutely fine. Just need to be clear what we do when creating/adapting the unittests. Don't think this aspect was tested before...

simonrp84:
@adybbroe This old PR of mine might be relevant here: https://github.com/pytroll/satpy/pull/1157

The purpose of the PR was to compute the pixel size to enable blending of various datasets based on whichever had the smallest pixel size.

I chose that method as, for example, in some cases GOES/ABI has a larger zenith angle but smaller pixel size than SEVIRI as ABI has a higher native pixel resolution.

adybbroe:
> @adybbroe This old PR of mine might be relevant here: #1157
> 
> The purpose of the PR was to compute the pixel size to enable blending of various datasets based on whichever had the smallest pixel size.
> 
> I chose that method as, for example, in some cases GOES/ABI has a larger zenith angle but smaller pixel size than SEVIRI as ABI has a higher native pixel resolution.

Oh, that's interesting! Didn't quite thought of that, might be useful also when blending AVHRR, MODIS and VIIRS (and eventually MetImage) for instance. I will have a look, but that PR was never merged, so work is "half" finished right?

satpy/multiscene.py
```diff
@@ -364,7 +374,7 @@ def blend(self, blend_function=stack):
         common_datasets = self.shared_dataset_ids
         for ds_id in common_datasets:
             datasets = [scn[ds_id] for scn in self.scenes if ds_id in scn]
-            new_scn[ds_id] = blend_function(datasets)
+            new_scn[ds_id] = blend_function(datasets, **kwargs)
```
djhoese, line 377: This shouldn't be needed if `functools.partial` is used for the `blend_function`. Using `partial` you could create a callable object with the kwargs (or args) already applied I believe.

satpy/multiscene.py
```diff
@@ -45,6 +45,16 @@
 log = logging.getLogger(__name__)
 
 
+def weighted(datasets, weights=None):
```
djhoese, line 48: `weighted_stack`? Weighted what?

djhoese (changes requested):
A couple questions inline. Otherwise, regarding the attrs from the first DataArray being used...no probably no real reason that I can think of. I'm not sure I'm even the one who added that function originally. I could see there being a chance of performance differences in that I think the first Scene is always generated first (if Scenes were created via a generator), but I think once this blend function is called it is already a list (maybe?) so it might not matter. I'm open to changing it.

simonrp84:
` I will have a look, but that PR was never merged, so work is "half" finished right?`

It was pretty much finished in terms of functionality, but I didn't persue it as Dave said it might not be suitable for satpy and I no longer needed the PR itself for my project. It's something I'd be interested in picking up though!

adybbroe:
And here two more images, now adding another polar scene, so two polar and one geo. Here the satz angles for the 2nd polar scene (NOAA-19) and the difference between the weighted composite of three scenes and the previous (with only one geo and one polar):
noaa19_satz_angles
diff_stack_weighted_one_and_two_polars

One can see that the new polar scene in the north complements the previous polar adding where it has better viewing compared to previous.

satpy/multiscene.py
djhoese: Or maybe we only have one `stack` and weights is an optional kwarg?

djhoese:
I'm not sure I like the final stacked result of your first example better, but I see the usefulness. It is kind of amazing how much can be done just by throwing on some weights and/or masks.

adybbroe:
@djhoese `test_multiscene` is now a bit long, CodeScene health check complains it is more than 600 lines long. I am sure my additions could be improved, but think we might have to think of splitting it in two. When testing off line all the mp4 tests complaint though I didn't touch them, guess it is some 3rd library I am missing? I ended up running: `pytest -s satpy/tests/test_multiscene.py -k 'test_ and not mp4'`. Yes, I am yet/still not using Pycharm... ;-)
But could we move the mp4 and timeseries stuff to a separate test file perhaps?

satpy/multiscene.py
adybbroe: > I'm not sure I like the final stacked result of your first example better, but I see the usefulness. It is kind of amazing how much can be done just by throwing on some weights and/or masks.

Yes, it is amazing, I agree! Satpy is fantastic!
Concerning what is better or worse in this case stacking with our without weights, is of course not important. The thing is we want to achieve the best possible cloud analysis every hour over our area, and normally (this is an assumption but can also be documented using validation data) we would get better cloud mask, type and height from the polar orbiters if we have them. Especially in the north. The sat-zenith angle is a good proxy to use here. I intend to mix in also the observation time, so that observations far away from the nominal time gets lower weight. Also, I will want to weight VIIRS cloud products the highest, higher than AVHRR. So, if you have a VIIRS cloud product at 10:45 and an AVHRR cloud product at 11:15, the 12:00 analysis will have more of VIIRS than of AVHRR, maybe even a high zenith angle for VIIRS is better than a low one for AVHRR... You can go on make it more and more complicated/sophisticated.

adybbroe: > 

@djhoese I am new to `partial`. But was reading the documentation and trying to understand what you are after but couldn't make sense of it. Sorry, can you elaborate a bit further or show me what you had in mind and I can test it...?

satpy/tests/test_multiscene.py
```diff
@@ -552,49 +572,217 @@ def test_crop(self):
...
+class TestBlendFuncs():
     """Test individual functions used for blending."""
 
-    def setUp(self):
-        """Set up test data."""
-        from datetime import datetime
+    @pytest.fixture(autouse=True)
+    def _get_line_column(self):
+        self._line = 2
+        self._column = 3
```
djhoese, line 581: How about a `setup_method` method that assigns these to `self`? I'm a little scared of using an autouse fixture that may or may not apply to all tests run in satpy.

djhoese:
> When testing off line all the mp4 tests complaint though I didn't touch them, guess it is some 3rd library I am missing? 

If it is the ones talking about "distributed" then it is probably the dask `distributed` library. Otherwise it is `imageio` and `imageio-ffmpeg` that you are missing.

I am OK splitting the tests if you make it a `test_multiscene/` directory and put tests in `test_save_animation.py` and `test_blend.py` and `test_misc.py` or something like that.

satpy/multiscene.py
djhoese: As far as I can tell here you are passing `**kwargs` from the user. Nothing from the keyword arguments is coming "dynamically" from the data being loaded. So a user could do something like this:

```
In [1]: from functools import partial

In [2]: def my_func(a=1, b=2):
   ...:     print(a, b)
   ...:

In [3]: pfunc = partial(my_func, a=3)

In [4]: pfunc(b=3)
3 3

In [5]: pfunc(a=2)
2 2

In [6]: pfunc(a=1)
1 2
```

The point being, by using `partial` we can **attach** args and keyword arguments to any function and then pass it as a callback to the blend function. So users could do `stack_with_weights = partial(stack, weight=my_weights)` and then `mscn.blend(blend_function=stack_with_weights)`. This is "better" in the sense that we avoid accepting `**kwargs` in the `blend` method. At the very least, if we don't want users to use `partial` we should make this a `blend_kwargs=None` keyword argument that gets passed if provided so we don't have to use the wildcard `**kwargs`.

adybbroe: Thanks @djhoese I think I got it now. I have followed your suggestion.

satpy/multiscene.py
adybbroe: @djhoese I have now only one function `stack` accessible for the user. In the tests I show how to use it in `test_blend_two_scenes_using_stack_weighted` for instance. 
Right now it calls a private function `_stack_weighted` if a sequence of weights are provided. I kept the two functions separate as the logic is more simple when no weighting is needed, and then the "old" code in stack is much easier to follow.

adybbroe:
> > When testing off line all the mp4 tests complaint though I didn't touch them, guess it is some 3rd library I am missing?
> 
> If it is the ones talking about "distributed" then it is probably the dask `distributed` library. Otherwise it is `imageio` and `imageio-ffmpeg` that you are missing.
> 
> I am OK splitting the tests if you make it a `test_multiscene/` directory and put tests in `test_save_animation.py` and `test_blend.py` and `test_misc.py` or something like that.

I forgot this, should I try split the test functions as suggested. I can have a go, if that's all which you think is left?

satpy/tests/test_multiscene.py
adybbroe: Done

djhoese:
If you're bored go for splitting the tests (`test_multiscene/` -> `test_save_animation.py`, `test_blend.py`, and `test_multiscene.py` maybe?).

djhoese (changes requested):
The main logic here seems good. I had some questions and suggestions though. Lastly, I'm wondering if `multiscene_tests/__init__.py` functions should be put in a `helpers.py` or `utils.py` or something? I see you have `make_dataid`, but isn't that defined somewhere else in Satpy's test infrastructure? These other functions are to create fake data so maybe instead of `utils.py` it could be a more descriptive like `fake_data.py` or something like that? Could even be `conftest.py` and make them pytest fixtures if you want to get real crazy.

satpy/multiscene.py
```diff
@@ -45,14 +45,45 @@
 log = logging.getLogger(__name__)
 
 
-def stack(datasets):
-    """Overlay series of datasets on top of each other."""
+def stack(datasets, weights=None):
+    """Combine a series of datasets together.
```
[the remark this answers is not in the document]
adybbroe: Yeah, better, thx!

satpy/multiscene.py
```diff
@@ -45,14 +45,45 @@
...
+    if weights:
+        return _stack_weighted(datasets, weights)
+
     base = datasets[0].copy()
     for dataset in datasets[1:]:
-        base = base.where(dataset.isnull(), dataset)
+        try:
+            base = base.where(dataset == dataset._FillValue, dataset)
+        except AttributeError:
+            base = base.where(dataset.isnull(), dataset)
```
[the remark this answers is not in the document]
adybbroe: Hmm, I thought it should be fine. You mean the way we do here with a try-except, rather than more carefully checking/analysing what kind of data we are working with?

satpy/multiscene.py
```diff
@@ -45,14 +45,45 @@
...
+    # Go through weights and set to zero where corresponding datasets have a value equals _FillValue or nan
+    for i, dataset in enumerate(datasets):
+        try:
+            weights[i] = xr.where(dataset == dataset._FillValue, 0, weights[i])
+        except AttributeError:
+            weights[i] = xr.where(dataset.isnull(), 0, weights[i])
+
+    indices = da.argmax(da.dstack(weights), axis=-1)
+    dims = datasets[0].dims
+    attrs = datasets[0].attrs
```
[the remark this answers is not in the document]
adybbroe: Yes I think we briefly discussed it. It is not obvious to me what you should do to make it intuitive and right. Which is why I would prefer to keep it as simple as possible. It feels like it will depend on the use case. It maybe that some scenes are not contributing at all to the final product, though they have been part of the stacking/blending process. Sounds like we should perhaps make a function that does this blending of attributes, and that way at least it will be more explicit in the code for future adaptations/changes? Then I would suggest we keep it simple what that function then actually does for now. I would go for a special treatment of the scene start and end times, and take from either first or last for the rest of the attributes? What else, besides times need special care?

satpy/etc/readers/nwcsaf-pps_nc.yaml
```diff
@@ -120,25 +120,25 @@ datasets:
     name: ct
     file_type: nc_nwcsaf_ct
     coordinates: [lon, lat]
-    standard_name: cloudtype
+    standard_name: ct
```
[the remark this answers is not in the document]
adybbroe: I think it was because I wanted to distinguish between the composite name and the dataset name, but probably I am not mastering the use/need of the `standard_name`? I can keep the `cloudtype` but not sure of the impact.

satpy/multiscene.py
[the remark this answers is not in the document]
djhoese: Sorry, I should have been more clear. I meant checking the `_FillValue`. I'd be afraid that there is some reader somewhere producing floating point data with a `_FillValue` where that `_FillValue` came from the file attributes and is no longer valid (after the data was unscaled with `scale_factor` and `add_offset`). Probably not worth it too worry about it now but instead fix it when we find the issue.

satpy/multiscene.py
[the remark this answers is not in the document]
djhoese: https://github.com/pytroll/satpy/blob/9a496eb6d963b69e3ac50e11e4d7dd8f3ec37d3c/satpy/dataset/metadata.py#L30-L48

The function already exists ^. But it looks like it doesn't handle start/end time the way I thought it would. That special handling is apparently only in the internal reader code.

satpy/etc/readers/nwcsaf-pps_nc.yaml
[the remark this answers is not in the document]
djhoese: If there is no CF standard name then this name only matters for matching enhancements at this point. If you update this standard_name here then an enhancement configuration should be added to match this standard_name (or have one for this specific product name `ct`). Or you can just let it find the default enhancement.

satpy/multiscene.py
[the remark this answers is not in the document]
adybbroe: Nice!

Now I have changed the tests so that the blended scene has the start and end times set to the oldest `start_time` and the newest `end_time` of the scenes in the weighted stack. And the test pass with this code in the stack function:

```
    attrs = combine_metadata(*[x.attrs for x in datasets])

    start_time = datetime.now()
    for s_time in [x.attrs['start_time'] for x in datasets if 'start_time' in x.attrs]:
        if s_time < start_time:
            start_time = s_time

    end_time = datetime.fromtimestamp(0)
    for e_time in [x.attrs['end_time'] for x in datasets if 'end_time' in x.attrs]:
        if e_time > end_time:
            end_time = e_time

    attrs['start_time'] = start_time
    attrs['end_time'] = end_time
```

I should of course put that time calculation in a separate function. But I suppose you would like to have a `combine_metadata` function that also handle the times as expected?
Here I am not averaging but looking for the time span that describes all the data resulting in the blended scene.

Also, do you think I should handle the case when a scene does not have a start and an end time, and how?
I would tend to require in that case that the user provides scenes that have start/end times! Or should all scenes have that (requirement of all readers)?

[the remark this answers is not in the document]
adybbroe: @djhoese I feel like this is growing, so hope I can finish it soon and then we could add more complexity and more clever handling of attributes etc in other/future PRs when that becomes needed, does that sound reasonable?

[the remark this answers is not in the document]
djhoese: I am OK keeping this as the first Scene metadata for now. If you have the time I think `combine_metadata` would at least be more accurate. If you also have the time a future PR to combine_metadata to min/max start/end time would be appreciated. I know the conversation has come up in the past so hopefully I'm not forgetting some reason for why it isn't already that way.

adybbroe:
@djhoese I have done a bit around the attributes and some minor refactoring and documentation. I have a feeling you will have an opinion. Maybe my new function should be `get_combined_start_end_times` should be in the `metadata.py` module?
And actually rather be built into the `combine_metadata` function?
So, not sure how to finish this nicely? Should I drop it for now perhaps? But it does make sense to make the combination of start and end times as I do I think. I thought it was more intuitive than an average in this case?

Also, I think my section in the doc pages should be improved, but have to run for now.

Otherwise I perhaps only have left to settle for the `standard_name` as you mentioned above. Just need to make some system testing first.

djhoese:
> Otherwise I perhaps only have left to settle for the standard_name as you mentioned above. Just need to make some system testing first.

Are you saying you're not sure if the standard_name changes broke something?

doc/source/multiscene.rst
```diff
@@ -85,6 +85,49 @@ iteratively overlays the remaining datasets on top.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is also possible to blend scenes together in a bit more sophisticated manner
+using pixel based weighting instead of just stacking the scenes on top of each
+other as described above. This can for instance be useful to make a cloud
+parameter (cover, height, etc) composite combining cloud parameters derived
+from both geostationary and polar orbiting satellite data over aa given area
+(for instance at high latitudes where geostatioonary data degrade quickly with
+latitude and polar data are more frequent) scenes valid close in time to each
+other.
```
djhoese, line 99:
```suggestion
from both geostationary and polar orbiting satellite data close in time and
over a given area. This is useful for instance at high latitudes where
geostationary data degrade quickly with latitude and polar data are more
frequent.
```

doc/source/multiscene.rst
```diff
@@ -85,6 +85,49 @@ iteratively overlays the remaining datasets on top.
...
+
+This weighted blending can be accomplished via the use of the builtin
+:func:`~functools.partial` function (see `Partial
+<https://docs.python.org/3/library/functools.html#partial-objects>`_) and the
+default :func:`~satpy.multiscene.stack` function. The
+:func:`~satpy.multiscene.stack` function can take the optional argument
+`weights` (`None` on default) which should be a sequence (of length equal to
+the number of scenes being blended) of arrays with pixel weights.
+
+The code below gives an example of how two clouud scenes can be blended using
```
djhoese, line 109:
```suggestion
The code below gives an example of how two cloud scenes can be blended using
```

satpy/multiscene.py
```diff
@@ -45,14 +46,70 @@
...
+    for i, dataset in enumerate(datasets):
+        try:
+            weights[i] = xr.where(dataset == dataset.attrs["_FillValue"], 0, weights[i])
+        except KeyError:
+            weights[i] = xr.where(dataset.isnull(), 0, weights[i])
+
+    return weights
+
+
+def get_combined_start_end_times(*metadata_objects):
```
djhoese, line 98: I'd say make this private with a `_`.

satpy/multiscene.py
```diff
@@ -45,14 +46,70 @@
...
+    """Get the start and end times attributes valid for the entire dataset series."""
+    start_time = datetime.now()
+    for md_obj in metadata_objects:
+        if md_obj['start_time'] < start_time:
+            start_time = md_obj['start_time']
+
+    end_time = datetime.fromtimestamp(0)
+    for md_obj in metadata_objects:
+        if md_obj['end_time'] > end_time:
+            end_time = md_obj['end_time']
```
djhoese, line 108: Based on CodeScene's comments maybe this should be one for loop or maybe put these in two separate sub-functions. I think performance-wise a single for loop might be smarter at the cost of "cleaner" code.

djhoese (changes requested):
Just some small changes. Otherwise this looks good assuming the standard_name thing isn't an issue.

satpy/etc/readers/nwcsaf-pps_nc.yaml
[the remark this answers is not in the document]
adybbroe: I retverted back using `cloudtype`.

satpy/multiscene.py
adybbroe: Yeah I fixed thx!

satpy/multiscene.py
adybbroe: Done

adybbroe:
@djhoese Perhaps you saw I reverted back on the `standard_name`? So, unless you have suggestions on the documentation, I think we could be ready to merge this?
Don't know if someone else wants to have a look?

djhoese:
Merge it and let's see what breaks. I'll let you click the button.

PR #2279, opened 2022-11-16
Add suppression of AttributeError when trying to close the netcdf file.

### LINKED ISSUE #2269: Occasional errors when closing FCI L2 NetCDF file
**Describe the bug**
The FCI L2 netcdf reader occasionally fails because it does not sufficiently suppress relevant errors when trying to close the file during routine operations/reading.

**Expected behavior**
The reader will close the file if still open.

**Actual results**
Occasionally the reader tries to close a file that is already closed, leading to an error.

**Additional context**
To be fixed by @strandgren during PCW in Athens

strandgren:
This PR adds the suppression of `AttributeError` when trying to close the netcdf file in the FCI L2 NetCDF reader, in case the file was never opened. The lack of this suppression has lead to occasional failures of the reader during routine reading/monitoring of FCI L2 test data.

 - [x] Closes #2269

### DIFF
--- a/satpy/readers/fci_l2_nc.py
+++ b/satpy/readers/fci_l2_nc.py
@@ -113,7 +113,7 @@ class FciL2CommonFunctions(object):
 
     def __del__(self):
         """Close the NetCDF file that may still be open."""
-        with suppress(OSError):
+        with suppress(AttributeError, OSError):
             self.nc.close()
 
 


djhoese (approved):
This doesn't have to be done in this PR, but we should maybe move to adding `.close` methods on file handlers that do this equivalent behavior (try to close the file and suppress any errors). There is a feature request somewhere for a close-like method on the Scene that would attempt to close all of its connected readers.

pnuu (approved):
To me looks like the `self.nc` is set to return value of `xr.open_dataset()`, so maybe the root cause is `open_dataset()` failing and returning `None` or something?

Anyway, this is what the other readers have too, so LGTM.

pnuu:
pre-commit.ci autofix

PR #2287, opened 2022-11-18
Feature: add support for HRFI imagery in the FCI L1c reader

### LINKED ISSUE #2273: Add support for HRFI imagery in the FCI L1c reader
## Feature Request

The current reader is able to read only FDHSI (normal resolution) data. It would be great to expand it to read also high-resolution HRFI data, which will be disseminated via EUMETCast Terrestrial and the Data Store. 

Expand the `fci_l1c_nc` reader to accept HRFI data.

No changes, just expanded capabilities.

ameraner:
This PR adds support for high-resolution full-disc HRFI imagery in the FCI L1c reader.

It was decided to keep the same channel names as in the FDHSI case, ignoring the added `_hr` inside the HRFI file channel names. This has the advantage that HRFI and FDHSI files can be used interchangeably: when HRFI data is provided together with FDHSI files, Satpy will automatically select the finest resolution version of a channel (similarly e.g. to the MODIS L1b reader). If needed, a user can manually define the wished resolution by defining the resolution kwarg in the dataset load call (e.g. `scn.load(['vis_06'], resolution=1000)`)

The PR also refactors the FCI L1c and GEOSegmentVariableYAMLReader tests, to improve the usage of fixtures, to have clearer class names and to be able to use parametrised tests for double FDHSI and HRFI checks.

 - [x] Closes #2273 
 - [x] Tests added 
 - [x] Fully documented 


satpy/readers/fci_l1c_nc.py
```diff
@@ -204,21 +204,31 @@ def end_time(self):
...
+    def get_channel_measured_group_path(self, channel):
+        """Get the channel's measured group path."""
+        if self.filetype_info['file_type'] == 'fci_l1c_hrfi':
+            channel += '_hr'
+        measured_group_path = 'data/{}/measured'.format(channel)
+
+        return measured_group_path
+
     def get_segment_position_info(self):
         """Get the vertical position and size information of the chunk (aka segment) for both 1km and 2km grids.
```
gerritholl, line 216: Should this be extended for 500m in addition to 1km and 2km grids?

Also: I know this was not added by this PR, but what is the "vertical position"?  That's something sounders would have.  I suppose here we mean "latitude"?

satpy/tests/reader_tests/test_fci_l1c_nc.py
```diff
@@ -282,236 +292,220 @@ def _get_reader_with_filehandlers(filenames, reader_configs):
...
+                 channels["solar"] + channels["terran"]], pad_data=False)
+            assert expected_res_n == len(res)
+            for ch in channels["solar"] + channels["terran"]:
+                assert res[ch].shape == (200, 11136)
+                assert res[ch].dtype == np.float64
+                assert res[ch].attrs["calibration"] == "radiance"
+                assert res[ch].attrs["units"] == 'mW m-2 sr-1 (cm-1)-1'
+                assert res[ch].attrs["radiance_unit_conversion_coefficient"] == 1234.56
+                if ch == 'ir_38':
+                    numpy.testing.assert_array_equal(res[ch][~0], 15)
```
gerritholl, line 393: Why `~0`?  That's just `-1`?

satpy/tests/reader_tests/test_fci_l1c_nc.py
```diff
@@ -282,236 +292,220 @@ def _get_reader_with_filehandlers(filenames, reader_configs):
...
+                assert caplog.text == ""
+            assert expected_res_n == len(res)
+            for ch in channels["terran"]:
+                assert res[ch].shape == (200, 11136)
+                assert res[ch].dtype == np.float64
+                assert res[ch].attrs["calibration"] == "brightness_temperature"
+                assert res[ch].attrs["units"] == "K"
+
+                if ch == 'ir_38':
+                    numpy.testing.assert_array_almost_equal(res[ch][~0], 209.68274099)
```
gerritholl, line 442: Same here with `~0`, see above.

satpy/tests/reader_tests/test_fci_l1c_nc.py
```diff
@@ -282,236 +292,220 @@ def _get_reader_with_filehandlers(filenames, reader_configs):
...
-
-    def test_load_radiance(self, reader_configs):
+        with mocked_basefilehandler(filehandler):
+            reader = _get_reader_with_filehandlers(filenames, reader_configs)
+            res = reader.load(
+                [make_dataid(name=name, calibration="counts") for name in
+                 channels["solar"] + channels["terran"]], pad_data=False)
+            assert expected_res_n == len(res)
+            for ch in channels["solar"] + channels["terran"]:
+                assert res[ch].shape == (200, 11136)
```
gerritholl, line 362: Would this be 22272 for the solar HR channels?

gerritholl (changes requested):
Thanks for the good work.  I left some inline comments.

I think we also need another area in areas.yaml, `mtg_fci_fdss_500m`.  Or is this part of another PR?

satpy/readers/fci_l1c_nc.py
ameraner: Good point yes, thanks. 
By vertical position I mean the position of the chunk (segment) inside the full-disc in the vertical image direction, i.e. the number of the start and end row. I can clarify this in the docstring.

gerritholl: I know that's what you mean, and I know that in 2-dimensional Cartesian coordinates the x-coordinates is sometimes called horizontal and the y-coordinate is sometimes called vertical, but in an Earth observation context I find it not the best choice of words, because of the competing definition of vertical to mean "along the local gravity direction".  The satellite height is a vertical coordinate.

satpy/tests/reader_tests/test_fci_l1c_nc.py
ameraner: Maybe I thought ~0 is cooler than -1 :) No idea to be honest, it's been a while. Changed now.

satpy/tests/reader_tests/test_fci_l1c_nc.py
ameraner: Changed now, see above.

satpy/tests/reader_tests/test_fci_l1c_nc.py
ameraner: The current tests are assuming the same shape and geolocation parameters for both FDHSI and HRFI, since the computation is done exactly the same. But I agree that it would make the tests a bit nicer.. I'll look into it.

satpy/readers/fci_l1c_nc.py
pnuu: Maybe `y_position` then?

satpy/readers/fci_l1c_nc.py
```diff
@@ -204,21 +204,31 @@ def end_time(self):
...
                     'segment_width': 11136},
-            '2km': {'start_position_row': self['data/ir_105/measured/start_position_row'].item(),
-                    'end_position_row': self['data/ir_105/measured/end_position_row'].item(),
-                    'segment_height': self['data/ir_105/measured/end_position_row'].item() -
-                    self['data/ir_105/measured/start_position_row'].item() + 1,
+            '2km': {'start_position_row': self[ir_105_measured_path+'/start_position_row'].item(),
+                    'end_position_row': self[ir_105_measured_path+'/end_position_row'].item(),
+                    'segment_height': self[ir_105_measured_path+'/end_position_row'].item() -
+                    self[ir_105_measured_path+'/start_position_row'].item() + 1,
                     'segment_width': 5568}
```
pnuu, line 232: Hmm. You might need some if/elif/else here to add only the resolutions available in the file. For FDHSI there are 1 km and 2 km data, for HRFI 500 m and 1 km data (I guess HRFI doesn't have any of the low res channels in them?).

satpy/tests/reader_tests/test_fci_l1c_nc.py
ameraner: Updated now to use realistic resolutions and shapes for each channel and data type input

satpy/readers/fci_l1c_nc.py
ameraner: Updated now, I tried to clarify the concept in the docstring a bit more, see https://github.com/pytroll/satpy/pull/2287/files#diff-5e6ace3bb73933d0b7fb85a692590ac424b905677f665f84a6e73f2dc3cf7097R224-R229

ameraner:
@gerritholl thanks for your review, I also added the new AreaDefinition for the 500m grid to the areas.yaml already in this PR now

ameraner:
In the process of handling the different grid types coming from the FCI L1c reader depending on the file type, I also now optimised/refactored a bit the handling of the missing segments computation in the `GEOVariableSegmentYAMLReader`. In particular, the collection of the position of the available segments is now postponed to the `get_dataset` call, instead of being done early during the `Scene` initialisation. As this was the only part of code that was requiring information from the file during the `Scene` init, this opens the door now for a new optimisation PR that avoids the file opening alltogether in the `FCIL1cNCFileHandler`/`NetCDF4FileHandler` init, that would significantly reduce the time needed to initialise a `Scene`.

satpy/readers/fci_l1c_nc.py
```diff
@@ -204,22 +220,42 @@ def end_time(self):
...
+
+        return measured_group_path
+
     def get_segment_position_info(self):
-        """Get the vertical position and size information of the chunk (aka segment) for both 1km and 2km grids.
+        """Get information about the size and the position of the chunk inside the final image array.
 
-        This is used in the GEOVariableSegmentYAMLReader to compute optimal chunk sizes for missing chunks.
+        As the final array is composed by stacking chunks (aka segments) vertically, the position of a chunk
+        inside the array is defined by the numbers of the start (lowest) and end (highest) row of the chunk.
```
pnuu, line 235: Should we use "segment" all over to reduce confusion between Dask chunks? The term "segment" is also familiar from SEVIRI, so using the same terms would be nice.

satpy/readers/fci_l1c_nc.py
```diff
@@ -204,22 +220,42 @@ def end_time(self):
...
+        The row numbering is assumed to start with 1.
+        This info is used in the GEOVariableSegmentYAMLReader to compute optimal chunk sizes for missing chunks.
```
pnuu, line 237: Here I'm not anymore sure if we are talking about optimal Dask chunk size or segments :joy: 
```suggestion
        This info is used in the GEOVariableSegmentYAMLReader to compute optimal sizes for missing segments.
```

satpy/_compat.py
```diff
@@ -88,3 +88,9 @@ def __get__(self, instance, owner=None):  # noqa
     # numpy <1.20
     from numpy import dtype as DTypeLike  # noqa
     from numpy import ndarray as ArrayLike  # noqa
+
+
+try:
+    from functools import cache  # type: ignore
+except ImportError:
+    from functools import lru_cache as cache  # noqa
```
pnuu, line 96: Maybe these could be removed, they are a leftover from my cache testings I think.

pnuu (approved):
Just few comments/suggestions, otherwise LGTM.

satpy/readers/fci_l1c_nc.py
ameraner: Agreed, I swapped all "chunk" for "segment" in  d73bdde

satpy/readers/fci_l1c_nc.py
ameraner: Fixed, see  d73bdde

satpy/_compat.py
ameraner: I'm actually using the `self.method = cache(self._method)` trick around here https://github.com/pytroll/satpy/blob/d73bdde89d5ba21a5ef9ceaf75aa97650e47783c/satpy/readers/yaml_reader.py#L1383, see also https://github.com/pytroll/satpy/pull/2287/commits/132f9a32e21323b4dcef8c399c25cdbfdd6428c3, so we should keep it.

pnuu: Oooh, yes! I missed that.

PR #2291, opened 2022-11-20
Fix native resampler not working for some chunk sizes

### LINKED ISSUE #2245: `native` resampler fails for some chunk sizes
**Describe the bug**
When resampling Himawari full disk data with the `native` resampler while using a user-defined chunk size the resampling operation fails. Thisis because the aggregation factor of the native resampler is not being a factor of the chunk size.

**To Reproduce**

```python

import os
os.environ['PYTROLL_CHUNK_SIZE'] = '550'

from satpy import Scene
from glob import glob

scn = Scene(glob('/my_himawari_data/*.DAT'), reader='ahi_hsd')
scn.load(['true_color_nocorr', 'B13'])

scn2 = scn.resample(scn['B13'].attrs['area'], resampler='native')

```

**Expected behavior**
The scene is resampled onto the B13 area.

**Actual results**
An error:

```
  File "/home/users/srproud/Software/satpy/satpy/scene.py", line 955, in resample
    self._resampled_scene(new_scn, destination, resampler=resampler,
  File "/home/users/srproud/Software/satpy/satpy/scene.py", line 871, in _resampled_scene
    res = resample_dataset(dataset, destination_area, **kwargs)
  File "/home/users/srproud/Software/satpy/satpy/resample.py", line 1418, in resample_dataset
    new_data = resample(source_area, dataset, destination_area, fill_value=fill_value, **kwargs)
  File "/home/users/srproud/Software/satpy/satpy/resample.py", line 1381, in resample
    res = resampler_instance.resample(data, **kwargs)
  File "/home/users/srproud/Software/satpy/satpy/resample.py", line 981, in resample
    return super(NativeResampler, self).resample(data,
  File "/home/users/srproud/Software/satpy/satpy/resample.py", line 439, in resample
    return self.compute(data, cache_id=cache_id, **kwargs)
  File "/home/users/srproud/Software/satpy/satpy/resample.py", line 1084, in compute
    d_arr = self._expand_reduce(data.data, repeats)
  File "/home/users/srproud/Software/satpy/satpy/resample.py", line 1051, in _expand_reduce
    return cls._aggregate(d_arr, y_size, x_size)
  File "/home/users/srproud/Software/satpy/satpy/resample.py", line 999, in _aggregate
    raise ValueError("Aggregation requires arrays with "
ValueError: Aggregation requires arrays with shapes and chunks divisible by the factor
```

This is because the aggregation factor for the B03 channel (4) is not a factor of the chunk size (550). The problem can be resolved by the user if they change the chunk size, but ideally satpy should be able to sort this out for itself.

djhoese:
This PR fixes the issue described in #2245 where a user's chosen chunk size for their data may not be divisible by the aggregation factor, but the overall shape of the array is. There is no reason except avoiding rechunking that this shouldn't work. This PR makes it so rechunking is done if needed.

I also removed the unnecessary rechunking done in the replication step. My hope is that dask will figure out this rechunking automatically and that this won't be needed.

@simonrp84 any testing you can do would be greatly appreciated.

 - [x] Closes #2245 
 - [x] Tests added

### DIFF
--- a/satpy/resample.py
+++ b/satpy/resample.py
@@ -140,6 +140,7 @@ defined.
 """
 import hashlib
 import json
+import math
 import os
 import warnings
 from logging import getLogger
@@ -154,6 +155,8 @@ import zarr
 from packaging import version
 from pyresample.ewa import fornav, ll2cr
 from pyresample.geometry import SwathDefinition
+
+from satpy.utils import PerformanceWarning
 
 try:
     from pyresample.resampler import BaseResampler as PRBaseResampler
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM, thanks for also migrating to pytest!

PR #2304, opened 2022-11-30
Add additional OLCI L2 datasets.

simonrp84:
I noticed that satpy doesn't support all the OLCI L2 datasets. This PR edits the YAML file for `olci_l2` to add some of those that're missing:
 - Green isntantaneous fAPAR
 - Integrated water vapor
 - Terrestrial chlorophyll index
 
Plus their various uncertainties and ancillary datasets.


mraspaud (approved):
LGTM, thanks for adding these!

PR #2305, opened 2022-12-02
Remove file reading for FCI L1c

pnuu:
This PR supersedes #2182 in bringing remote file reading for FCI L1c data:

```python
fnames = ["simplecache::s3://satellite-data-fci-test-data-2022/*DEV_20170920112*.nc"]
scn = Scene(reader='fci_l1c_nc', filenames=fnames)
scn.load(['ir_123'])
scn.save_datasets()
```

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/fci_l1c_nc.py
+++ b/satpy/readers/fci_l1c_nc.py
@@ -361,10 +361,11 @@ class FCIL1cNCFileHandler(NetCDF4FsspecFileHandler):
         actual_subsat_lon = float(np.nanmean(self._get_aux_data_lut_vector('subsatellite_longitude')))
         actual_subsat_lat = float(np.nanmean(self._get_aux_data_lut_vector('subsatellite_latitude')))
         actual_sat_alt = float(np.nanmean(self._get_aux_data_lut_vector('platform_altitude')))
-        mtg_geos_proj = self.get_and_cache_npxr("data/mtg_geos_projection")
-        nominal_and_proj_subsat_lon = float(mtg_geos_proj.longitude_of_projection_origin)
+        nominal_and_proj_subsat_lon = float(
+            self.get_and_cache_npxr("data/mtg_geos_projection/attr/longitude_of_projection_origin"))
         nominal_and_proj_subsat_lat = 0
-        nominal_and_proj_sat_alt = float(mtg_geos_proj.perspective_point_height)
+        nominal_and_proj_sat_alt = float(
+            self.get_and_cache_npxr("data/mtg_geos_projection/attr/perspective_point_height"))
 
         orb_param_dict = {
             'orbital_parameters': {
@@ -447,8 +448,7 @@ class FCIL1cNCFileHandler(NetCDF4FsspecFileHandler):
         logger.debug('Row/Cols: {} / {}'.format(nlines, ncols))
 
         # Calculate full globe line extent
-        mtg_geos_proj = self.get_and_cache_npxr("data/mtg_geos_projection")
-        h = float(mtg_geos_proj.perspective_point_height)
+        h = float(self.get_and_cache_npxr("data/mtg_geos_projection/attr/perspective_point_height"))
 
         extents = {}
         for coord in "xy":
@@ -515,12 +515,11 @@ class FCIL1cNCFileHandler(NetCDF4FsspecFileHandler):
         if key['resolution'] in self._cache:
             return self._cache[key['resolution']]
 
-        mtg_geos_proj = self.get_and_cache_npxr("data/mtg_geos_projection")
-        a = float(mtg_geos_proj.semi_major_axis)
-        h = float(mtg_geos_proj.perspective_point_height)
-        rf = float(mtg_geos_proj.inverse_flattening)
-        lon_0 = float(mtg_geos_proj.longitude_of_projection_origin)
-        sweep = str(mtg_geos_proj.sweep_angle_axis)
+        a = float(self.get_and_cache_npxr("data/mtg_geos_projection/attr/semi_major_axis"))
+        h = float(self.get_and_cache_npxr("data/mtg_geos_projection/attr/perspective_point_height"))
+        rf = float(self.get_and_cache_npxr("data/mtg_geos_projection/attr/inverse_flattening"))
+        lon_0 = float(self.get_and_cache_npxr("data/mtg_geos_projection/attr/longitude_of_projection_origin"))
+        sweep = str(self.get_and_cache_npxr("data/mtg_geos_projection/attr/sweep_angle_axis"))
 
         area_extent, nlines, ncols = self.calc_area_extent(key)
         logger.debug('Calculated area extent: {}'
--- a/satpy/readers/fci_l1c_nc.py
+++ b/satpy/readers/fci_l1c_nc.py
@@ -114,6 +114,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import logging
 from functools import cached_property
 
+import dask.array as da
 import numpy as np
 import xarray as xr
 from netCDF4 import default_fillvals
@@ -122,7 +123,7 @@ from pyresample import geometry
 from satpy.readers._geos_area import get_geos_area_naming
 from satpy.readers.eum_base import get_service_mode
 
-from .netcdf_utils import NetCDF4FileHandler
+from .netcdf_utils import NetCDF4FsspecFileHandler
 
 logger = logging.getLogger(__name__)
 
@@ -173,7 +174,7 @@ def _get_channel_name_from_dsname(dsname):
     return channel_name
 
 
-class FCIL1cNCFileHandler(NetCDF4FileHandler):
+class FCIL1cNCFileHandler(NetCDF4FsspecFileHandler):
     """Class implementing the MTG FCI L1c Filehandler.
 
     This class implements the Meteosat Third Generation (MTG) Flexible
@@ -290,8 +291,11 @@ class FCIL1cNCFileHandler(NetCDF4FileHandler):
         measured = self.get_channel_measured_group_path(key['name'])
         data = self[measured + "/effective_radiance"]
 
-        attrs = data.attrs.copy()
+        attrs = dict(data.attrs).copy()
         info = info.copy()
+        if not isinstance(data, xr.DataArray):
+            data = xr.DataArray(
+                da.from_array(data), dims=data.dimensions, attrs=attrs, name=data.name)
 
         fv = attrs.pop(
             "FillValue",
@@ -396,7 +400,10 @@ class FCIL1cNCFileHandler(NetCDF4FileHandler):
     def _get_aux_data_lut_vector(self, aux_data_name):
         """Load the lut vector of an auxiliary variable."""
         lut = self.get_and_cache_npxr(AUX_DATA[aux_data_name])
-
+        if not isinstance(lut, xr.DataArray):
+            attrs = dict(lut.attrs.items()).copy()
+            lut = xr.DataArray(
+                da.from_array(lut), dims=lut.dimensions, attrs=attrs, name=lut.name)
         fv = default_fillvals.get(lut.dtype.str[1:], np.nan)
         lut = lut.where(lut != fv)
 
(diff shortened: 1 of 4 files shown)


pnuu:
Initial timing tests using local files and test script from https://github.com/pytroll/satpy/pull/2237#issuecomment-1282332336:

main:
```
Done. It took 3.73s  to start Scene.
Done. It took 2.63s to load datasets.
Done. It took 11.24s  to save datasets.
Done. It took 17.59s  for all ops.
```

this PR:
```
Done. It took 27.46s  to start Scene.
Done. It took 5.02s  to load datasets.
Done. It took 11.81s  to save datasets.
Done. It took 44.29s  for all ops.
```

I'll run some tests with data on S3 next week.

pnuu:
The test failure is some kind of race condition, which I saw also locally. When the failing tests are individually, the pass. Will investigate some time next week.

ameraner:
Do you have a plan for switching (automagically?) between `NetCDF4FsspecFileHandler` and `NetCDF4FileHandler`, depending if the data is remote or not, to avoid the performance penalty with local data?

pnuu:
Oh yeah, thanksbfor reminding! I think I have that already somewhere 🤔

pnuu:
This is one way I found for the automatic backend selection: https://github.com/pnuu/satpy/blob/feature_optimise_fcil1c_only_necessary_h5/satpy/readers/netcdf_utils.py#L363

pnuu:
I now transferred the code I linked to this PR. Will need unit tests and testing with real data. I hope to have time for those on Wednesday. I might have a look at the currently failing tests even before that.

pnuu:
The file handler caches were not cleared, so the values from previous tests were re-used in the later tests. Well, at least the caching works :joy:

pnuu:
Tested with data on internal CEPH/S3, and the processing works. Due to network delay, the processing time isn't great. Looking at the graph, the downloading part was 72 s of 94 s full processing time. I'll see if I can test again on a server closer to the bucket. Here I just created single `airmass` composite and resampled to 500 m EPSG:3035 grid (7836 x 7736 pixels).

Screenshot 2022-12-07 at 16-34-23 Bokeh Plot

pnuu:
Damn, tested the updated code with local files, and it fails in platform name handling. I'll have a look tomorrow (I hope).

pnuu:
Ok, the problem is that `self['attr/platform']` is used to get the translation from short platform name to full platform name, and the the value of it is this dictionary:

```python
ipdb> print(self['attr/platform'])
{'Conventions': 'CF-1.7', 'title': 'FCI Level 1C Rectified Radiance Product', 'summary': 'Body chunk of the MTG FCI Level 1c dataset for the FDHSI subset of spectral channels', 'keywords': 'MTG FCI FDHSI level_1c', 'keywords_vocabulary': 'original generated file', 'history': 'Original generated version', 'institution': 'EUMETSAT', 'location_indicator': 'XX-EUMETSAT-Darmstadt', 'data_designator': 'IMG+SAT', 'platform': 'MTI1', 'data_source': 'FCI', 'processing_level': '1C', 'type': 'RRAD', 'subtype': 'FDHSI', 'component1': 'CHK', 'component3': 'null', 'product_id': 'FCI1CRRADFDHSI', 'baseline_version': 'n.x.y', 'release_version': 'original', 'processor_version': '<configured_value>', 'algorithm_version': '<configured_value>', 'processing_mode': 'N', 'subsetting': 'null', 'source': 'IDPFI', 'runtime_data': '<runtime_value>', 'parent_data': '<runtime_value>', 'linked_data': '<runtime_value>', 'references': 'www.eumetsat.int', 'comment': 'None', 'group_tag': 'MTI1_FCI_1C_RRAD_FDHSI_YYY_DDD_NNNN_original', 'instrument_configuration_id': '<runtime_value>', 'instrument_configuration_id_version': '<runtime_value>', 'subsettable_groups': 'vis_04 vis_05 vis_06 vis_08 vis_09 nir_13 nir_16 nir_22 ir_38 wv_63 wv_73 ir_87 ir_97 ir_105 ir_123 ir_133', 'subsettable_groups_present': 'vis_04 vis_05 vis_06 vis_08 vis_09 nir_13 nir_16 nir_22 ir_38 wv_63 wv_73 ir_87 ir_97 ir_105 ir_123 ir_133', 'mtg_name': '<runtime_value>', 'alternative_name': 'null', 'format': 'NC4E', 'id': '<configured_value>', 'naming_authority': '<configured_value>', 'creator_type': 'institution', 'creator_institution': 'EUMETSAT', 'creator_url': 'www.eumetsat.int', 'license': 'null', 'standard_name_vocabulary': 'CF Standard Name Table v27', 'project': 'MTG', 'time_coverage_duration': 'null', 'time_coverage_resolution': 'null', 'cdm_datatype': 'null', 'coverage': 'FD', 'component2': 'BODY', 'format_version': '4c', 'time_coverage_start': '20170920120008', 'time_coverage_end': '20170920120015', 'disposition_mode': 'T', 'facility_or_tool': 'GTT', 'environment': 'DEV', 'date_created': '20170920120515', 'repeat_cycle_in_day': '0073', 'processed_count_in_repeat_cycle': '0001', 'count_in_repeat_cycle': '0001', 'purpose': array([], dtype=float64), 'date_time_position': '20170920120000', 'time_position': '120000', 'geospatial_lat_min': -81.19571706772024, 'geospatial_lat_max': -69.26566506254072, 'geospatial_lon_min': -52.67815684599577, 'geospatial_lon_max': 52.67815684599568, 'creator_name': 'EUMETSAT', 'creator_email': 'ops@eumetsat.int', 'special_compression': 'JLS'}
```

I'll try to hunt the origin of this change. Tomorrow. Hopefully.

pnuu:
It was a stupid mistake, using `super()._get_object_attrs(obj)` instead of `super()._get_attr(obj, key)` when not using `h5netcdf` backend.

pnuu:
I don't think I can add more meaningful tests without replacing all the fake file handlers with actual files that are written by the tests. And that is such a big task that I'm not willing to do that within this PR.

pnuu:
Here are the tests running on a server. Only two cores, `airmass` resampled to the 7836 x 7736 pixel area.

Using files NAS mounted as local file system (so using `netCDF4` backend):
Screenshot 2022-12-08 at 12-35-44 Bokeh Plot

Reading the files from CEPH/S3 (same network speed as for the NAS):
Screenshot 2022-12-08 at 12-39-24 Bokeh Plot

The bit until 100 seconds includes the file transfer and (slow) file handler creation. Using `h5netcdf` backend for local files (commenting out the try/except in `NetCDF4FsspecFileHandler._get_file_handle()`) shows that the CEPH/S3 reading doesn't slow down the reading/file handling at all:
Screenshot 2022-12-08 at 12-57-36 Bokeh Plot

pnuu:
After merging `main` also the unstable tests now pass.

The remaining question is [this](https://github.com/pytroll/satpy/pull/2305#issuecomment-1342297927): should the tests write a file that mimics the relevant parts of the file? If yes, that should be done in a separate PR as it's a rather large task.

mraspaud:
> should the tests write a file that mimics the relevant parts of the file?

Imo, for the tests to be comprehensive while still readable, nothing beats having a stub file on temp storage.
Agreed that this should be in a different PR though.

mraspaud (approved):
Thanks for this, it looks good to me.
The way the fsspec-based file handler is done feels a bit hacked, but unfortunately the way the reader works doesn't allow polymorphism here, so nothing to do here.
Just one small comment, no big deal, as I said, LGTM.

satpy/readers/fci_l1c_nc.py
```diff
@@ -396,7 +401,10 @@ def _get_dataset_index_map(self, dsname):
     def _get_aux_data_lut_vector(self, aux_data_name):
         """Load the lut vector of an auxiliary variable."""
         lut = self.get_and_cache_npxr(AUX_DATA[aux_data_name])
-
+        if not isinstance(lut, xr.DataArray):
+            attrs = dict(lut.attrs.items()).copy()
+            lut = xr.DataArray(
+                da.from_array(lut), dims=lut.dimensions, attrs=attrs, name=lut.name)
```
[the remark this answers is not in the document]
pnuu: Refactored in 561fbcc98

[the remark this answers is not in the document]
mraspaud: thanks!

satpy/readers/fci_l1c_nc.py
```diff
@@ -652,3 +647,10 @@ def calibrate_rad_to_refl(self, radiance, key):
 
         res = 100 * radiance * np.pi * sun_earth_distance ** 2 / cesi
         return res
+
+
+def _ensure_dataarray(arr):
+    if not isinstance(arr, xr.DataArray):
+        attrs = dict(arr.attrs.items()).copy()
```
djhoese, line 654: Is the `.items()` and `dict` needed to convert this to a regular dictionary? Does doing `dict(arr.attrs).copy()` not work? Or `arr.attrs.copy()`?

satpy/readers/netcdf_utils.py
```diff
@@ -349,3 +377,56 @@ def _compose_replacement_names(variable_name_replacements, var, variable_names):
...
+    def _get_file_handle(self):
+        try:
+            # Default to using NetCDF4 backend for local files
+            return super()._get_file_handle()
+        except OSError:
+            # The netCDF4 lib raises either FileNotFoundError or OSError for remote files. OSError catches both.
+            import h5netcdf
+            f_obj = open_file_or_filename(self.filename)
+            self._use_h5netcdf = True
+            return h5netcdf.File(f_obj, 'r')
```
djhoese, line 394: This seems prone to failure or unexpected results. I've recently discovered that depending on the build of NetCDF-C, netcdf can handle different types of files and URLs. It also sometimes depends on if `#mode=bytes` is appended to the URL. So what I'm saying is maybe there are sometimes when the NetCDF C library will be able to handle remote files and this file handler won't be used. Are we OK with that? Should we be more explicit with `if self.filename.startswith("s3://")` or something?

Is there anything special being done in this new file handler that should also be in the parent file handler?

@mraspaud said:

> The way the fsspec-based file handler is done feels a bit hacked, but unfortunately the way the reader works doesn't allow polymorphism here, so nothing to do here.

I'm not sure I understand. Is there a way to put this try/except and basically all off this logic in the main parent class?

mraspaud: > @mraspaud said:
> 
> > The way the fsspec-based file handler is done feels a bit hacked, but unfortunately the way the reader works doesn't allow polymorphism here, so nothing to do here.
> 
> I'm not sure I understand. Is there a way to put this try/except and basically all off this logic in the main parent class?

No, that's the problem. I would have liked satpy internals to be able to choose between two classes depending eg on the uri, but the way it works now we can't. If we had an interface like xarray's open_dataset (or open_datatree) for that matter, that function would be able to instantiate the right class, but now we hard code the filehandler class in the yaml file.

djhoese: This may be more work than this PR can support right now, but can we:

Split the base class and this new class into 3 total classes. Essentially a "netcdf4 handler" and an "h5netcdf file handler" and then a single class that uses both of these classes as "mixins"? Or maybe just sub-objects? The creation of the object can be determined when the file handler tries to open `self.filename`. It still has the try/except but instead of calling NetCDF4-python directly it creates an instance of the "netcdf4 handler" class and if exception then it makes the h5netcdf handler class. They would be essentially two classes implementing the file-reading interfaces that the Satpy NetCDF4 utility handler needs to get attributes/data.

mraspaud: Indeed, that would work, but probably something for another PR.

satpy/readers/fci_l1c_nc.py
pnuu: Hmm. I wonder why I've had the `.items()` in there. Plain `dict(arr.attrs).copy()` works, but with `h5netcdf`
```python
In [5]: fid.attrs.copy()
AttributeError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 fid.attrs.copy()

AttributeError: 'Attributes' object has no attribute 'copy'
```
I'll remove the `.items()`.

satpy/readers/netcdf_utils.py
pnuu: I'm OK removing this new class, if the NetCDF4 library some day _always_ supports the remote reading. Relying on a specific source (`conda-forge`, for example) for the library won't be enough, as not everybody can use it.

I'm open to ideas how the `.startswith("s3://")` could be enhanced.

satpy/readers/fci_l1c_nc.py
pnuu: Removed in 9e5186c0e

pnuu:
The lower test coverage is the only remaining bit I think, and as discussed above, that requires another PR where these tests are converted to use a real file with all the required stuff in it.

satpy/readers/netcdf_utils.py
djhoese: The `try/except` might be the best option. I think we're generally OK with the behavior of "do whatever you have to do to read this file", but we'll have to keep it in mind when debugging things for us and user bug reports ("wait, do you only have h5netcdf installed?").

The only reason *not* to do the try/except is that NetCDF4-C may be able to read URLs (HTTP for sure, s3:// some times I think) but may not perform as well as h5netcdf depending on what modes/methods are being used to read the data (as a stream of bytes versus HTTP byte range requests). The worst part about this would be that it would be hard to detect because things would still be working, just slower.

I think my overall vote is for merging this stuff into one user-level class (likely constructed from 2 or more other classes) and leaving the try/except.

pnuu: I created a new issue for this: https://github.com/pytroll/satpy/issues/2362

PR #2306, opened 2022-12-02
fix description for 500m resolution

gerritholl:
Fix the description for the 500m resolution, which said it was 1 km.  Also clarify that all those resolutions are SSP resolutions.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


ameraner:
Sorry I missed it, thanks for spotting this!

Issue #1940, opened 2021-12-14
worldeqc3km has resolution of 10 km


gerritholl:
**Describe the bug**

The AreaDefinition `worldeqc3km`, that is shipped with satpy, implies in the name and claims in the description that it has a resolution of 3 km.  In reality, it has a resolution closer to 10 km.

**To Reproduce**

```python
from satpy.resample import 
ar = get_area_def("worldeqc3km")
print(ar.resolution)
```

**Expected behavior**

Something close to 3 km.

**Actual results**

```(9783.939620507812, 9783.939620507812)```

**Environment Info:**
 - Satpy Version: v0.33.0-6-ga2b1e82e
 - PyResample Version: v1.22.3

**Additional context**

See also #1207 and #819.

Issue #1207, opened 2020-05-14
Descriptions in areas.yaml need cleanup


gerritholl:
**Describe the bug**

Many descriptions of areas defined in `areas.yaml` are incorrect.  They appear copy-pasted from other descriptions.

**To Reproduce**

Some examples:

https://github.com/pytroll/satpy/blob/b8bd9ab48ac5b42f097ae1844a39a6fdbfb48823/satpy/etc/areas.yaml#L129-L140

(this is not the world in 3 km, and it cannot be 3 km worldwide if it's plate caree)

https://github.com/pytroll/satpy/blob/b8bd9ab48ac5b42f097ae1844a39a6fdbfb48823/satpy/etc/areas.yaml#L239-L253

(not Kuwait!)

https://github.com/pytroll/satpy/blob/b8bd9ab48ac5b42f097ae1844a39a6fdbfb48823/satpy/etc/areas.yaml#L510-L523

(not at 4 km)

https://github.com/pytroll/satpy/blob/b8bd9ab48ac5b42f097ae1844a39a6fdbfb48823/satpy/etc/areas.yaml#L1093-L1106

(1 km or 10 km?)

There are several others.

**Expected behavior**

I expect that descriptions are accurate and useful.

**Actual results**

Actual descriptions are sometimes accurate, sometimes inaccurate.  Sometimes useful, sometimes not useful.

gerritholl:
(And another thing: the names of `eurol` and `euro1` are terribly similar in many fonts, but that would be impossible to fix without breaking backward compatibility.)

PR #2310, opened 2022-12-05
Naming of Area definitions in yaml file

BENR0:
While working on #2167 I noticed that there were two areas defined in the yaml file with the same name, at least when ignoring capitalization. While this is suboptimal with respect to aforementioned PR (sphinx complaining about duplicate labels), I think this also should be avoided in general. 

This PR should be base for a discussion about naming conventions for area definitions. Furthermore, at least suggested from the comments in the area yaml file, it seems there are some "deprecated" area definitions present.

To get the discussion started I think it would be nice to have something like the following for area names:
\<region\>_<projection?>_<resolution?>

Related Issues:
- #1207
- #1248 
- #819 
- #1940

### DIFF
--- a/satpy/etc/areas.yaml
+++ b/satpy/etc/areas.yaml
@@ -301,7 +301,7 @@ EastEurope:
     lower_left_xy: [654112.8864287604, 2989901.7547366405]
     upper_right_xy: [4553111.804127298, 5390224.287390241]
 
-AfHorn:
+AfHorn_geos:
   description: Eastern disk MSG image 0 degrees
   projection:
     proj: geos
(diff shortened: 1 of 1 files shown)


djhoese:
I'm surprised you even noticed this. Doesn't YAML just overwrite the previously loaded key as it loads the dictionary/mapping?

I agree the names should not match or be confuse-able in general. There is another issue regarding builtin satellite-based area definition naming that has a specific "scheme" being worked on. The other problem is that a lot of these are based on legacy areas from mpop I think. We've also talked about in some of the monthly meetings I think about have builtin areas for specific regions (I don't think there are any US ones currently).

BENR0:
Yes I think duplicate keys get overwritten but in this case the due to different capitalization and keys being case sensitive they both are included. I think regions in general are a good idea at least in the description somewhere. To sort the yaml file by but also from a user perspective if they are included in the area definition table one could sort through those by region.



[capitalization](https://dict.leo.org/englisch-deutsch/capitalization) in the name

PR #2312, opened 2022-12-05
Add config option for temporary directory

sfinkens:
Add config option `tmp_dir` for a temporary directory. That is, for example, where decompressed input files are stored. Default: `tempfile.gettempdir()`

The test is pretty weak, but I didn't see a way for a meaningful test, except "the code I wrote is the code I wrote". Let me know if you have a better idea.

 - [X] Tests added 
 - [X] Fully documented

### DIFF
--- a/doc/source/config.rst
+++ b/doc/source/config.rst
@@ -260,6 +260,20 @@ correction.
 
 .. _component_configuration:
 
+
+Temporary Directory
+^^^^^^^^^^^^^^^^^^^
+
+* **Environment variable**: ``SATPY_TMP_DIR``
+* **YAML/Config Key**: ``tmp_dir``
+* **Default**: `tempfile.gettempdir()`_
+
+Directory where Satpy creates temporary files, for example decompressed
+input files. Default depends on the operating system.
+
+.. _tempfile.gettempdir(): https://docs.python.org/3/library/tempfile.html?highlight=gettempdir#tempfile.gettempdir
+
+
 Component Configuration
 -----------------------
 
--- a/doc/source/config.rst
+++ b/doc/source/config.rst
@@ -258,9 +258,6 @@ as part of the :func:`~satpy.modifiers.angles.get_angles` and
 used by multiple modifiers and composites including the default rayleigh
 correction.
 
-.. _component_configuration:
-
-
 Temporary Directory
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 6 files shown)


djhoese:
FYI technically it isn't at import time, but is when the first temporary directory is requested:

https://github.com/python/cpython/blob/124ecd657646f808d1d3282c37ee19aae6bcb47f/Lib/tempfile.py#L283-L295

After that it is cached and can't be changed.

satpy/tests/test_config.py
```diff
@@ -480,6 +480,22 @@ def test_bad_str_config_path(self):
...
+            'cache_dir',
+            'cache_lonlats',
+            'cache_sensor_angles',
+            'config_path',
+            'data_dir',
+            'demo_data_dir',
+            'download_aux',
+            'sensor_angles_position_preference',
+        }
+        assert set(satpy.config.to_dict().keys()) == keys_exp
```
djhoese, line 497: I'm not sure I like tests like these. We've had a couple tests like these in the past with stuff like DataArray.attrs checks and I always get annoyed when I change one small thing and have to update a ton of tests that weren't actually doing anything very useful. The overall contents of the config object doesn't really matter as long as it is usable. Technically users can add their own keys just by defining `SATPY_*` environment variables. We only define defaults in `_config.py`. This test will fail if anything else is defined.

How about a test that gets the tmp dir value and makes sure it is a directory? Maybe even a writable directory?

sfinkens:
> FYI technically it isn't at import time, but is when the first temporary directory is requested

Aha, interesting! That makes sense. So in my case cartopy is using `gettempdir` for the first time (imported by pyresample which is imported by Satpy). That's why I would need to set `TMPDIR` before importing Satpy.

satpy/tests/test_config.py
sfinkens: Good idea, I'll try that!

doc/source/config.rst
```diff
@@ -260,6 +260,20 @@ correction.
 
 .. _component_configuration:
```
djhoese, line 261: This label needs to be moved back to the correct section (the `Component Configuration` section below the temporary directory section).

djhoese (changes requested):

doc/source/config.rst
sfinkens: Done, thanks for checking!

mraspaud (approved):
LGTM

PR #2317, opened 2022-12-08
Fix for bug regarding masking of good scan lines with seviri hrit reader introduced by #1693

BENR0:
The refactor of `mask_bad_scan_lines` from `np.choose` to `dataarray.where` inverted the logic of the line mask which lead to good scan lines being masked.

Furthermore this fixes an undiscovered bug regarding the optional line masking in the hrit reader introduced in in PR 1693 where the corresponding reader kwarg was not recognized.

This is a fix which should return the hrit reader to work the same as before #1693.

Notes/Todos:
- during the search for the bug I noticed the second bug mentioned above. It is fixed but there obviously is no unit test which is why the bug was not discovered earlier. I will check how this is tested with other reader_kwargs and make another PR so this fix does not have to wait for it.
- I am embarrassed to say that in #1693 I reused the quality flag setup of the tests which was already there even though I thought the different line masking cases were severly undertested. I will have another look but I think the masking logic is not completely right (this obviously also depends on which flags the user thinks are still considered good). So there should be tests for more different flag combinations.

### DIFF
--- a/satpy/readers/seviri_base.py
+++ b/satpy/readers/seviri_base.py
@@ -948,7 +948,7 @@ def _create_bad_quality_lines_mask(line_validity, line_geometric_quality, line_r
             Quality flags with shape (nlines,).
 
     Returns:
-        numpy.ndarray
+        numpy.ndarray: Indicating if the scan line is bad.
     """
     # Based on missing (2) or corrupted (3) data
     line_mask = line_validity >= 2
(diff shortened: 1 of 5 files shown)


mraspaud (commented):
Looks good here, but tests are failing

sfinkens (approved):
LGTM, thanks for the quick fix!

mraspaud (approved):
LGTM

PR #2327, opened 2022-12-18
Update himawari8/9 JMA True Color Reproduction composite

mherbertson:
This update aligns the `true_color_reproduction` composite closer towards JMA. This includes:

- adding an uncorrected blend at the terminator with day/night compositor
- removal of cira_strech and using default log stretch with custom min/max values for color balance and contrast.

TCR_202212122100

TCR_202212130220

TCR_202212130840

### DIFF
--- a/satpy/etc/composites/ahi.yaml
+++ b/satpy/etc/composites/ahi.yaml
@@ -70,8 +70,7 @@ composites:
     standard_name: toa_bidirectional_reflectance
 
   reproduced_green:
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
This looks good to me, although I am not to familiar with the composite. Maybe @simonrp84 wants to have a look here?

simonrp84:
Looks good to me, images this produces are excellent and the YAML files good too.

simonrp84 (approved):

PR #2335, opened 2022-12-21
Fix Scene modifying user-provided reader_kwargs

djhoese:
While working with the `MultiScene` I noticed that only the first Scene creation was working and failing on the second because `reader_kwargs` passed by me were being modified by the Scene. This is the simplest fix I could think of for fixing this.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/tests/test_scene.py
+++ b/satpy/tests/test_scene.py
@@ -710,6 +710,31 @@ class TestScene:
                 assert mock.call(filenames["reader3"], **storage_options_3) in open_files.mock_calls
                 assert reader_kwargs == orig_reader_kwargs
 
+    def test_storage_options_from_reader_kwargs_per_reader_and_global(self):
+        """Test getting storage options from reader kwargs.
+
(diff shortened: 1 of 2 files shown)


pnuu (commented):
Codecov seems to think the fix isn't covered by the modified tests.

djhoese:
I never liked codecov anyway :wink: 

If I click on the link I get this:

image

But then if I expand `utils.py` it looks like it isn't being covered:

image

djhoese:
I don't know exactly what codecov's problem is. Everything is tested now and when you click on the codecov Details it shows that things are covered...I think.

As discussed on slack with @pnuu, this PR now also clarifies/fixes the case where storage_options are defined at the "global" reader_kwargs level and the per-reader level (use global as defaults, overwrite with reader-specific).

pnuu (approved):
LGTM

PR #2337, opened 2022-12-22
Add ATMS SDR HDF5 format reader

adybbroe:
This PR adds a reader for the ATMS SDR HDF5 format as it may be produced by the CSPP package.

The format is described in the JPSS COMMON DATA FORMAT CONTROL BOOK (CDFCB):

Joint Polar Satellite System (JPSS) Common Data Format Control Book -
External (CDFCB-X) Volume III - SDR/TDR Formats

(474-00001-03_JPSS-CDFCB-X-Vol-III_0124C.pdf)

https://www.nesdis.noaa.gov/about/documents-reports/jpss-technical-documents/jpss-science-documents

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented


adybbroe:
Here an example loading all channels and available composites and generating a humidity channel RGB
```
scn = Scene(filenames=glob(os.path.join(ATMS_DIR, '*h5')), reader='atms_sdr_hdf5')
scn.load(scn.available_composite_names() + scn.available_dataset_names())
local = scn.resample(areaid)
local.show('mw183_humidity_surface')
```
atms_mw183_humidity_surface_20221220_0910_nsper_swe

For comparison I here add a corresponding MHS RGB from a nearby NOAA-19 scene:
mhs_mw183_humidity_surface_20221220_0847_nsper_swe

adybbroe:
I am using the `viirs_sdr` reader code quite extensively, by making minimal changes to it. So, now there are a couple of ATMS things in the `viirs_sdr.py`. Maybe we should rename it to make it clear that it can also deal with ATMS, and possibly add the new atms class to the same file? But then it starts getting quite long...

I also need to add some tests. I found it difficult starting with the tests this time as I needed to read the actual data before understanding what to adapt to the code...

satpy/readers/viirs_sdr.py
```diff
@@ -107,8 +107,10 @@ def _get_invalid_info(granule_data):
                 'SVM14': 'VIIRS-M14-SDR',
                 'SVM15': 'VIIRS-M15-SDR',
                 'SVM16': 'VIIRS-M16-SDR',
-                'IVCDB': 'VIIRS-DualGain-Cal-IP'
-                }
+                'IVCDB': 'VIIRS-DualGain-Cal-IP',
+                'SATMS': 'ATMS-SDR',
+                'GATMO': 'ATMS-SDR-GEO',
+                'TATMS': 'ATMS-TDR'}
```
djhoese, line 113: Can we maybe make this `DATASET_KEYS` part of the class object, then refactor the class to be more of a base class and have a VIIRS subclass and an ATMS subclass?

adybbroe: I have made a JPSS SDR base class as suggested but still keeping the `DATASET_KEYS` as a common variable outside the class. Should I include it in the base class?

I tried adding some test coverage. Not sure about this probably rather old (> 7 years) try-except:

https://github.com/pytroll/satpy/blob/11708d0d4cfd39b9eb62e4177ca3eb4a63c49653/satpy/readers/viirs_sdr.py#L151

If that is ever being used? But it was not tested. It is now, but I guess this line is from the time before Dask? Therefore the change here.

satpy/tests/reader_tests/test_atms_sdr_hdf5.py
```diff
@@ -0,0 +1,337 @@
...
+                self._add_data_info_to_file_content(file_content, filename, prefix3,
+                                                    self._num_test_granules)
+            elif filename[0] == 'G':
+                self._add_geolocation_info_to_file_content(file_content, filename, prefix3,
+                                                           self._num_test_granules)
+            final_content.update(file_content)
+        self._convert_numpy_content_to_dataarray(final_content)
+        return final_content
+
+    def _adjust_scaling_factors(self, factors, file_units, output_units):
```
mraspaud, line 244: According to codecove, this function isn't use at all, can we remove it?

satpy/tests/reader_tests/test_atms_sdr_hdf5.py
```diff
@@ -0,0 +1,337 @@
...
+        """Adjust scaling factors ."""
+        if file_units == output_units:
+            return factors
+        else:
+            factors = np.array([1.0, 0], dtype=np.float32)
+            factors = xr.DataArray(da.from_array(factors, chunks=1))
+        return factors
+
+
+class TestATMS_SDR_Reader(unittest.TestCase):
```
mraspaud, line 254: We try to move away from unittest and use pytest-style tests nowadays. Can this class be updated to follow this?

satpy/tests/reader_tests/test_atms_sdr_hdf5.py
```diff
@@ -0,0 +1,337 @@
...
+        r = load_reader(self.reader_configs)
+        loadables = r.select_files_from_pathnames([
+            'SATMS_j01_d20221220_t0910240_e0921356_b26361_c20221220100456348770_cspp_dev.h5',
+            'GATMO_j01_d20221220_t0910240_e0921356_b26361_c20221220100456680030_cspp_dev.h5',
+        ])
+        r.create_filehandlers(loadables)
+        ds = r.load(ATMS_CHANNEL_NAMES)
+        self.assertEqual(len(ds), 22)
+        for d in ds.values():
+            self._assert_bt_properties(d, with_area=True)
```
mraspaud, line 324: As Codescene says, this test and the next could be refactored together through parametrization.

mraspaud (commented):
Thanks for adding this reader and refactoring the common parts out of the viirs reader. The logic seems sane, I just have a few inline comments.

satpy/readers/atms_sdr_hdf5.py
```diff
@@ -0,0 +1,128 @@
...
+    def _get_scaling_factors(self, factor_var_path, ch_idx):
+        """Get file scaling factors and scale according to expected units."""
+        if ch_idx is None:
+            return self.get(factor_var_path)
+
+        if ch_idx == 21:
+            ch_idx = 20  # The BrightnessTemperatureFactors array is only
+            # 42 long!? But there are 22 ATMS bands to be scaled! We assume
+            # the scale/pffset values are the same for all bands!
+            # FIXME!
```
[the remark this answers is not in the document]
adybbroe: Yes, good question. I think I need consulting someone who might know what is going on. Feels a bit uncomfortable - could I have misunderstood something?
I am asking around now. So let's wait a bit before closing this PR, hopefully I will get some clarification on the issue.

satpy/readers/viirs_atms_sdr_utils.py
```diff
@@ -0,0 +1,371 @@
...
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Common utilities for reading VIIRS and ATMS SDR data."""
```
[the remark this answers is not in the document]
adybbroe: True, agree. Will do.

satpy/readers/viirs_sdr.py
```diff
@@ -515,13 +231,23 @@ def filter_filenames_by_info(self, filename_items):
...
         if geo_keep:
             filename_items = self._remove_geo_datasets_from_files(filename_items, geo_del)
-        return super(VIIRSSDRReader, self).filter_filenames_by_info(filename_items)
+
+        filename_items = self._remove_non_viirs_datasets_from_files(filename_items, viirs_del)
+        return super().filter_filenames_by_info(filename_items)
+
+    def _remove_non_viirs_datasets_from_files(self, filename_items, files_to_edit):
+        no_viirs = ATMS_DATASET_KEYS
+        return self._remove_datasets_from_files(filename_items, files_to_edit, no_viirs)
```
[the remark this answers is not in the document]
adybbroe: Na, the thing is that there may be ATMS SDR files in the same directory, and the names are following exactly same naming convention. This means the patterns specified in the `viirs_sdr.yaml` will also pick up the ATMS files, e.g.:
```
GATMO_j01_d20221220_t0910240_e0921356_b26361_c20221220100456680030_cspp_dev.h5
SATMS_j01_d20221220_t0910240_e0921356_b26361_c20221220100456348770_cspp_dev.h5
TATMS_j01_d20221220_t0910240_e0921356_b26361_c20221220100456524427_cspp_dev.h5
```
The file patterns specified in the yaml file is like this:
```
    file_patterns: ['{datasets}_{platform_shortname}_d{start_time:%Y%m%d_t%H%M%S%f}_e{end_time:%H%M%S%f}_b{orbit:5d}_c{creation_time:%Y%m%d%H%M%S%f}_{source}.h5']
```

In the atms counterpart it is much easier to be more specific as there are only three types of files (as above). For VIIRS there are many, so I think it would be cumbersome to specify this full list in the yaml file?

satpy/tests/reader_tests/test_atms_sdr_hdf5.py
adybbroe: Ahh, yes. Good, thx

satpy/tests/reader_tests/test_atms_sdr_hdf5.py
adybbroe: Well yes, I agree. But I am very much building upon how it was already done for VIIRS. So this is no consistent with how the viirs-sdr tests are formulated. I am either not completely comfortable with how the mock part works in the `setUp`, so it it feels not trivial to me how to migrate it to using pytest and fixtures. I would rather leave it for another refactoring PR, and then doing both VIIRS and ATMS together.

satpy/tests/reader_tests/test_atms_sdr_hdf5.py
adybbroe: Ahh, yes, fixed.

satpy/tests/reader_tests/test_atms_sdr_hdf5.py
adybbroe: I did as suggested now, thought it required more work, was not acquainted with the `setup_method`/`teardown_method` methods. Thanks!

satpy/readers/atms_sdr_hdf5.py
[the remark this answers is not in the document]
adybbroe: Nigel, UK, pointed out for me that the data in the `SDR_All` section applies to all granules in the file. Thus the dimension of 42 of the `BrightnessTemperatureFactors` dataset fits is determined by the number of granules in the file. There are 21 granules, and the data holds 21 pairs of scale and offset, one set for each granule.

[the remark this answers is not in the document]
adybbroe: This should be fixed now. Now also the code for the VIIRS and ATMS scaling is common to both.

satpy/readers/viirs_atms_sdr_base.py
```diff
@@ -0,0 +1,366 @@
...
+        start_orbit_path = self.filetype_info.get('start_orbit', default).format(dataset_group=dataset_group)
+        return int(self[start_orbit_path])
+
+    @property
+    def end_orbit_number(self):
+        """Get end orbit number."""
+        dataset_group = DATASET_KEYS[self.datasets[0]]
+        default = 'Data_Products/{dataset_group}/{dataset_group}_Aggr/attr/AggregateEndingOrbitNumber'
+        end_orbit_path = self.filetype_info.get('end_orbit', default).format(dataset_group=dataset_group)
+        return int(self[end_orbit_path])
```
mraspaud, line 149:
```suggestion
    @property
    def end_orbit_number(self):
        """Get end orbit number."""
        end_orbit_path = self._get_aggr_path("end_orbit", "AggregateEndingOrbitNumber")
        return int(self[end_orbit_path])
        
    def _get_aggr_path(self, fileinfo_key, aggr_default):
        dataset_group = DATASET_KEYS[self.datasets[0]]
        default = 'Data_Products/{dataset_group}/{dataset_group}_Aggr/attr/' + aggr_default
        return self.filetype_info.get(fileinfo_key, default).format(dataset_group=dataset_group)
```

We could introduce this new method and reduce the code duplication in the four previous methods

satpy/tests/test_readers.py
```diff
@@ -510,7 +521,15 @@ def test_no_parameters(self, viirs_file):
...
         assert ri['viirs_sdr'] == [viirs_file]
 
-    def test_bad_sensor(self, viirs_file):
+    def test_no_parameters_atms(self, atms_file):
+        """Test with no limiting parameters."""
+        from satpy.readers import find_files_and_readers
+
+        ri = find_files_and_readers()
+        assert list(ri.keys()) == ['atms_sdr_hdf5']
+        assert ri['atms_sdr_hdf5'] == [atms_file]
```
mraspaud, line 530: I don't think this is needed, because it's basically the same think as the function above (or that could be parametrized).
But maybe it would be important to test what `find_files_and_readers` returns when both a viirs file and an atms file are in the same directory?

mraspaud (commented):
Looks better and better, just a couple of comments this time.

satpy/readers/viirs_sdr.py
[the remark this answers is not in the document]
mraspaud: yes, it would be cumbersome, especially since some viirs files can contain multiple channels and thus have highly variable names... The only worry then is what happens when there are both atms and viirs files in the same directory, to they get assigned to the readers correctly? Or should the viirs reader raise some kind of error when being assigned an atms file?

satpy/tests/test_readers.py
adybbroe: Indeed, yes! I just tested this and the `find_files_and_readers` seem to never finish! :-(

satpy/readers/viirs_atms_sdr_base.py
mraspaud: You can use `_get_aggr_path` in the `start_orbit_number`, `start_time` and `end_time` methods too.

satpy/etc/enhancements/atms.yaml
```diff
@@ -0,0 +1,29 @@
...
+  mw183_humidity:
+    standard_name: mw183_humidity
+    operations:
+    - name: inverse
+      method: !!python/name:satpy.enhancements.invert
+      args:
+      - [true, true, true]
+    - name: stretch
+      method: !!python/name:satpy.enhancements.stretch
+      kwargs: {stretch: linear}
```
pnuu, line 12: Would it be possible to have a static stretch for this and the other composite? That way the inversion wouldn't be needed and the whole data (which might not be that large?) wouldn't need to be readed for determining the stretching.

satpy/tests/test_readers.py
adybbroe: No, my bad. It seemed so, but it actually does terminate and gives a nice (expected) output.

satpy/etc/enhancements/atms.yaml
adybbroe: Good point, let me see...

satpy/readers/viirs_sdr.py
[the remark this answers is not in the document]
adybbroe: I have added a test for the case of both ATMS and VIIRS SDRs in the same directory. See below.
I think that should be enough?

satpy/etc/enhancements/atms.yaml
adybbroe: Okay, I agree t could make sense here to make the stretch static. But please be aware that I have not spend a great deal of research on how to combine the different channels in the best way to make it useful for a forecaster. But thought it was good to have something included here, then someone else (maybe a later me) can elaborate more and suggest adequate RGB types. I included one with the window channels and one with only the absorbing channels.
And I think I did as you suggested?

pnuu: Looking great for such a quick addition, thanks!

Are there any references for these composites, or did you just invent them on-the-fly?

satpy/readers/viirs_atms_sdr_base.py
adybbroe: Fixed!

satpy/etc/enhancements/atms.yaml
adybbroe: My invention so far. Haven't seen such RGBs before...
That is also why I am sure that more "research" is possible here!

pnuu: Well invented in that case!

mraspaud (approved):
LGTM! Well done!

PR #2340, opened 2023-01-04
update to SEVIRI grib and bufr readers

samain-eum:
We updated the SEVIRI readers yaml files to update / add the standard_name and long_names attributes in order to be CF compliant

Olivier Samain
On behalf of the EUMETSAT Optical Imaging team

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,62 @@
+## Version 0.39.0 (2022/12/30)
+
+### Issues Closed
+
+* [Isse2326](https://github.com/pytroll/satpy/issues/2326) - satpy installation failed with pip
+* [Isse2325](https://github.com/pytroll/satpy/issues/2325) - VIIRS SDR DNB units and file_units are conflicting
+* [Isse2323](https://github.com/pytroll/satpy/issues/2323) - use area def raise value different when process himawari datasets
+* [Isse2311](https://github.com/pytroll/satpy/issues/2311) - scatsat1_l2b and caliop_l2_cloud readers fail to initialise due to failed `Dataset` import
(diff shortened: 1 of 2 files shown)


satpy/etc/readers/seviri_l2_grib.yaml
```diff
@@ -276,41 +302,46 @@ datasets:
...
 
   num_accumulations:
     name: num_accumulations
+    long_name: CRM Product Number of Accumulations
     resolution: 3000.403165817
     file_type: grib_seviri_crm
     parameter_number: 6
     units: "1"
-    long_name:  num_accumulations
+    standard_name: number_of_accumulations
```
sjoro, line 320: can you please move this up under the `long_name` so that the order is same for all the datasets in this file.

satpy/etc/readers/seviri_l2_bufr.yaml
```diff
@@ -58,6 +58,7 @@ datasets:
     key: 'latitude'
     resolution: [48006.450653072,9001.209497451]
     file_type: [seviri_l2_bufr_asr,seviri_l2_bufr_cla,seviri_l2_bufr_csr,seviri_l2_bufr_gii,seviri_l2_bufr_thu,seviri_l2_bufr_toz]
+    long_name: Latitude
```
sjoro, line 61: i'd like to see the order here as in the GRIB-yaml file, i.e. `name`, `long_name`, `standard_name`, and then the other keys. please move these up.

sjoro (changes requested):
a few small nitpicks on the order of the keys, please see comments. otherwise this looks good.

PR #2342, opened 2023-01-05
Fix typos in Changelog Isse -> Issue

gerritholl:
Fix typos in CHANGELOG.md, where several issues were referred to as "Issexxx" rather than "Issue xxx".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,26 +2,26 @@
 
 ### Issues Closed
 
-* [Isse2326](https://github.com/pytroll/satpy/issues/2326) - satpy installation failed with pip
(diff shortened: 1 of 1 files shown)


gerritholl:
I found it remarkable that the same typo is repeated many times.  Is there a script that is used to generate those typos?

djhoese:
See the RELEASING.md instructions. My guess is that either me or Martin have an old version of the changelog generating package installed so the typo only shows up when that person makes releases. I'll have to look more when I'm more awake.

djhoese:
So this was apparently something wrong in my vim config. I commented out everything related to code folding and now pasting into a file of this size seems to work again. I'm not really sure what is/was going on. Thanks for cleaning this up.

PR #2360, opened 2023-01-20
Fix "p_id" to be compatible with Himawari-9

tsukada-cs:
The Projection ID (`"p_id"`) of this reader was fixed to `"geosh8"`, assuming only input of the Himawari-8 satellite. However, it was modified to `"geosh8"` for the Himawari-8 and `"geosh9"` for the Himawari-9.

### DIFF
--- a/satpy/readers/ahi_hsd.py
+++ b/satpy/readers/ahi_hsd.py
@@ -505,7 +505,7 @@ class AHIHSDFileHandler(BaseFileHandler):
 
         pdict['a_name'] = self.observation_area
         pdict['a_desc'] = "AHI {} area".format(self.observation_area)
-        pdict['p_id'] = 'geosh8'
+        pdict['p_id'] = f'geosh{self.basic_info["satellite"][0].decode()[-1]}'
 
         return get_area_definition(pdict, aex)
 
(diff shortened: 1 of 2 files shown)


mraspaud (changes requested):
Thanks for the fix! Could you just add a line to the tests, for example here: https://github.com/pytroll/satpy/blob/main/satpy/tests/reader_tests/test_ahi_hsd.py#L195
just to make sure that this works as expected in the future?

tsukada-cs:
> Thanks for the fix! Could you just add a line to the tests, for example here: https://github.com/pytroll/satpy/blob/main/satpy/tests/reader_tests/test_ahi_hsd.py#L195 just to make sure that this works as expected in the future?

I have added a line to test for my fix. It was needed to add a test with the `_fake_hsd_handler` function, so I added it to this position.

mraspaud (approved):
LGTM!

djhoese:
@tsukada-cs Just curious, do you use the projection ID stored in the AreaDefinition for something? We have plans for this information to be removed/unused in a future major release of Pyresample (pyresample 2.0).

tsukada-cs:
> @tsukada-cs Just curious, do you use the projection ID stored in the AreaDefinition for something? We have plans for this information to be removed/unused in a future major release of Pyresample (pyresample 2.0).

No, I don't use it. In handling the Himawari-9 data, I just checked to see if there were any lines where only Himawari-8 was assumed.

PR #2372, opened 2023-02-01
add flash_area composite for LI L2

TAlonglong:
Add simple composite to be able to demostrate LI L2 capabilities of satpy

### DIFF
--- a/satpy/etc/composites/li.yaml
+++ b/satpy/etc/composites/li.yaml
@@ -1,0 +1,9 @@
+---
+sensor_name: visir/li
+
+composites:
+  flash_area:
+    compositor: !!python/name:satpy.composites.SingleBandCompositor
+    standard_name: flash_area
+    prerequisites:
+    - accumulated_flash_area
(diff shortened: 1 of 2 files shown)


djhoese (approved):
This seems fine to me, but I know nothing about LI so someone on the EU side should really review this.

mraspaud:
@sjoro @seeno opinions on this PR?

gerritholl:
What does it look like?  Could you perhaps add an image to this PR?

gerritholl:
For lightning it might be nice if 0 is transparent rather than yellow.  I don't know if there is a different way to do that apart from making an RGBA version of the indicated colormap.  It seems all the named colormaps built into trollimage are RGB and none are RGBA.

pnuu (approved):
LGTM. I've already used this as a basis on highlighting lightning on top of a SEVIRI composite.

TAlonglong:
flash_area

pnuu:
The failing unstable tests are something in `holoviews`. I saw that mentioned somewhere, and in any case should be handled in a PR touching code. Merging.

TAlonglong:
flash_area

pnuu:
> For lightning it might be nice if 0 is transparent rather than yellow.

For zero flash area there is no data, so those areas are transparent.

PR #2375, opened 2023-02-02
Add initial support for EUMETSAT's climatological calibration

simon-sat:
EUMETSAT are working on long-term calibration coefficients for SEVIRI, which will be available as external files specifying the gain and offset for each satellite over their lifetime.

This PR adds initial support for these calibration coefficients as an additional calibration method, `EUMCLIM`. Currently, I've only added support for the `native` reader, as this is the data format for historical images on the data store and eoportal. I will also add support to the HRIT and netCDF readers at a later point, as this should be reasonably straightforward.

This is a draft for now as I need to add the docs.

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/seviri_base.py
+++ b/satpy/readers/seviri_base.py
@@ -197,6 +197,16 @@ VISIR_NUM_LINES = 3712
 HRV_NUM_COLUMNS = 11136
 HRV_NUM_LINES = 11136
 
+CDS_EPOCH = np.datetime64('1958-01-01')
+
+EUMCLIM_COEF_MAP = {'WV_062': 'IR062',
+                    'WV_073': 'IR073',
+                    'IR_087': 'IR087',
+                    'IR_097': 'IR097',
+                    'IR_108': 'IR108',
+                    'IR_120': 'IR120',
+                    'IR_134': 'IR134'}
+
 CHANNEL_NAMES = {1: "VIS006",
                  2: "VIS008",
                  3: "IR_016",
@@ -371,9 +381,9 @@ def get_cds_time(days, msecs):
         days = np.array([days], dtype='int64')
         msecs = np.array([msecs], dtype='int64')
 
-    time = (np.datetime64('1958-01-01').astype('datetime64[ms]') +
+    time = (CDS_EPOCH.astype('datetime64[ms]') +
             days.astype('timedelta64[D]') + msecs.astype('timedelta64[ms]'))
-    time[time == np.datetime64('1958-01-01 00:00')] = np.datetime64("NaT")
+    time = np.where(time == CDS_EPOCH,  np.datetime64("NaT"), time)
 
     if len(time) == 1:
         return time[0]
@@ -390,18 +400,10 @@ def load_eumclim_nc(fname, ref_time):
             jul_time = jdays(ref_time)
             jul_idx = np.argmin(abs(jul_time - fid['julian_time']))
 
-            coef_map = {'WV_062': 'IR062',
-                        'WV_073': 'IR073',
-                        'IR_087': 'IR087',
-                        'IR_097': 'IR097',
-                        'IR_108': 'IR108',
-                        'IR_120': 'IR120',
-                        'IR_134': 'IR134'}
-
             eum_coef = {}
-            for chan in coef_map.keys():
-                eum_coef[chan] = {'gain': float(fid[f'b_{coef_map[chan]}'][jul_idx]),
-                                  'offset': float(fid[f'a_{coef_map[chan]}'][jul_idx])}
+            for chan in EUMCLIM_COEF_MAP.keys():
+                eum_coef[chan] = {'gain': float(fid[f'b_{EUMCLIM_COEF_MAP[chan]}'][jul_idx]),
+                                  'offset': float(fid[f'a_{EUMCLIM_COEF_MAP[chan]}'][jul_idx])}
     except OSError:
         raise OSError(f'Error: EUM calibration file {fname} does not exist.')
 
--- a/satpy/readers/seviri_base.py
+++ b/satpy/readers/seviri_base.py
@@ -383,7 +383,7 @@ def get_cds_time(days, msecs):
 
     time = (CDS_EPOCH.astype('datetime64[ms]') +
             days.astype('timedelta64[D]') + msecs.astype('timedelta64[ms]'))
-    time = np.where(time == CDS_EPOCH,  np.datetime64("NaT"), time)
+    time = np.where(time == CDS_EPOCH, np.datetime64("NaT"), time)
 
     if len(time) == 1:
         return time[0]
--- a/satpy/readers/seviri_base.py
+++ b/satpy/readers/seviri_base.py
@@ -398,6 +398,10 @@ def load_eumclim_nc(fname, ref_time):
     try:
         with Dataset(fname, 'r') as fid:
             jul_time = jdays(ref_time)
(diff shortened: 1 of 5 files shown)


satpy/readers/seviri_base.py
```diff
@@ -371,15 +371,43 @@ def get_cds_time(days, msecs):
...
+            coef_map = {'WV_062': 'IR062',
+                        'WV_073': 'IR073',
+                        'IR_087': 'IR087',
+                        'IR_097': 'IR097',
+                        'IR_108': 'IR108',
+                        'IR_120': 'IR120',
+                        'IR_134': 'IR134'}
+
+            eum_coef = {}
+            for chan in coef_map.keys():
```
pnuu, line 402: `.keys()` isn't needed.

satpy/readers/seviri_base.py
```diff
@@ -371,15 +371,43 @@ def get_cds_time(days, msecs):
...
+            jul_time = jdays(ref_time)
+            jul_idx = np.argmin(abs(jul_time - fid['julian_time']))
+
+            coef_map = {'WV_062': 'IR062',
+                        'WV_073': 'IR073',
+                        'IR_087': 'IR087',
+                        'IR_097': 'IR097',
+                        'IR_108': 'IR108',
+                        'IR_120': 'IR120',
+                        'IR_134': 'IR134'}
```
pnuu, line 399: Move outside as a constant. `EUMCLIM_COEF_MAP` or something.

satpy/readers/seviri_l1b_native.py
```diff
@@ -120,6 +122,12 @@ def __init__(self, filename, filename_info, filetype_info,
         self._read_trailer()
         self.image_boundaries = ImageBoundaries(self.header, self.trailer, self.mda)
 
+        # Read the EUMETSAT climate calibration coefficients if available
+        if self.calib_mode == 'EUMCLIM':
+            self.eumclim_cal_coef = load_eumclim_nc(self.calib_file, self.observation_start_time)
+        else:
+            self.eumclim_cal_coef = {}
```
pnuu, line 129: Set `self.eumclim_cal_coef = {}` by default so the `else` branch isn't needed.

satpy/readers/seviri_l1b_native.py
```diff
@@ -554,6 +571,13 @@ def _get_calib_coefs(self, channel_name):
             'RadiometricProcessing']['MPEFCalFeedback']
         radiance_types = self.header['15_DATA_HEADER']['ImageDescription'][
                 'Level15ImageProduction']['PlannedChanProcessing']
+
+        # Merge potential external and EUMETSAT climatology coefficients
+        if self.calib_mode == 'EUMCLIM':
+            tmp_dict = self._sort_ext_eum_coefs()
+        else:
+            tmp_dict = self.ext_calib_coefs
```
pnuu, line 579: Could the `self.ext_calib_coefs` be set with the sorted EUMCLIM values already in the `if self.calib_mode == 'EUMCLIM` branch in `__init__()`? Then `tmp_dict` and the if/else wouldn't be needed.

pnuu (commented):
Some initial comments inline.

pnuu:
pre-commit.ci autofix

pnuu:
pre-commit.ci run

pnuu:
pre-commit.ci autofix

mraspaud (changes requested):
LGTM logic-wise, just some documentation about the new parameter missing, and maybe a link to the place to download/get some info about the calibration files?

satpy/tests/reader_tests/test_seviri_base.py
```diff
@@ -149,6 +175,41 @@ def test_get_padding_area_int():
...
+        np.testing.assert_allclose(retval['IR_120']['offset'], DUMMY_DATA['IR_120_offset'][0])
+
+    def test_load_eumclim_nc_futurefile(self, make_dummy_ncdf):
+        """Test we get last value if we specify time in distant past."""
+        with pytest.warns(UserWarning):
+            retval = load_eumclim_nc(make_dummy_ncdf, datetime(2040, 1, 1, 12, 0, 0))
+        np.testing.assert_allclose(retval['WV_073']['gain'], DUMMY_DATA['WV_073_gain'][-1])
+        np.testing.assert_allclose(retval['IR_120']['offset'], DUMMY_DATA['IR_120_offset'][-1])
+
+    def test_load_eumclim_nc_goodfile(self, make_dummy_ncdf):
```
sfinkens, line 200: This test could be simplified using a parametrization

```python
pytest.mark.parametrize(
    "timestamp,index_exp",
    [
        (datetime(2020, 1, 12, 15, 36), 0),
        (datetime(2020, 1, 12, 15, 46), 1),
        ...
)
def test_load_eumclim_nc_goodfile(self, make_dummy_ncdf, timestamp, index_exp):
    retval = load_eumclim_nc(make_dummy_ncdf, timestamp)
    np.testing.assert_allclose(retval['WV_073']['gain'], DUMMY_DATA['WV_073_gain'][index_exp])
    np.testing.assert_allclose(retval['IR_120']['offset'], DUMMY_DATA['IR_120_offset'][index_exp])
```

satpy/tests/reader_tests/test_seviri_base.py
```diff
@@ -35,16 +36,41 @@
...
+
+    fname = f'{tmp_path}/MSG1-COEFs.nc'
+    with Dataset(fname, 'w') as fid:
+        fid.createDimension('time', 3)
+        fid.createVariable('julian_time', np.float64, ('time',))[:] = DUMMY_DATA['time']
+        for val in EUMCLIM_COEF_MAP:
+            fid.createVariable(f'a_{EUMCLIM_COEF_MAP[val]}', np.float32, ('time',))[:] = DUMMY_DATA[f'{val}_offset']
+            fid.createVariable(f'b_{EUMCLIM_COEF_MAP[val]}', np.float32, ('time',))[:] = DUMMY_DATA[f'{val}_gain']
+
+    print(fname)
```
sfinkens, line 65: Debugging leftover

satpy/readers/seviri_base.py
```diff
@@ -371,15 +381,39 @@ def get_cds_time(days, msecs):
...
+    from pyorbital.astronomy import jdays
+
+    try:
+        with Dataset(fname, 'r') as fid:
+            jul_time = jdays(ref_time)
+            min_tdiff = np.min(abs(jul_time - fid['julian_time']))
+            if min_tdiff > 1:
+                warnings.warn("Closest EUMCLIM calibration coefficients differ from image time by more than one day.",
+                              UserWarning)
+            jul_idx = np.argmin(abs(jul_time - fid['julian_time']))
```
sfinkens, line 405: Can the index calculation be extracted to a separate method?

sfinkens (commented):
Oh, that's good news! Thanks for adding support for this.

I'm a bit concerned that the number of calibration options is getting too large. What if we delegate the loading of EUMCAL coefficients to the user? For example:

```python
from satpy.readers.seviri_base import load_eumcal_nc
eum_coefs = load_eumcal_nc(filename)
scene = Scene(..., ext_calib_coefs=eum_coefs)
```
And, if they wanted to merge them with other coefficients, provide a helper method:

```python
from satpy.readers.seviri_base import merge_coefs
merged_coefs = merge_coefs(other_coefs, eum_coefs)
```

Of course that would require the reader to be aware of time dependent coefficients, but that's a nice feature I think.

PR #2378, opened 2023-02-03
Adapt unit test between categorical and continuous

gerritholl:
Adapt unit test for pmode imagery to test the different behaviour expected for float or int data.

This doesn't actually fix any bug, because the fix for #2376 turns out to be in trollimage, where it's fixed by https://github.com/pytroll/trollimage/pull/123.  However, I think it's still useful to test for this in satpy because of the closely tied interface between the two.

Depends on https://github.com/pytroll/trollimage/pull/123.

Also changes the spelling of an import in `dataset/__init__.py` to make isort happy.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/dataset/__init__.py
+++ b/satpy/dataset/__init__.py
@@ -19,7 +19,5 @@
 
 from .anc_vars import dataset_walker, replace_anc  # noqa
 from .data_dict import DatasetDict, get_key  # noqa
-from .dataid import (  # noqa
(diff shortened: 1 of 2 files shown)


gerritholl:
I don't know why pre-commit complains about isort; this PR doesn't touch any imports.

PR #2380, opened 2023-02-05
Add an interface for reducing Rayleigh correction at high solar zenith angles in composites YAML

### LINKED ISSUE #2288: Add support for reducing Rayleigh correction at large solar zenith angles

yukaribbba:
As @strandgren requested in #2288, this PR will make it possible in a yaml config file. Also thanks to @simonrp84 for his original contribution [https://github.com/pytroll/pyspectral/pull/141](https://github.com/pytroll/pyspectral/pull/141).

 - [ ] Closes #2288 
 - [x] Tests added
 - [x] Fully documented 

With this PR, the YAML may look like:

```
  rayleigh_corrected_reduced:
    modifier: !!python/name:satpy.modifiers.PSPRayleighReflectance
    atmosphere: us-standard
    aerosol_type: rayleigh_only
    reduced_correction: True
    lim_low: 70
    lim_high: 95
    strength: 0.5
    prerequisites:
    - name: B03
      modifiers: [sunz_corrected]
    optional_prerequisites:
    - satellite_azimuth_angle
    - satellite_zenith_angle
    - solar_azimuth_angle
    - solar_zenith_angle

  true_color_reduced_corr:
    compositor: !!python/name:satpy.composites.spectral.GenericCompositor
    prerequisites:
    - name: Red
      modifiers: [sunz_corrected, rayleigh_corrected_reduced]
    - name: Green
      modifiers: [sunz_corrected, rayleigh_corrected_reduced]
    - name: Blue
      modifiers: [sunz_corrected, rayleigh_corrected_reduced]
    standard_name: true_color_reproduction
```

Some tests from Himawari-8 AHI data:
**Full disk overview**
image
**Edge detail**
image

Further, this can be combined with a un-corrected image for DayNightCompositor to make the day/night edge a little more fluent.
image


yukaribbba:
@djhoese Do we need a test for this one? I couldn’t find where to add it.

satpy/modifiers/atmosphere.py
```diff
@@ -77,6 +117,13 @@ def __call__(self, projectables, optional_datasets=None, **info):
             refl_cor_band = corrector.get_reflectance(sunz, satz, ssadiff,
                                                       vis.attrs['wavelength'][1],
                                                       red.data)
+        if reduced_correction:
```
djhoese, line 120: Is it possible to remove `reduced_correction` and just check if `lim_low`, `lim_high`, and `strength` are all specified? I guess I'm not sure what their defaults would be and which one would be required... :thinking:

djhoese (commented):
> Do we need a test for this one? I couldn’t find where to add it.

I...I...how are there no tests for this modifier? :confused: 

@adybbroe @mraspaud do you remember if there are tests for this somewhere? And @yukaribbba yes there needs to be tests added. I made some other comments inline.

satpy/modifiers/atmosphere.py
yukaribbba: Then I’ll just set `strength` to 0 as default. No reduction will be applied under it.

yukaribbba:
Maybe all the function are imported from pyspectral so we don't need to test them again? That could be the reason. I'm not sure...

satpy/tests/test_modifiers.py
```diff
@@ -392,6 +392,104 @@ def test_compositor(self, calculator, apply_modifier_info, sza):
...
+        assert res.attrs['calibration'] == 'reflectance'
+        assert res.attrs['units'] == '%'
+        assert res.attrs['wavelength'] == wavelength
+        assert res.attrs['name'] == name
+        assert res.attrs['resolution'] == resolution
+        assert res.attrs['sensor'] == 'abi'
+        assert res.attrs['start_time'] == '2017-09-20 17:30:40.800000'
+        assert res.attrs['end_time'] == '2017-09-20 17:41:17.500000'
+        assert res.attrs['area'] == area
+        assert res.attrs['ancillary_variables'] == []
```
djhoese, line 483: I don't think we need to check all of these attributes unless we assume they are going to be broken. This seems mostly like clutter. I know some of the other tests do stuff like this, but I disagree with those checks there too.

satpy/tests/test_modifiers.py
```diff
@@ -392,6 +392,104 @@ def test_compositor(self, calculator, apply_modifier_info, sza):
                                       masking_limit=NIRReflectance.MASKING_LIMIT)
 
 
+class TestPSPRayleighReflectance(unittest.TestCase):
```
djhoese, line 395: What do you think about moving this test to a new module in `tests/modifier_tests/test_atmosphere.py`? Additionally, could you make this class not based on `unittest.TestCase`? That will make it a "pytest" class and should make your parametrize functionality work.

satpy/modifiers/atmosphere.py
```diff
@@ -77,6 +115,13 @@ def __call__(self, projectables, optional_datasets=None, **info):
             refl_cor_band = corrector.get_reflectance(sunz, satz, ssadiff,
                                                       vis.attrs['wavelength'][1],
                                                       red.data)
+
+        if lim_low > lim_high:
+            lim_low = lim_high
+        logger.info("Reducing Rayleigh correction at high zenith angles by strength %s.", strength)
```
strandgren, line 121: I would probably remove this, since a similar logging message will come form pyspectral: https://github.com/pytroll/pyspectral/blob/main/pyspectral/rayleigh.py#L294.

satpy/modifiers/atmosphere.py
```diff
@@ -77,6 +115,13 @@ def __call__(self, projectables, optional_datasets=None, **info):
             refl_cor_band = corrector.get_reflectance(sunz, satz, ssadiff,
                                                       vis.attrs['wavelength'][1],
                                                       red.data)
+
+        if lim_low > lim_high:
```
strandgren, line 119: If I understand it correctly you decided to drop the key `reduced_correction`, e.g. `reduced_correction: True` as you have in your PR description? And as far as I can see the default values will be `lim_low = 70`, `lim_high = 95` and `strength = 0`, meaning that this will only be become active if the user defines at least the `strength` in the YAML recipe?

 In that case I would add an if statement here to continue only if the strength is greater than 0. I guess the computations are not that heavy, but it might be confusing for the user the log says that Rayleigh is being reduced when they have not asked for it.

satpy/modifiers/atmosphere.py
```diff
@@ -31,7 +31,41 @@
...
+          - name: B03
+            modifiers: [sunz_corrected]
+        optional_prerequisites:
+          - satellite_azimuth_angle
+          - satellite_zenith_angle
+          - solar_azimuth_angle
+          - solar_zenith_angle
+
+    In the case above, rayleigh correction is reduced gradually starting at solar zenith angle 70°.
+    When reaching 95°, the correction will only remain half of its initial strength.
```
strandgren, line 66: to make it more clear it should rather say: "..., the correction will only remain half of its initial strength **at lim_high**."

satpy/modifiers/atmosphere.py
```diff
@@ -31,7 +31,41 @@
...
+
+    .. code-block:: yaml
+
+      rayleigh_corrected_reduced:
+        modifier: !!python/name:satpy.modifiers.PSPRayleighReflectance
+        atmosphere: us-standard
+        aerosol_type: rayleigh_only
+        lim_low: 70
+        lim_high: 95
+        strength: 0.5
```
strandgren, line 55: I'm wondering whether these shall be renamed to make it more clear what their meaning is. Perhaps simply `reduce_lim_low`, `reduce_lim_high` and `reduce_strength`?

yukaribbba:
@djhoese @mraspaud CI tests passed

strandgren (commented):
Thanks a lot for working on this, it looks very good! I left a few comments inline. 

Another more general comment: I like the result including a blend with uncorrected data using the `DayNightCompositor` (although a bit too blue), but in principle, the method `pyspectral.rayleigh.Rayleigh.reduce_rayleigh_highzenith` is doing the same thing as this blend with uncorrected. The only, yet significant, difference is that the `DayNightCompositor` does the scaling across the blending area in `cos(sunz)`-space, whereas in this PR the scaling is in `sunz`-space. Could you try and see what happens if you convert the `sunz`, `lim_low` and `lim_high` to cosine-space before sending them them to `reduce_rayleigh_highzenith`?

@simonrp84 what did you actually have in mind for the implementation in pyspectral? To do the scaling with `sunz` or `cos(sunz)`?

satpy/modifiers/atmosphere.py
yukaribbba: I'll add that if.

satpy/modifiers/atmosphere.py
yukaribbba: And to make things clearer, I'll choose a 0.4/0.6 example instead of this half/half one.

yukaribbba:
@strandgren I'm trying the cosine part you mentioned but I think this is more like something in the pyspectral, and here we only import it and use it...

strandgren:
> @strandgren I'm trying the cosine part you mentioned but I think this is more like something in the pyspectral, and here we only import it and use it...

That's what I thought too, so I started to write the comment as just a note for the future that we might consider modifying/adding this in pyspectral. But then when I looked in pyspectral I saw that it's a `staticmethod` with no statement that the angles shall be in either `sunz` or `cos(sunz)`, which could allow us the freedom to pass `cos(sunz)` angles when calling the method from satpy (if this gives better results). Still we should not break any logic that @simonrp84 had in mind, so I'd be curious to hear his opinion.

simonrp84:
To be honest I wrote the pyspectral PR so long ago that I can't remember what I had in mind! My gut feeling is that I preferred zenith angle rather than cosine space as this is easier for users to understand. Beyond that, I can't offer much insight I'm afraid!

Regarding what the DayNightCompositor does, this wasn't something I considered in my PR as I was focusing on daytime composites only and attempting to create a more natural daylight-to-darkness transition, so didn't test with composites that also have a night component.

Lastly, I have one review comment on this - otherwise it looks good!

satpy/modifiers/atmosphere.py
```diff
@@ -60,6 +94,10 @@ def __call__(self, projectables, optional_datasets=None, **info):
 
         atmosphere = self.attrs.get('atmosphere', 'us-standard')
         aerosol_type = self.attrs.get('aerosol_type', 'marine_clean_aerosol')
+        reduce_lim_low = abs(self.attrs.get('reduce_lim_low', 70))
+        reduce_lim_high = abs(self.attrs.get('reduce_lim_high', 95))
```
simonrp84, line 98: I'm a bit concerned about this hard limit. Why was 95 degrees chosen? In some conditions you can have meaningful radiances beyond 95 degrees so if there's no important reason for this choice then I'd like to see the hard upper limit bumped to something like 105 degrees.

yukaribbba: Ok we'll set lim_high 105° by default

yukaribbba:
@strandgren I tried passing cos_zenith instead of zenith to the function and unfortunately it didn't have the correct thing as expected. I guess it's because angles are increasing from 0 to 180 but cosine is decreasing, so if you want to pass cosine you neet to invert the entire `factor = 1. - strength * where(zenith < thresh_zen, 0, (zenith - thresh_zen) / (maxzen - thresh_zen))` in pyspectral.

yukaribbba:
Maybe someday in the future we can look into pyspectral and use the DayNightCompositor technique to modify `reduce_rayleigh_highzenith`. But now for me I think I'm satisfied with the results here.

strandgren:
> To be honest I wrote the pyspectral PR so long ago that I can't remember what I had in mind! My gut feeling is that I preferred zenith angle rather than cosine space as this is easier for users to understand. Beyond that, I can't offer much insight I'm afraid!
> 
> Regarding what the DayNightCompositor does, this wasn't something I considered in my PR as I was focusing on daytime composites only and attempting to create a more natural daylight-to-darkness transition, so didn't test with composites that also have a night component.

Alright thanks, in that case we should perhaps make some examples with `sunz` and `cos(sunz)` and agree on a common solution. Lookin at the images in your pyresample PR I think that a scaling in the cos(sunz)-space could improve the results, by reducing the amount of blue at higher values of `strength`. I can try to have a look at this in the afternoon.

Just to clarify the idea of the DayNightCompositor; the way we talk about it here it's not used to blend a daytime composite with a nighttime composite, but instead it blends two daytime composites, with an without Rayleigh correction in order to reduce the correction at high zenith angels. You can see an example here: https://github.com/pytroll/satpy/blob/main/satpy/etc/composites/ahi.yaml#L303. In my opinion this achieves the same result as your (and this) PR, with the only difference that the DayNightCompositor does the scaling using the cosine of the angles.

strandgren:
> Maybe someday in the future we can look into pyspectral and use the DayNightCompositor technique to modify `reduce_rayleigh_highzenith`. But now for me I think I'm satisfied with the results here.

Ah I see, thanks for testing though! Yes, I will keep it in mind. Although I would rather say that we should use the methodology in this PR for reducing the Rayleigh correction rather than the DayNightCompositor technique.

yukaribbba:
@djhoese Any other suggestions? We’re almost done here.

djhoese (approved):
Looks good to me. Thanks for sticking with it and handling all of our reviews.

yukaribbba:
Thanks to all!

PR #2388, opened 2023-02-13
Fix gap between day/night-only composites

pnuu:
This PR was inspired by the discussion in https://github.com/pytroll/satpy/pull/2358#issuecomment-1400650541

With the current Satpy `main` there is discontinuity between a pair of day-only and night-only composites causing a dark band between them when stacked (in this case in Gimp using the mode "Merge"):

current_merged

The two composites were done using `DayNightCompositor` with `day_night: {day,night}_only` and `include_alpha: True`.

With the version of this PR, the result of stacking the two sub-composites is the following:

new_merged

The change introduced here is that *only* the transparency is adjusted on the SZA transition zone, while previously also the image date were changed. The scaling of image data caused stacking/blending outside of Satpy to in essence scale the data twice: once in Satpy and the other time when weighing the data with the transparencies of the overlapping images.

 - [x] Tests added (and adjusted)
 - [ ] Fully documented 

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -731,7 +731,7 @@ class DayNightCompositor(GenericCompositor):
 
         return coszen.clip(0, 1)
 
-    def _get_data_for_single_side_product(self, foreground_data, coszen):
+    def _get_data_for_single_side_product(self, foreground_data, weights):
         # Only one portion (day or night) is selected. One composite is requested.
         # Add alpha band to single L/RGB composite to make the masked-out portion transparent when needed
         # L -> LA
@@ -739,19 +739,19 @@ class DayNightCompositor(GenericCompositor):
         if self.include_alpha:
             foreground_data = add_alpha_bands(foreground_data)
         else:
-            coszen = self._mask_coszen(coszen)
+            weights = self._mask_weights(weights)
 
-        day_data, night_data = self._weight_single_side_data(foreground_data, coszen)
-        return day_data, night_data, coszen
+        day_data, night_data = self._weight_single_side_data(foreground_data, weights)
+        return day_data, night_data, weights
 
-    def _mask_coszen(self, coszen):
+    def _mask_weights(self, weights):
         if "day" in self.day_night:
-            return da.where(coszen != 0, coszen, np.nan).compute()
-        return da.where(coszen != 1, coszen, np.nan).compute()
+            return da.where(weights != 0, weights, np.nan).compute()
+        return da.where(weights != 1, weights, np.nan).compute()
 
-    def _weight_single_side_data(self, foreground_data, coszen):
+    def _weight_single_side_data(self, foreground_data, weights):
         if self.include_alpha:
-            weight = coszen if "day" in self.day_night else (1 - coszen)
+            weight = weights if "day" in self.day_night else (1 - weights)
             foreground_data[-1, :, :] = foreground_data[-1, :, :] * weight
         if "day" in self.day_night:
             return foreground_data, 0
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -689,79 +689,23 @@ class DayNightCompositor(GenericCompositor):
         # At least one composite is requested.
         foreground_data = projectables[0]
 
-        lim_low = np.cos(np.deg2rad(self.lim_low))
-        lim_high = np.cos(np.deg2rad(self.lim_high))
-        try:
-            coszen = np.cos(np.deg2rad(projectables[2 if self.day_night == "day_night" else 1]))
-        except IndexError:
-            from satpy.modifiers.angles import get_cos_sza
-            LOG.debug("Computing sun zenith angles.")
-            # Get chunking that matches the data
-            coszen = get_cos_sza(foreground_data)
-        # Calculate blending weights
-        coszen -= np.min((lim_high, lim_low))
-        coszen /= np.abs(lim_low - lim_high)
-        coszen = coszen.clip(0, 1)
+        weights = self._get_coszen_blending_weights(projectables)
 
-        # Apply enhancements
+        # Apply enhancements to the foreground data
         foreground_data = enhance2dataset(foreground_data)
 
         if "only" in self.day_night:
-            # Only one portion (day or night) is selected. One composite is requested.
-            # Add alpha band to single L/RGB composite to make the masked-out portion transparent when needed
-            # L -> LA
-            # RGB -> RGBA
-            if self.include_alpha:
-                foreground_data = add_alpha_bands(foreground_data)
-            else:
-                if "day" in self.day_night:
-                    coszen = da.where(coszen != 0, coszen, np.nan).compute()
-                else:
-                    coszen = da.where(coszen != 1, coszen, np.nan).compute()
-
-            # No need to replace missing channel data with zeros
-            # Get metadata
             attrs = foreground_data.attrs.copy()
-
-            # Determine the composite position
-            if "day" in self.day_night:
-                foreground_data[-1, :, :] = foreground_data[-1, :, :] * coszen
-                day_data = foreground_data
-                night_data = 0
-            else:
-                foreground_data[-1, :, :] = foreground_data[-1, :, :] * (1 - coszen)
-                night_data = foreground_data
-                day_data = 0
-
+            day_data, night_data, weights = self._get_data_for_single_side_product(foreground_data, weights)
         else:
-            # Both day and night portions are selected. Two composites are requested. Get the second one merged.
-            background_data = projectables[1]
-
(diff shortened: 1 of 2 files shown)


pnuu:
@yukaribbba could you have test if this would still work for you, as you added the `include_alpha` functionality in #2358 ?

djhoese:
It'd be interesting to see this comparison with a true_color as the day side RGB. I think my only concern with this result is that the resulting pixels in the transition range could be a combination of the two images rather than just one of the images. That would mean you could look at a pixel and it could be something that doesn't exist in either image...I think.

I guess it depends on how the user combines the images later on. If the software doing the merging does it based on the alpha then it might do some kind of weighting so you get this mixed/unrealistic pixel. If it does some kind of threshold `if image_a_alpha_value > image_b_alpha_value: ...` to choose which single RGB value to use.

djhoese:
Oh CodeScene doesn't like the complexity anymore :grimacing:

pnuu:
Yeah, I saw that there were places to do some refactoring, but wanted to show the first working version before starting on that.

yukaribbba:
@pnuu Had a quick test with AHI data. Two `day/night_only` composites, `include_alpha=True` -- True color for day and an IR band for night. However the gap still exists when you stack these two images together.

pnuu:
I noticed that with Gimp, the stacking mode made a difference. With "Normal" there is some left-over transparency, but with "Merge" the gap disappears with the images from this PR, like above. So I guess the way the stacking is done affects the result a lot.

My use-case would be with OpenGeoWeb, so maybe I need to ask the devs at work how transparencies in overlapping imagery are handled.

yukaribbba:
Well I use Photoshop to do that so I’m not sure about web services. Maybe your environment could have a different mechanism for reading data and transparency.

pnuu:
Well, there are still multiple ways to do the blending :-)

Gimp [documentation](https://docs.gimp.org/en/gimp-concepts-layer-modes.html) says this about the "Merge" mode:

> it assumes the source and destination are two parts of an original whole, and are therefore mutually exclusive. This is useful for blending cut and pasted content without artifacts, or for replacing erased content in general. 

This is essentially what we do in Satpy with `result = foreground * weight + background * (1 - weight)`. So as I now get the expected result (same as doing the day+night directly with Satpy) with that blending mode, I think this is the "correct" way.

But in any case, I need check how the merging is handled by other software.

pnuu:
Ah, now I got it why the "Normal" blending mode leaves the gap:

> The layer on top covers the layers below it. If you want to see anything below the top layer when you use this mode, the layer must have some transparent areas.

Lets say both of the layers have 50 % transparency. This means that 50 % of the bottom layer value is "seen through", which for the alpha channel would mean the result would still have a transparency of 75 %. And as the same would happen for the R, G and B channels, the overlap will have 75 % brightness compared to full addition, plus the transparency would cause more visible discontinuity.

For the current way in Satpy, the cos(SZA) transition is applied also to RGB channels, making the overlap even darker when the images are blended with either of these two method.

djhoese:
Another thing we might want to be careful of is invalid data where an alpha channel might show full transparency in the middle of the image. We don't want a fully valid image behind it to show up in this invalid area of the "top" image. I suppose if this was a problem we could zero-out RGB when A == 0.0.

pnuu:
We can already do that with `fill_value=0`, right?

djhoese:
:thinking: hm but then you wouldn't have the alpha band to do the merging properly in gimp/external program, right?

yukaribbba:
I start to understand what @pnuu needs. On your website, If a user chooses to stack a `day_only` image and a `night_only` image together he will see a perfectly merged day/night image instead of the current gap one, right? So all you need is two full images without being masked-out and a DayNight mask (which is this PR supposed to do). And the web engine will do the transparency calculation instead of satpy. The code be like:

```
If user chooses both "only" mode images:
    only apply DayNight mask on the upper layer, leave the bottom one alone, then calculate transparency
else:
    .............
```

yukaribbba:
Last night I was busy heading to bed:laughing:so I just had a simple test. And today comes the real problem, when `include_alpha=False`, I got some weird results:
image
Despite the yellow part, what I really expect is:
image

pnuu:
I created one day-only and one night-only layer on our test Geoserver with this PR and requested them together via the Layer preview system. It merged them just like if I had created the day/night composite directly with Satpy. This makes me very hopeful that also Geoweb will merge these layers correctly. I haven't yet been able to test this via Geoweb, possibly because the test Geoserver isn't visible in the internet.

pnuu:
> Last night I was busy heading to bedlaughingso I just had a simple test. And today comes the real problem, when `include_alpha=False`, I got some weird results

Damn. I'll have a look.

pnuu:
Meh, in my haste I also updated the `without_alpha` tests. I'll revert those and see how to make the code work again for those.

pnuu:
@yukaribbba maybe it now works properly also for `include_alpha=False`?

yukaribbba:
Looks like L mode is back to normal.
image

But rgb one still has that yellow ribbon.
image

pnuu:
Damn. Apparently the RGB variant wasn't tested. I'll get the correct values from `main` and update the tests.

pnuu:
@yukaribbba one more test please :-)

yukaribbba:
Both `include_alpha=True/False` and **L/RGB** mode are safe now. Well done! :smile:
image

pnuu:
Thanks! Now I just need to check what happens if there're no data areas in one of the composites on the transition zone.

pnuu:
This is what happens when there's a missing SEVIRI segment on the day-side composite.
hole

The result is the same whether done directly on Satpy or stacking in Gimp. And same result on `main` and in this PR.

I think the no-data area on day side should be transparent and the night-side should fade smoothly to transparent. Other thoughts on how it should look like?

pnuu:
And here's the day-only product that was used above.

hole_day-side

mraspaud (approved):
LGTM, thanks for refactoring the call of the day/night compositor along the way!

pnuu:
I'll leave this as-is. If someone comes up with a better way to handle missing data we can then revisit this in a new PR.

The last commit was just to rename a variable and a method to more meaningful name.

pnuu:
And merged `main` for the nc warning fix.

pnuu:
There's more to be done. This is what happens with the day-only and `include_alpha=True` for a polar satellite scene:

20221111_0854_Suomi-NPP_day_microphysical_day_only

The out-of-swath data should be transparent instead of being black.

pnuu:
Ok, if I use `scn.load(..., generate=True)` the out-of-swath areas are transparent as they should. But that's very wasteful when creating a lot of composites.

pnuu:
Forgot to mention that the black area is there also with `main`.

pnuu:
If I define the day-only and night-only composites with the SZAs as prerequisite for the polar case the images are fine:

20221111_0854_Suomi-NPP_truecolor_day_only

So adding masking to the SZAs computed on-demand should fix this.

pnuu:
Yep, adding the masking fixed the black area. Just need to come up with a test for this.

pnuu:
This actually fixes also the missing data area for the day-only image:
day-only_hole

And when merged externally produces the image I was hopingo for:
merged_hole

The downside is that the direct day/night isn't the same and has the missing data in black. I'll see if I can get this to work the same way with transparent missing data.

pnuu:
Making the direct day/night version to match the externally merged one is getting quite complicated, I need to work on this more tomorrow.

pnuu:
Ok, I _think_ this finally does what we want.

@yukaribbba if you could run one more test round, please :-)

If there is more need for cleanup or fixes, it's unlikely that I get to it in the next two weeks.

yukaribbba:
`include_alpha=True` passed. Unfortunately it seems `False` doesn't do any DayNight things at all...
image

pnuu:
Damn. Thanks for testing. I'll have a look in couple of weeks :see_no_evil:

yukaribbba:
Don't worry. We'll figure it out finally.:smile_cat:

pnuu:
Maybe now, @yukaribbba?

I noticed that:
- in few places I was passing the datasets as `comp((dset),)` instead of `comp((dset,))`
- the day/night only products were not masked at all
- the day-only product with `include_alpha: False) only seemed to work, because there's no light when it's dark
- the missing data coincided with the night-only pixel of `self.sza`, so the tests were not checking the correct thing. I swapped to the other dataset

yukaribbba:
Good. I'm going to bed now so I'll test it tomorrow.

yukaribbba:
@pnuu For me it's done. :smile:
image
image

pnuu:
@yukaribbba thanks again for testing!

PR #2391, opened 2023-02-14
Increase warnings stacklevel

sfinkens:
Increase warnings stacklevel  `warnings.warn(..., stacklevel=2)` as recommended by flake8 ([example](https://github.com/pytroll/satpy/actions/runs/4173011657/jobs/7224821751))

### DIFF
--- a/satpy/composites/config_loader.py
+++ b/satpy/composites/config_loader.py
@@ -131,9 +131,12 @@ class _ModifierConfigHelper:
             loader = modifier_info.pop('modifier', None)
             if loader is None:
                 loader = modifier_info.pop('compositor')
-                warnings.warn("Modifier '{}' uses deprecated 'compositor' "
-                              "key to point to Python class, replace "
-                              "with 'modifier'.".format(modifier_name))
+                warnings.warn(
+                    "Modifier '{}' uses deprecated 'compositor' "
+                    "key to point to Python class, replace "
+                    "with 'modifier'.".format(modifier_name),
+                    stacklevel=2
+                )
         except KeyError:
             raise ValueError("'modifier' key missing or empty for '{}'. Option keys = {}".format(
                 modifier_name, str(modifier_info.keys())))
--- a/satpy/composites/config_loader.py
+++ b/satpy/composites/config_loader.py
@@ -135,7 +135,7 @@ class _ModifierConfigHelper:
                     "Modifier '{}' uses deprecated 'compositor' "
                     "key to point to Python class, replace "
                     "with 'modifier'.".format(modifier_name),
-                    stacklevel=2
+                    stacklevel=5
                 )
         except KeyError:
             raise ValueError("'modifier' key missing or empty for '{}'. Option keys = {}".format(
--- a/satpy/modifiers/angles.py
+++ b/satpy/modifiers/angles.py
@@ -186,7 +186,7 @@ class ZarrCacheHelper:
                 "future calculations. "
                 f"Original chunks: {arg_chunks}; New chunks: {new_chunks}",
                 PerformanceWarning,
-                stacklevel=2
+                stacklevel=3
             )
 
     def _cache_results(self, res, zarr_format):
--- a/satpy/modifiers/angles.py
+++ b/satpy/modifiers/angles.py
@@ -185,7 +185,9 @@ class ZarrCacheHelper:
                 "has been rechunked for caching, but this is not optimal for "
                 "future calculations. "
                 f"Original chunks: {arg_chunks}; New chunks: {new_chunks}",
-                PerformanceWarning)
+                PerformanceWarning,
+                stacklevel=2
+            )
 
     def _cache_results(self, res, zarr_format):
         os.makedirs(os.path.dirname(zarr_format), exist_ok=True)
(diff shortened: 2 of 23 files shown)


satpy/composites/config_loader.py
```diff
@@ -131,9 +131,12 @@ def _get_modifier_loader_from_config(modifier_name, modifier_info):
...
             if loader is None:
                 loader = modifier_info.pop('compositor')
-                warnings.warn("Modifier '{}' uses deprecated 'compositor' "
-                              "key to point to Python class, replace "
-                              "with 'modifier'.".format(modifier_name))
+                warnings.warn(
+                    "Modifier '{}' uses deprecated 'compositor' "
+                    "key to point to Python class, replace "
+                    "with 'modifier'.".format(modifier_name),
+                    stacklevel=2
```
djhoese, line 138: This should probably be at least stacklevel 3 shouldn't it? This is a helper class inside the config loading so we probably want the caller of the caller of this method? Oh actually this method is called from the `_load_config_modifier` method in this same class and that one from another method so maybe stacklevel 5?

satpy/modifiers/angles.py
```diff
@@ -185,7 +185,9 @@ def _warn_if_irregular_input_chunks(args, modified_args):
                 "has been rechunked for caching, but this is not optimal for "
                 "future calculations. "
                 f"Original chunks: {arg_chunks}; New chunks: {new_chunks}",
-                PerformanceWarning)
+                PerformanceWarning,
+                stacklevel=2
```
djhoese, line 189: Since this is called from `__call__`, maybe do 3?

satpy/modifiers/parallax.py
```diff
@@ -389,7 +389,9 @@ def _check_overlap(self, cth_dataset):
         """
         warnings.warn(
             "Overlap checking not impelemented. Waiting for "
-            "fix for https://github.com/pytroll/pyresample/issues/329")
+            "fix for https://github.com/pytroll/pyresample/issues/329",
+            stacklevel=2
```
djhoese, line 393: 3?

satpy/modifiers/parallax.py
```diff
@@ -389,7 +389,9 @@ def _check_overlap(self, cth_dataset):
         """
         warnings.warn(
             "Overlap checking not impelemented. Waiting for "
```
djhoese, line 391: While we're at it:

```suggestion
            "Overlap checking not implemented. Waiting for "
```

satpy/readers/__init__.py
```diff
@@ -180,10 +180,12 @@ def _get_file_keys_for_reader_files(reader_files, group_keys=None):
...
                     warnings.warn(
-                            f"Found matching file {f:s} for reader "
-                            "{reader_name:s}, but none of group keys found. "
-                            "Group keys requested: " + ", ".join(group_keys),
-                            UserWarning)
+                        f"Found matching file {f:s} for reader "
+                        "{reader_name:s}, but none of group keys found. "
+                        "Group keys requested: " + ", ".join(group_keys),
+                        UserWarning,
+                        stacklevel=2
```
djhoese, line 187: 3

satpy/readers/ahi_hsd.py
```diff
@@ -458,7 +458,10 @@ def _modify_observation_time_for_nominal(self, observation_time):
         """
         timeline = "{:04d}".format(self.basic_info['observation_timeline'][0])
         if not self._is_valid_timeline(timeline):
-            warnings.warn("Observation timeline is fill value, not rounding observation time.")
+            warnings.warn(
+                "Observation timeline is fill value, not rounding observation time.",
+                stacklevel=2
```
djhoese, line 463: 3 maybe? That way it shows the function using the start_time property, right?

djhoese (commented):
I made a couple suggestions and hoping the same logic can be used for the rest of the changes. I think a lot of these warnings show up in helper methods or non-user-facing functions so doing stacklevel=2 still doesn't give the user much context. What do you think?

sfinkens:
Thanks for your suggestions! You're right, just blindly setting the level to 2 isn't helpful. I'll look into it again.

sfinkens:
After playing around with the stacklevel I'm a bit undecided. I actually find stacklevel=1 helpful sometimes, because I can click on the trace and my IDE opens exactly that line where the warning came from. With stacklevel > 1 that place is harder to find. On the other hand more context is also valuable. 

I tried to come up with a compromise: Set stacklevel to the next public (or __special__) method up the calltree that's still in the same file. So when clicking on the trace you get to the relevant file at least. If the method is called from multiple places, keep stacklevel=2.

Obviously this also depends on the implementation and I'm not sure I like that. Does anyone check stacklevels of warnings when refactoring?

pnuu:
Is this something that will fix the suddenly appearing failures on CI for recently updated PRs?

djhoese:
> Is this something that will fix the suddenly appearing failures on CI for recently updated PRs?

Yes @pnuu.

> Obviously this also depends on the implementation and I'm not sure whether I like that. Does anyone check stacklevels of warnings when refactoring?

No, I know I should have used stacklevels more, but you're right I definitely don't do it when refactoring although I'm not sure I encounter warnings too much.

Thinking about your comment about IDE clicking makes me rethink how we approach this. Are the warnings for developers of the library or for users. I think it is for users who we don't want to have to dive into the low-level code. If the warning is clear enough they probably shouldn't need to worry about the actual code (ex. "the thing you provided is bad" -> "Oh, I shouldn't pass that thing like that").

mraspaud (approved):
LGTM

PR #2393, opened 2023-02-16
Enhanced selecting with weights (PR 2275) to handle composites. Added…

lobsiger:
… true blending with weights for single channels and composites.

Multiscene blending revisited

The code of PR 2275 appeared first in Satpy 0.40.
I was mislead by its titel as this was actually
about selecting (choosing) pixels with weights.

This has been developed by Adam for combining 'ct'
data of GEO and LEO satellites up north. My use of the
code was for multipass images of LEOs with identical
sensors. I have enhanced PR 2275 to handle composites.

I added a function for true blending of overlapping
datasets. This also works with common RGB composites.

I propose this naming:

default)
multiscene "stacking"  
(the original way, datasets put on top of each other)

1)
multiscene "selecting" with weights 
(pixel with highest weight is selected from one dataset)

2)
multiscene "blending" with weights 
(pixels blended,  more than one dataset may contribute)

"Stacking" is the default. When working with multiscene and
a list with weights is provided the user has an additional
way to select case 1) or 2) (and maybe future options ...?).
This is handled by integer keyword argument 'blending_type'.

For 1) I propose to document/use the weighting function
w = 180 - satellite_zenith_angle

For 2) I propose a function similar to the flight track
dispersion formula as used in the German 'AzB'
 
         weights=[]
            for s in new_mscn.scenes:
                w = s['satellite_zenith_angle']
                w = w / w.max(dim = ['x', 'y'])
                w = (1. - w*w)
                weights.append(w*w)

Where new_mscn is the multiscene resampled to the area.
This will certainly not work for 'ct', 'cma' type data.

NOTE: I'm absolutely new to this array stuff and will
certainly not be able to write meaningful tests. I just
hope that Adam's tests for PR 2275 can be used somehow.

NOAA20-20230213-NIG-0214-night_overview-westminster-multi-blended
NOAA20-20230213-NIG-0214-night_overview-westminster-multi-selected
NOAA20-20230213-NIG-0214-night_overview-westminster-multi-stacked

Cheers,
Ernst

### DIFF
--- a/satpy/multiscene.py
+++ b/satpy/multiscene.py
@@ -139,13 +139,13 @@ def set_weights_to_zero_where_invalid(datasets, weights):
                 weights[i] = xr.where(dataset[0] == dataset.attrs["_FillValue"], 0, weights[i])
             except KeyError:
                 weights[i] = xr.where(dataset[0].isnull(), 0, weights[i])
-    else:
-        for i, dataset in enumerate(datasets):
-            try:
-                weights[i] = xr.where(dataset == dataset.attrs["_FillValue"], 0, weights[i])
-            except KeyError:
-                weights[i] = xr.where(dataset.isnull(), 0, weights[i])
+    return weights
 
+    for i, dataset in enumerate(datasets):
+        try:
+            weights[i] = xr.where(dataset == dataset.attrs["_FillValue"], 0, weights[i])
+        except KeyError:
+            weights[i] = xr.where(dataset.isnull(), 0, weights[i])
     return weights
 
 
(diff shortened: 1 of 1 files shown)


lobsiger:
@mraspaud @djhoese it's some time back I made a contribution. Sorry but I do not understand the failing tests. Do I have to provide two functions _stack_blended(), one for single channels and one for composites? Or do I simply have to move the "if dims[0] == 'bands':" outside the nested loops and make two blocks,  one for single channels and one for composites? I can certainly not lower the code line count of multiscene.py below 600! Any help welcome.

lobsiger:
@mraspaud @djhoese I will (try to close, delete?) this and begin from scratch with an all flat version that I test again locally (the way I understand) . I screwed up my branch to the point github does not accept commits anymore but ends in errors. I think I understood your tests do not like nested and complex stuff.

lobsiger:
O,K. I press the button and will delete my %&*?^£ branch .

PR #2394, opened 2023-02-17
Add weighted blended stacking to MultiScene (fixes multi-band handling)

lobsiger:
Multiscene blending revisited

This is a second try after my PR 2393 went wrong.

The code of PR 2275 appeared first in Satpy 0.40.
I was mislead by its titel as this was actually
about selecting (choosing) pixels with weights.

This has been developed by Adam for combining 'ct'
data of GEO and LEO satellites up north. My use of the
code was for multipass images of LEOs with identical
sensors. I have enhanced PR 2275 to handle composites.

I added a function for true blending of overlapping
datasets. This also works with common RGB composites.

I propose this naming:

default)
multiscene "stacking"  
(the original way, datasets put on top of each other)

1)
multiscene "selecting" with weights 
(pixel with highest weight is selected from one dataset)

2)
multiscene "blending" with weights 
(pixels blended,  more than one dataset may contribute)

"Stacking" is the default. When working with multiscene and
a list with weights is provided the user has an additional
way to select case 1) or 2) (and maybe future options ...?).
This is handled by integer keyword argument 'blending_type'.

For 1) I propose to document/use the weighting function
w = 180 - satellite_zenith_angle

For 2) I propose a function similar to the flight track
dispersion formula as used in the German 'AzB'
 
         weights=[]
            for s in new_mscn.scenes:
                w = s['satellite_zenith_angle']
                w = w / w.max(dim = ['x', 'y'])
                w = (1. - w*w)
                weights.append(w*w)

Where new_mscn is the multiscene resampled to the area.
This will certainly not work for 'ct', 'cma' type data.

NOTE: I'm absolutely new to this xarray stuff and will
certainly not be able to write meaningful tests. I just
hope that Adam's tests for PR 2275 can be used somehow.

I attach 3 examples of MetopB multipass over 'eurol10'.
A single "channel" image of weights has been has been
produced at the same time (old busy Xeon, 24GB of RAM).

"stacked"     was produced in    2 Minutes 23 Seconds
"selected"    was produced in    9 Minutes 33 Seconds
"blended"    was produced in    2 Minutes 16 Seconds

Cheers,
Ernst

MetopB-20230216-DAY-0929-natural_color-eurol10-multi-stacked
MetopB-20230216-DAY-0929-satellite_zenith_angle-eurol10-multi-stacked
MetopB-20230216-DAY-0929-natural_color-eurol10-multi-selected
MetopB-20230216-DAY-0929-satellite_zenith_angle-eurol10-multi-selected
MetopB-20230216-DAY-0929-natural_color-eurol10-multi-blended
MetopB-20230216-DAY-0929-satellite_zenith_angle-eurol10-multi-blended


lobsiger:
@mraspaud @djhoese @adybbroe so far so good. I cannot help with the 600 line limit. Ans *YES* this was complex (for me). Regarding "code duplication": In my first (and closed) PR 2393 the problem was if if nesting stuff in two functions. Now I took those functions apart to four to flatten nesting and the new complaint is about code duplication. As I already said: This is all at the edge of my knowledge. I can certainly not help should more multiscene unit tests have to be added. I need your help.

lobsiger:
I wrote a few lines for blending GOES-16 and GOES-18 images to get a full view of North America.
I wonder what   get_satellite_zenith_angle()   from   satpy.modifiers.angles   gives back in free space?
For my AzB weighting function wmax = w.max(dim = ['x', 'y']) did not work. I had to set wmax = 90.

GOESX-true_color-namerica5km-blended
GOESX-true_color-namerica5km-selected
GOESX-true_color-namerica5km-stacked

satpy/multiscene.py
```diff
@@ -69,8 +80,86 @@ def stack(datasets, weights=None, combine_times=True):
...
+    dims = datasets[0].dims
+    attrs = combine_metadata(*[x.attrs for x in datasets])
+
+    if combine_times:
+        if 'start_time' in attrs and 'end_time' in attrs:
+            attrs['start_time'], attrs['end_time'] = _get_combined_start_end_times(*[x.attrs for x in datasets])
+
+    total = weights[0].copy() + 1.e-9
+    for n in range(1, len(weights)):
+        total += weights[n]
```
mraspaud, line 96:
```suggestion
    total = sum(weights)
```
the sum function even has a "start" argument, where you could put the offset you have here, like `total = sum(weights, start=1.e-9)`
However, I think this is not necessary. A division by zero later on will result in a NaN value, which will then be replaced with the `fillna` call.

satpy/multiscene.py
```diff
@@ -69,8 +80,86 @@ def stack(datasets, weights=None, combine_times=True):
...
+
+    datasets0 = []
```
mraspaud, line 98: this variable should be renamed to something more understandable...

satpy/multiscene.py
```diff
@@ -69,8 +80,86 @@ def stack(datasets, weights=None, combine_times=True):
...
+    for n in range(0, len(datasets)):
+        weights[n] /= total
+        datasets0.append(datasets[n].fillna(0))
+        datasets0[n][0] *= weights[n]
+        datasets0[n][1] *= weights[n]
+        datasets0[n][2] *= weights[n]
```
mraspaud, line 104:
```suggestion
    for weight, dataset in zip(weights, datasets):
        weight /= total
        datasets0.append(dataset * weight)
```
I think the array broadcasting mechanism should take care of the weight multiplication. If that works, it would probably mean that we can merge the single and mulitple band cases into one function, right?

Regarding the `fillna` call, I think we should be careful here about what we use as a fill value. In particular, it's probably good to check if the dataset as a defined `_FillValue` attribute already.

satpy/multiscene.py
```diff
@@ -69,8 +80,86 @@ def stack(datasets, weights=None, combine_times=True):
...
+
+    base = datasets0[0].copy()
+    for n in range(1, len(datasets0)):
+        base += datasets0[n]
```
mraspaud, line 108:
```suggestion
    base = sum(datasets0)
```

mraspaud (changes requested):
@lobsiger thanks a lot for adding this feature, it is very nice results you get!
I have just started with the review, but maybe we can start with that?
There is a lot of places which are flagged as not tested indeed. I understand that Xarray can be a bit frightening, how about starting with tests using regular numpy arrays first, that we can then adapt to Xarray?

lobsiger:
@mraspaud   thanks a lot for taking the time to look into my code. Be aware that I started with Python 2018 and I normally do my hobby programming in C the way I did in the 1970es. I have never coded a Python class so far and just had a first glance at numpy and xarray. So my results might look good but the code is mostly try and error and shows my missing knowledge in Python (I just yesterday understood enumerate()). I also have to confess that I do not understand all the (impressive) tests and that the way I understand and use git is a botch. So feel free to hack my code to bits. All I can add is that with using my weights w, w*w (AzB), w*w*w or w*w*w*w you can tighten the blending zone probably with Adam's selection images as limiting case.

mraspaud:
> @mraspaud thanks a lot for taking the time to look into my code. Be aware that I started with Python 2018 and I normally do my hobby programming in C the way I did in the 1970es. I have never coded a Python class so far and just had a first glance at numpy and xarray. So my results might look good but the code is mostly try and error and shows my missing knowledge in Python (I just yesterday understood enumerate()). I also have to confess that I do not understand all the (impressive) tests and that the way I understand and use git is a botch. So feel free to hack my code to bits. All I can add is that with using my weights w, w_w (AzB), w_w_w or w_w_w_w you can tighten the blending zone probably with Adam's selection images as limiting case.

No problem @lobsiger, we all have our backgrounds and I understand that you come from the C world, and that's perfectly fine. Hence my comments in the review regarding how to do things best in python and numpy/xarray. Feel free to check them out, python can be a fantastic tool once you tame it :)

As for the tests, I was thinking something as simple as a creating two small weight arrays and two small datasets by hand, and ensuring that the result is as you expect.
For example something like this:
```python
weights_1 = np.array([[0, 1, 2], [0, 1, 2], [0, 1, 2]])
weights_2 = np.array([[1.1, 1.1, 1.1], [1.1, 1.1, 1.1], [1.1, 1.1, 1.1]])
dataset_1 = np.full((3, 3), 1)
dataset_2 = np.full((3, 3), 2)

expected = np.array([[2, 2, 1], [2, 2, 1], [2, 2, 1]])
result = stack([dataset_1, dataset_2], [weights_1, weights_2]])

np.testing.assert_array_equal(result, expected)
```
Then you can do the same with the other variations of stack, blend, select.

lobsiger:
@mraspaud why are the tests cancelled? Did I press some wrong button or is it just some overload of github?
I have commited your suggestions and some more. Unfortunately I still don't see how I can merge the two set_weights_to_zero_where_invalid() functions as these treat datasets with different dimensions from weights. 
I'm interested in your proposal below especially (I have no clue) if this can replace the use of partial() in scripts.
Regarding the .fillna(0) : If I omit that only the portion where all datasets overlap is returned with ugly edges.

"
However, I wonder if it would be clearer or cleaner to have separate methods for this, so that instead of calling eg
multiscene.stack(datasets, weights=five_kilos, blend_type="select_with_weights")
we could just call
multiscene.select(datasets, weights=five_kilos)
"

MSGX-overview-moll-stacked
MSGX-overview-moll-blended
MSGX-overview-moll-selected

pnuu:
> @mraspaud why are the tests cancelled? Did I press some wrong button or is it just some overload of github?

The rest of the tests were cancelled after the first one crashed.

You can see the failure messages for this run here: https://github.com/pytroll/satpy/actions/runs/4234069951/jobs/7355886611#step:10:191

It seems the expected and actual result matrices are a bit different, and `'sensor'` entry isn't in the `result.attrs` dictionary.

lobsiger:
@mraspaud quite sure it will fail again. Is this because a test of Adam for selected weighting is used that was designed for his 'ct' , 'cma' integer data only and had no coords=coords that I must add for datasets with bands? This is completely over my head ...

lobsiger:
@pnuu yes I think this is Adam's test for 'ct' data. It worked when I had two functions for selected weighting, one for single channel data that probably passed Adam's test and one for data with bands where no test existed yet? Not 100% sure what's going on though ...

pnuu:
Yeah, sorry, haven't really followed this PR so don't know what's going on either :sweat_smile: 

The existing tests shouldn't start failing on new PRs, which would mean that the current behavior is changed. Unless that's the purpose, this shouldn't happen, and if this is the purpose, these tests should be updated.

@adybbroe could you check this?

lobsiger:
I'm stuck.

djhoese:
I'm really swamped this week, but I will try to look at this when I get a chance. So far the related mailing list thread and this pull request have moved too fast for the available time I have so I'm very behind. I'll need to find time to read some of the conversation.

lobsiger:
@djhoese thanks a lot for your time. In a nutshell: Adam started this as PR2275 for his blending with weights 'ct' and 'cma' type files for GEO and LEO satellites up north. I wanted to use it for multi pass RGB composites of LEO satellites  and had a longer monologue on the google list. I made PR2093 messed it up, closed it and reopened this PR2094  with a lot of try and error. At last most of the tests seemed to pass but there were complaints about too many code lines, too much complexity and untested function lines. @mraspaud made a couple of very useful suggestions that enabled me to write the much cleaner code we have now. I think that this is the first time Adam's stacked multiscene with weights unit tests are invoked and of course fail as they are not designed for what I do here. I cannot be of further help regarding these unit tests as all that code is well over my head.

lobsiger:
@mraspaud @djhoese @pnuu I think the main remaining problem with codecov/patch is the renaming/replacing of Adam's stacking with weights function _stack_weighted() with the new name _stack_selected(), the addition of _stack_blended() and my inability to combine set_weights_to_zero_where_invalid() with the case with bands set_weights_to_zero_where_invalid_red(), where I only check the 'R' band. I have described the problem and asked for help on the google February 6th 2023 list:

https://groups.google.com/g/pytroll/c/N20hwYVYI2s/m/t3qjXxyUAgAJ

An of course the main problem is my kind of PR that replaces just too much at a time and thus is also hard to follow :-\.

satpy/multiscene.py
```diff
@@ -46,18 +46,26 @@
...
     By default, datasets are stacked on top of each other, so the last one applied is
-    on top. If a sequence of weights arrays are provided the datasets will
-    be combined according to those weights. The result will be a composite
-    dataset where the data in each pixel is coming from the dataset having the
-    highest weight.
+    on top. If a sequence of weights (with equal shape) is provided, the datasets will
+    be combined according to those weights. Datasets can be integers like 'ct', 'cma',
+    or radiances single channel or RGB composites. In the later case weights is applied
+    to each 'R', 'G', 'B' coordinate in the same way. The result will be a composite
+    dataset where each pixel is constructed in a way depending on variable blend_type.
```
djhoese, line 57:
```suggestion
    be combined according to those weights. Datasets can be integer category products
    (ex. cloud type), single channels (ex. radiance), or RGB composites. In the
    latter case, weights is applied
    to each 'R', 'G', 'B' coordinate in the same way. The result will be a composite
    dataset where each pixel is constructed in a way depending on ``blend_type``.
```

I'd prefer if the documentation here avoided non-standard/reader-specific abbreviations like "ct" (I don't even know what "cma" is). I reworded your original paragraph, but I think it gets the same point across.

Could we also document the individual blend types here?

Note: I'm still reviewing everything here.

lobsiger: @djhoese AFAIK 'cma' is cloud mask 'ctth' cloud top temperature height (?). Cloud type 'ct' is basically what Adam started with. I think the current documentation needs some overhaul as well. At least it was not easy to follow for me. Two things come to my mind: I propose that we do not use negative weights. If we allow for those Adam can simply set his weights to -satellite_zenith_angle  or to 90-satellite_zenith_angle which is satellite_elevation. If only positive weights are allowed Adam could use 180-satellite_zenith_angle. All these give the same results if the pixels are selected but these weights are more understandable than 1./satellite_zenith_angle. Sorry, for the missing tests I can probably be of no help. But I think it should be possible to start with Adam test as a basis.

djhoese (commented):
Ok so it looks like all the old/existing tests pass. The "experimental" environment which installs unstable/dev versions of some of our dependencies is currently expected to fail due to an issue that someone else is/was going to fix (I might just fix that first before doing anything on this PR). Or are there still tests that aren't being run?

The CodeScene stuff...yeah I can clean that up.

Tests: Yeah I can try to add some eventually. Is there anything missing from this summary of this PR:

This PR fixes support for MultiScene blending with multi-band products. It also adds the ability to stack layers with a blending effect between the layers based on weights. This is essentially a weighted average, right? Previously only a single pixel was chosen based on weight value.

satpy/multiscene.py
djhoese: Ok on the abbreviations. My preference still stands. I'd prefer not to have docstrings refer to products that aren't self-explanatory or are reader-specific. Like `true_color` is obviously (at least hopefully for most Satpy users) expresses the general idea of an RGB composite.

Regarding negative weights, I haven't looked at the actual math, but does it currently (as written in this PR) work with negative weights or is that broken? What would change code-wise in multiscene.py if only positive weights were supported? What does this buy us (cleaner code? better performance?)? Would this then only require users (this is a brand new feature so I'm not worried about backwards compatibility much) to change how they generate weights? You mentioned it being more understandable than `1 / satellite_zenith_angle`, is this only used in the documentation for this weighting functionality or is it used somewhere else in Satpy?

lobsiger: As I understand/assume 1./satellite_zenith_angle was just "invented" by Adam because he wanted a weight that decreases with increasing satellite_zenith_angle.  First I began with 1./(1.+ satellite_zenith_angle) to avoid division by zero. But finally I found out that any function that decreases with increasing satellite_zenith_angle leads to the same 'selected' result. No idea whether 1./satellite_zenith_angle is used elsewhere in satpy.

lobsiger: Just tested weights = - satellite_zenith_angle with Adams selected version enhanced for bands: DOES NOT WORK :-). For blending negative weights this might lead to negative radiance or maybe byte underflow  (?). Certainly no option I think

MetopB-20230301-DAY-0959-natural_color-eurol10-multi-blended
.

lobsiger: The same three MetopB passes with weights = (180. - satellite_zenith_angle) work fine for blend_type='select_with_weights'.

MetopB-20230301-DAY-0959-natural_color-eurol10-multi-blended

satpy/multiscene.py
```diff
@@ -46,18 +46,26 @@
...
 
     """
-    if weights:
-        return _stack_weighted(datasets, weights, combine_times)
+    bands = datasets[0].dims[0] == 'bands'
```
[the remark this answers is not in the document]
lobsiger: @djhoese  regarding your question above "**_This is essentially a weighted average, right?_** " the answer is YES. But for blend_type = "blend_with_weights" I also use my special "AzB" type weights that go to zero towards the boundaries of available data. Otherwise we would still see e.g. the ends of the scan lines for multi pass LEO images. In other words for blend_type = "blend_with_weights" a setting like (above) weights = (180. - satellite_zenith_angle") makes no sense.

lobsiger:
@djhoese @mraspaud I have put some short blending test scripts on wetransfer

https://groups.google.com/g/pytroll/c/N20hwYVYI2s/m/YkBEyKLPAwAJ

NWCSAF still does not work. I wonder whether this is somehow related to #2403

lobsiger:
The results where I only get a triangle on top are not very reproducible. They seem to depend on what I did before or whether I have weights strictly <= 1.0 . Maybe an integer rounding or a scene caching thing (just thinking out loud).

https://groups.google.com/g/pytroll/c/N20hwYVYI2s/m/xhkUHyqIAgAJ

lobsiger:
It seems the timing of #2275, #2394 and #2403 was done by Murphy. I hesitate to open a new satpy issue for
NWCSAF cloud data at the moment. I have documented my findings regarding the problems we currently face:

https://groups.google.com/g/pytroll/c/n_wvjJwbOsY

Unless it is possible to make NWCSAF LEO composites of
Metop-B/C work and look like NWCSAF GEO MSG2/3/4 
composites I propose to leave those aside of this #2394.

satpy/multiscene.py
```diff
@@ -69,20 +78,52 @@ def stack(datasets, weights=None, combine_times=True):
...
+    return selected_array
+
+
+def set_weights_to_zero_where_invalid_red(datasets, weights):
+    """Go through the weights and set to pixel values to zero where corresponding datasets[0] are invalid."""
+    for i, dataset in enumerate(datasets):
+        try:
+            weights[i] = xr.where(dataset[0] == dataset.attrs["_FillValue"], 0, weights[i])
+        except KeyError:
+            weights[i] = xr.where(dataset[0].isnull(), 0, weights[i])
```
djhoese, line 124: @lobsiger I'm refactoring this while also writing tests, but noticed that this says `datasets[0]`, should this be `datasets[i]`? Don't worry about fixing it yourself if this is wrong, I'll do it.

djhoese: Oh! This is only looking at the red channel. Why?

satpy/multiscene.py
```diff
@@ -46,54 +48,99 @@
...
 
-def _stack_weighted(datasets, weights, combine_times):
-    """Stack datasets using weights."""
-    weights = set_weights_to_zero_where_invalid(datasets, weights)
 
+def _stack_select_by_weights(datasets, weights, combine_times):
+    """Stack datasets selecting pixels using weights."""
     indices = da.argmax(da.dstack(weights), axis=-1)
+    if "bands" in datasets[0].dims:
+        indices = [indices] * datasets[0].sizes["bands"]
```
djhoese, line 122: @lobsiger I refactored this so that regardless of the number of bands in the DataArray (L, LA, RGB, RGBA) it would have indices for each one which I thought was what you wanted here. I'm realizing now, should this ignore alpha bands?

djhoese:
I'll try to work on this PR more tonight or maybe this weekend if I find the time. My plan is to add tests and get good coverage then refactor the multiscene into a subpackage with something like:

```
satpy/
  multiscene/
    __init__.py
    _multiscene.py
    _blend_funcs.py
```

and maybe I'll separate the animation and save_datasets handling into a separate module too. The point is, I'll have at least a couple more commits to do here.

lobsiger:
@djhoese thanks for looking into this. I have been pretty much absorbed for days testing your overlay cache improvements.

**I refactored this so that regardless of the number of bands in the DataArray (L, LA, RGB, RGBA) it would have indices for each one which I thought was what you wanted here. I'm realizing now, should this ignore alpha bands?**

@djhoese TBH I was thinking about RGBA but had no clue under what condition I would get such or even other kinds of bands. IIRC all my LEO satellites worked and I found only RGB data. The NWC cloud data seems still be changing after the move of the compositors. But of course this PR should (if possible at all) finally work for all sorts of bands that we get in a resampled scene.

satpy/multiscene.py
lobsiger: > Oh! This is only looking at the red channel. Why?

@djhoese this is simply because I was unable to write an accepted function that treated all available bands properly. So I just copied the code @adybbroe wrote for single band cloud data and looked at the RED band only. My (probably wrong) assumption was that either none or all bands would be null or have the _FillValue. At least it seemed to correct OLCI ERR data images. Without this function those images had black lines at the ends (left/right) of the swath of each satellite pass.

satpy/multiscene.py
```diff
@@ -46,54 +48,99 @@
 log = logging.getLogger(__name__)
 
 
-def stack(datasets, weights=None, combine_times=True):
-    """Overlay a series of datasets together.
+def stack(datasets, weights=None, combine_times=True, blend_type='select_with_weights'):
```
djhoese, line 51: I'm going through this code and trying to consolidate logic. I'm having trouble with this `combine_times` argument from @adybbroe's PR. @adybbroe any memory of why you/we had this `combine_times` keyword argument instead of just always doing it? I think @gerritholl had filed a bug report about regular stacking not combining the times already. I think I'll need to at least make the "no weights" case use this.

satpy/multiscene.py
```diff
@@ -46,66 +48,140 @@
...
-def _stack_weighted(datasets, weights, combine_times):
-    """Stack datasets using weights."""
-    weights = set_weights_to_zero_where_invalid(datasets, weights)
+    overlays = []
+    for weight, overlay in zip(weights, datasets):
+        overlays.append(overlay.fillna(0) * weight)
 
-    indices = da.argmax(da.dstack(weights), axis=-1)
-    attrs = combine_metadata(*[x.attrs for x in datasets])
+    base = sum(overlays) / sum(weights, start=1.e-9)
```
djhoese, line 123: @lobsiger I'm not sure I agree with this `start` parameter as a way to handle dividing by 0. What do we think the value should be if weights result to `0`?

There are different ways to handle a divide by 0, but I'm kind of leaning towards letting it be NaN in the final result.

djhoese:
Ok remaining TODO:

- [ ] Add additional tests using pytest fixtures
  - [ ] Add RGB/A tests
  - [ ] Add float input tests
- [ ] Decide on behavior of `/ 0` weights - NaN or other?
- [ ] Refactor multiscene.py into sub-modules (_stack_funcs.py, _iterate.py, etc)

satpy/multiscene.py
lobsiger: @djhoese  I agree that we should not have start=1e-6 when summing the weights in the denominator.
Instead we should divide by 0 outside the data area of passes, probably producing 0/0=None (?).
I took the latest version of your multiscene.py and tested all possible combinations of three
MetopB passes over area eurol10 similar to what I did here:

https://github.com/pytroll/pycoast/issues/95#issuecomment-1494668797

I made images for composites 'natural_color', 'ir108_3d' and channel '4' (IR 10.8, used for 'ir108_3d').
I allowed for all 3 defined blend types, generate=True/False, fill_value=0/255/None. This resulted in
3 x 3 x 2 x 3 = 54 different image files without setting start in the denominator. All these images
looked as expected. Then I set sum(weights, start=1e-6) and made 18 more images for 'blend_with_weights'.
All these 18 files are problematic, not discovered so far because I mainly looked at RGB composites
with fill_value=0. I attach 3 examples left is image as expected, right is wrong image with start=1e-6.
Original files produced are *.png, but I reduced them in size and changed them to *.jpg to save space.


Problems with 'blend_with_weights', sum(weights, start=1e-6), generate=False/True does not matter:

'4' Data Pale (IR 10.8, almost white) fill_value0     No_Data region is BLACK (O.K., but caused by 1e-6)
'4' Data Pale (IR 10.8, almost white) fill_value255   No_Data region is BLACK (instead of white)
'4' Data Pale (IR 10.8, almost white) fill_valueNone  No_Data region is BLACK (instead of transparent)

'ir108_3d' Data Dark (almost black)   fill_value0     No_Data region is WHITE (instead of black, reversed)
'ir108_3d' Data Dark (almost black)   fill_value255   No_Data region is WHITE (O.K., but caused by 1e-6)
'ir108_3d' Data Dark (almost black)   fill_valueNone  No_Data region is WHITE (instead of transparent)

'natural_color' Data look as expected fill_value0     No_Data region is BLACK (O.K., but caused by 1e-6)
'natural_color' Data look as expected fill_value255   No_Data region is BLACK (instead of white)
'natural_color' Data look as expected fill_valueNone  No_Data region is BLACK (instead of transparent)

MetopB-4-eurol10-blend_with_weights-generateTrue-fill_value255-startNoneMetopB-4-eurol10-blend_with_weights-generateTrue-fill_value255-start1e-6

MetopB-ir108_3d-eurol10-blend_with_weights-generateTrue-fill_value0-startNoneMetopB-ir108_3d-eurol10-blend_with_weights-generateTrue-fill_value0-start1e-6

MetopB-natural_color-eurol10-blend_with_weights-generateTrue-fill_value255-startNoneMetopB-natural_color-eurol10-blend_with_weights-generateTrue-fill_value255-start1e-6

djhoese: Divide by zero with numpy will generally produce NaNs. In Satpy/trollimage these are considered fill values so they either become transparent via the Alpha band or they get set with your fill_value keyword argument. Based on your comment I think this is what you're seeing. For the start value sum cases, yeah I don't see how that would work properly in real world cases. So I'm not concerned. I think I can continue with the rest of my TODO list then.

djhoese:
New question @lobsiger: If you provide an RGB DataArray, do your weights have to also have 3 bands? Or can they be just 2? Theoretically, your weights could be solar/sensor angle based or are specific to the AreaDefinition so they really don't care about per-band (R, G, or B) "stuff". Right?

lobsiger:
@djhoese weights is provided by the user as a list of 2d (y,x) arrays that have the shape of the area
that the multiscene was resampled to. If I start with 3 LEO passes in scenes, weights will have one entry
for each pass. Weights have no bands. All bands per pass are treated with the same 2d weight array.

Meaningful weights for LEO passes are not defined area specific but rather instrument swath
specific. The easyest way is to base them on 'satellite_zenith_angle' (satz). The starting idea
of Adam was to give pixels with small satz (around SSP) the biggest weight. When you load your
multiscene you must make sure satz is loaded too. Then it will be resampled to your choosen
area just as all other channels needed. After resampling Adam used weights = 1./ satz.
I proposed to use weights = 1. / (1 + satz) to prevent division by 0. But this was all about
what we now call 'select_with_weights'. I finally found most of that that out here:

https://groups.google.com/g/pytroll/c/N20hwYVYI2s/m/IOKXxZXcAwAJ

In Adam's code -- that is now 'select_with_weights' -- very different definition of weights
based on satz lead to the same result as the values of weights are only **compared** per pixel.
In my 'blend_with_weights' different constructions of weights will lead to different results.

If your question is only for building test cases you can probably directly provide weights as list of (y,x) arrays with
the shape of your area. Values in weights can be anything >=0 that should be defined where you have radiance values.


P.S. Regarding latest multiscene.py:

In my 'blend_with_weights' code there is one place where sum(weights) is broadcasted (if "bands") in the line

    base = sum(overlays) / sum(weights)

and there is two lines in 'select_with_weights' because broadcasting indices did not work with Adam's code

    if "bands" in datasets[0].dims:
        indices = [indices] * datasets[0].sizes["bands"]

djhoese:
@lobsiger @mraspaud @adybbroe I think this PR is as done as I can make it without further feedback. Let me know what you think. Things should overall be more consistent and tests should be more thorough.

lobsiger:
@djhoese  first of all I want to thank you very much for bringing this PR to the point where we are now!
As you made it ready for all kind of input data sets (RGBA attached below) maybe the following doc string:

```
 """Combine a series of datasets in different ways.
    By default, datasets are stacked on top of each other, so the last one applied is
    on top. If a sequence of weights (with equal shape) is provided, the datasets will
    be combined according to those weights. Datasets can be integer category products
    (ex. cloud type), single channels (ex. radiance), or RGB composites. In the
    latter case, weights is applied
    to each 'R', 'G', 'B' coordinate in the same way. The result will be a composite
    dataset where each pixel is constructed in a way depending on ``blend_type``.

 """

```
could be slightly enhanced. Adam's ancestor #2275 started with cloud data and while this PR has
been developed the processing of cloud data also changed a lot and more changes are requested.
I therefore recommend (as I also stated here https://github.com/pytroll/satpy/pull/2440#issuecomment-1503935134 ) to review this PR as is and come
back later (if necessary) to the problems that might remain with NWCSAF cloud data. This LGTM!

I also want to thank @mraspaud for his encouragement and for telling the old dog some new tricks.
Attached below GOES18 and GOES16 satellites, 'blend_with_weights' of composite 'night_ir_alpha'.

GOESXX-night_ir_alpha-northamerica_10km-blend_with_weights

djhoese:
Thanks @lobsiger. Yeah, I'll try updating that docstring. And yeah I agree, we should avoid testing this PR and this functionality in general with the NWCSAF products until we're sure they're working properly.

djhoese:
For the docstring, what did you have in mind? You think I should mention RGBA images?

djhoese:
Ok docstring has been updated, but I realized to make the API docs (on the HTML docs) show this new multiscene structure I needed to tell sphinx/api-docs to generate API docs for private members (things prefixed with `_`). This revealed a few docstring issues in some other parts of Satpy. I could also see people not wanting some of this stuff documented but :man_shrugging: 

@mraspaud thoughts?

I think this PR is ready for review and merge.

mraspaud (approved):
LGTM, thanks for reorganising the code!

PR #2403, opened 2023-03-01
Fix failing NWCSAF/PPS composites

### LINKED ISSUE #2392: The cloud_top_height composite broken for PPS
**Describe the bug**
The CloudTopHeightCompositor is not working. It is used in the composit cloud_top_height in visir.yaml. The problem happend with PR #2301. Replacing  CloudTopHeightCompositor with SingleBandCompositor gives an image that no longer handles the fill_value_color attribute in the ctth_alti_pal. It could be fixed if the 
colormap in trollimage  (in function from_array_with_metadata) reads and apply the attribute. Like this:

```python

       color_array = np.concatenate((values[:, np.newaxis], squeezed_palette), axis=1)
       fill_color =palette.attrs.get('fill_value_color', None)
        if fill_color is not None:
            fill_color = np.array(fill_color)
            if valid_range is not None:
                lower = np.append( np.array(valid_range[0]) * scale_factor + add_offset - 1, fill_color)
                upper = np.append( np.array(valid_range[1]) * scale_factor + add_offset + 1, fill_color)
                color_array = np.concatenate((lower[np.newaxis,:], color_array, upper[np.newaxis,:]), axis=0)
```

Not sure how that would affect palettes without palette_meanings but with  fill_value_color attribute though. Might be better ways to fix it.

**To Reproduce**

```python
# Your code here
filename2 = "S_NWC_CTTH_alti_noaa18_00000_20230213T1034434Z_20230213T1050239Z..nc"
scn1.load(['cloud_top_height'])
scn1.show( 'cloud_top_height')
```

**Expected behavior**
Expected the cloud_top_height composite to load without errors.

**Actual results**

[DEBUG: 2023-02-14 13:28:58 : satpy.writers] Data for DataID(name='cloud_top_height') will be enhanced with options:
	[{'name': 'palettize', 'method': <function palettize at 0x7f06c8ff9670>, 'kwargs': {'palettes': [{'dataset': 'ctth_alti_pal', 'color_scale': 255}]}}]
  File "/home/a001865/git/foua_python_scripts/nina/test_satpy_nwcsaf.py", line 22, in <module>
    scn1.show( 'cloud_top_height')
  File "/home/a001865/miniconda3/envs/ppsepssg/lib/python3.9/site-packages/satpy/scene.py", line 987, in show
  File "/home/a001865/miniconda3/envs/ppsepssg/lib/python3.9/site-packages/satpy/writers/__init__.py", line 441, in get_enhanced_image
    enhancer.apply(img, **dataset.attrs)
  File "/home/a001865/miniconda3/envs/ppsepssg/lib/python3.9/site-packages/satpy/writers/__init__.py", line 1216, in apply
    fun(img, *args, **kwargs)
  File "/home/a001865/miniconda3/envs/ppsepssg/lib/python3.9/site-packages/satpy/enhancements/__init__.py", line 349, in palettize
    img.palettize(full_cmap)
  File "/home/a001865/git/trollimage/trollimage/xrimage.py", line 1377, in palettize
    raise ValueError("Image should be grayscale to colorize")
ValueError: Image should be grayscale to colorize

**Screenshots**

**Environment Info:**
 - Satpy Version: [0.40.0]
 - PyResample Version:

**Additional context**

gerritholl:
Fix failing NWCSAF/PPS composites.

Current status: added a unit test reproducing the failing problem.  No solution yet.

 - [x] Closes #2392 
 - [x] Tests added

### DIFF
--- a/satpy/tests/enhancement_tests/test_enhancements.py
+++ b/satpy/tests/enhancement_tests/test_enhancements.py
@@ -603,8 +603,9 @@ def test_nwcsaf_comps(fake_area, tmp_path, data):
                 "source": "satpy unit test",
                 "time_coverage_start": "0001-01-01T00:00:00Z",
                 "time_coverage_end": "0001-01-01T01:00:00Z",
-                } | id_,
+                }
             )
+    ds.attrs.update(id_)
     ds.to_netcdf(fk)
     sc = Scene(filenames=[os.fspath(fk)], reader=[reader])
     sc[palettename] = xr.DataArray(
--- a/satpy/tests/enhancement_tests/test_enhancements.py
+++ b/satpy/tests/enhancement_tests/test_enhancements.py
@@ -502,67 +502,67 @@ def fake_area():
 
 
 _nwcsaf_geo_props = {
-     'cma_geo': ("geo", "cma", 'cma_pal', None, 'cloudmask', 'CMA', "uint8"),
-     'cma_pps': ("pps", "cma", 'cma_pal', None, 'cloudmask', 'CMA', "uint8"),
-     'cma_extended_pps': ("pps", "cma_extended", 'cma_extended_pal', None,
+     'cma_geo': ("geo", "cma", None, 'cma_pal', None, 'cloudmask', 'CMA', "uint8"),
+     'cma_pps': ("pps", "cma", None, 'cma_pal', None, 'cloudmask', 'CMA', "uint8"),
+     'cma_extended_pps': ("pps", "cma_extended", None, 'cma_extended_pal', None,
                           'cloudmask_extended', 'CMA', "uint8"),
-     'cmaprob_pps': ("pps", "cmaprob", 'cmaprob_pal', None, 'cloudmask_probability',
+     'cmaprob_pps': ("pps", "cmaprob", None, 'cmaprob_pal', None, 'cloudmask_probability',
                      'CMAPROB', "uint8"),
-     'ct_geo': ("geo", "ct", 'ct_pal', None, 'cloudtype', 'CT', "uint8"),
-     'ct_pps': ("pps", "ct", 'ct_pal', None, 'cloudtype', 'CT', "uint8"),
-     'ctth_alti_geo': ("geo", "ctth_alti", 'ctth_alti_pal', None, 'cloud_top_height',
+     'ct_geo': ("geo", "ct", None, 'ct_pal', None, 'cloudtype', 'CT', "uint8"),
+     'ct_pps': ("pps", "ct", None, 'ct_pal', None, 'cloudtype', 'CT', "uint8"),
+     'ctth_alti_geo': ("geo", "ctth_alti", None, 'ctth_alti_pal', None, 'cloud_top_height',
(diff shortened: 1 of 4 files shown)


gerritholl:
Does this look correct?

Cloud top height:

NOAA-20-viirs-cloud_top_height-202303011200-test

Cloud type:

NOAA-20-viirs-cloudtype-202303011200-test

Some others look quite empty but I haven't compared the images with 0.39 yet, so maybe that's not Satpys fault.

satpy/tests/enhancement_tests/test_enhancements.py
```diff
@@ -501,75 +501,145 @@ def fake_area():
...
-                "ancillary_variables": [sc[palette]],
+                "ancillary_variables": ancvars,
                 "valid_range": rng})
-    sc.load([comp])
+
+    def _fake_get_varname(info, info_type="file_key"):
+        return altname or dvname
+    with mock.patch("satpy.readers.nwcsaf_nc.NcNWCSAF._get_varname_in_file") as srnN_:
+        srnN_.side_effect = _fake_get_varname
+        sc.load([comp])
```
gerritholl, line 635: This mock is not pretty, but otherwise the reader is actually looking into the NetCDF file, and I wanted to avoid creating a fully realistic fake NetCDF file.  Trying to write tests without testing on real data files is already creating enough work as it is.

gerritholl:
For easier comparison, I've generated with `euron1` as well now:


NOAA-20-viirs-cloud_top_height-202303011200-test

NOAA-20-viirs-cloudtype-202303011200-test

ninahakansson:
Thanks @gerritholl for fixing this! Yes the RGB looks as expected with the black where it is cloud free for CTTH. Have you tested with a VIIRS scene too? It is important that the remapping are not treating cloud free nodata the same way as nodata because of missing data.

mraspaud:
> Have you tested with a VIIRS scene too?

This is a viirs scene :)

ninahakansson:
> 

Excellent!

mraspaud:
@gerritholl I'm wondering if the area around ct which it grey in your examples should be transparent?

gerritholl:
> @gerritholl I'm wondering if the area around ct which it grey in your examples should be transparent?

It's black in v0.39 and grey in v0.40, in main, or with this PR.  The NetCDF source file shows

```ct_pal:fill_value_color = 100LL, 100LL, 100LL ;```

so grey would seem to be correct.  Not sure why it's black with v0.39.

mraspaud:
Ok, then it sounds sensible with grey.

gerritholl:
I think this is due to how Satpy resampling works.  For float data such as CTH, it puts NaN which becomes transparent when saved as a GeoTIFF in RGBA format, but for int data, it becomes 255, which gets the fill value color and is not set to transparent.  It would seem this problem existed already without this PR.

mraspaud (approved):
Thanks for the fix on PPS composites. I'm not really happy about having standard name of geo and pps, but I don't think satpy would support any other way of doing it for now, so we just go for it. Thanks also for adding tests for this.

PR #2404, opened 2023-03-01
Convert TestYAMLFiles to pytest

djhoese:
This should hopefully fix some leaking monkeypatch modifications. My theory based on the failures seen in the unstable/test environment is that this monkeypatch'd import was being used in other tests besides the one it was implemented in (tracebacks include the `real_import` name). It may not be the actual solution to the failures in the experimental environment, but it is a start.

CC @BENR0

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/tests/enhancement_tests/test_ahi.py
+++ b/satpy/tests/enhancement_tests/test_ahi.py
@@ -22,10 +22,10 @@ import numpy as np
 import xarray as xr
 
 
-class TestAHIEnhancement():
+class TestAHIEnhancement:
     """Test the AHI enhancement functions."""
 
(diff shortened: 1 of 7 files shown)


djhoese:
I thought I mentioned it here, but I guess I only commented on slack. The CI failure is caused by an incompatibility between holoviews and dev numpy. See:

https://github.com/numpy/numpy/issues/23307

and:

https://github.com/holoviz/holoviews/issues/5635

mraspaud (approved):
LGTM, thanks for the modernisation!

PR #2408, opened 2023-03-13
Remove DataID attribute access in seviri_l2_grib reader

strandgren:
The recently merged PR https://github.com/pytroll/satpy/pull/2396, removed the DataID attribute access and modified affected readers accordingly. However, two instances in the `seviri_l2_grib` reader were not modified leading to failure when using this reader with the current main. This PR resolves this issue.

The initial commit fixed the attribute access, but broke the reader's unit test since a mocked `DataID` was used, which lead to `TypeError` since the mock object was not subscriptable. This was resolved by using `make_dataid` for the test instead.

### DIFF
--- a/satpy/readers/seviri_l2_grib.py
+++ b/satpy/readers/seviri_l2_grib.py
@@ -65,8 +65,8 @@ class SeviriL2GribFileHandler(BaseFileHandler):
 
     def get_area_def(self, dataset_id):
         """Return the area definition for a dataset."""
-        self._area_dict['column_step'] = dataset_id.resolution
-        self._area_dict['line_step'] = dataset_id.resolution
+        self._area_dict['column_step'] = dataset_id["resolution"]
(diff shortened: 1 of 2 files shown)


djhoese:
> The initial commit fixed the attribute access, but broke the reader's unit test since a mocked DataID was used, which lead to TypeError since the mock object was not subscriptable. This was resolved by using make_dataid for the test instead.

Over-mocking in tests gets us again.

djhoese (approved):
If there is any type of urgency to this PR for your/EUM's work, I'm OK merging it into main if you can confirm that all tests pass locally. We'll deal with the NetCDF test failure in CI in another PR.

strandgren:
@djhoese I'm missing packages needed to run all the tests, but I can confirm that `test_seviri_l2_grib` passes locally. But unless we're talking weeks, it's not that urgent.

PR #2412, opened 2023-03-13
Repair CF writer tests failing with libnetcdf >= 4.9.0

sfinkens:
CF writer tests fail with `libnetcdf4 >= 4.9.0`, because they use the `compression` keyword argument, which is ignored by xarray (https://github.com/pydata/xarray/issues/7388). To fix this, tie usage of the `compression` keyword to the xarray version. There's already an open PR, so we can assume that this will be fixed in the next minor release (`xarray >= 2023.03`).

My bad, sorry I missed that in #2390 .

### DIFF
--- a/satpy/tests/writer_tests/test_cf.py
+++ b/satpy/tests/writer_tests/test_cf.py
@@ -29,11 +29,9 @@ from unittest import mock
 import numpy as np
 import pytest
 import xarray as xr
-from packaging.version import Version
 
 from satpy import Scene
 from satpy.tests.utils import make_dsq
-from satpy.writers.cf_writer import _get_backend_versions
 
 try:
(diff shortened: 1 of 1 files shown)


djhoese:
I assume that once the xarray issue is fixed this will start failing? Can we check the netcdf version and the xarray version to have better control of this?

sfinkens:
I guess they would first issue a deprecation warning before dropping support for the `zlib` keyword. `netcdf4-python` currently supports both.

sfinkens:
I don't know if that's a valid assumption :D But I think it's a good idea. If that PR doesn't make it into the next xarray release, our tests will remind us.

djhoese (approved):

PR #2415, opened 2023-03-15
Fix dataid attribute access in JMA HRIT readers

sfinkens:
JMA HRIT readers still use dataid attribute access, which was removed in 0.41.0. Use key access instead. 

A mock in the unit test was masking the problem...

 - [X] Tests added


mraspaud (approved):
LGTM!

djhoese (approved):

PR #2416, opened 2023-03-16
Drop support for Python 3.8 and add 3.11 tests

gerritholl:
Drop support for Python 3.8 and add tests for Python 3.11.  Remove a note that Python 2 support was dropped as this information is no longer relevant except for historians.

PR #2417, opened 2023-03-16
Fix NUCAPS reader compatibility with new versions of xarray

djhoese:
First noticed in https://github.com/pytroll/satpy/pull/2416, the unreleased version of xarray doesn't like using boolean dask arrays to mask/drop data of other arrays since it creates DataArrays of unknown shapes. The simple workaround in the NUCAPS reader is to just compute the small condition array/mask and then use it.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


gerritholl (approved):
Looks good to me, even if I don't understand exactly what's going on :)

There also exists `.compute_chunk_sizes()` but I suppose that would not be good enough here?

djhoese:
I don't think it would be worth it, plus the error is happening in xarray before we have access to the dask array calculation.

PR #2421, opened 2023-03-23
Add MODIS 05 L2 datasets to `modis_l2` reader

bornagain1981:
Hi guys, below a couple of product modis05 datasets .  content of water vapor with retrieval in infrared and near infrared bands, correction factor , vza, ad acquisition time datasets (for mod06 and mod05 both)
I think the tests are enough for those that already exist

### DIFF
--- a/satpy/etc/readers/modis_l2.yaml
+++ b/satpy/etc/readers/modis_l2.yaml
@@ -207,16 +207,6 @@ datasets:
 
 # file contents: https://atmosphere-imager.gsfc.nasa.gov/sites/default/files/ModAtmo/MOD06_L2_CDL_fs.txt
 
-  scan_start_time:
-    name: scan_start_time
-    long_name: TAI time at start of scan replicated across the swath
-    units: seconds since 1993-1-1 00:00:00.0 0
-    file_type: mod06_hdf
-    coordinates: [longitude, latitude]
-    resolution:
-      5000:
-        file_key: Scan_Start_Time
-
   brightness_temperature:
     name: brightness_temperature
     long_name: Observed Brightness Temperature from Averaged Radiances in a 5x5 1-km Pixel Region
--- a/satpy/etc/readers/modis_l2.yaml
+++ b/satpy/etc/readers/modis_l2.yaml
@@ -9,6 +9,11 @@ reader:
   sensors: [modis]
 
 file_types:
+  mod05_hdf:
+    file_patterns:
+    - 'M{platform_indicator:1s}D05_L2.A{start_time:%Y%j.%H%M}.{collection:03d}.{production_time:%Y%j%H%M%S}.hdf'
+    - '{platform_indicator:1s}1.{start_time:%y%j.%H%M}.mod05.hdf'
+    file_reader: !!python/name:satpy.readers.modis_l2.ModisL2HDFFileHandler
   mod35_hdf:
     file_patterns:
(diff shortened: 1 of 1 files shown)


bornagain1981:
not accepted 2 file_key having same name: "Scan_Start_Time"
This is the name of variable in hdf, correct?
So this is the same name in MOD05 and MOD06, how to solve it?

simonrp84:
I think you need to delete that key from the MOD05 section. Then, in the MOD06 section, change file type to:
`file_type: [mod05_hdf, mod06_hdf]`

satpy/etc/readers/modis_l2.yaml
```diff
@@ -143,12 +148,75 @@ datasets:
...
+    units: seconds since 1993-1-1 00:00:00.0 0
+    file_type: mod05_hdf
+    coordinates: [longitude, latitude]
+    resolution:
+      5000:
+        file_key: Scan_Start_Time
+
+  sensor_zenith:
+    name: sensor_zenith
+    long_name: Sensor Zenith Angle, Cell to Sensor
```
djhoese, line 166: For consistency with the rest of Satpy, what do people think about naming this `satellite_zenith_angle`?

djhoese (changes requested):
Looks pretty good. One small name change request for the sensor angle. Otherwise, I agree with Simon that you need to combine the scan_start_time using the `file_type`.

satpy/etc/readers/modis_l2.yaml
simonrp84: Yes, good idea.

bornagain1981: done

satpy/etc/readers/modis_l2.yaml
```diff
@@ -143,6 +148,59 @@ datasets:
...
+    file_type: [mod05_hdf, mod06_hdf]
+    coordinates: [longitude, latitude]
+    resolution:
+      5000:
+        file_key: Sensor_Zenith
+
+  water_vapor_correction_factor:
+    name: water_vapor_correction_factor
+    long_name: Aerosol Correction Factor for Water Vapor - Near Infrared Retrieval
+    units: None
```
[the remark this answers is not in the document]
bornagain1981: using Panoply  shows this
 :units = "none" 

`short Water_Vapor_Correction_Factors(Cell_Along_Swath_1km=2030, Cell_Across_Swath_1km=1354);
  :long_name = "Aerosol Correction Factor for Water Vapor - Near Infrared Retrieval";
  :add_offset = 0.0; // double
  :Parameter_Type = "Output";
  :Cell_Along_Swath_Sampling = 1, 2030, 1; // int
  :Cell_Across_Swath_Sampling = 1, 1354, 1; // int
  :Geolocation_Pointer = "Internal geolocation arrays";
  :valid_range = 0S, 20000S; // short
  :_FillValue = -9999S; // short
  :units = "none";
  :scale_factor = 0.0010000000474974513; // double`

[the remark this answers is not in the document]
djhoese: 1. `"none"` != `None`
2. I wouldn't look to MODIS data as a source for CF standard metadata. I'm not sure CF was even used/existed when HDF-EOS/MODIS files were first created.
3. We're trying to be consistent with the rest of Satpy, not necessarily with MODIS L2 processing.
4. If valid units are in the data file we (the python code) should be pulling that metadata from the file, not configuring it in the YAML. In this case I don't think the units are valid though.

djhoese (approved):

PR #2429, opened 2023-03-27
Cleanup and separate Scene unit tests

djhoese:
I wanted to work on some stuff for the Scene, but decided to refactor the tests first. I'm not really sure if the new module names are the best choices. Any other ideas?

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


djhoese:
@mraspaud @pnuu maybe this is enough work just for refactoring the scene tests that it should be its own PR. Thoughts?

djhoese:
I've updated the original description. This is now a cleanup/refactor PR only. As mentioned above, I'm not sure if there are better names for the new test module names. Let me know what you think.

pnuu (approved):
I was going to make a ton (stopped at 13 and cancelled the review) of "this is N separate cases" comments, but this is more on restructuring the tests.

One thing I'd like to see is the mention of "This test module uses the following external fixtures: foo, bar, baz" or something like that at the top of each module. Or maybe the `@pytest.mark.usefixtures('foo', 'bar')` for the test classes is close enough :thinking: 

LGTM, thanks for the work!

djhoese:
> I was going to make a ton (stopped at 13 and cancelled the review) of "this is N separate cases" comments, but this is more on restructuring the tests.

Yeah, as mentioned on slack, as far as the contents of the tests I didn't really change anything beyond what was needed to make CodeScene happy and reorganizing tests to be grouped better.

> One thing I'd like to see is the mention of "This test module uses the following external fixtures: foo, bar, baz" or something like that at the top of each module. Or maybe the @pytest.mark.usefixtures('foo', 'bar') for the test classes is close enough thinking

I thought about it and couldn't decide. As you know I'm not a huge fan of the comment at the top of the module, but somehow the more explicit "usefixtures" is almost worse than just using it as an argument everywhere. For these tests, especially with how many there are, it can be hard to understand why they actually work without first knowing that this fixture is being enabled at the top of the class (possibly hundreds of lines of code above the test method).

FYI I use this `usefixtures` instead of making it an argument because some linters (like ruff/flake8) don't understand pytest fixtures as being unused arguments but still having an effect on the test.

djhoese:
There, I added the comments.

PR #2433, opened 2023-04-01
Make mask dataset a kwarg to resample

ninahakansson:
Make mask dataset a kwarg to resample. And make sure the remapper is cached.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/resample.py
+++ b/satpy/resample.py
@@ -407,7 +407,7 @@ class BaseResampler(object):
         """
         raise NotImplementedError
 
-    def resample(self, data, cache_dir=None, mask_area=None, **kwargs):
+    def resample(self, data, cache_dir=None, mask_area=None, mask_array=None, **kwargs):
         """Resample `data` by calling `precompute` and `compute` methods.
 
         Only certain resampling classes may use `cache_dir` and the `mask`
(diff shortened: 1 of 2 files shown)


djhoese:
I read your issue in #2434, but I still don't understand how you expect to use this keyword argument? Assuming the mask is the same as the geo mask (`mask_area`), is this a workaround for some dask-related issue that isn't sharing computations? Or are you saying you want to assume that the invalid mask between all bands is the same? Is that a safe assumption?

ninahakansson:
In operation I have 10 polar nwcsaf products on the same swath. They are produced with trollflow2. I want images of all products for three formats and three areas. So almost 100 images. And I want them finished in a few minutes (less than 4 would be nice :). 

The same remapper and mask (from the cloudmask) is safe to use for all my 10 products. Using the same mask can also solve some problems "with cloud-free is also nodata" that is present in some of the products. I guess the mask is the same as the invalid mask? So only three remappers (one per area) need to be calculated. This is not possible in satpy currently. With this PR I can do that, but maybe I also broke something and it need to be done in a different way? So this is a feature I would like to add.

I expect to use the mask_array_name like this:
```
  scn2 = scn1.resample("euron1", compute=False, mask_array_name="cloudmask")
 ```
Also the lat/lon for the area should only be calculated one time per area (discussed in the issue). That the calculation is done several times I think is a bug.

djhoese:
Do you have some files for this reader that I could test with?

ninahakansson:
With a non-buggy conda environment and this PR https://github.com/pytroll/satpy/pull/2440 with the new pps-omposites I get my 100 images in less than 4 minutes. As the new compositor uses the same mask when remapping the remapper is cached and reused automatically.

PR #2437, opened 2023-04-03
Update MODIS composites with specific channel for Rayleigh correction

simonrp84:
Currently, the `rayleigh_corrected` modifier in `visir.yaml` requires a wavelength of `0.67μm`. This can be a problem for MODIS as it has a band with an upper wavelength matching this value as well as several other bands with similar wavelengths.
This can produce errors such as:
`Too many possible datasets to load for DataQuery(wavelength=0.67, modifiers=(), calibration='reflectance')`

This PR adds a MODIS-specific Rayleigh modifier that directly specifies which channel to use by name rather than by wavelength, which resolves the issue.

### DIFF
--- a/satpy/etc/composites/modis.yaml
+++ b/satpy/etc/composites/modis.yaml
@@ -11,6 +11,22 @@ modifiers:
     - name: solar_azimuth_angle
     - name: solar_zenith_angle
 
+  rayleigh_corrected:
+    modifier: !!python/name:satpy.modifiers.PSPRayleighReflectance
+    atmosphere: us-standard
+    aerosol_type: rayleigh_only
+    prerequisites:
+    - name: '1'
+      modifiers: [sunz_corrected]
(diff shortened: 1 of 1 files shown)


djhoese (commented):
This seems reasonable to me. From our conversation on slack, I'm pretty sure the `0.67` dependency in the generic `visir.yaml` config is hitting some floating point precision issues since the MODIS reader specifies the upper bound of band 1 as `0.67`. Put another way, floating point doesn't guarantee `0.67 == 0.67`.

Otherwise, @simonrp84 maybe the angles should be required and moved into `prerequisites`? I *think* the rayleigh modifier can handle that (I think I do this for VIIRS). This way people don't waste time trying to load rayleigh for MODIS with dynamically generated angles. Thoughts?

mraspaud (approved):
LGTM

PR #2441, opened 2023-04-07
Add channel aliases to the CLAVRx reader to facilitate composites

joleenf:
This PR updates the clavrx reader with aliases for some visible channels to allow the use of the cimss cloud type composite and color bar (4-Level) for the cloud mask.  Test updates have been made to verify that the aliases are being created correctly.


djhoese:
I wonder if maybe we could have a more descriptive title to this PR? :stuck_out_tongue_winking_eye:

satpy/readers/clavrx.py
```diff
@@ -126,24 +140,21 @@ def _get_data(data, dataset_id: dict) -> xr.DataArray:
...
             attrs.pop('_FillValue', None)
-
-        if all(valid_range):
-            valid_min = _CLAVRxHelper._scale_data(valid_range[0], factor, offset)
-            valid_max = _CLAVRxHelper._scale_data(valid_range[1], factor, offset)
-            if flags:
-                data = data.where((data >= valid_min) & (data <= valid_max), fill)
-            else:
+            if isinstance(valid_range, np.ndarray):
+                valid_min = _CLAVRxHelper._scale_data(valid_range[0], factor, offset)
```
joleenf, line 151: What is the proper way to patch this _scale_data function and use an assert_called() test?

djhoese: Easiest might be to extract `_scale_data` outside the class and then do the mocking. To do it "proper" (read: the way I would prefer) do:

```
from satpy.readers.clavrx import _scale_data
with mock.patch("satpy.readers.clavrx._scale_data", wraps=_scale_data) as scale_data:
    ...
scale_data.assert_called()
```

But overall it may be even easier if you just check the final result and make sure it has the scaled values you expect.

joleenf:
I am not sure I want to do anything about the CodeFactor fail.  I am not certain there is anything I should do about the CI/test (3.11, ubuntu-latest) test.

djhoese:
Don't worry about the CI failure. The "true" at the end of that is that it is using experimental versions of dependencies. With numpy 2.0 and some other dependencies we know this is failing.

The CodeFactor thing I would like you to fix. It is complaining about the spacing you used. Let's remove any unnecessary spacing.

satpy/readers/clavrx.py
```diff
@@ -94,9 +108,29 @@ def _get_rows_per_scan(sensor: str) -> Optional[int]:
     return None
 
 
+def _scale_data(data_arr: Union[xr.DataArray, int], scale_factor: float, add_offset: float) -> xr.DataArray:
```
djhoese, line 111: I'm not requesting a change here, but FYI if you do `from __future__ import annotations` at the top of the module you can then use `xr.DataArray | int` here instead of the `Union`.

satpy/readers/clavrx.py
```diff
@@ -94,9 +108,29 @@ def _get_rows_per_scan(sensor: str) -> Optional[int]:
...
+    @staticmethod
+    def _get_nadir_resolution(sensor, resolution_from_filename_info):
+        """Get nadir resolution."""
+        for k, v in NADIR_RESOLUTION.items():
+            if sensor.startswith(k):
+                return v
+        res = resolution_from_filename_info
+        if res.endswith("m"):
+            return int(res[:-1])
+        elif res is not None:
```
djhoese, line 131: `elif` is unnecessary. Not sure why the linter didn't complain about this. Maybe it isn't enabled on satpy or ruff isn't checking it yet.

```suggestion
        if res is not None:
```

But also...if `res` can be `None` then won't the previous `res.endswith` fail?

satpy/readers/clavrx.py
```diff
@@ -94,9 +108,29 @@ def _get_rows_per_scan(sensor: str) -> Optional[int]:
     return None
 
 
+def _scale_data(data_arr: Union[xr.DataArray, int], scale_factor: float, add_offset: float) -> xr.DataArray:
+    """Scale data, if needed."""
+    scaling_needed = not (scale_factor == 1.0 and add_offset == 0.0)
+    if scaling_needed:
+        data_arr = data_arr * scale_factor + add_offset
```
djhoese, line 115: Do you/we need `np.float32(scale_factor)` (and on the offset) to make sure 64-bit floats aren't created unnecessarily?

satpy/readers/clavrx.py
```diff
@@ -123,28 +149,31 @@ def _get_data(data, dataset_id: dict) -> xr.DataArray:
...
-                data = data.where((data >= valid_min) & (data <= valid_max), fill)
-            else:
+            data = _scale_data(data, factor, offset)
+
+            if valid_range[0] is not None:
+                valid_min = _scale_data(valid_range[0], factor, offset)
+                valid_max = _scale_data(valid_range[1], factor, offset)
                 data = data.where((data >= valid_min) & (data <= valid_max))
-            attrs["valid_range"] = [valid_min, valid_max]
+                attrs["valid_range"] = [valid_min, valid_max]
```
djhoese, line 176: I don't need this to be cleaned up before merging, but this code feels like it wants to be refactored. For example, the `fill` is created in both `if` branches in the same way. Also you're doing `data.where` with the same types of conditions and a fill value. Granted if you specify the `fill` for the float case you might have to make sure to do `np.float32(np.nan)` to preserve the 32-bit float dtype.

Eh maybe keep this the way it is. It is a lot of code in the middle of this method though.

satpy/readers/clavrx.py
```diff
@@ -235,14 +268,24 @@ def _read_axi_fixed_grid(filename: str, l1b_attr) -> geometry.AreaDefinition:
...
-            np.asarray(area_extent))
+        if sensor == "abi":
+            area = geometry.AreaDefinition(
+              "abi_geos",
+              "ABI L2 file area",
+              "abi_geos",
+              proj,
+              ncols,
+              nlines,
+              np.asarray(area_extent))
```
djhoese, line 279: Not sure why this `np.asarray` was ever a thing, but area_extent should always be a tuple when passed to AreaDefinition.

satpy/readers/clavrx.py
```diff
@@ -450,29 +509,34 @@ def available_datasets(self, configured_datasets=None):
...
+            matches = self.file_type_matches(ds_info["file_type"])
+            if matches and ds_info.get("resolution") != self.resolution:
+                # reader knows something about this dataset (file type matches)
+                # add any information that this reader can add.
+                new_info = ds_info.copy()
+                new_info["resolution"] = self.resolution
                 handled_vars.add(ds_info["name"])
-            yield self.file_type_matches(ds_info["file_type"]), ds_info
-        yield from self._available_new_datasets(handled_vars)
+                yield True, new_info
```
djhoese, line 520: Should the resolution check here be checking for `is None`? What is the case where we, the file handler, are configured to handle this variable but the resolution is defined as something other than what we contain (currently resolution != self.resolution)?

djhoese (changes requested):
A couple suggestions, but otherwise this looks mostly good to me. I'm wondering also if the tests (which we may have talked about) should be in a separate `test_clavrx/` directory or something just to group them together. That way it might be easier to have any shared logic in some kind of `_shared.py` or `_utils.py` or something. Just an idea.

satpy/readers/clavrx.py
joleenf: Yes.  It would.

satpy/readers/clavrx.py
joleenf: Yes.

satpy/readers/clavrx.py
joleenf: Other than setting global attributes, is the nadir resolution used in remapping?  I am wondering if it is okay if the nadir resolution is returned as none?  There should be enough here to always have a result.  I am just wondering if I should just let the reader fail if there is a None value.

djhoese: It looks like `self.resolution` is used to fill in `ds_info["resolution"]` in the datasets. This becomes part of the `DataArray.attrs` but also part of the `DataID` that Satpy uses to identify each dataset/product. I think it is allowed to be None in most cases when it is unknown or not applicable.

joleenf: This is an odd mix.  Not sure if I should also re-work how the resolution value is extracted from the string.  Sometimes I get too fancy, but I was thinking, hypothetically, resolution could be a different unit system.  It is highly unlikely, but not impossible.

Is this overboard?
```python
def _get_nadir_resolution(sensor, resolution_from_filename_info):
        """Get nadir resolution."""
        for k, v in NADIR_RESOLUTION.items():
            if sensor.startswith(k):
                return v
        if resolution_from_filename_info is None:
            return None
        elif isinstance(resolution_from_filename_info, int):
            return int(resolution_from_filename_info)
        elif isinstance(resolution_from_filename_info, str):
            print(resolution_from_filename_info)
            result = ''.join([i for i in resolution_from_filename_info if i.isdigit()])
            return int(result)
```

djhoese: A little simpler:

```python
    def _get_nadir_resolution(sensor, resolution_from_filename_info):
        """Get nadir resolution."""
        for k, v in NADIR_RESOLUTION.items():
            if sensor.startswith(k):
                return v
        if resolution_from_filename_info is None:
            return None
        if isinstance(resolution_from_filename_info, str):
            resolution_from_filename_info = ''.join([i for i in resolution_from_filename_info if i.isdigit()])
        return int(resolution_from_filename_info)
```

Right now Satpy doesn't support units other than meters in the resolution field...although it also doesn't use it beyond the user requesting a specific resolution. I'd say removing the "m" is probably good enough unless you have evidence that a "123 degrees" or something might exist. Oh I guess there could be "km" but in that case you need to actually convert that or specify the units when assigning the resolution in some way. So maybe it is best to let it fail automatically by only removing the "m" (`if isinstance(resolution_from_filename_info, str) and resolution_from_filename_info.startswith("m"):`)

satpy/readers/clavrx.py
joleenf: I would like to handle this separately.  I remember working on this when some of the AWIPS displays were not displaying correctly. I would want to test any changes here with the awips_tiled writer and that is a little more work.  If the values are flags, the fill value is retained in the data array attributes, otherwise it isn't.  I not completely convinced this had anything to do with how the data in AWIPS displayed, but I don't want to take the chance at this time.

djhoese: Sounds good. Understood. Leave it as it is.

satpy/readers/clavrx.py
joleenf: This part has been tricky for me to get correct.  The structure is essentially following [Example 1](https://satpy.readthedocs.io/en/stable/api/satpy.readers.file_handlers.html#satpy.readers.file_handlers.BaseFileHandler.available_datasets).  Are you suggesting:

```python
matches = self.file_type_matches(ds_info["file_type"])
            if matches and ds_info.get("resolution") != self.resolution:
                # reader knows something about this dataset (file type matches)
                # add any information that this reader can add.
                new_info = ds_info.copy()
                if self.resolution is not None:
                    new_info["resolution"] = self.resolution
                handled_vars.add(ds_info["name"])
                yield True, new_info
        yield from self._available_file_datasets(handled_vars)
```
It has become a little complex trying to get the aliases to work within the available datasets, but that has nothing to do with the question of resolution, it just adds layers of yield statements.

djhoese: The git diff is a little complicated now, but I kind of see the reasoning behind the way I made "Example 1". I think the idea is something like MODIS L1b files where each resolution is in a different file (250m, 500m, 1km). So in that case it is very likely that we could be looking at a `ds_info` that has a resolution different than our own and I suppose the file matches is just a "shortcut" condition to say "yeah, we definitely don't know how to handle this dataset".

I guess what I'm saying is leave it the way you have it if it works for you.

satpy/readers/clavrx.py
```diff
@@ -236,13 +272,13 @@ def _read_axi_fixed_grid(filename: str, l1b_attr) -> geometry.AreaDefinition:
...
 
         area = geometry.AreaDefinition(
-            "ahi_geos",
-            "AHI L2 file area",
-            "ahi_geos",
-            proj,
-            ncols,
-            nlines,
-            np.asarray(area_extent))
+          f"{sensor}_geos",
```
joleenf, line 275: Removes the type np.asarray as requested.  Simplifies the description using f-strings.

satpy/readers/clavrx.py
```diff
@@ -94,9 +110,30 @@ def _get_rows_per_scan(sensor: str) -> Optional[int]:
     return None
 
 
+def _scale_data(data_arr: xr.DataArray | int, scale_factor: float, add_offset: float) -> xr.DataArray:
+    """Scale data, if needed."""
+    scaling_needed = not (scale_factor == 1.0 and add_offset == 0.0)
+    if scaling_needed:
+        data_arr = data_arr * np.float32(scale_factor) + np.float32(add_offset)
```
joleenf, line 117: Added np.float32 on scale_factor and add_offset as requested

satpy/readers/clavrx.py
```diff
@@ -236,13 +272,13 @@ def _read_axi_fixed_grid(filename: str, l1b_attr) -> geometry.AreaDefinition:
...
-            ncols,
-            nlines,
-            np.asarray(area_extent))
+          f"{sensor}_geos",
+          f"{sensor.upper()} L2 file area",
+          f"{sensor}_geos",
+          proj,
+          ncols,
+          nlines,
+          area_extent)
```
joleenf, line 281: This resolved the np.asarray issue.

djhoese (changes requested):
I think you need a `/__init__.py` in your new clavrx test directory. Maybe not because pytest will find it anyway, but we generally think of the test directories as part of the package so they need the init module.

djhoese:
I have a feeling ruff tried to change your apostrophes to double quotes when it thought they were single quotes. I'm hoping it will be OK with my latest commit.

djhoese (approved):

djhoese:
Thanks for sticking with it. I figure at this point you are the owner of this code so if anything goes wrong you're likely to be the one to identify it and fix it. I tried to review what I could and what I understood. Nice job.

PR #2448, opened 2023-04-13
Add section to FAQ to describe how `generate=False` may speed up RGBs

simonrp84:
Based on a comment from @djhoese on slack, I added `generate=False` to some of my `load` calls. This resulted in a massive speed up when generating composites, with processing going nearly three times faster than with the `generate=True` option (default).

In this PR, I add a short section to the FAQ docs to describe how adding `generate=False` may help users produce images more rapidly. Am not sure if I've described it right, or whether I've missed anything out, so there's probably scope for improving the text.

 - [x] Fully documented

### DIFF
--- a/doc/source/faq.rst
+++ b/doc/source/faq.rst
@@ -29,12 +29,12 @@ load your composite:
     scn.load(['true_color'], generate=False)
     scn_res = scn.resample(...)
 
-By default, `generate=True` which means that Satpy will create as many
+By default, ``generate=True`` which means that Satpy will create as many
 composites as it can with the available data. In some cases this could
 a lot of intermediate products (ex. rayleigh corrected data using dynamically
 generated angles for each band resolution) that will then need to be
 resampled.
-By setting `generate=False`, Satpy will only load the necessary dependencies
+By setting ``generate=False``, Satpy will only load the necessary dependencies
 from the reader, but not attempt generating any composites or applying any
 modifiers. In these cases this can save a lot of time and memory as only one
 resolution of the input data have to be processed. Note that this option has
--- a/doc/source/faq.rst
+++ b/doc/source/faq.rst
@@ -13,7 +13,7 @@ an issue on GitHub or talk to us on the Slack team or mailing list. See the
     :local:
 
 
-How can I speed up image production for composites that need resampling?
+How can I speed up creation of composites that need resampling?
 ------------------------------------------------------------------------
 
 Satpy performs some initial image generation on the fly, but for composites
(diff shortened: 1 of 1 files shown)


pnuu:
Another case where `generate=False` will speed up processing significantly is when some channels are used in several composites and the data are resampled to some area. With `generate=True` all the RGBs are resampled individually, which means the multi-use channels are resampled several times. With `generate=False` each of the input channels are first resampled (only once!) and then used in creating the composites.

lobsiger:
@simonrp84 see also the speed up discussion in this tread (starting with 1'37'' down to 21''):

https://groups.google.com/g/pytroll/c/Zi5x-Y0b9GE/m/38mWf1EUAgAJ

It might be useful to note that not all composites will work with generate=False.
Notably GEO composites with backgrounds like 'night_ir_with_background_hires'
never worked for me when setting generate=False. And there is one LEO (Aqua/Terra)
MODIS composite 'true_color_thin' that comes out ugly with generate=False but
good looking with generate=True (probably/maybe a channel resolution effect?).

doc/source/faq.rst
```diff
@@ -12,6 +12,30 @@ an issue on GitHub or talk to us on the Slack team or mailing list. See the
...
+
+    scn = Scene(filenames=filenames, reader='abi_l1b')
+    scn.load(['true_color'], generate=False)
+    scn_res = scn.resample(...)
+
+By default, `generate=True` which means that Satpy will compute some initial
+information about the scene at *each* spatial resolution.
+By setting `generate=False`, Satpy will only compute this information when the
+final image is generated. This can save a lot of time, and can result in images
+being generated up to 3x faster.
```
djhoese, line 36:
```suggestion
By default, `generate=True` which means that Satpy will create as many
composites as it can with the available data. In some cases this could
a lot of intermediate products (ex. rayleigh corrected data using dynamically
generated angles for each band resolution) that will then need to be
resampled.
By setting `generate=False`, Satpy will only load the necessary dependencies
from the reader, but not attempt generating any composites or applying any
modifiers. In these cases this can save a lot of time and memory as only one
resolution of the input data have to be processed. Note that this option has
no effect when only loading data directly from readers (ex. IR/visible bands
directly from the files) and where no composites or modifiers are used. Also
note that in cases where most of your composite
inputs are already at the same resolution and you are only generating a limited
number of composites, ``generate=False`` may actually hurt performance.
```

doc/source/faq.rst
```diff
@@ -12,6 +12,30 @@ an issue on GitHub or talk to us on the Slack team or mailing list. See the
     :depth: 1
     :local:
 
+
+How can I speed up image production for composites that need resampling?
```
djhoese, line 16:
```suggestion
How can I speed up creation of composites that need resampling?
```

I'd like to avoid language that assumes we're generate "images". I don't want to confuse people who aren't generating images (PNG, JPEG) as their final output that this is only an option when writing to disk. Some people may also not consider creating interactive plots as generating images. Also, this doesn't only apply to RGB/A composites but any composite and any modifier.

djhoese (changes requested):
Thanks for putting this together. This really makes me want to rewrite the docs similar to how I started to in pyresample with an entire high-level no-code section about "Concepts". Since we don't have that yet and it would take too long to do, this is a really good start to at least get the information out there. I've rewritten parts of this to be more explicit about the details of how/why this works. Hopefully these changes avoid confusion and/or future questions.

doc/source/faq.rst
```diff
@@ -12,6 +12,39 @@ an issue on GitHub or talk to us on the Slack team or mailing list. See the
...
+load your composite:
+
+.. code-block:: python
+
+    scn = Scene(filenames=filenames, reader='abi_l1b')
+    scn.load(['true_color'], generate=False)
+    scn_res = scn.resample(...)
+
+By default, ``generate=True`` which means that Satpy will create as many
+composites as it can with the available data. In some cases this could
```
djhoese, line 33:
```suggestion
composites as it can with the available data. In some cases this could mean
```

djhoese: Hhhmm or maybe `this could create`?

djhoese (approved):

PR #2449, opened 2023-04-14
Generalise the `true_color_reproduction` composite and enhancement

simonrp84:
Currently, the `true_color_reproduction` composite can only be used with Himawari/AHI. In this PR, I generalise both the composite and its enhancement so that can be more widely used.

This process isn't straightforward, as the enhancement requires a conversion matrix to convert the satellite RGB channels through CIE XYZ colorspace and into sRGB.
For AHI, a conversion matrix is given in [this paper](https://www.jstage.jst.go.jp/article/jmsj/96B/0/96B_2018-049/_pdf/-char/en) but that cannot be used for other sensors.

Therefore, I also wrote a utility to calculate these matrices. It is not included in this PR, however.

So, right now the PR is a draft. I need to add tests and also the matrixes for a few more sensors (at least the GEO ones). I also have some questions:

1) What shall I do with the utility script? Commit it somewhere in satpy? Commit it somewhere in pyspectral? Don't commit it anywhere?
2) Is my current solution for selecting the right matrix a sensible one? See lines 607 onwards of `enhancements/__init__.py`
3) Right now the matrices assume a central wavelength for the green channel of 0.555 micron (same as MODIS). Because none of the GEO satellite have an actual green channel we have a bunch of synthetic greens. I have no idea what the simulated central wavelength of these are, does anyone have documentation about that?
 - [x] Tests added
 - [x] Fully documented


simonrp84:
I'd like @mherbertson to take a look at this if he has time, but I can't add him as a reviewer...so tagging here.

simonrp84:
I've added the 0.42 milestone to this, as it would be nice to include it. That said, it depends on feedback as there could still be quite a bit to do here...

mraspaud:
I agree that it would be nice to see @mherbertson have a look at this.

1. For the script, you could create a gist and link it here or in the code as a comment maybe? Or how big is it actually?
2. how about using just the platform name? There is always only one imager on each platform, right?

satpy/etc/composites/fci.yaml
```diff
@@ -64,3 +64,34 @@ composites:
       - name: ndvi_hybrid_green_raw
       - name: vis_04
     standard_name: true_color_raw
+
+  true_color_reproduction:
```
mraspaud, line 68: I don't see a matrix for fci, will it still work?

simonrp84: Not added it yet, will do soon.

mherbertson:
Great stuff @simonrp84 

1. I wonder if a short example/methodology in a comment might suffice rather than attaching the entire script?
2. For readability I'd just go with one key, it's always possible to change naming convention in the future if required.
3. I don't have any information but I am happy to help, I'll try to make some time and see if I can find any specs online.

Thanks for starting this PR!

mraspaud:
1. Alternatively, do you think the script would fit better in pyspectral?
2. I would say remove the sensor part, but I don't really care that much, so keep it like this if you prefer.

mherbertson:
3. I've looked at numerous research papers on-line and unfortunately couldn't find hard data. It appears to me at least in the ABI simulated green band case that the aim is to get as close to 0.55 µm as possible.

simonrp84:
Ok, I've updated this to include MTG and GK-2A matrices and have sorted out the tests. I think this is more or less ready now, and I'll have a think about what to do with the matrix generation script. Probably put it in pyspectral, but that's a separate PR.

simonrp84:
Updated this to add some better tests.

Below are example images showing the original `true_color` and new `true_color_reproduction` composites for AMI:
true_color_20230509_001032 png
true_color_reproduction_20230509_001032 png

And simulated data FCI:

true_color_20170920_120008 png
true_color_reproduction_20170920_120008 png

mraspaud (approved):
LGTM

PR #2452, opened 2023-04-20
Add initial Scene/DataTree conversion via `Scene.to_xarray_datatree`

djhoese:
Related to #2352, I thought I'd add a `Scene.to_xarray_datatree` method to the Scene. I want to use this as a stepping stone into more DataTree usage in Satpy and see where it takes us. I also wanted a reason to play around with DataTree more and see how it goes.

At the time of writing this is a pretty simple implementation, but also provides a lot of flexibility. When discussing this concept in #2352 I brought up the ideas of "schemes" or predefined tree-structuring methods that users could choose from. For example, "area"-based or "DataID"-based organizing. As I started working on this (TDD for the win!) I realized I could just structure it as "what metadata keys should I use for each 'group'". This means users can select the level of the hierarchy of the tree. For example, group by platform_name and sensor by specifying `group_keys=("platform_name", "sensor")`.

So at this point, this works as described above, but I'm looking for feedback for others. I seem to remember @mraspaud and @gerritholl being part of this conversation and @ghiggi doing some reworking of `Scene.to_xarray_dataset` related to CF compatibility. I'm curious what you guys think of this.

There is still a lot TODO like:

* handling grouping by area. For example, what do I use as the node name? area hash? Area name (if it exists, pyresample 2.0 will have optional names)?.
* handling of other non-string properties including resolution (units?), calibration enum, etc.
* What attributes to define on each group
* Documentation

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/_scene_converters.py
+++ b/satpy/_scene_converters.py
@@ -28,27 +28,46 @@ except ImportError:
     DataTree = None
 
 
-def to_xarray_datatree(data_arrays: Scene | Iterable[xr.DataArray], **kwargs) -> DataTree:
+def to_xarray_datatree(
+        data_arrays: Scene | Iterable[xr.DataArray],
+        group_keys: Iterable[str] = ("platform_name", "sensor"),
+        **kwargs
+) -> DataTree:
     """Convert this Scene into an Xarray DataTree object."""
     if DataTree is None:
         raise ImportError("Missing 'xarray-datatree' library required for DataTree conversion")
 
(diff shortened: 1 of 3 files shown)


BENR0:
I think that this could also be really useful for #2171. Currently the sorting is done by platform/sensor and the by area because this made the most sense when I started. But I think it could be useful to display the `Scene` by a different hierarchy. So this would make that task easier I think.

PR #2453, opened 2023-04-22
Fix various warnings encountered when running tests

djhoese:
We have a lot of warnings when the tests are run...let's get rid of some of them.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #2454, opened 2023-04-23
Change optional rayleigh prerequisites to required for MODIS

djhoese:
For MODIS data these should never really be optional in my opinion. I noticed this change from @simonrp84 after it started making my upstream tests for Polar2Grid fail because my custom rayleigh corrector (which uses crefl) was pulling in hard requirements and optional requirements.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/etc/composites/modis.yaml
+++ b/satpy/etc/composites/modis.yaml
@@ -18,7 +18,6 @@ modifiers:
     prerequisites:
     - name: '1'
       modifiers: [sunz_corrected]
-    prerequisites:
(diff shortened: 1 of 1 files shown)


simonrp84:
Aah sorry, I meant to include this in my PR for the other MODIS issue (you mentioned it there) but it slipped my mind!

djhoese:
Given the simplicity and the approval from Simon I'm going to merge this now.

PR #2456, opened 2023-04-28
Fix Adaptive DNB composite bug introduced from refactoring

djhoese:
This fixes a bug I introduced when contributing to @simonrp84's #2124 PR when I refactored the VIIRS composites module. One simplification I did resulted in allowing negative indices to be used when accessing data when they were actually supposed to be considered out of bounds.

I haven't added a test for this explicitly because the difference it makes with our test dataset seems to be extremely small (0.9989 to 0.9990) and even in real world cases I didn't see anything obvious. The only reason I caught this is because of some automated testing I have in Polar2Grid which compared two images and it discovered that some pixels had changed.

Otherwise, this PR also fixes a few issues discovered in the tests (name changes and one expected data not being checked).

 - [ ] Closes #xxxx 
 - [ ] Tests added

### DIFF
--- a/satpy/composites/viirs.py
+++ b/satpy/composites/viirs.py
@@ -644,6 +644,9 @@ def _get_cumul_bin_info_for_tile(
     # current tile)
     calculated_row = num_row_tile - 1 + weight_row
     calculated_col = num_col_tile - 1 + weight_col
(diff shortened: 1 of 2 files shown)


djhoese:
Looks like there is a coveralls outage that is causing tests to fail. It isn't the actual tests failing.

PR #2462, opened 2023-05-03
Specify target branch in CI badge URL

sfinkens:
I noticed that the CI badge is red if the CI pipeline in one of the pull requests is red. To fix this, specify the target branch (`main`) in the badge URL.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@ Satpy
 =====
 
 .. image:: https://github.com/pytroll/satpy/workflows/CI/badge.svg?branch=main
-    :target: https://github.com/pytroll/satpy/actions?query=workflow%3A%22CI%22
+    :target: https://github.com/pytroll/satpy/actions?query=workflow%3A%22CI%22+branch%3Amain
 
 .. image:: https://coveralls.io/repos/github/pytroll/satpy/badge.svg?branch=main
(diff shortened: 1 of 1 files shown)


djhoese:
I don't think I understand. You're talking about the image being red, but what you've changed here is just the target of the hyperlink when you click on the image. The image was already specific to the branch. It sometimes lags behind the actual state due to github caching the badge images, but I was pretty sure it was just `main` since we were specifying the branch.

sfinkens:
So I clicked on the red badge and that took me to [this page](https://github.com/pytroll/satpy/actions?query=workflow%3A%22CI%22). The most recent job in the list was failing (and from a PR), so I thought that's the reason for the badge being red. But now the most recent job is green (from this PR) and the badge is still red. Then it's probably the caching that you mentioned.

djhoese:
I wonder if the badge includes the failing "allowed failure" in it's failure determination. Or maybe the last main CI run had a network issue and failed when it shouldn't have?

djhoese:
So I looked and it looks like the last job on `main` was cancelled. Likely as part of the release process where my concurrent job configuration on the GitHub Actions workflow said "we're already running this job for the git tag so don't run it for `main` too". I've restarted the job now.

sfinkens:
Aha, it worked! So I'll close this one.

PR #2466, opened 2023-05-05
Remove compression for the mitiff writer

### LINKED ISSUE #2464: MITIFF writer using pillow: turn off compression due to rowsperstrip issues
## Feature Request

The satpy mitiff writer uses pillow to write the tiff file with special IMAGEDESCRIPTION tag used by DIANA https://github.com/metno/diana to handle geolocation etc. for visualization.

When compression is applied using pillow number of rowsperstrip is calculated by pillow and not possible to give by the user. https://github.com/python-pillow/Pillow/blob/main/src/PIL/TiffImagePlugin.py#L1721-L1729

DIANA can only handle raw strips, ie. the whole dataset as on strip.

So several solutions is possible.
1: Turn off the hard coded compression in the mitiff writer
2: Use other library than pillow that allow user set rows per strip
3: Implement TIFFReadEncodedStrip in DIANA.

I really, really really don't want to do 3.
For 2, I don't have any idea right now.
So then I end up with 1. This will lead to slighter bigger files, but I think this is acceptable.

Any other ideas before I starting doing this?

Please assign this to me.

TAlonglong:
Remove compression for the mitiff writer. The intended usage (DIANA) does not support the encoded stripes created by PIL when compression is applied.

 - [x] Closes #2464 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/writers/mitiff.py
+++ b/satpy/writers/mitiff.py
@@ -640,7 +640,7 @@ class MITIFFWriter(ImageWriter):
                 LOG.error("In a mitiff palette image a color map must be provided: palette_color_map is missing.")
                 return
 
-            img.save(tmp_gen_filename, compression='tiff_deflate', compress_level=9, tiffinfo=tiffinfo)
+            img.save(tmp_gen_filename, compression='raw', compress_level=9, tiffinfo=tiffinfo)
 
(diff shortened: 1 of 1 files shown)


mraspaud (approved):
LGTM

TAlonglong:
Could this be merged?

PR #2473, opened 2023-05-10
Fix rayleigh correction not handling angles as required inputs

djhoese:
This fixes an issue identified by Dmitry on slack. This was an accidental bug introduced in #2454. I tested things in this way with the CREFL rayleigh correction that I use in Polar2Grid but not with the pyspectral rayleigh correction. The pyspectral/PSP version of the modifier does not properly handle angles being provided as hard requirements (`prerequisites`) instead of optionals (`optional_prerequisites`).

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #2478, opened 2023-05-12
Fix eccodes package names in setup.py, update documentation for setting up development environment.

sjoro:
This PR fixes the package names for `eccodes` in `setup.py`. Old package names `eccodes-python` and `python-eccodes` are deprecated, see:

https://pypi.org/project/eccodes-python/
https://confluence.ecmwf.int/display/UDOC/How+to+install+ecCodes+with+Python+bindings+in+conda+-+ecCodes+FAQ

For added confusion both `eccodes` and `python-eccodes` are available in conda-forge and can be installed, but the new installation of the package should be:
```
conda install eccodes
pip install eccodes
```
The `conda` installation installs the C/Fortran library and the command line tools, `pip` installs the python bindings to this `eccodes`-library. For this reason the documentation is also updated to reflect the correct procedure.

The new docs built without warnings on my machine.

### DIFF
--- a/doc/source/dev_guide/index.rst
+++ b/doc/source/dev_guide/index.rst
@@ -68,6 +68,12 @@ should be run from the root of the cloned Satpy repository (where the
 You can now edit the python files in your cloned repository and have them
 immediately reflected in your conda environment.
 
+All the required dependencies for a full development environment, i.e. running the
+tests and building the documentation, can be installed with::
+
+    conda install eccodes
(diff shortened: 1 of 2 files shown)


djhoese (approved):
I know this isn't your fault, but this is unacceptable to me. Why would they deprecate the python-eccodes package on PyPI when that's the name of the conda package? Ugh. Fine.

PR #2489, opened 2023-05-23
Add "neutral_resolution_band" kwarg to RatioSharpenedRGB/SelfSharpenedRGB

yukaribbba:
Current RatioSharpenedRGB/SelfSharpenedRGB compositor is based on one assumption: there's only one high-res band. However in some composites, like 'false_red_color' of VIIRS, there could be a second high-res band. And it will be wrongfully sharpened in current compositor. So this PR will fix it by adding a new kwarg called "neutral_resolution_band". 

Here's a comparison.

The current one
**R: I02            -- 375m            -- another high-res band (no need for sharpening but still got it)
G: M05/I01 -- 750m/375m -- high-res band (calculate ratio)
B: M04          -- 750m            -- low-res band (need sharpening)**
```
  false_red_color:
    compositor: !!python/name:satpy.composites.RatioSharpenedRGB
    prerequisites:
    - name: I02
      modifiers: [sunz_corrected_iband]
    - name: M05
    - name: M04
    optional_prerequisites:
    - name: I01
    standard_name: crefl
    high_resolution_band: green

```
(100% enlarged)
image
(100% enlarged -- I02)
image

This PR
**R: I02         -- 375m            -- another high-res band (marked as "neutral_resolution_band" so won't get involved)
G: M05/I01 -- 750m/375m -- high-res band (calculate ratio)
B: M15       -- 750m            -- low-res band (need sharpening)**
```
  false_red_color:
    compositor: !!python/name:satpy.composites.RatioSharpenedRGB
    prerequisites:
    - name: I02
      modifiers: [sunz_corrected_iband]
    - name: M05
    - name: M04
    optional_prerequisites:
    - name: I01
    standard_name: crefl
    high_resolution_band: green
    neutral_resolution_band: red
```
(100% enlarged)
image
(100% enlarged -- I02)
image

Additionally, if there's no high-resolution file provided, a non-sharpened, low-res image will come out instead.
**R: I02         -- 375m
G: M05  -- 750m
B: M15       -- 750m**
```
  false_red_color:
    compositor: !!python/name:satpy.composites.RatioSharpenedRGB
    prerequisites:
    - name: I02
      modifiers: [sunz_corrected_iband]
    - name: M05
    - name: M04
    optional_prerequisites:
    - name: I01
    standard_name: crefl
    high_resolution_band: green
    neutral_resolution_band: red
```
image

Note that "neutral_resolution_band" should be considered as a complementary to "high_resolution_band". So only when the high-resolution file is given and "high_resolution_band" is set, this kwarg takes effect.

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1114,7 +1114,7 @@ class RatioSharpenedRGB(GenericCompositor):
 
         return bands['red'], bands['green'], bands['blue'], new_attrs
 
-    def _sharpen_bands_with_high_res(self, bands, high_res):
+    def _sharpen_bands_with_high_res(self, bands, high_res): 
         ratio = da.map_blocks(
             _get_sharpening_ratio,
             high_res.data,
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1047,7 +1047,8 @@ class RatioSharpenedRGB(GenericCompositor):
         R_hi -  500m resolution - shape=(4000, 4000)
 
     To avoid the green band getting involved in calculating ratio or sharpening,
-    specify it by "neutral_resolution_band: green" in YAML config file. Then::
+    specify it by "neutral_resolution_band: green" in YAML config file. Therefore,
+    only blue band will get sharpened::
 
         ratio = R_hi / R_lo
         new_R = R_hi
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1103,41 +1103,30 @@ class RatioSharpenedRGB(GenericCompositor):
             if 'rows_per_scan' in high_res.attrs:
                 new_attrs.setdefault('rows_per_scan', high_res.attrs['rows_per_scan'])
             new_attrs.setdefault('resolution', high_res.attrs['resolution'])
-            colors = ['red', 'green', 'blue', None]
-            low_resolution_index = colors.index(self.high_resolution_color)
-            neutral_resolution_index = colors.index(self.neutral_resolution_color)
-            neutral_res = datasets[neutral_resolution_index] if self.neutral_resolution_color is not None else None
 
         else:
             LOG.debug("No sharpening band specified for ratio sharpening")
             high_res = None
-            neutral_res = None
-            low_resolution_index = 0
-            neutral_resolution_index = 0
 
         if high_res is not None:
-            bands = locals()
-            colors = ['red', 'green', 'blue']
-            bands["low_res_" + self.high_resolution_color] = high_res
-            colors.remove(self.high_resolution_color)
-            low_res = (low_res_red, low_res_green, low_res_blue)[low_resolution_index]
+            bands = {'red': low_res_red, 'green': low_res_green, 'blue': low_res_blue}
+
             ratio = da.map_blocks(
                 _get_sharpening_ratio,
                 high_res.data,
-                low_res.data,
+                bands[self.high_resolution_color].data,
                 meta=np.array((), dtype=high_res.dtype),
                 dtype=high_res.dtype,
                 chunks=high_res.chunks,
             )
+
+            bands[self.high_resolution_color] = high_res
+
             with xr.set_options(keep_attrs=True):
-                if neutral_res is not None:
-                    if low_resolution_index != neutral_resolution_index:
-                        colors.remove(self.neutral_resolution_color)
-
-                for color in colors:
-                    bands["low_res_" + color] = bands["low_res_" + color] * ratio
-
-            return bands["low_res_red"], bands["low_res_green"], bands["low_res_blue"], new_attrs
+                for color in bands.keys():
+                    if color != self.neutral_resolution_color and color != self.high_resolution_color:
+                        bands[color] = bands[color] * ratio
+                return bands['red'], bands['green'], bands['blue'], new_attrs
 
         else:
             return low_res_red, low_res_green, low_res_blue, new_attrs
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1039,15 +1039,35 @@ class RatioSharpenedRGB(GenericCompositor):
         new_G = G * ratio
         new_B = B * ratio
 
+    In some cases, there could be another high resolution band:
+
+        R_lo -  1000m resolution - shape=(2000, 2000)
+        G_hi - 500m resolution - shape=(4000, 4000)
+        B - 1000m resolution - shape=(2000, 2000)
+        R_hi -  500m resolution - shape=(4000, 4000)
+
+    To avoid the green band getting involved in calculating ratio or sharpening,
+    specify it by "neutral_resolution_band: green" in YAML config file. Then:
+
+        ratio = R_hi / R_lo
+        new_R = R_hi
+        new_G = G_hi
+        new_B = B * ratio
+
     """
 
     def __init__(self, *args, **kwargs):
         """Instanciate the ration sharpener."""
         self.high_resolution_color = kwargs.pop("high_resolution_band", "red")
+        self.neutral_resolution_color = kwargs.pop("neutral_resolution_band", "red")
         if self.high_resolution_color not in ['red', 'green', 'blue', None]:
             raise ValueError("RatioSharpenedRGB.high_resolution_band must "
                              "be one of ['red', 'green', 'blue', None]. Not "
                              "'{}'".format(self.high_resolution_color))
+        if self.neutral_resolution_color not in ['red', 'green', 'blue', None]:
+            raise ValueError("RatioSharpenedRGB.neutral_resolution_band must "
+                             "be one of ['red', 'green', 'blue', None]. Not "
+                             "'{}'".format(self.neutral_resolution_color))
         super(RatioSharpenedRGB, self).__init__(*args, **kwargs)
 
     def __call__(self, datasets, optional_datasets=None, **info):
@@ -1082,12 +1102,22 @@ class RatioSharpenedRGB(GenericCompositor):
             if 'rows_per_scan' in high_res.attrs:
                 new_attrs.setdefault('rows_per_scan', high_res.attrs['rows_per_scan'])
             new_attrs.setdefault('resolution', high_res.attrs['resolution'])
-            low_res_colors = ['red', 'green', 'blue']
-            low_resolution_index = low_res_colors.index(self.high_resolution_color)
+            colors = ['red', 'green', 'blue']
+            low_resolution_index = colors.index(self.high_resolution_color)
+            high_resolution_index = low_resolution_index
+            if self.neutral_resolution_color is not None:
+                neutral_resolution_index = colors.index(self.neutral_resolution_color)
+                neutral_res = datasets[neutral_resolution_index]
+            else:
+                neutral_res = None
+                neutral_resolution_index = 0
         else:
             LOG.debug("No sharpening band specified for ratio sharpening")
             high_res = None
+            neutral_res = None
             low_resolution_index = 0
+            high_resolution_index = 0
+            neutral_resolution_index = 0
 
         if high_res is not None:
             low_res = (low_res_red, low_res_green, low_res_blue)[low_resolution_index]
@@ -1100,9 +1130,27 @@ class RatioSharpenedRGB(GenericCompositor):
                 chunks=high_res.chunks,
             )
             with xr.set_options(keep_attrs=True):
-                low_res_red = high_res if low_resolution_index == 0 else low_res_red * ratio
-                low_res_green = high_res if low_resolution_index == 1 else low_res_green * ratio
-                low_res_blue = high_res if low_resolution_index == 2 else low_res_blue * ratio
+                if neutral_res is not None:
+                    if high_resolution_index == 0:
+                        low_res_red = high_res
+                        low_res_green = neutral_res if neutral_resolution_index == 1 else low_res_green * ratio
+                        low_res_blue = neutral_res if neutral_resolution_index == 2 else low_res_blue * ratio
+
+                    elif high_resolution_index == 1:
+                        low_res_red = neutral_res if neutral_resolution_index == 0 else low_res_red * ratio
+                        low_res_green = high_res
+                        low_res_blue = neutral_res if neutral_resolution_index == 2 else low_res_blue * ratio
+
+                    elif high_resolution_index == 2:
+                        low_res_red = neutral_res if neutral_resolution_index == 0 else low_res_red * ratio
+                        low_res_green = neutral_res if neutral_resolution_index == 1 else low_res_green * ratio
+                        low_res_blue = high_res
+
+                else:
+                    low_res_red = high_res if high_resolution_index == 0 else low_res_red * ratio
+                    low_res_green = high_res if high_resolution_index == 1 else low_res_green * ratio
+                    low_res_blue = high_res if high_resolution_index == 2 else low_res_blue * ratio
+
         return low_res_red, low_res_green, low_res_blue, new_attrs
 
     def _combined_sharpened_info(self, info, new_attrs):
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1109,27 +1109,27 @@ class RatioSharpenedRGB(GenericCompositor):
             high_res = None
 
         bands = {'red': low_res_red, 'green': low_res_green, 'blue': low_res_blue}
-        self._sharpen_bands_with_high_res(bands, high_res)
+        if high_res is not None:
+            self._sharpen_bands_with_high_res(bands, high_res)
 
         return bands['red'], bands['green'], bands['blue'], new_attrs
 
     def _sharpen_bands_with_high_res(self, bands, high_res):
-        if high_res is not None:
-            ratio = da.map_blocks(
-                _get_sharpening_ratio,
-                high_res.data,
-                bands[self.high_resolution_color].data,
-                meta=np.array((), dtype=high_res.dtype),
-                dtype=high_res.dtype,
-                chunks=high_res.chunks,
-            )
+        ratio = da.map_blocks(
+            _get_sharpening_ratio,
+            high_res.data,
+            bands[self.high_resolution_color].data,
+            meta=np.array((), dtype=high_res.dtype),
+            dtype=high_res.dtype,
+            chunks=high_res.chunks,
+        )
 
-            bands[self.high_resolution_color] = high_res
+        bands[self.high_resolution_color] = high_res
 
-            with xr.set_options(keep_attrs=True):
-                for color in bands.keys():
-                    if color != self.neutral_resolution_color and color != self.high_resolution_color:
-                        bands[color] = bands[color] * ratio
+        with xr.set_options(keep_attrs=True):
+            for color in bands.keys():
+                if color != self.neutral_resolution_color and color != self.high_resolution_color:
+                    bands[color] = bands[color] * ratio
 
     def _combined_sharpened_info(self, info, new_attrs):
         combined_info = {}
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1109,6 +1109,11 @@ class RatioSharpenedRGB(GenericCompositor):
             high_res = None
 
         bands = {'red': low_res_red, 'green': low_res_green, 'blue': low_res_blue}
+        self._sharpen_bands_with_high_res(bands, high_res)
+
+        return bands['red'], bands['green'], bands['blue'], new_attrs
+
+    def _sharpen_bands_with_high_res(self, bands, high_res):
         if high_res is not None:
             ratio = da.map_blocks(
                 _get_sharpening_ratio,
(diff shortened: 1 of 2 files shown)


djhoese:
This is also the case for some true/false/natural color RGBs with ABI or MODIS or VIIRS so it is definitely common. However, I don't think I needed to use it because of the ways I was separating the bands in my own processing. I'm OK with this feature if you get the tests passing.

Should the value be a list to specify multiple bands? I suppose the only case where that would be needed is if this compositor also supports RGBA generation with 4 bands as inputs.

yukaribbba:
> However, I don't think I needed to use it because of the ways I was separating the bands in my own processing.

Yeah I also did it that way but sometimes I still need this for a quicklook. Well if this compositor can produce RGBA images then the kwarg should be a list. But I haven't found such one so I think I'll leave it for future. For now this is just a minor update.

I can't get through the test but it looks like not my fault. It seems other people's PRs also have the same errors.

```
FAILED satpy/tests/scene_tests/test_conversions.py::TestSceneConversions::test_geoviews_basic_with_area - ImportError: cannot import name 'LinkCallback' from 'holoviews.plotting.bokeh.callbacks' (/usr/share/miniconda3/envs/test-environment/lib/python3.11/site-packages/holoviews/plotting/bokeh/callbacks.py)
FAILED satpy/tests/scene_tests/test_conversions.py::TestSceneConversions::test_geoviews_basic_with_swath - ImportError: cannot import name 'LinkCallback' from 'holoviews.plotting.bokeh.callbacks' (/usr/share/miniconda3/envs/test-environment/lib/python3.11/site-packages/holoviews/plotting/bokeh/callbacks.py)
FAILED satpy/tests/writer_tests/test_cf.py::TestEncodingKwarg::test_encoding_kwarg[True] - assert 0 == 7
FAILED satpy/tests/writer_tests/test_cf.py::TestEncodingAttribute::test_encoding_kwarg[True] - assert 0 == 7
FAILED satpy/tests/writer_tests/test_cf.py::TestEncodingAttribute::test_encoding_attribute[True] - assert 0 == 7
```

djhoese:
I just merged a PR that fixed the issues with CI. Try rebasing your main with the pytroll main (or merging) and things should start passing. In the future I highly recommend making a *new* branch specific to the PR you're going to make. Using `main` makes it difficult for cases like this where you need to synchronize with upstream `main`.

yukaribbba:
Got it.

satpy/composites/__init__.py
```diff
@@ -1039,15 +1039,36 @@ class RatioSharpenedRGB(GenericCompositor):
...
+    In some cases, there could be another high resolution band::
+
+        R_lo -  1000m resolution - shape=(2000, 2000)
+        G_hi - 500m resolution - shape=(4000, 4000)
+        B - 1000m resolution - shape=(2000, 2000)
+        R_hi -  500m resolution - shape=(4000, 4000)
+
+    To avoid the green band getting involved in calculating ratio or sharpening,
+    specify it by "neutral_resolution_band: green" in YAML config file. Therefore,
+    only blue band will get sharpened::
```
djhoese, line 1051:
```suggestion
    To avoid the green band getting involved in calculating ratio or sharpening,
    add "neutral_resolution_band: green" in the YAML config file. This way
    only the blue band will get sharpened::
```

satpy/composites/__init__.py
```diff
@@ -1082,12 +1103,22 @@ def _get_and_sharpen_rgb_data_arrays_and_meta(self, datasets, optional_datasets)
...
-            low_resolution_index = low_res_colors.index(self.high_resolution_color)
+            colors = ['red', 'green', 'blue']
+            low_resolution_index = colors.index(self.high_resolution_color)
+            high_resolution_index = low_resolution_index
+            if self.neutral_resolution_color is not None:
+                neutral_resolution_index = colors.index(self.neutral_resolution_color)
+                neutral_res = datasets[neutral_resolution_index]
+            else:
+                neutral_res = None
+                neutral_resolution_index = 0
```
djhoese, line 1114: Is there a way to clean this up? What if `None` was added to the `colors` list? That way you could just do:

```python
neutral_resolution_index = colors.index(self.neutral_resolution_color)
neature_res = datasets[neutral_resolution_index] if neutral_resolution_index is not None else None
```

Not sure if that actually cleans anything up...but yeah.

satpy/composites/__init__.py
```diff
@@ -1100,9 +1131,27 @@ def _get_and_sharpen_rgb_data_arrays_and_meta(self, datasets, optional_datasets)
...
+
+                    elif high_resolution_index == 2:
+                        low_res_red = neutral_res if neutral_resolution_index == 0 else low_res_red * ratio
+                        low_res_green = neutral_res if neutral_resolution_index == 1 else low_res_green * ratio
+                        low_res_blue = high_res
+
+                else:
+                    low_res_red = high_res if high_resolution_index == 0 else low_res_red * ratio
+                    low_res_green = high_res if high_resolution_index == 1 else low_res_green * ratio
+                    low_res_blue = high_res if high_resolution_index == 2 else low_res_blue * ratio
```
djhoese, line 1153: What if we put this ternary operation (`X if index == Y else Z * ratio`) in a separate method and just called it for each band:

```
low_res_red = self._choose_sharpest_band(low_res_red, high_res, high_resolution_index, ratio)
```

Ugh nevermind, that doesn't handle the neutral band check...but that's ugly too. Oh but it also seems like you added the `high_resolution_index` and stopped using `low_resolution_index`. That shouldn't be needed, right?

I'm really unsure of this block of if/else statements. There has to be a cleaner way, but I'm having trouble thinking of it.

satpy/tests/test_composites.py
```diff
@@ -216,17 +221,72 @@ def test_basic_no_sharpen(self):
...
+        """Test that high resolution blue with red neutral band can be passed."""
+        from satpy.composites import RatioSharpenedRGB
+        comp = RatioSharpenedRGB(name='true_color', high_resolution_band='blue', neutral_resolution_band='red')
+        res = comp((self.ds1, self.ds2, self.ds3), optional_datasets=(self.ds4,))
+        res = res.values
+        self.assertEqual(res.shape, (3, 2, 2))
+        np.testing.assert_allclose(res[0], np.array([[5.0, 5.0], [np.nan, 0.0]], dtype=np.float64))
+        np.testing.assert_allclose(res[1], np.array([[0.75, 0.75], [np.nan, 0.75]], dtype=np.float64))
+        np.testing.assert_allclose(res[2], np.array([[1.0, 1.0], [np.nan, 1.0]], dtype=np.float64))
+
```
djhoese, line 289: Maybe we should convert this class to a pytest-based class (no `unittest.TestCase` subclass), remove the `self.assertX` statements, and then use `@pytest.mark.parametrize` to combine all these tests together. They are all very similar.

satpy/composites/__init__.py
djhoese: I think the main confusion here is that `neutral_res` doesn't need to be treated as an "action" that we are performing on the band. We are just not doing anything extra to that band. In the case of the high res band/index, we are swapping. I'm still not exactly sure how to clean the logic up though.

djhoese (changes requested):
Thanks for putting this together. I had a lot of comments and things I'm unsure of that I commented inline. Another thought I had while reviewing this though is that it takes away from the automatic choosing/handling of the cases where the high resolution band isn't available. So for example in the VIIRS case if you didn't provide the I-band resolution data files then asking for a ratio sharpened true color would just get you the low resolution version. With this neutral resolution functionality, maybe it would be best if there were two optional datasets available: the high resolution band used for ratios and the neutral one used inplace of the low resolution option. Hhhmmm but that makes it ugly in cases where there is no low resolution version of the "neutral" band.

For now I'll post this review that is likely very confusing and I'll have to think about these things offline. The functionality is good, but I think there are cleaner ways to handle some of the logic.

satpy/composites/__init__.py
```diff
@@ -1039,15 +1039,36 @@ class RatioSharpenedRGB(GenericCompositor):
...
+        new_R = R_hi
+        new_G = G_hi
+        new_B = B * ratio
+
     """
 
     def __init__(self, *args, **kwargs):
         """Instanciate the ration sharpener."""
         self.high_resolution_color = kwargs.pop("high_resolution_band", "red")
+        self.neutral_resolution_color = kwargs.pop("neutral_resolution_band", "red")
```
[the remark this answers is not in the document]
yukaribbba: For back-compatibility `None` could be more appropriate.

yukaribbba:
> Another thought I had while reviewing this though is that it takes away from the automatic choosing/handling of the cases where the high resolution band isn't available.

Well the "neutral" block is completely under "high res", so if you don't give any high res files it will produce a low res image, just like current one. I've updated the examples above so you can see that case.
I want this PR have minor impact on the existed script/yaml files. So having two "optional datasets" could be confusing for some users....

djhoese:
pre-commit.ci autofix

satpy/composites/__init__.py
```diff
@@ -1082,14 +1103,23 @@ def _get_and_sharpen_rgb_data_arrays_and_meta(self, datasets, optional_datasets)
...
         else:
             LOG.debug("No sharpening band specified for ratio sharpening")
             high_res = None
+            neutral_res = None
             low_resolution_index = 0
+            neutral_resolution_index = 0
 
         if high_res is not None:
+            bands = locals()
+            colors = ['red', 'green', 'blue']
```
djhoese, line 1120: Can we combine these in a more explicit dictionary? `bands = {"red": low_res_red, ...}`? Or actually, you never store anything except low res in it so could it just be `bands = {"red": low_res_red, "green": low_res_green, "blue": low_res_blue}` then you do the same overwrite for the high res color. If you did the high res assignment after the `ratio` is created, then the ratio `map_blocks` call could use `bands[self.high_resolution_color]` and `high_res` (decently clean...I think?).

Then...in the `with xr.set_options` block you could iterate over the colors (`for color in bands.keys()`) and do `if color != neutral_resolution_color and color != high_resolution_color: bands[color] = bands[color] * ratio`. Right? Thoughts? I think your idea to use dictionary-style access really makes this simple and then with my suggestions I think you can get rid of the `_index` and `neutral_res` variables.

satpy/tests/test_composites.py
```diff
@@ -129,7 +129,7 @@ def test_nondimensional_coords(self):
         self.assertNotIn('acq_time', ret_datasets[0].coords)
 
 
-class TestRatioSharpenedCompositors(unittest.TestCase):
+class TestRatioSharpenedCompositors:
     """Test RatioSharpenedRGB and SelfSharpendRGB compositors."""
 
     def setUp(self):
```
djhoese, line 135: Change this to `def setup_method` and add the `self.ds1 =`-style lines back. Unless this data is really only used by one test in this class. If so, then it should be renamed to something like `_create_fake_data`.

satpy/tests/test_composites.py
```diff
@@ -170,86 +169,147 @@ def setUp(self):
...
+    @pytest.mark.parametrize(
+        ("case", "exp"),
+        [
+            ("high bad color", ValueError),
+            ("neutral bad color", ValueError),
+            ("match_data_arrays", satpy.composites.IncompatibleAreas),
+            ("more than three datasets", ValueError),
+            ("no high res band in self sharpened", ValueError)
+        ]
+    )
```
djhoese, line 195: At this point it might be better to just have separate tests, but...

What if you did this as parameters `("init_kwargs", "call_kwargs", "init_exc", "call_exc")`. But this would require some "fun" context managers to wrap `pytest.raises` with the boolean like `if init_exc is not None:`. Or maybe you can provide an empty tuple to `pytest.raises`...I'm not sure.

Or you could split these into `test_init_errors` and `test_call_errors`. Or split them into separate tests.

If you can't describe the cases as changes in input (even if you use a data producing function as one of the parameters) then they should probably be separate tests.

djhoese (changes requested):
I had a couple more ideas. Let me know what you think. We're getting there.

djhoese:
Oh also, I had pre-commit (the service) fix your code a bit so you'll need to make sure to do a `git pull` before you make any more commits/updates. Otherwise the commit history will be a little wonky.

satpy/composites/__init__.py
yukaribbba: Now it's really cleaned up.

yukaribbba:
Ready

satpy/composites/__init__.py
```diff
@@ -1082,28 +1103,33 @@ def _get_and_sharpen_rgb_data_arrays_and_meta(self, datasets, optional_datasets)
...
             new_attrs.setdefault('resolution', high_res.attrs['resolution'])
-            low_res_colors = ['red', 'green', 'blue']
-            low_resolution_index = low_res_colors.index(self.high_resolution_color)
+
         else:
             LOG.debug("No sharpening band specified for ratio sharpening")
             high_res = None
-            low_resolution_index = 0
 
         if high_res is not None:
```
djhoese, line 1111: This condition and the `else` could be swapped to clean up the indentation. So `if high_res is None: return low_res_red, ...` and then no `else`, but instead dedent the version with the high resolution. Or you could extract the `bands = {...}` definition outside the if statement and do `return bands['red'], ...` outside of the `if` statement.

I'm just trying to think of ways to make CodeScene happier about the code quality.

satpy/tests/test_composites.py
```diff
@@ -173,83 +175,165 @@ def setUp(self):
...
+                                               (-2000, -2000, 2000, 2000))
+        self.ds4_big = ds4_big
 
-    def test_bad_color(self):
+    @pytest.mark.parametrize(
+        "exp",
+        [
+            ValueError,
+        ]
+    )
```
djhoese, line 194: This could be removed as-is since there is only one parameter. You could combine these two "bad color" tests where you have parameters `("init_kwargs", "exp_exception")`. Actually, it is `ValueError` in both cases so I guess it could be removed in that case too.

satpy/tests/test_composites.py
```diff
@@ -173,83 +175,155 @@ def setUp(self):
...
+            "bad"
+        ]
+    )
+    def test_bad_colors(self, init_kwarg):
         """Test that only valid band colors can be provided."""
         from satpy.composites import RatioSharpenedRGB
-        self.assertRaises(ValueError, RatioSharpenedRGB, name='true_color', high_resolution_band='bad')
+        with pytest.raises(ValueError):
+            RatioSharpenedRGB(name='true_color', high_resolution_band="red", neutral_resolution_band=init_kwarg)
+            RatioSharpenedRGB(name='true_color', high_resolution_band=init_kwarg, neutral_resolution_band="red")
```
djhoese, line 201: I'm sorry, I don't understand what you're trying to do here. `"bad"` in the init_kwarg should be a dictionary of kwargs to pass to `RatioSharpenedRGB`. There should only be one call/creation of this class inside this single `raises`.

yukaribbba: my fault:sweat_smile:

satpy/composites/__init__.py
```diff
@@ -1082,28 +1103,31 @@ def _get_and_sharpen_rgb_data_arrays_and_meta(self, datasets, optional_datasets)
...
+
             with xr.set_options(keep_attrs=True):
-                low_res_red = high_res if low_resolution_index == 0 else low_res_red * ratio
-                low_res_green = high_res if low_resolution_index == 1 else low_res_green * ratio
-                low_res_blue = high_res if low_resolution_index == 2 else low_res_blue * ratio
-        return low_res_red, low_res_green, low_res_blue, new_attrs
+                for color in bands.keys():
+                    if color != self.neutral_resolution_color and color != self.high_resolution_color:
+                        bands[color] = bands[color] * ratio
+                return bands['red'], bands['green'], bands['blue'], new_attrs
```
djhoese, line 1128: This line isn't needed.

satpy/tests/test_composites.py
```diff
@@ -173,83 +175,153 @@ def setUp(self):
...
+        ds4_big.attrs['area'] = AreaDefinition('test', 'test', 'test',
+                                               {'proj': 'merc'}, 4, 4,
+                                               (-2000, -2000, 2000, 2000))
+        self.ds4_big = ds4_big
 
-    def test_bad_color(self):
+    @pytest.mark.parametrize(
+        "init_kwargs",
+        [
+            {'name': "true_color", "high_resolution_band": "bad", "neutral_resolution_band": "bad"},
```
djhoese, line 192: You want two cases in this list. One is the bad high resolution band, the other (a separate dictionary) is the bad neutral resolution band. So a list with two dictionaries in it.

satpy/tests/test_composites.py
```diff
@@ -173,83 +175,153 @@ def setUp(self):
...
+        with pytest.raises(ValueError):
+            RatioSharpenedRGB(**init_kwargs)
 
-    def test_match_data_arrays(self):
+    @pytest.mark.parametrize(
+        "exp",
+        [
+            satpy.composites.IncompatibleAreas,
+        ]
+    )
```
djhoese, line 206: This test case and all others that have a parametrize with only one case should have the parameters removed.

satpy/tests/test_composites.py
```diff
@@ -173,83 +175,153 @@ def setUp(self):
...
 
-    def test_basic_no_sharpen(self):
-        """Test that color None does no sharpening."""
+    @pytest.mark.parametrize(
+        ("case", "exp"),
+        [
+            ("without optional high res", (3, 2, 2)),
+            ("high res band is None", (3, 2, 2))
+        ]
+    )
```
djhoese, line 246: These string cases should be removed and instead pass init_kwargs and maybe a boolean for whether or not optional_datasets should be passed to the compositor.

```
call_kwargs = {}
if pass_opt:
    call_kwargs["optional_datasets = (self.ds4,)
comp = RatioSharpenedRGB(name="true_color", **init_kwargs)
res = comp((self.ds1, self.ds2, self.ds3), **call_kwargs)
```

satpy/tests/test_composites.py
djhoese: Since `name` isn't different between the cases, this can be hardcoded in the test as `name="true_color"` and then followed by `**init_kwargs`.

satpy/tests/test_composites.py
yukaribbba: I decide to split them just like orginal ones for better understanding

djhoese (changes requested):
I had some questions about some of the tests and one more request on cleaning up the code to make CodeScene a little happier. It also seems like pre-commit wasn't run so pre-commit.ci is complaining about some small things. I can fix those after the other changes are made though.

satpy/tests/test_composites.py
```diff
@@ -173,83 +176,130 @@ def setUp(self):
...
+            comp((self.ds1, self.ds2, self.ds3), optional_datasets=(self.ds4_big,))
 
     def test_more_than_three_datasets(self):
         """Test that only 3 datasets can be passed."""
-        from satpy.composites import RatioSharpenedRGB
-        comp = RatioSharpenedRGB(name='true_color')
-        self.assertRaises(ValueError, comp, (self.ds1, self.ds2, self.ds3, self.ds1),
-                          optional_datasets=(self.ds4_big,))
+        from satpy.composites import SelfSharpenedRGB
+        comp = SelfSharpenedRGB(name='true_color', high_resolution_band=None)
```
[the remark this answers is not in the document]
yukaribbba: Sorry that's a copy paste error

satpy/tests/test_composites.py
```diff
@@ -173,83 +176,130 @@ def setUp(self):
...
+        with pytest.raises(ValueError):
+            comp((self.ds1, self.ds2, self.ds3))
+
+    def test_self_sharpened_no_high_res(self):
+        """Test for exception when no high_res band is specified."""
+        from satpy.composites import SelfSharpenedRGB
+        comp = SelfSharpenedRGB(name='true_color', high_resolution_band=None)
+        with pytest.raises(ValueError):
+            comp((self.ds1, self.ds2, self.ds3))
```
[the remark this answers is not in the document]
yukaribbba: No it's okay. The above one is the problem.

satpy/composites/__init__.py
```diff
@@ -1082,28 +1103,33 @@ def _get_and_sharpen_rgb_data_arrays_and_meta(self, datasets, optional_datasets)
...
+            bands[self.high_resolution_color] = high_res
+
             with xr.set_options(keep_attrs=True):
-                low_res_red = high_res if low_resolution_index == 0 else low_res_red * ratio
-                low_res_green = high_res if low_resolution_index == 1 else low_res_green * ratio
-                low_res_blue = high_res if low_resolution_index == 2 else low_res_blue * ratio
-        return low_res_red, low_res_green, low_res_blue, new_attrs
+                for color in bands.keys():
+                    if color != self.neutral_resolution_color and color != self.high_resolution_color:
+                        bands[color] = bands[color] * ratio
```
djhoese, line 1132: I was thinking `if high_res is not None:` would stay outside the method. If you want it inside the method could you invert the condition and do:

```
if high_res is None:
    return
... high res code ...
```

yukaribbba:
CI keeps failing in coveralls stage.

djhoese:
I think coveralls was just having a hard time. I've restarted the jobs and we'll see how things go now.

PR #2495, opened 2023-05-25
Add support for temporal composites

### LINKED ISSUE #1748: Define and calculate composites with temporal dependencies
## Feature Request

Satpy can calculate composites from datasets within a single `Scene`.  For some use cases, it would be desirable to calculate composites from datasets from multiple scenes, such as multiple time steps in a `MultiScene`.  In nowcasting, the temporal component contains important information for forecasting the near future: rising air parcels may indicate convective initiation, a vertically growing cloud may indicate a rapidly developing thunderstorm, a cloud growing rapidly horizontally may tell us something about the expected lifetime of a convective system, and a rapid increase in lightning density likewise.  Currently, it is difficult to build such time-dependent "composites" into Satpy.

I would like to be able to define composites that need multiple time steps of a certain variable.  Such composites could only be loaded from a `MultiScene`, or _maybe_ from a blended `Scene` where the datasets have a time component.  

In `composites.yaml`, we would define what time range or time steps are needed.  Maybe a product that needs 6.2 µm and 7.3 µm for the present and the previous scene could be defined as:

```
prerequisites:
    - wavelength: 6.2
    - wavelength: 6.2
      time: -1
    - wavelength: 7.3
    - wavelength: 7.3
      time: -1
```

The compositor would be passed four inputs: two from the present scene, and two from the previous scene.

Or a composite that needs the last five minutes of flash extent density:

```
prerequisites:
    - name: flash_extent_density
      time_from: -5 minutes  # anything understood by `pandas.Timedelta`?
      time_to: 0  # alternative: now
```

The compositor would be passed a `DataArray` with a time dimension, containing the last five minutes of `flash_extent_density`.

Inevitably, there will be time steps, probably the first time steps, where those conditions cannot be met.  For those time steps, the composite would be filled with fill values.

Compositors could also reduce the data.  If we have a `MultiScene` with sixty `Scene`s containing `flash_extent_density`, one `Scene` for every minute, the second example above could be used to calculate a running average starting at t=5 minutes.  But what if, instead of a running average, we want to get a `MultiScene` with only 12 scenes, each containing the summed `flash_extent_density` for a block of 5 minutes?  It would be interesting if we could "load" this in Satpy as well, but this has more potential caveats than non-reducing operations.  Maybe rather than a `.load(...)` operation that reduces the number of scenes in the `MultiScene`, this should have an API that returns a new, reduced `MultiScene` leaving the original intact.

In principle, there should be no backward compatibility concerns, as this concerns a new feature.  However, it's possible that it would require such an overhaul of `MultiScene` and perhaps `Scene`, that user workflow or existing unit tests would need to be changed.  I don't foresee right now what those changes would be.

**Additional context**

I have code that takes a `MultiScene` and calculates a value based on the Δ between two subsequent scenes (akin to the first example above), and then manually assigns this to each `Scene`.  That works, but it does not very well fit in the `Satpy` framework.  The `Scene` class is not really aware of the resulting dataset (see also, for example, #1747), and I cannot define this procedure in a composites file.  An even dirtier workaround that would be possible in present satpy is to create a reader that reads multiple time steps and defines datasets such as "BT73_1_minute_ago".  That would be a bad solution for various reasons.

See also #1747 and #1469.

gerritholl:
Add a framework for temporal composites.

Current status
===

Implementation:
---

Draft.

When loading a temporal composite from a multiscene, the first iteration is failing as expected, because time dimension information is missing.  The `MultiScene` blend function then calls `generate_possible_composites` after timeseries blending, and the composite is loaded.

See below for a list of open tasks.

Unit tests
---

Passing.

Documentation
---

Present.

 - [x] Closes #1748 
 - [x] Tests added 
 - [x] Fully documented 
 - [x] Fix implementation
 - [ ] Don't throw a warning for every single scene
 - [ ] Disambiguate between "needs resampling" and "needs blending", as the latter is going to be very rare and such a message will just confuse users
 - [x] Add functionality to handle the temporal component of the prerequisites
 - [x] Add one or more composites that use the new temporal rgb compositor (this also helps as documentation).
 - [x] Add time tolerances or slicing, for the (common) case where start times are not rounded

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1701,3 +1701,28 @@ class LongitudeMaskingCompositor(SingleBandCompositor):
 
         masked_projectable = projectable.where(lon_min_max)
         return super().__call__([masked_projectable], **info)
+
+
+class MissingTime(Exception):
+    """Raised when temporal composite building lacks time dimension."""
+
+
+class BaseTemporalCompositor(CompositeBase):
+    """Compositors combining multiple time steps.
+
+    Base class for compositors that combine inputs from two or more time steps.
+    """
+
+    def __call__(self, projectables, nonprojectables=None, **info):
+        """Build the composite."""
+        self._check_time_dimension(projectables)
+        return super().__call__(projectables, nonprojectables, **info)
+
+    def _check_time_dimension(self, projectables):
+        """Make sure all projectables have a time dimension."""
+        for projectable in projectables:
+            if "time" not in projectable.dims:
+                raise MissingTime(
+                    "Creating temporal composite needs time dimensions. "
+                    "Typically, this comes from starting with a MultiScene "
+                    "and then performing timeseries blending.")
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1770,6 +1770,7 @@ class TemporalRGB(BaseTemporalCompositor):
 
     def __call__(self, projectables, optional_datasets=None, **info):
         """Build the composite."""
+        projectables = self.match_data_arrays(projectables)
         self._check_time_dimension(projectables)
         new_projectables = self._apply_temporal_prerequisites(projectables)
         # need to drop the time coordinate, otherwise we can't concatenate
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1722,7 +1722,7 @@ class BaseTemporalCompositor(CompositeBase):
     the generation should work.
     """
 
-    def __init__(self, name, prerequisites, optional_prerequisites, **kwargs):
+    def __init__(self, name, prerequisites, optional_prerequisites=None, **kwargs):
         """Initialise a temporal compositor."""
         self._ensure_prerequisites_have_times(prerequisites)
         super().__init__(name, prerequisites, optional_prerequisites, **kwargs)
@@ -1766,7 +1766,11 @@ class BaseTemporalCompositor(CompositeBase):
 
 
 class TemporalRGB(BaseTemporalCompositor):
-    """Make an RGB where different timesteps of the same channel go in different bands."""
+    """Make an RGB where different timesteps of the same channel go in different bands.
+
+    See the note in the parent class :class:`BaseTemporalCompositor` for
+    usage instructions.
+    """
 
     def __call__(self, projectables, optional_datasets=None, **info):
         """Build the composite."""
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1760,7 +1760,7 @@ class BaseTemporalCompositor(CompositeBase):
         ref_time = projectables[0]["time"][-1]
         new_projectables = []
         for (dq, proj) in zip(self.attrs["prerequisites"], projectables):
-            new_da = proj.sel(time=ref_time+pd.Timedelta(dq["time"]))
+            new_da = proj.sel(time=ref_time+pd.Timedelta(dq["time"]), method='nearest')
             new_projectables.append(new_da)
         return new_projectables
 
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -22,6 +22,7 @@ import warnings
 
 import dask.array as da
 import numpy as np
+import pandas as pd
 import xarray as xr
 from trollimage.colormap import Colormap
 
@@ -245,6 +246,16 @@ class CompositeBase:
                       "'{}'".format(self.attrs['name']))
             raise IncompatibleAreas("Areas are different")
 
+    def _concat_datasets(self, projectables, mode):
+        try:
+            data = xr.concat(projectables, 'bands', coords='minimal')
+            data['bands'] = list(mode)
+        except ValueError as e:
+            LOG.debug("Original exception for incompatible areas: {}".format(str(e)))
+            raise IncompatibleAreas
+
+        return data
+
 
 class DifferenceCompositor(CompositeBase):
     """Make the difference of two data arrays."""
@@ -396,16 +407,6 @@ class GenericCompositor(CompositeBase):
         if 'bands' in data_arr.coords and isinstance(data_arr.coords['bands'][0].item(), str):
             return ''.join(data_arr.coords['bands'].values)
         return cls.modes[data_arr.sizes['bands']]
-
-    def _concat_datasets(self, projectables, mode):
-        try:
-            data = xr.concat(projectables, 'bands', coords='minimal')
-            data['bands'] = list(mode)
-        except ValueError as e:
-            LOG.debug("Original exception for incompatible areas: {}".format(str(e)))
-            raise IncompatibleAreas
-
-        return data
 
     def _get_sensors(self, projectables):
         sensor = set()
@@ -1711,12 +1712,20 @@ class BaseTemporalCompositor(CompositeBase):
     """Compositors combining multiple time steps.
 
     Base class for compositors that combine inputs from two or more time steps.
+
+    To use this, the user must start with a :class:`MultiScene` and load the
+    temporal composite from there.  Composite generation will fail due to missing
+    temporal information in the containing scenes, but it will add the temporal
+    composite to the scene wishlists.  Now we run ``MultiScene.blend(timeseries)``,
+    which will create DataArrays with a time dimension.  After blending, the
+    blend method tries again to generate the composites, and on this second run
+    the generation should work.
     """
 
-    def __call__(self, projectables, nonprojectables=None, **info):
-        """Build the composite."""
-        self._check_time_dimension(projectables)
-        return super().__call__(projectables, nonprojectables, **info)
+    def __init__(self, name, prerequisites, optional_prerequisites, **kwargs):
+        """Initialise a temporal compositor."""
+        self._ensure_prerequisites_have_times(prerequisites)
+        super().__init__(name, prerequisites, optional_prerequisites, **kwargs)
 
     def _check_time_dimension(self, projectables):
         """Make sure all projectables have a time dimension."""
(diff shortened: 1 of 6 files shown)


satpy/multiscene/_multiscene.py
```diff
@@ -359,6 +359,10 @@ def blend(
             datasets = [scn[ds_id] for scn in self.scenes if ds_id in scn]
             new_scn[ds_id] = blend_function(datasets)
 
+        new_scn._wishlist = self.first_scene.wishlist.copy()
+        # without copying this, there is a KeyError in dependency_tree.trunk
+        new_scn._dependency_tree = self.first_scene._dependency_tree.copy()
+        new_scn.generate_possible_composites(True)
```
djhoese, line 365: Yeah maybe we do basically need to copy the first Scene just like `Scene.copy` does:

https://github.com/pytroll/satpy/blob/3fd4c458cab81df54043b16de7e43640a6b6e72f/satpy/scene.py#L562-L576

gerritholl: The failure I get if not copying the dependency tree:

```
../../multiscene/_multiscene.py:365: in blend
    new_scn.generate_possible_composites(True)
../../scene.py:1430: in generate_possible_composites
../../scene.py:1446: in _generate_composites_from_loaded_datasets
    trunk_nodes = self._dependency_tree.trunk(limit_nodes_to=self.missing_datasets,
../../dependency_tree.py:97: in trunk
    child_node = self._all_nodes[child_id]
../../dependency_tree.py:615: in __getitem__
    key = self.get_key(item)
../../dependency_tree.py:607: in get_key
    return get_key(match_key, self.keys())
_



        key = create_filtered_query(key, query)

        res = key.filter_dataids(key_container)
        if not res:
>           raise KeyError("No dataset matching '{}' found".format(str(key)))
E           KeyError: "No dataset matching 'DataQuery(name='temporal')' found"

../../dataset/data_dict.py:107: KeyError
```

gerritholl:
Although I still have some open ends, a first round of review would be already valuable, in case developers have doubt with the direction this is going.

pnuu:
No comments on the code, just a question: is this functional already? For example, can the "temporal RGB" in https://github.com/pytroll/satpy/pull/2488 already created? If yes, we could close that in favor of this more flexible approach. A simple example built-in temporal composite would be great, and at some point a short section to `MultiScene` documentation with a code example how to create the said composite even better :-)

gerritholl:
```python
import hdf5plugin
from glob import glob
from satpy.multiscene import MultiScene, timeseries

ms = MultiScene.from_files(
        glob("/media/nas/x21308/MTG_test_data/2022_05_MTG_Testdata/RC007[012]/*BODY*.nc"),
        reader="fci_l1c_nc",
        group_keys=["repeat_cycle_in_day"])
ms.load(["temporal_rgb_vis06"])
sc = ms.blend(blend_function=timeseries)
ls = sc.resample("eurol")
ls.save_datasets()
```

temporal_rgb_vis06_20170920_114008

pnuu:
Hmm. I tried this composite:
```yaml
  temporal_rgb_hrv:
    compositor: !!python/name:satpy.composites.TemporalRGB
    prerequisites:
      - name: HRV
        time: 0
      - name: HRV
        time: -60 min
      - name: HRV
        time: -120 min
    standard_name: temporal_rgb_hrv
```
with this code:
```python
import sys

from satpy import MultiScene
from satpy.multiscene import timeseries


fnames = sys.argv[1:]

mscn = MultiScene.from_files(
    fnames,
    reader='seviri_l1b_hrit',
    group_keys=['start_time'],
)
mscn.load(['temporal_rgb_hrv'])
blended = mscn.blend(blend_function=timeseries)
scn = blended.resample('euron1', resampler='gradient_search')
scn.save_datasets(filename="/tmp/{start_time:%Y%m%d_%H%M}_{platform_name}_{name}.png")
```
and get this trace:
```python
The following datasets were not created and may require resampling or temporal blending to be generated: DataID(name='temporal_rgb_hrv')
  File "pandas/_libs/index.pyx", line 548, in pandas._libs.index.DatetimeEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 2263, in pandas._libs.hashtable.Int64HashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 2273, in pandas._libs.hashtable.Int64HashTable.get_item
KeyError: 1683705309504000000

  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 3802, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas/_libs/index.pyx", line 516, in pandas._libs.index.DatetimeEngine.get_loc
  File "pandas/_libs/index.pyx", line 550, in pandas._libs.index.DatetimeEngine.get_loc
KeyError: Timestamp('2023-05-10 07:55:09.504000')

The above exception was the direct cause of the following exception:

  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/pandas/core/indexes/datetimes.py", line 736, in get_loc
    return Index.get_loc(self, key, method, tolerance)
  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 3804, in get_loc
    raise KeyError(key) from err
KeyError: Timestamp('2023-05-10 07:55:09.504000')

The above exception was the direct cause of the following exception:

  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/xarray/core/indexes.py", line 486, in sel
    indexer = self.index.get_loc(label_value)
  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/pandas/core/indexes/datetimes.py", line 738, in get_loc
    raise KeyError(orig_key) from err
KeyError: numpy.datetime64('2023-05-10T07:55:09.504000000')

The above exception was the direct cause of the following exception:

  File "/home/lahtinep/bin/test_temporal_rgb_timeseries.py", line 17, in <module>
    blended = mscn.blend(blend_function=timeseries)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/multiscene/_multiscene.py", line 365, in blend
    new_scn.generate_possible_composites(True)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/scene.py", line 1430, in generate_possible_composites
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/scene.py", line 1449, in _generate_composites_from_loaded_datasets
    return self._generate_composites_nodes_from_loaded_datasets(needed_comp_nodes)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/scene.py", line 1455, in _generate_composites_nodes_from_loaded_datasets
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/scene.py", line 1513, in _generate_composite
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/composites/__init__.py", line 1775, in __call__
    new_projectables = self._apply_temporal_prerequisites(projectables)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/composites/__init__.py", line 1763, in _apply_temporal_prerequisites
    new_da = proj.sel(time=ref_time+pd.Timedelta(dq["time"]))
  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/xarray/core/dataarray.py", line 1531, in sel
    ds = self._to_temp_dataset().sel(
  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/xarray/core/dataset.py", line 2580, in sel
    query_results = map_index_queries(
  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/xarray/core/indexing.py", line 189, in map_index_queries
    results.append(index.sel(labels, **options))
  File "/home/lahtinep/mambaforge/envs/py310/lib/python3.10/site-packages/xarray/core/indexes.py", line 488, in sel
    raise KeyError(
KeyError: "not all values found in index 'time'. Try setting the `method` keyword argument (example: method='nearest')."
```

Could this be something data related?

gerritholl:
Good question.  Maybe something about the different start times (nominal, actual, etc.).  The time matching as I've currently implemented it needs to be exact, which would only work with nominal start times.  I'll have a look.

gerritholl:
We probably need a way to add tolerance or slicing capability, "search between T-65 and T-55" or so.

satpy/composites/__init__.py
```diff
@@ -1701,3 +1702,85 @@ def __call__(self, projectables, nonprojectables=None, **info):
...
+        steps, we will be passed the same 3-D data array three times.
+
+        This method matches the temporal prerequisites as defined in the
+        compositor with the time dimension associated with the dataarrays.
+        """
+        # reference time is the newest time
+        ref_time = projectables[0]["time"][-1]
+        new_projectables = []
+        for (dq, proj) in zip(self.attrs["prerequisites"], projectables):
+            new_da = proj.sel(time=ref_time+pd.Timedelta(dq["time"]))
```
pnuu, line 1763:
```suggestion
            new_da = proj.sel(time=ref_time+pd.Timedelta(dq["time"]), method='nearest')
```
This resolves the issue I saw with SEVIRI data.

gerritholl: That was much easier than I thought it might be :)

satpy/composites/__init__.py
```diff
@@ -1701,3 +1702,85 @@ def __call__(self, projectables, nonprojectables=None, **info):
...
+                    "and then performing timeseries blending.")
+
+    def _ensure_prerequisites_have_times(self, prerequisites):
+        """Make sure all prerequisites have times defined."""
+        for preq in prerequisites:
+            if "time" not in preq.to_dict():
+                raise KeyError(
+                    "In a temporal composite, all prerequisites should have "
+                    "time information defined.  Time information is missing for "
+                    f"{preq!s}")
```
djhoese, line 1746: Should this also make sure that `pd.Timedelta` understands the time value? Or should that wait until `__call__` time as you have it now?

satpy/etc/composites/visir.yaml
```diff
@@ -642,3 +642,20 @@ composites:
...
+      the MultiScene, then apply timeseries blending. The resulting Scene
+      should have the composites.
+    compositor: !!python/name:satpy.composites.TemporalRGB
+    prerequisites:
+      - wavelength: 0.64
+        time: 0
+      - wavelength: 0.64
+        time: -10 min
+      - wavelength: 0.64
+        time: -20 min
```
djhoese, line 660: So this means green and blue are *before* the red channel in time? Just curious, is this a standard way of looking at this or would +10, +20, should something equivalent?

satpy/scene.py
```diff
@@ -1521,8 +1521,9 @@ def _generate_composite(self, comp_node: CompositorNode, keepables: set):
                 self._wishlist.remove(comp_node.name)
                 self._wishlist.add(cid)
             self._dependency_tree.update_node_name(comp_node, cid)
-        except IncompatibleAreas:
-            LOG.debug("Delaying generation of %s because of incompatible areas", str(compositor.id))
+        except (IncompatibleAreas, MissingTime):
+            LOG.debug("Delaying generation of %s because of incompatible areas "
+                      "or missing time dimension", str(compositor.id))
```
djhoese, line 1526: Can we split this debug message or make it specific to the case (temporal versus area)?

djhoese (changes requested):
Lots of questions, but this is looking pretty good. One major high-level concern though.

PR #2503, opened 2023-06-07
Install meson-python in unstable CI environment

djhoese:
Numpy, scipy, and pandas are starting to use "meson" to build their packages. This is an additional dependency that isn't already installed in our CI environments when we install the packages from conda. This PR adds a manual installation of this package so that it can be used by the unstable installation which uses `--no-deps`. An alternative might be to change the installation to additionally look at the anaconda PyPI repository instead of *replacing* PyPI as the only package repository searched (2 repositories versus 1). However, I think this would make it easier for us to get packages mixed up between stable and unstable environments.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -86,9 +86,12 @@ jobs:
       - name: Install unstable dependencies
         if: matrix.experimental == true
         shell: bash -l {0}
+        # Install meson-python manually since it is a new build dependencies
+        # not already included by the existing installed conda packages.
         # We must get LD_PRELOAD for stdlibc++ or else the manylinux wheels
         # may break the conda-forge libraries trying to use newer glibc versions
         run: |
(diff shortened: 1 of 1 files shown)


djhoese:
I thought this might happen. The meson-python manual installation didn't work because build environments are isolated from the main package environment. I'll try the `--extral-index-url` idea before trying to deal with `--no-build-isolation`.

djhoese:
Ok, maybe we just wait for the new versions to be released or something. I'm not exactly sure how to get this to pull in the dependency while also not installing things from PyPI. This last change ignored most dependencies and said they were already installed. I've seen some other projects (pandas) include some pip flags in their `pip:` section of their conda environment YAML. I'm not sure I want to completely rewrite the dependency installation for now.

PR #2507, opened 2023-06-12
Fix HDF4 support in geocat reader with hardcoded engine

### LINKED ISSUE #2506: Add xarray_kwargs capability to the geocat reader
## Feature Request

**Problem:**
The geocat reader is cannot load the hdf geocat files for 2 reasons:

1. The netcdf4 engine currently needs to be specified for xarray so that geocat hdf files currently being generated can read with xarray.
2. The time units (for example, the metadata for the variable scan_line_time) are not cf compliant.  They are currently 
units="seconds_since_jan_1_"

**Solution** (to assist in reading data from current geocat files):
Add an `__init__` to the GEOCATFileHandler so that xarray_kwargs can be passed through the reader_kwargs when the scene is loaded:

```python
scene = satpy.Scene(filenames, reader='geocat', reader_kwargs={'decode_times': True})
```

Set the defaults of these arguments to {"engine": "netcdf4", "decode_times": False} which will still allow users to set decode_times and engines as needed if the units are changed in the geocat software.  

**Workflow Changes**
As I understand this, it will not affect backward compatibility.  It might force the addition of future reader_kwargs if the geocat software package is updated.

**Additional context**
The question of updating the metadata for the units has been presented to the geocat team.  However, this software is already in use by many teams and it appears that this bug in the metadata has not been reported.  In the end, it would be nice if this reader could work regardless of the final decision by the team concerning the units metadata.  Old files will have this unit specification and the software has been in use for a while.

joleenf:
Adds an `__init__` method to the GEOCATFileHandler to allow the addition of xarray_kwargs to the reader.  This has been done so that the engine to the xarray_opendataset can be set as well as setting decode_times=False since the current metadata does not allow xarray to decode the times.  (Even when the engine is set, decode_times=True will throw an error because xarray cannot interpret the units metadata as written by GEOCAT).  

 - [ ] Closes #2506
 - [ ] Tests added:  A basic test that adds an xarray_kwarg and checks to see that datasets have been loaded with the reader.

Relates to the discussion on https://github.com/pydata/xarray/pull/7905/
I wanted to isolate this pull request from later work which will rely on being able to read these GEOCAT files with SatPy (i.e. adding aliases to facilitate RGB composites with existing recipes).


satpy/readers/geocat.py
```diff
@@ -56,7 +56,30 @@
...
+
+    **Loading data with decode_times=True**
+
+    By default, this reader will use xarray_kwargs={"engine": "netcdf4", "decode_times": False}.
+    to match behavior of xarray when the geocat reader was first written.  To use different options
+    use reader_kwargs when loading the Scene:
+
+        scene = satpy.Scene(filenames,
+                            reader='geocat',
+                            reader_kwargs={'decode_times': True})
```
djhoese, line 69: Correct me if I'm wrong but this isn't how this would work. The user would be doing `reader_kwargs={"xarray_kwargs": {"decode_times": True}}`.

satpy/readers/geocat.py
```diff
@@ -56,7 +56,30 @@
...
+    """
+
+    def __init__(self, filename, filename_info, filetype_info,
+                 **kwargs):
+        """Open and perform initial investigation of NetCDF file."""
+        kwargs.setdefault('xarray_kwargs', {}).setdefault(
+            'decode_times', False)
+        kwargs.setdefault('xarray_kwargs', {}).setdefault(
+            'engine', "netcdf4")
```
djhoese, line 78: Thinking about it, these can just be hardcoded can't they? So `kwargs.setdefault("xarray_kwargs", {"decode_times": False, "engine": "netcdf4"})`?

joleenf: At this point, I can't think of a use case where they could be anything else.

satpy/readers/geocat.py
```diff
@@ -58,6 +58,12 @@
 class GEOCATFileHandler(NetCDF4FileHandler):
     """GEOCAT netCDF4 file handler."""
 
+    def __init__(self, filename, filename_info, filetype_info):
+        """Open and perform initial investigation of NetCDF file."""
+        super(GEOCATFileHandler, self).__init__(
+            filename, filename_info, filetype_info,
+            xarray_kwargs={"engine": "netcdf4", "decode_times": False})
```
djhoese, line 65: I guess the only change I'd like to consider is allowing the user to pass `xarray_kwargs` in a `**kwargs`. I think we had done something like `kwargs.setdefault("xarray_kwargs, {...})`.

satpy/readers/geocat.py
joleenf: You are correct fixed in  510c24

satpy/readers/geocat.py
```diff
@@ -56,7 +56,30 @@
...
+                 **kwargs):
+        """Open and perform initial investigation of NetCDF file."""
+        kwargs.setdefault('xarray_kwargs', {}).setdefault(
+            'decode_times', False)
+        kwargs.setdefault('xarray_kwargs', {}).setdefault(
+            'engine', "netcdf4")
+
+        super(GEOCATFileHandler, self).__init__(
+            filename, filename_info, filetype_info,
+            xarray_kwargs=kwargs["xarray_kwargs"])
```
djhoese, line 82: I think this should just be `**kwargs` in case the user is passing other kwargs down to the utility file handler.

satpy/readers/geocat.py
```diff
@@ -56,7 +56,30 @@
...
+
+    **Loading data with decode_times=True**
+
+    By default, this reader will use xarray_kwargs={"engine": "netcdf4", "decode_times": False}.
+    to match behavior of xarray when the geocat reader was first written.  To use different options
+    use reader_kwargs when loading the Scene:
+
+        scene = satpy.Scene(filenames,
+                            reader='geocat',
+                            reader_kwargs={'xarray_kwargs': {'decode_times': True}})
```
djhoese, line 69:
```suggestion
    use reader_kwargs when loading the Scene::

        scene = satpy.Scene(filenames,
                            reader='geocat',
                            reader_kwargs={'xarray_kwargs': {'decode_times': True}})
```

satpy/readers/geocat.py
```diff
@@ -56,7 +56,30 @@
 
 
 class GEOCATFileHandler(NetCDF4FileHandler):
-    """GEOCAT netCDF4 file handler."""
+    """GEOCAT netCDF4 file handler.
+
+    **Loading data with decode_times=True**
+
+    By default, this reader will use xarray_kwargs={"engine": "netcdf4", "decode_times": False}.
```
djhoese, line 63:
```suggestion
    By default, this reader will use ``xarray_kwargs={"engine": "netcdf4", "decode_times": False}``.
```

satpy/readers/geocat.py
joleenf: If the geocat code changes, and the unit metadata becomes compliant, I think it would be nice to be able to interpret the scan line times when reading them.  I might be mistaken, but 75-78 as written would allow the user to set the decode_times back to True if it is possible to decode them?

djhoese: Yeah, it looks like it would work. My suggestion for doing the entire `xarray_kwargs` dictionary would also work, but would require users to make sure they include engine and decode_times if they were specifying xarray_kwargs at all. This is such a small thing and not sure it matters too much.

satpy/tests/reader_tests/test_geocat.py
```diff
@@ -135,6 +135,18 @@ def test_init(self):
...
+        """Test basic init with extra parameters."""
+        from satpy.readers import load_reader
+        r = load_reader(self.reader_configs, xarray_kwargs={"decode_times": True})
+        loadables = r.select_files_from_pathnames([
+            'geocatL2.GOES-13.2015143.234500.nc',
+        ])
+        self.assertEqual(len(loadables), 1)
+        r.create_filehandlers(loadables, fh_kwargs={"xarray_kwargs": {'decode_times': True}})
+        # make sure we have some files
+        self.assertTrue(r.file_handlers)
```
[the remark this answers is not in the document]
joleenf: Would you want that changed at lines 133 and 136 as well?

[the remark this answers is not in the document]
djhoese: Only if you're bored. I guess maybe don't worry about this last request...unless you're bored. We can merge this test as is.

satpy/readers/geocat.py
joleenf: Circling back to this again, sorry.... if the [kwargs are set as in a previous commit](https://github.com/joleenf/satpy/blob/b8a42dee445be65bc0db9e8a99feaf9cdd8e61c6/satpy/readers/geocat.py#L74-L78)

Doesn't that leave the engine as "netcdf4" if the user only enters a value for decode_times?  If no xarray_kwargs are provided, the values would be ```{"engine": "netcdf4", "decode_times": False}```?   

I think I am not understanding your point fully on this requested change to 
```kwargs.setdefault("xarray_kwargs", {"decode_times": False, "engine": "netcdf4"})```.

djhoese: You are correct. Yours is more flexible. That's why I said it doesn't matter much. My main reason for my suggestion was that it treated xarray_kwargs as an all-or-nothing keyword argument rather than individual things. Sorry for all the confusion for something so dumb (on my part). Let me know if you want to put it back the way you had it or leave it as is. Either way I'm fine with it and we can merge regardless of the pre-commit issue...which I think is a hiccup on pre-commit's side of things.

joleenf: It is fine.  I think either way works, so to avoid errors that can be introduced with too much more fiddling, I think I will leave it as is.

djhoese:
pre-commit.ci autofix

satpy/readers/geocat.py
joleenf: I think I understand where you want kwargs, in the  `__init__` correct?  The user would still send the xarray specific kwargs with the key "xarray_kwargs" so the documentation should reflect that.

djhoese: Correct. This call to `__init__` should use kwargs to pass everything to the base class. That doesn't change what the user uses.

djhoese:
Closing and reopening to hopefully pick up any commits from pre-commit.

djhoese:
pre-commit.ci autofix

djhoese:
@joleenf I checked the pre-commit.ci log and it looks like it is having some trouble merging/pushing fixes to your branch. It says to pull/merge with upstream main to fix it. You could also do the fixes yourself (after merging with main) by running `pre-commit run -a` and that should make the changes pre-commit.ci would have made anyway.

djhoese (approved):

PR #2513, opened 2023-06-16
Add a new fire RGB using both L1 and L2 data.

simon-sat:
This PR creates a new composite that blends L1 data with an L2 product to highlight regions with fires.
Thus far, it combines GOES/ABI `true_color_reproduction` L1 data with the L2 `FDCC` product that contains fire radiative power.

Example output:
Untitled

right now, it doesn't look great. We need some better method to blend the edges of the fire with the background data. I've tried to do this in the maskingcompositor by defining various thresholds, but they're obviously not right yet.

 - [ ] Tests added
 - [ ] Fully documented


simonrp84:
pre-commit.ci autofix

PR #2514, opened 2023-06-20
Fix argument name in DayNightComposite example document

marty-sullivan:
This fixes an argument name in the example documentation for `DayNightComposite`. This wrong value in the example caused me a headache in figuring out since it did not throw an error but also did not do the expected behavior. I found in the API doc for `DayNightComposite` that `need_alpha=False` was supposed to be `include_alpha=False`.


djhoese (approved):
Nice catch. Yeah, sorry about that. I missed this changed when I asked the contributor to change the kwarg name.

PR #2516, opened 2023-06-21
Fix SEVIRI native reader failing when missing main header

ameraner:
This PR fixes a bug introduced by https://github.com/pytroll/satpy/pull/2198 (cc @simonrp84 ) that causes the SEVIRI reader to fail for files that do not have a `15_MAIN_PRODUCT_HEADER` -this header is read, optionally, here:
https://github.com/pytroll/satpy/blob/f44473f905fa07615f4f17e5b41657645a3ce7d1/satpy/readers/seviri_l1b_native_hdr.py#L84-L86
Files that do not contain the archive header were missing the `15_MAIN_PRODUCT_HEADER`, which was causing a `KeyError`.

After this PR, an info log message is issued specifying that the quality check could not be performed in case of the missing header.
 - [x] Tests added

### DIFF
--- a/satpy/readers/seviri_l1b_native.py
+++ b/satpy/readers/seviri_l1b_native.py
@@ -370,13 +370,16 @@ class NativeMSGFileHandler(BaseFileHandler):
         self.mda['hrv_number_of_lines'] = int(sec15hd["NumberLinesHRV"]['Value'])
         self.mda['hrv_number_of_columns'] = cols_hrv
 
-        if self.header['15_MAIN_PRODUCT_HEADER']['QQOV']['Value'] == 'NOK':
-            warnings.warn(
-                "The quality flag for this file indicates not OK. "
-                "Use this data with caution!",
-                UserWarning,
-                stacklevel=2
-            )
+        if '15_MAIN_PRODUCT_HEADER' not in self.header:
+            logger.info("Quality flag check was not possible due to missing 15_MAIN_PRODUCT_HEADER.")
+        else:
+            if self.header['15_MAIN_PRODUCT_HEADER']['QQOV']['Value'] == 'NOK':
+                warnings.warn(
+                    "The quality flag for this file indicates not OK. "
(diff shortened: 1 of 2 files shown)


satpy/readers/seviri_l1b_native.py
```diff
@@ -370,13 +370,16 @@ def _read_header(self):
...
 
-        if self.header['15_MAIN_PRODUCT_HEADER']['QQOV']['Value'] == 'NOK':
-            warnings.warn(
-                "The quality flag for this file indicates not OK. "
-                "Use this data with caution!",
-                UserWarning,
-                stacklevel=2
-            )
+        if '15_MAIN_PRODUCT_HEADER' not in self.header:
+            logger.info("Quality flag check was not possible due to missing 15_MAIN_PRODUCT_HEADER.")
```
djhoese, line 374: info or debug? Put another way, is this something you want to see most SIFT users seeing in an INFO-level log?

satpy/readers/seviri_l1b_native.py
```diff
@@ -370,13 +370,16 @@ def _read_header(self):
...
+        else:
+            if self.header['15_MAIN_PRODUCT_HEADER']['QQOV']['Value'] == 'NOK':
```
djhoese, line 376: elif?

ameraner: sure, changed to elif (in "add test" commit, message wrong by mistake, sorry)

satpy/readers/seviri_l1b_native.py
ameraner: seems to me that info is more suited here, since it's not a debug message (useful for troubleshooting or so), it's just informing about a state of the code, and no futher action is needed in most cases, nor done in the code. But if there are stronger guidelines on this, I can change it of course.

mraspaud (approved):
LGTM

PR #2518, opened 2023-06-22
Generic reader for EUMETSAT Data Tailor simplified NetCDF output

gerritholl:
This PR will add a generic reader for EUMETSAT Data Tailor simplified NetCDF output files.

This is an early work in progress with incomplete and incorrect tests and no implementation or documentation.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/etc/readers/datatailor_nc.yaml
+++ b/satpy/etc/readers/datatailor_nc.yaml
@@ -1,0 +1,21 @@
+reader:
+  name: datatailor_nc
+  short_name: EUMETSAT Data Tailor NetCDF
+  long_name:  Generic reader for simplified NetCDF files coming from the EUMETSAT Data Tailor
+  description: >
+    The EUMETSAT Data Tailor is a tool published by EUMETSAT that allows tailoring
+    satellite data to and from various file formats, among other things.
+    This is a generic reader for the simplified NetCDF output format.
+    For more information on the data tailor, see https://www.eumetsat.int/data-tailor
+  status: Alpha
+  supports_fsspec: True
+  reader: !!python/name:satpy.readers.yaml_reader.FileYAMLReader
+  sensors: [many]
+  default_datasets:
+
+file_types:
+  datatailor_simple_nc:
+    file_reader: !!python/name:satpy.readers.datatailor_nc.DataTaliorNC
+    file_patterns:
+      - '{product_key}{start_sensing_time}{stop_sensing_time}epct{process_uuid}_{steps_acronym}.nc'
+      - "{stem}.nc"
--- a/satpy/tests/reader_tests/test_datatailor_nc.py
+++ b/satpy/tests/reader_tests/test_datatailor_nc.py
@@ -1,0 +1,140 @@
+# Copyright (c) 2023 Satpy developers
+#
+# This file is part of satpy.
+#
+# satpy is free software: you can redistribute it and/or modify it under the
+# terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# satpy is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
+"""Test module for the datatailor netcdf reader."""
+
+import numpy as np
+import pytest
+import xarray
+
+
+@pytest.fixture
+def fake_datatailor_dataset():
+    """Generate the contents for a fake data tailor NetCDF file."""
+    nx = 10
+    ny = 10
+    nz = 5
+    lats = np.linspace(-89, 89, nx*ny).reshape(nx, ny)
+    lons = np.linspace(-189, 189, nx*ny).reshape(nx, ny)
+    iwv = np.linspace(0, 1, nx*ny).reshape(nx, ny)
+    temp = np.linspace(200, 330, nx*ny*nz).reshape(nx, ny, nz)
+
+    ds = xarray.Dataset(
+        {"lat":
+            ((yx := ("along_track", "across_track")),
+             lats,
+             {"long_name": "grid_latitude",
+              "standard_name": "latitude",
+              "units": "degrees_north"}),
+         "lon":
+             (yx, lons,
+              {"long_name": "grid_longitude",
+               "standard_name": "longitude",
+               "units": "degrees_east"}),
+         "integrated_water_vapour":
+            (yx, iwv,
+             {'long_name': 'integrated water vapour',
+              'standard_name': 'atmosphere_mass_content_of_water_vapour',
+              'units': 'kg.m^-2',
+              'comment': 'Integrated water vapour (for 120 IFOV)'}),
+         "atmospheric_temperature":
+            (yx + ("nlt",), temp,
+             {'standard_name': 'air_temperature',
+              'comment': 'Temperature (for 120 IFOV with up to 101 vertical levels)',
+              'units': 'degree(K)'})},
+        {"nlt": np.arange(nz)},
+        {'Conventions': 'CF-1.7',
+         'Metadata_Conventions': 'Unidata Dataset Discovery v1.0',
+         'title': 'IASI 2b the Infrared Atmospheric Sounding Interferometer',
+         'title_short_name': 'IASI L2',
+         'summary': 'NA',
+         'source': 'MetOp-B IASI',
+         'references': 'https://www.eumetsat.int/website/home/Data/DataDelivery/EUMETSATDataCentre/index.html',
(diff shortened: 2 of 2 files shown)


gerritholl:
The reader seems to work, but maybe some metadata is not being set correctly, because resampling fails with `IndexError`.

```python
from satpy import Scene
fn = "/media/nas/x21308/IASI/IASSND02-Metop-B-IASI-20230602T193252-20230602T211452-tailored.nc"

sc = Scene(filenames=[fn], reader=["datatailor_nc"])
print(sc.available_dataset_names())
datasets = ["atmospheric_temperature", "integrated_water_vapour"]
sc.load(datasets)
ls = sc.resample("eurol")
print(ls["atmospheric_temperature"].compute())
```

fails with

```python-traceback
  File "/data/gholl/checkouts/protocode/satpy-read-iasi-l2-tailor.py", line 13, in <module>
    print(ls["atmospheric_temperature"].compute())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/xarray/core/dataarray.py", line 1101, in compute
    return new.load(**kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/xarray/core/dataarray.py", line 1075, in load
    ds = self._to_temp_dataset().load(**kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/xarray/core/dataset.py", line 781, in load
    evaluated_data = chunkmanager.compute(*lazy_data.values(), **kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/xarray/core/daskmanager.py", line 70, in compute
    return compute(*data, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/base.py", line 595, in compute
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/threaded.py", line 89, in get
    results = get_async(
              ^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/local.py", line 511, in get_async
    raise_exception(exc, tb)
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/local.py", line 319, in reraise
    raise exc
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/local.py", line 224, in execute_task
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/core.py", line 121, in _execute_task
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/optimization.py", line 992, in __call__
    return core.get(self.dsk, self.outkey, dict(zip(self.inkeys, args)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/core.py", line 151, in get
    result = _execute_task(task, cache)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/core.py", line 121, in _execute_task
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py311/lib/python3.11/site-packages/dask/utils.py", line 73, in apply
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/checkouts/pyresample/pyresample/future/resamplers/nearest.py", line 100, in _my_index
    res = data_arr[vii_slices][ia_slices]
          ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
IndexError: index -1 is out of bounds for axis 0 with size 0
```

gerritholl:
Applying the scale factors solves the resampling problem.

satpy/etc/readers/datatailor_nc.yaml
```diff
@@ -0,0 +1,21 @@
...
+  long_name:  Generic reader for simplified NetCDF files coming from the EUMETSAT Data Tailor
+  description: >
+    The EUMETSAT Data Tailor is a tool published by EUMETSAT that allows tailoring
+    satellite data to and from various file formats, among other things.
+    This is a generic reader for the simplified NetCDF output format.
+    For more information on the data tailor, see https://www.eumetsat.int/data-tailor
+  status: Alpha
+  supports_fsspec: True
+  reader: !!python/name:satpy.readers.yaml_reader.FileYAMLReader
+  sensors: [many]
```
mraspaud, line 13: I understand the idea here, but will this work in practice? Satpy has some parts dependent on the sensor name being correct (composites and enhancement). But maybe this is anyway superseded by the reader itself?

satpy/etc/readers/datatailor_nc.yaml
```diff
@@ -0,0 +1,21 @@
...
+  default_datasets:
+
+file_types:
+  datatailor_simple_nc:
+    file_reader: !!python/name:satpy.readers.datatailor_nc.DataTailorNC
+    file_patterns:
+      - '{product_key}{start_sensing_time}{stop_sensing_time}epct{process_uuid}_{steps_acronym}.nc'
+      - "{stem}.nc"
```
mraspaud, line 21: This could conflict with the satpy nc reader which also accepts everything...

mraspaud (commented):
Just a couple of comments inline, but looks good overall

satpy/etc/readers/datatailor_nc.yaml
gerritholl: I copied `many` from the `satpy_cf_nc` reader. I have no idea if it works.

satpy/etc/readers/datatailor_nc.yaml
gerritholl: It does. It means that if a user has a file `foo.nc`, that Satpy cannot automatically determine the reader. Maybe such a limitation is acceptable — that if the filename is not precise enough, that the user must provide the reader?

So far Satpy takes whatever is the first possible reader that fits. That's understandable from an optimisation point of view, but it's not very stable and violates the principle "In the face of ambiguity, refuse the temptation to guess" (Zen of Python).

mraspaud: I the filename output by the data tailor really free or random?

gerritholl: It depends. When using the Python API like I do, it outputs a stream. There is a server-side filename in the metadata, but of course users can name the file what they want client-side. The command-line client does create a file with the same name as on the server.

The server-side filename in the metadata follows the first of the file patterns that I indicated (source: answer from EUMETSAT Secretariat Helpdesk).  Example filename:

`IASISND02_20230602060852Z_20230602075052Z_epct_f2846918_F.nc`

I'm not too keen on forcing this, in particular due to the `process_uuid`, which is different even if the same file is downloaded with the same customisation multiple times. The other problem is that this filename pattern is not publicly documented and there is no public API to obtain the filename except by generating the product. It might be just  an implementation detail and subject to change.

(It also seems the filename pattern I was given by the Help Desk does not actually match the files I have obtained so far.)

mraspaud: Ok, so nothing much we can do here when using the api to force the filename. One cool solution though would be to pass the stream directly to satpy somehow (without going through a filename) but I have not thought of the implications).

Back to filenames, so what I'm thinking right now is actually to have satpy raise an error when there is a filename that could fit multiple readers to force the user into specifying the reader to use. Would that be possible?

satpy/etc/readers/datatailor_nc.yaml
gerritholl: If you mean, "will it find composites", the answer is no.

satpy/etc/readers/datatailor_nc.yaml
gerritholl: Maybe that would be possible.  Should that be done in this PR?

mraspaud: Well if the filename matches multiple readers, I think we should raise an error, yes.
> In the face of ambiguity, refuse the temptation to guess.

satpy/etc/readers/datatailor_nc.yaml
mraspaud: can we leave this parameter out of the yaml, and make it only available at run time?

PR #2529, opened 2023-07-17
Add reader for LSA-SAF data

simon-sat:
This PR adds a reader for data produced by EUMETSAT's LSA-SAF. Right now, it only supports land surface temperature files but I plan to add additional ones soon.
Tests need to be added as well, but I wanted to make a PR so gain some feedback on the way I've coded this before I do the testing for it.

 - [ ] Tests added
 - [ ] Fully documented


simonrp84:
pre-commit.ci autofix

simonrp84:
Kinda curious that there pre-commit.ci shows failures, I ran the pre-commit checks locally before pushing!

satpy/etc/readers/lsasaf-geo.yaml
```diff
@@ -0,0 +1,44 @@
+reader:
+  name: lsasaf-geo
```
djhoese, line 2: Is the name of the software "LSASAF GEO"? If not, then I'd say an underscore in the reader name would be more consistent. Or even "lsasaf" if there is no LEO version.

satpy/readers/lsasaf_h5.py
```diff
@@ -0,0 +1,231 @@
...
+            if "a" in proj_dict:
+                proj_dict["a"] *= 1000.
+            if "b" in proj_dict:
+                proj_dict["b"] *= 1000.
+            if "R" in proj_dict:
+                proj_dict["R"] *= 1000.
+            proj_dict["h"] *= 1000.
+            area_extent = tuple([val * 1000. for val in area_extent])
+            crs = CRS.from_dict(proj_dict)
+        return crs, area_extent
```
djhoese, line 194: Unused method? If not, I'm tempted to say we should just deal with the projection in kilometers if possible.

satpy/readers/lsasaf_h5.py
```diff
@@ -0,0 +1,231 @@
...
+            return INST_NAMES[inst_id]
+        except KeyError:
+            raise KeyError(f'Unknown instrument ID: {inst_id}')
+
+    def get_dataset(self, dsid, info):
+        """Load a dataset."""
+        dsid_name = dsid['name']
+        if dsid_name in self.cache:
+            logger.debug('Get the data set from cache: %s.', dsid_name)
+            return self.cache[dsid_name]
```
djhoese, line 81: 1. Is the cache ever actually written to?
2. If it was, would it really provide a performance improvement?
3. In what cases is the same file handler asked for the same DataArray multiple times?

satpy/etc/composites/visir.yaml
```diff
@@ -647,3 +647,9 @@ composites:
       - wavelength: 3.9
       - wavelength: 6.2
       - wavelength: 1.6
+
+  land_surface_temperature:
+    compositor: !!python/name:satpy.composites.SingleBandCompositor
+    prerequisites:
+      - name: 'lst'
+    standard_name: land_surface_temperature
```
djhoese, line 655: I'm tempted to say this shouldn't be in `visir` yet. I'm also tempted to say this compositor should not exist...could you add the standard_name in the reader YAML and/or reader python code? Then this composite doesn't conflict with other readers who already have a product called "land_surface_temperature", but your enhancement based on `standard_name` would still be found.

PR #1972, opened 2022-01-11
Add reader for VIIRS level 2 products produced by NOAA enterprise

simonrp84:
This PR adds (initial / basic) support for the VIIRS JRR level 2 products produced by the NOAA enterprise processing suite.
The following products exists and, where ticked, are supported:
 - [x] Aerosol products
 - [x] Cloud mask
 - [ ] Cloud base height
 - [ ] Cloud top height
 - [ ] Cloud optical and microphysical properties (day)
 - [ ] Cloud optical and microphysical properties (night)
 - [ ] Ice concentration
 - [ ] Land surface temperature
 - [ ] Snow cover
 - [ ] Ice age
 - [x] Surface reflectance
 - [ ] Surface albedo
 - [ ] Daily vegetation indices
 - [ ] Weekly vegetation indices

I do not intend to add support for all these products in this PR, but want to make a list of what's available for future reference.

This is currently a draft while I work on it.

 - [ ] Tests added
 - [ ] Fully documented

PR #1483, opened 2020-12-10
Add reader support for VIIRS Land Surface Reflectance

nbearson:
This PR adds a reader for the VIIRS Land Surface Reflectance product, with additional NDVI/EVI fields provided by the (upcoming) CSPP VIIRS LSR release.

 - [ ] Tests added 
 - [ ] Passes ``flake8 satpy`` 
 - [ ] Fully documented 
 - [ ] Add your name to `AUTHORS.md` if not there already

PR #2531, opened 2023-07-18
Add "viirs_edr" reader

djhoese:
This is a replacement of #1972, #1483, and may be a replacement for the `viirs_l2_cloud_mask_nc` reader added in #2316. Basically I need to be able to read the NOAA Enterprise EDR files and all of these PRs share a little bit of logic but were never completed. My priority right now is the surface reflectance products and the vegetation indices. There are cloud top temp and height products that I may add here or in a separate PR.

The existing readers or PRs for readers all use the "L2" name. These are "EDRs". Functionally the same, but we should be consistent with the majority of NOAA documentation (which admittedly also occasionally refers to them as L2). There are also existing readers that I had added in the past called "viirs_edr_X". Given the nature of these products, I think this new reader can be called a generic `viirs_edr` name (no suffix). I haven't looked into most of the files I hope to support, but my hope is that the enterprise algorithms are relatively similar in file structure. Even if they aren't, they are NetCDF files and represent the same level of product for the same instrument. Anyone object?

Side note: Kathy Strabala and I thought maybe the surface reflectance products could be called "srf_i01" and so on. Reflecting that they are different than the SDR bands (TOA versus surface), but also being consistent with the "i01" style naming rather than the "i1" naming that the existing PRs are using. I could also see doing "I01" (uppercase letter) to be consistent with the `viirs_sdr` reader, but then I'd have to uppercase SRF which I don't really want to do.

I'll have to decide later whether or not it is appropriate to deprecate the `viirs_l2_cloud_mask_nc` reader in favor of this one.

Opinions? Please comment.

CC @pnuu @simonrp84 @nbearson

Edit: I see @simonrp84 used `surf_refl_I01` for his surface reflectance products. Maybe I'll stick with that. Again, opinions welcome.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


simonrp84:
I prefer uppercase band names with leading zeros (`I01` etc) for consistency with the SDR datasets. Regarding the choice of name, I chose `surf_refl_` simply for lack of another obvious option.

My gut feeling is that `srf_` could be confused with other variables, or even spectral response function, so to me spelling it out more clearly is the better option.

Seems fine to call the whole lot `EDR` rather than `L2`.

djhoese:
@simonrp84 @pnuu Ignoring the surface reflectance files, what do you think about a dynamic `viirs_edr` reader that dynamically looks at the variables in the files and supports any 2D variable? Most of the files I'm looking at are pretty accurate and CF compliant so most of the stuff is as easy as just returning the DataArray xarray provides (sometimes converting "unitless" units to "1" for consistency and renaming dimensions). This would mean that we'd have to use the variable names provided in the file, but maybe that's for the best? For example, `Ash` instead of `ash`.

djhoese:
Ok I think I'm ready for people to start reviewing this. A lot more could be added, but I think the main functionality is complete. The main remaining question for me is the question in my previous comment: do I change all these hardcoded-in-YAML variables to be dynamically discovered so the reader can basically load anything?

pnuu (approved):
LGTM as-is.

> do I change all these hardcoded-in-YAML variables to be dynamically discovered so the reader can basically load anything?

The reader YAML is pretty long, so I'm not against it. The only downside I can think if is that to see what is supported (and by which name) one would need to first have the data and use `available_dataset_names()`.

djhoese:
Yeah that is true, but I think that comes with the territory of a generic reader that reads generic files (NOAA's enterprise algorithm output). @simonrp84 thoughts?

satpy/etc/composites/viirs.yaml
```diff
@@ -231,10 +231,10 @@ composites:
     - name: M05
       modifiers: [sunz_corrected, rayleigh_corrected]
     optional_prerequisites:
-    - name: I01
+    - name: I02
       modifiers: [sunz_corrected_iband, rayleigh_corrected_iband]
     standard_name: false_color
-    high_resolution_band: blue
+    high_resolution_band: green
```
simonrp84, line 237: Out of curiosity, what was the reason for changing bands here? I presume to match some external visualisation method?

satpy/readers/viirs_edr.py
```diff
@@ -0,0 +1,309 @@
...
+
+import numpy as np
+import xarray as xr
+
+from satpy import DataID
+from satpy.readers.file_handlers import BaseFileHandler
+from satpy.utils import get_chunk_size_limit
+
+LOG = logging.getLogger(__name__)
+M_COLS = 3200
```
simonrp84, line 61: Is this guaranteed to be constant? The documentation says a granule "typically" has this many columns.

satpy/readers/viirs_edr.py
```diff
@@ -0,0 +1,309 @@
...
+        self.sensor_name = 'viirs'
+
+    def rows_per_scans(self, data_arr: xr.DataArray) -> int:
+        """Get number of array rows per instrument scan based on data resolution."""
+        return 16 if data_arr.shape[1] == M_COLS else 32
+
+    def get_dataset(self, dataset_id: DataID, info: dict) -> xr.DataArray:
+        """Get the dataset."""
+        data_arr = self.nc[info['file_key']]
+        data_arr = self._mask_invalid(data_arr, info)
```
simonrp84, line 108: Should this be made optional? Some users may wish to access the unmasked data.

simonrp84 (commented):
A couple of minor comments. Personally, I'd like the masking of invalid data to be optional rather than hardcoded.
My comment about the number of columns is due to bitter experience of being hit by granules varying from the "typical" number of columns in the past!

satpy/readers/viirs_edr.py
djhoese: I've never seen anything different, but I also wasn't super comfortable with this as my main check for resolution either. I couldn't think of any other way to do it considering some files only have "Columns" and "Rows" dimensions *and* we currently rename the dimensions right away anyway.

satpy/readers/viirs_edr.py
djhoese: For the NDVI/EVI special case with QF checks, I think it could be optional since that's very complex and opinionated. Otherwise, I'm not sure I agree. This is checking fill values and valid ranges. Why would a user need to have that? The L2 products don't encode anything with different fill values as far as I know...I think, right?

satpy/etc/composites/viirs.yaml
djhoese: It matches what we do in Polar2Grid. The findings are basically that the image looks better when you do it on the green band. I don't remember the scientific reasoning, but can I use the excuse of "Liam Gumley told me to do it"?

satpy/readers/viirs_edr.py
simonrp84: Ok, fair enough!

satpy/etc/composites/viirs.yaml
simonrp84: Yeah, let's go with that ;-)

satpy/readers/viirs_edr.py
simonrp84: Sorry, I meant to comment on the *other* `_mask_invalid`, the one for the NDVI/EVI. If we could make that one optional then that'd be great. This one, that I actually comment on, is fine.

djhoese: This is done now. I had a very non-creative moment and went with `filter_veg` as the kwarg. Other ideas?

djhoese:
@pnuu I added your `viirs_l2_cloud_mask_nc` reader to the `PENDING_OLD_READER_NAMES` dictionary. This basically tells users "here's the name you used, but you should use this name. For now Satpy will automatically switch to the new reader". But in the case of our two readers aren't just a rename, the variable names are different. This is causing the tests for your reader to fail because they try to load variables that don't exist in the new reader. I feel like my only two options are:

1. Remove the code and tests for your reader. This is something I forgot about this deprecation functionality I added many years ago. The idea is that the old reader doesn't exist anymore.
2. Don't deprecate your reader.

What do you think?

pnuu:
Go with option 1.

I tried the reader, and currently I get this error:
```python
glbl.load(['CloudMaskBinary'])
*** KeyError: "No dataset matching 'DataQuery(name='longitude_jrr_cloudmask', resolution=750)' found"
```

Looking at the files I have, the coordinate variables are `Latitude` and `Longitude`.

djhoese:
I did detect one bug late last night that I'll work on this morning. I'll try to find the same files you're using and figure out the issues.

djhoese:
Oh FYI @pnuu that error message still makes sense. The variables for CloudMaskBinary actually makes sense because I name the  individual lon/lat datasets based on the file type they are coming from. I decided to change how Simon had done it in the original version of this reader where it assumed you were only ever loading one file type at a time and that all lon/lats were equivalent across files. So it was doing "oh we found longitude in this file type, use that for all longitude requests", but now it always loads lon/lat from the file for a particular dataset.

I'm still debugging things...

pnuu:
Thanks, seems to be working as before (after reader and channel name adjustments, ofc), good job!

djhoese:
Great. I think after I fix a few more bugs I've encountered this should be ready to merge.

PR #2533, opened 2023-07-19
Fix geolocation errors of the FY-4 AGRI reader base

### LINKED ISSUE #2532: FY-4 AGRI reader can't get the right geolocations
**Describe the bug**
There're two issues here.

1. Center coordinate is not at (0,0) 
image

2. Area extent is completely wrong
It is especially serious on REGC scan mode. REGC aims at China and surrounding areas. But the output looks like this (in ENVI):

image
image
image
You can clearly see that the REGC image is located at the southern hemisphere which is totally confusing.

I believe that this problem is due to the the calculations of the line/column offset in `fy4_base`. I'll start a PR to fix it.

yukaribbba:
Fix the offset calculations in FY-4 reader base according to HSD reader.

 - [x] Closes #2532 
 - [x] Tests added 

**1.Area extent of the GEOS projection**
This is calculated by column/line offset (COFF/LOFF), CFAC, width/height of the image using `get_area_extent` under `_geos_area`. But, this method begins at (0, 0) which is fit for full-disk image. When the scan mode is regional, the initial column/line position is also important. And that is why FY-4 REGC mode got the wrong area.

**2.Comparison between Himawari HSD with FY-4**
HSD reader, no matter whether the data is fulldisk / regional, also uses `get_area_extent` . The differences are:

**Himawari full-disk COFF/LOFF**
`[22000.5, 11000.5, 5500.5, 2750.5, 1375.5]`
**Himawari full-disk width/height**
`[44000, 22000, 11000, 5500, 2750]`

**FY-4 full-disk COFF/LOFF**
`[21983.5, 10991.5, 5495.5, 2747.5, 1373.5]`
**FY-4 full-disk width/height**
`[43968, 21984, 10992, 5496, 2748]`

Apparently,
`Himawari's offset = width(height) / 2 + 0.5`
While
`FY-4's offset = width(height) / 2 - 0.5`
That 1 pixel diff could explain why FY-4's center coord isn't at (0, 0).

And the calculation:
Himawari:
```
pdict['coff'] = np.float32(self.proj_info['COFF'])
pdict['loff'] = -np.float32(self.proj_info['LOFF']) + 1
pdict['loff'] = pdict['loff'] + (self.segment_number * pdict['nlines'])

pdict['ncols'] = width
pdict['nlines'] = height
```
When testing on a target-area scanning dataset `HS_H09_20230701_0000_B01_R301_R10_S0101.DAT`:
```
pdict['coff'] = 360.5
pdict['loff'] = -2279.5
```
The COFF/LOFF of HSD is stored inside the dataset, not fixed like FY-4. So it seems that these two numbers come from nowhere. But, if this regional image reprojected to full-disk frame, it is clear that the scanning start at **col 5140 / line 2220**.
image
This is a 1000x1000 B01 dataset. The fulldisk size of B01 is 11000 and the COFF/LOFF is 5500.5.

Therefore:
```
pdict['coff'] = fulldisk_COFF - start_column
pdict['loff'] = -fulldisk_LOFF + 1 + start_line + pdict['nlines']
```

**3.Back on FY-4**
Because of the 1-pixel difference:
```
pdict['coff'] = fulldisk_COFF - start_column + 1
pdict['loff'] = -fulldisk_LOFF + 1 + start_line + pdict['nlines'] - 1 = -fulldisk_LOFF + start_line + pdict['nlines']
```
Further:
`end_line + 1 = start_line + pdict['nlines']`
```
pdict['coff'] = fulldisk_COFF - start_column + 1
pdict['loff'] = -fulldisk_LOFF + end_line + 1
```
Luckily, both `start_column` and `end_line` are attributes in the FY-4 HDF files. So it is very easy to get them.

4.Results
The center point back to zero.
image

And the REGC image
image


djhoese:
Pinging @simonrp84 who I think added this reader or helped me add it.

simonrp84:
Yes, this is my reader. Will have a look in next few days.

yukaribbba:
@simonrp84 What can I do with this PR?

simonrp84:
Sorry, haven't had time to look at this. Will test this evening if possible.

For reference, the `coff` and `loff` numbers used in the reader were provided by the China Meteorological Administration. It wouldn't surprise me if the values are wrong as the CMA has been very inconsistent with their data over the years.

simonrp84:
Just tested this and it looks good for all modes of AGRI on FY-4A and -4B. Also works for GHI on FY-4B.

yukaribbba:
Great. @djhoese If you have any suggestions

satpy/readers/fy4_base.py
```diff
@@ -240,21 +244,12 @@ def get_area_def(self, key):
         pdict['a_name'] = f'{self.filename_info["observation_type"]}_{res}m'
         pdict['p_id'] = f'FY-4, {res}m'
 
-        pdict['nlines'] = pdict['nlines'] - 1
-        pdict['ncols'] = pdict['ncols'] - 1
-
-        pdict['coff'] = pdict['coff'] - 0.5
-        pdict['loff'] = pdict['loff'] + 1
-
```
djhoese, line 248: Am I understanding these changes correctly and that these lines are no longer needed because the changes to the coff/loff accomplish the same goal (but more accurately)?

yukaribbba: Yep. There's a more precise way so we don't need them any more.

djhoese:
Thanks!

PR #2537, opened 2023-07-31
Fix mismatch in shape for lat/lon when reading amv bufr

### LINKED ISSUE #2535: SEVIRI_L2_bufr reader is getting inconsistent output
**Describe the bug**
SEVIRI_L2_bufr reader is getting inconsistent output 
When reading AMV bufr the shape of the latittue and longitude as outputed by Scene is not the same as pressure, spead and direction.
Lat and lon are alsway larger by a few element that then data it is supposed to represent. 
 
The difference is rather samell but any attemps at 

**To Reproduce**

```python 
path = '/tcenas/home/manuelc/TEMP/FromGreg/MTG-I1_data/MSG_AMV_BUFR_products/'
   fname = glob.glob(path+'*')[1]
   scn = Scene(filenames=[fname],
               reader='seviri_l2_bufr',
               reader_kwargs={'with_area_definition': False}
               )
   list_pmtr = ['pressure', "speed", "direction",'latitude', 'longitude',]
   scn.load(list_pmtr)
        
   speed = scn['speed'].data
   dire = scn['direction'].data
   pres = scn['pressure'].data/100#conversion to hPa
   lat = scn['latitude'].data
   lon = scn['longitude'].data

```

**Expected behavior**
Latitude and Longitude shall have the same shape to allow for correct projection of the AMV into a map. 

**Actual results**
When investigating the shape we can see: 

```python  
lat
dask.array<where, shape=(52079,), dtype=float64, chunksize=(1541,), chunktype=numpy.ndarray>
```
While Pressure has a different value 
```python  
pres
dask.array<where, shape=(51944,), dtype=float64, chunksize=(1538,), chunktype=numpy.ndarray>
``` 
**Screenshots**
I decided to remove le last 135 values of the lat,lo not see if I could get something out of it non the less.

and I got this:
image

Which Does not make much sense(even when you are used to looking at GEO AMVs Barb wind plots 😉 )

I installed an instance of eccodes on my VM and write my own bufr reader.

```python 

  f = open(input_file, 'rb')
    lat_ = []
    lon_ = []
    speed_ = []
    wind_dir_ = []
    press_ = []
    freq_ = []
    i=1
    ibufr = codes_bufr_new_from_file(f)
    while ibufr != None:
        codes_set(ibufr, 'unpack', 1)
        lat_.append(codes_get_array(ibufr, '#1#latitude'))
        lon_.append(codes_get_array(ibufr, '#1#longitude'))
        speed_.append(codes_get_array(ibufr, 'windSpeed'))
        wind_dir_.append(codes_get_array(ibufr, 'windDirection'))
        press_.append(codes_get_array(ibufr, '#1#pressure'))
        codes_release(ibufr)
        i+=1
        ibufr = codes_bufr_new_from_file(f)
    f.close()

 lat = np.array(flat_list(lat))
 lon = np.array(flat_list(lon))
 speed = np.array(flat_list(speed))
 wind_dir = np.array(flat_list(wind_dir))
 pres = np.array(flat_list(press))
 freq = np.array(flat_list(freq))
```
First of all lat, lon, pressure, speed and direction all have the same amount of element 51944.
I can plot properly the MSG data for the 18/03/2023 12:00 and I get this:
image

**Potential Solution** 
I poked around and found that in the line
https://github.com/pytroll/satpy/blob/f099cd39189ea12c54ae167d34a40cd6b29625df/satpy/etc/readers/seviri_l2_bufr.yaml#L65C12-L65C12
https://github.com/pytroll/satpy/blob/f099cd39189ea12c54ae167d34a40cd6b29625df/satpy/etc/readers/seviri_l2_bufr.yaml#L75C16-L75C16 
that the key should have a `#1#` as a prefix. 
like so: 
```yaml
  latitude:
    name: latitude
    key: '#1#latitude'
    long_name: Latitude
    standard_name: latitude
    resolution: [48006.450653072,9001.209497451,72009.675979608]
    file_type: [seviri_l2_bufr_asr,seviri_l2_bufr_cla,seviri_l2_bufr_csr,seviri_l2_bufr_gii,seviri_l2_bufr_thu,seviri_l2_bufr_toz,seviri_l2_bufr_amv]
    units: degree_north
    fill_value: -1.e+100

  longitude:
    name: longitude
    key: '#1#longitude'
    resolution: [48006.450653072,9001.209497451,72009.675979608]
    file_type: [seviri_l2_bufr_asr,seviri_l2_bufr_cla,seviri_l2_bufr_csr,seviri_l2_bufr_gii,seviri_l2_bufr_thu,seviri_l2_bufr_toz,seviri_l2_bufr_amv]
    long_name: Longitude
    standard_name: longitude
    units: degree_east
    fill_value: -1.e+100
```
I tested this and it worked and the pytest is passing without issue.

YouvaEUMex:
This PR is to provide a fix for the issue described in https://github.com/pytroll/satpy/issues/2535 
The Fix is rather simple as the issue lies with the keys used to read the bufr for the latitude and longitude(as explained in the issue linked above)
```yaml
  latitude:
    name: latitude
#    key: 'latitude' 
    key: '#1#latitude'
    long_name: Latitude
    standard_name: latitude
    resolution: [48006.450653072,9001.209497451,72009.675979608]
    file_type: [seviri_l2_bufr_asr,seviri_l2_bufr_cla,seviri_l2_bufr_csr,seviri_l2_bufr_gii,seviri_l2_bufr_thu,seviri_l2_bufr_toz,seviri_l2_bufr_amv]
    units: degree_north
    fill_value: -1.e+100

  longitude:
    name: longitude
#    key: 'longitude'
    key: '#1#longitude'
    resolution: [48006.450653072,9001.209497451,72009.675979608]
    file_type: [seviri_l2_bufr_asr,seviri_l2_bufr_cla,seviri_l2_bufr_csr,seviri_l2_bufr_gii,seviri_l2_bufr_thu,seviri_l2_bufr_toz,seviri_l2_bufr_amv]
    long_name: Longitude
    standard_name: longitude
    units: degree_east
    fill_value: -1.e+100
```

 - [ ] Closes #2535


djhoese:
Do you need to update the tests? For example, places like https://github.com/pytroll/satpy/blob/b0473aa9b437db122a5d8fae79e9e87306ad1af9/satpy/tests/reader_tests/test_seviri_l2_bufr.py#L133-L134 so that it reflects the files you're using/seeing? I wonder if the coverage loss is because the YAML is not finding the longitude/latitude arrays and is just not supplying anything in `.attrs["area"]` in the result. If so, then we really need to have the tests updated to check that the area is not None. This is just my guess for the coverage change.

djhoese (approved):
This looks good to me, but I have nothing to do with this reader and no very little about it. I'd prefer if someone more familiar with it gave the final "OK".

At least the coverage looks good now. :wink:

strandgren:
I did some tests with other SEVIRI L2 BUFR products applying Youva's solution and it seems to work well and I did not see any unexpected behavior :)

PR #2555, opened 2023-08-24
Update AMI `true_color_reproduction` composites to use the hybrid NDVI green

simonrp84:
Currently, the AMI `true_color_reproduction` composites use the old, deprecated, green method. This PR updates the AMI YAML file to switch to the preferred hybrid NDVI scaling method for computing the green channel.


mraspaud (approved):
LGTM

PR #2556, opened 2023-08-24
Implement modifier for reducing signal as a function of sunz angle

strandgren:
This PR Implements a new modifier that reduces the signal within a defined sunz angle interval in order to avoid the excessive white glow at the limb in geostationary VIS/NIR imagery. An example on how to define it and use it is shown below:

```
modifiers:
  sunz_reduced:
    modifier: !!python/name:satpy.modifiers.SunZenithReducer
    correction_limit: 55
    max_sza: 90
    strength: 1.5
    optional_prerequisites:
      - solar_zenith_angle

composites:
   true_color:
    compositor: !!python/name:satpy.composites.SelfSharpenedRGB
    prerequisites:
      - name: vis_06
        modifiers: [sunz_corrected, rayleigh_corrected, sunz_reduced]
      - name: ndvi_hybrid_green
      - name: vis_04
        modifiers: [sunz_corrected, rayleigh_corrected, sunz_reduced]
    standard_name: true_color
```

The ability to reduce the signal as a function of the sunz angle is already implemented as part of e.g. the `SunZenithCorrector` modifier which generally works very well for reducing the correction at high angles. The problem occurs when applying Rayleigh correction on imagery where the sunz correction has been reduced at high angles, since pyspectral assumes regular reflectance data normalized by the sun zenith angle when estimating the Rayleigh contribution. This issue is clearly demonstrated in the examples below:

image

It was discussed whether the Rayleigh correction can be adapted to account for this, but after another round of investigation and testing I still did not find a proper way to do this. Hence, I suggest to add the proposed modifier and then when users have had a chance to test it, we can see if we can find a better implementation. A better implementation would mean that the user defines the reduction in the sunz correction and then the Rayleigh correction properly accounts for this (meaning that there would be no need for a third modifier).

The results with the modifier implemented here, following the example recipe above looks like this:
image

Another example with the terminator across the disk, without and with the sunz reduction modifier in this PR:
image

The default values for the reduction (`correction_limit: 55, max_sza: 90, strength: 1.5`) have been chosen to give an overall good trade-off in having a nice transition into deep-space at the limb and not loosing too much information at the terminator. Nevertheless these values can of course be modified by any user that want to use the modifier.

As can be seen in the last example the modifier remove some information at the terminator. Nevertheless, it shall be noted that:
- Without the modifier the colors in this area are generally not well represented.
- The effect get's less evident/relevant when blending the composite with a night-time layer
- It should be possible to restore some of the information by reducing the Rayleigh correction at high sunz angles.

 - [x] Tests added 
 - [X] Fully documented

### DIFF
--- a/satpy/etc/composites/ahi.yaml
+++ b/satpy/etc/composites/ahi.yaml
@@ -107,11 +107,11 @@ composites:
     compositor: !!python/name:satpy.composites.spectral.NDVIHybridGreen
     prerequisites:
     - name: B02
-      modifiers: [sunz_corrected, rayleigh_corrected]
+      modifiers: [sunz_corrected, rayleigh_corrected, sunz_reduced]
     - name: B03
-      modifiers: [sunz_corrected, rayleigh_corrected]
+      modifiers: [sunz_corrected, rayleigh_corrected, sunz_reduced]
     - name: B04
-      modifiers: [sunz_corrected]
+      modifiers: [sunz_corrected, sunz_reduced]
     standard_name: toa_bidirectional_reflectance
 
   airmass:
@@ -275,10 +275,10 @@ composites:
     compositor: !!python/name:satpy.composites.SelfSharpenedRGB
     prerequisites:
     - name: B03
-      modifiers: [sunz_corrected, rayleigh_corrected]
+      modifiers: [sunz_corrected, rayleigh_corrected, sunz_reduced]
     - name: ndvi_hybrid_green
     - name: B01
-      modifiers: [sunz_corrected, rayleigh_corrected]
+      modifiers: [sunz_corrected, rayleigh_corrected, sunz_reduced]
     high_resolution_band: red
     standard_name: true_color
 
--- a/satpy/etc/composites/fci.yaml
+++ b/satpy/etc/composites/fci.yaml
@@ -13,11 +13,11 @@ composites:
     limits: [0.15, 0.05]
     prerequisites:
       - name: vis_05
-        modifiers: [sunz_corrected, rayleigh_corrected]
+        modifiers: [sunz_corrected, rayleigh_corrected, sunz_reduced]
       - name: vis_06
-        modifiers: [sunz_corrected, rayleigh_corrected]
+        modifiers: [sunz_corrected, rayleigh_corrected, sunz_reduced]
       - name: vis_08
-        modifiers: [sunz_corrected ]
+        modifiers: [sunz_corrected, sunz_reduced ]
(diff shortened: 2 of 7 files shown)


strandgren:
Another example for AHI:

image

mraspaud:
I'm wondering if changing the sunz cutoff would have a similar effect?

strandgren:
> I'm wondering if changing the sunz cutoff would have a similar effect?

That's what I try to demonstrate in the PR description. It works well as long as you don't apply any Rayleigh correction, but as soon as the Rayleigh correction is applied, you get issues at the limb since (I believe) pyspectral assumes reflectance data where no such reduction has been applied.

mraspaud:
> That's what I try to demonstrate with the four images in the PR description. It works well as long as you don't apply any Rayleigh correction, but as soon as the Rayleigh correction is applied, you get issues at the limb since (I believe) pyspectral assumes reflectance data where no such reduction has been applied.

Ah, indeed, sorry.

satpy/composites/spectral.py
```diff
@@ -114,8 +114,8 @@ class NDVIHybridGreen(SpectralBlender):
 
     This green band correction follows the same approach as the HybridGreen compositor, but with a dynamic blend
     factor `f` that depends on the pixel-level Normalized Differece Vegetation Index (NDVI). The higher the NDVI, the
-    smaller the contribution from the nir channel will be, following a liner relationship between the two ranges
-    `[ndvi_min, ndvi_max]` and `limits`.
+    smaller the contribution from the nir channel will be, following a liner (default) or non-linear relationship
+    between the two ranges `[ndvi_min, ndvi_max]` and `limits`.
```
mraspaud, line 118: This is from another PR, right?

mraspaud (commented):
LGTM, but I think there are two PRs mixed in here.

satpy/composites/spectral.py
strandgren: hmm yes.. Seems like I may have branched out from the branch used for the other PR.. May I ask what's the best way to deal with this? Reset this branch or leave it as if for now and then pull in master and modify accordingly once the other PR has been merged?

mraspaud: I don't know if it's the right way, but I'd just remove here the code that is implemented in the other PR, and commit that. And then later merge the main branch.

strandgren:
> LGTM, but I think there are two PRs mixed in here.

I think this should be resolved now :)

simonrp84 (approved):

Issue #2559, opened 2023-08-25
MODIS fixtures are loaded for all tests


simonrp84:
**Describe the bug**
The `conftest.py` file in the `tests/reader_tests/` folder contains imports for a number of fixtures used in the MODIS tests.
But, `conftest.py` is loaded by pytest for *all* tests, not just the MODIS ones and hence the MODIS fixtures are unnecessarily loaded. The MODIS tests should be moved to a separate subdirectory to get around this.

djhoese:
To be clear, since it is in `tests/reader_tests/` only the tests in `reader_tests/` are loading these fixtures. They are also per testing session, not for every test so they are only created once. But yes for running individual tests that aren't MODIS this is very annoying an unnecessary.

The fix is to put modis based tests in a new subdirectory `tests/reader_tests/modis_tests/` or similar and move relevant fixtures from `reader_tests/conftest.py` to `reader_tests/modis_tests/conftest.py`

PR #2561, opened 2023-08-29
Fix invalid AVHRR product names in AWIPS config

djhoese:
I also through in the fixes for #2559.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese:
@simonrp84 I tested the change in MODIS testing and it looks like it works as expected (only runs the fixtures when running the modis tests).

simonrp84:
> @simonrp84 I tested the change in MODIS testing and it looks like it works as expected (only runs the fixtures when running the modis tests).

Great, thanks for checking!

PR #2563, opened 2023-08-31
Fix VIIRS EDR handling of multiple fill values in geolocation

djhoese:
Turns out VIIRS EDRs are fun. The fill value is specified as -999.9, but there are other fills in there too like -999.3. This PR fixes this by always masking with valid_range/valid_min/valid_max if available.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese:
This is a small fix. Merging without review.

PR #2565, opened 2023-09-05
Add level-1 readers for the arctic weather satelliter data

adybbroe:
This PR adds a reader for the ESA Arctic Weather Satellite (AWS) l1b format, as well as the corresponding EUMETSAT AWS Level-1b format. The latter to become the EPS-Sterna level-1b file format.

The ESA AWS file format is not documented anywhere but is netCDF and supposed to follow the EPS-SG MWS level-1b format. See  https://www.eumetsat.int/media/44139

Here a draft version of the EPS-Sterna level-1b file format, as od December 2024:

[EPS-Sterna Level 1B Product Format Specification (PFS) v1C [Approved].docx](https://github.com/user-attachments/files/18200814/EPS-Sterna.Level.1B.Product.Format.Specification.PFS.v1C.Approved.docx)

ESA has also developed a level-1c processor, mapping the data for all four channels groups (feed horns) to one of the groups, default is the 183 GHz group. This PR also includes a first reader for this L1c file format. The ESA L1c file format is neither documented, but it is following the L1b format closely.

This PR was started autumn 2023 where the ESA AWS level-1b file format had not matured. Now, the format is supposed to be the final, and we have used first real data processed and delivered to us (SMHI among other partners in the MWS/AWS SAG and EUMETSAT) in November 2024.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 


adybbroe:
Here is a view of channel 1 remapped to a near-ided perspective projection over northern Europe:

aws_testdata_50 3GHz_20230707_1200_nsper_swe

satpy/etc/readers/aws_l1b_nc.yaml
```diff
@@ -0,0 +1,425 @@
+reader:
+  name: aws_l1b_nc
+  short_name: AWS L1B RAD NetCDF4
+  long_name: AWS L1B Radiance (NetCDF4)
+  description: Reader for the ESA AWS (Arctic Weather Satellite) Sounder level-1b files in netCDF4.
```
mraspaud, line 5: Is there a link to online documentation of the format we could use?

satpy/etc/readers/aws_l1b_nc.yaml
```diff
@@ -0,0 +1,425 @@
...
+  reader: !!python/name:satpy.readers.yaml_reader.FileYAMLReader
+  sensors: [aws,]
+  status: Beta
+  default_channels: []
```
mraspaud, line 9: you could also add the remote reading status of the reader.

mraspaud (commented):
Just two comments for now.

adybbroe (commented):
Looks very nice, many thanks!

satpy/etc/readers/aws_l1b_nc.yaml
djhoese: I don't think `default_channels` is used so it could be removed.

satpy/etc/readers/aws_l1b_nc.yaml
```diff
@@ -0,0 +1,415 @@
...
+    frequency_double_sideband:
+      type: !!python/name:satpy.readers.pmw_channels_definitions.FrequencyDoubleSideBand
+    frequency_range:
+      type: !!python/name:satpy.readers.pmw_channels_definitions.FrequencyRange
+    resolution:
+    polarization:
+      enum:
+        - QH
+        - QV
+    horn:
```
djhoese, line 23: What is "horn"? Just curious.

satpy/etc/readers/aws_l1b_nc.yaml
mraspaud: good point, thanks

satpy/etc/readers/aws_l1b_nc.yaml
mraspaud: From what I understand, the instrument is using different "cameras", or horns, for different channels. Which in turn leads to slightly different longitudes/latitudes for each horn.

satpy/etc/readers/aws_l1b_nc.yaml
mraspaud: Fixed in [f039715](https://github.com/pytroll/satpy/pull/2565/commits/f03971533795e871342299e33fa5536dd88380b1)

satpy/etc/readers/aws_l1b_nc.yaml
adybbroe: @djhoese Yes, in MW remote sensing the different radiometer receivers are often called Feedhorns. The Arctic Weather Satellite radiometer has 4 feedhorns, corresponding to the four channel groups, temperature sounding @50-55 GHz, window channel @ 89 GHz, Humidity sounding channels @ 165-183 GHz and the 325 GHz band for humidity and clouds.
We use the same naming for the ICI reader

adybbroe:
So, just a comment/observation. The AWS testdata file delivered August 2023 has the datasets structured a bit odd and different from most other satellite datasets (e.g. the EPS-SG MWS testdata which format it is supposed to follow).

Here a phrase communicated to ESA abouth this layout issue:

"The ordering of the dimensions is unexpected, e.g. for “aws_toa_brightness_temperature” we have (19, 145, 4991) which is (n_channels, n_fovs, n_scans). But in netCDF the first dimension is always the one that varies the most slowly. For MWS the order is reversed, i.e. (n_scans, n_fovs, n_channels)."

This ordering currently has the unfortunate consequence that the optimal radius of influence used for remapping cannot be determined. The method `geocentric_resolution` in the `CoordinateDefinition` class in the `geometry.py` module of Pyresample assumes the first dimension is the scanlines, but here it is the pixel number across scan instead, so the points found are the points close to nadir of the first scanline. In the testdata set delivered the first 10-20 scanlines or so are missing-data in the geolocation (for some strange reason). 

I am preparing a PR for Pyresample on this.

Until now one has to provide a reasonable value (say 40000) for `radius_of_influence` parameter explicitly in the `resample` call on the scene object.

When doing this one may be able create an image like the below using code similar to this (I have here skipped the decoration/coast overlay part for simplification):

```
scn = Scene(filenames=FILENAMES, reader='aws_l1b_nc')
composite_name = 'mw183_humidity_surface'
scn.load([composite_name])

local = scn.resample(AREAID, radius_of_influence=40000)
local.show(composite_name)

```

aws_testdata_radsim_mw183_humidity_surface_20240115_1111_nsper_swe

pnuu (commented):
Some comments inline. I'd like some clarifiactions, but in general LGTM!

satpy/readers/aws_l1b.py
```diff
@@ -0,0 +1,171 @@
...
+#         for key in quality_group:
+#             # Add the values (as Numpy array) of each variable in the group
+#             # where possible
+#             try:
+#                 quality_dict[key] = quality_group[key].values
+#             except ValueError:
+#                 quality_dict[key] = None
+#
+#         quality_dict.update(quality_group.attrs)
+#         return quality_dict
```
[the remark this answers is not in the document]
mraspaud: This is a rest from the first version that @adybbroe wrote. I don't think it's needed yet, but I was waiting for him to give his opinion.

satpy/etc/readers/aws_l1b_nc.yaml
```diff
@@ -0,0 +1,415 @@
...
+        horn: "3"
+        coordinates: [longitude, latitude]
+        file_type: aws_l1b_nc
+        file_key: data/calibration/aws_toa_brightness_temperature
+    '16':
+        name: '16'
+        frequency_double_sideband:
+           central: 325.15
+           side: 1.2
+           bandwidth: 0.8
```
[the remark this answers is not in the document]
adybbroe: Exactly!

https://www.esa.int/Applications/Observing_the_Earth/Meteorological_missions/Arctic_Weather_Satellite/The_instrument

Also see the figure I made for MWS (which has both "quadruple side-bands" and "double side-bands") here:
https://user-images.githubusercontent.com/169069/171609031-a1a26b42-211e-411a-92bb-93d090eccb7c.png

satpy/readers/aws_l1b.py
```diff
@@ -0,0 +1,171 @@
...
+
+DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S.%f"
+
+AWS_CHANNEL_NAMES_TO_NUMBER = {"1": 1, "2": 2, "3": 3, "4": 4,
+                               "5": 5, "6": 6, "7": 7, "8": 8,
+                               "9": 9, "10": 10, "11": 11, "12": 12,
+                               "13": 13, "14": 14, "15": 15, "16": 16,
+                               "17": 17, "18": 18, "19": 19}
+
+AWS_CHANNEL_NAMES = list(AWS_CHANNEL_NAMES_TO_NUMBER.keys())
```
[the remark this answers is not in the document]
adybbroe: I suppose I was not sure if we should name the channels 1 to 19 or as indicated on the ESA web site by using two digits where the first one indicate the group...
https://www.esa.int/Applications/Observing_the_Earth/Meteorological_missions/Arctic_Weather_Satellite/The_instrument

adybbroe:
I somehow thought we had merged this long ago? 
But, it makes sense not to merge it now, as we will have new test data from ESA in about a month, where the issues with the format that we raised with the, should be solved. Then we can update the reader for the new format.
So, we can wait another month  or two perhaps, or what do you say @mraspaud and @pnuu ?

Sorry for not coming back to your comments/questions earlier!

mraspaud:
Any news on the new test dataset?

adybbroe:
Test data is supposed to come in ~2 weeks according to latest news from ESA (August 16). Let's hope this latest forecast is more accurate than the previous.

satpy/etc/readers/aws_l1b_nc.yaml
adybbroe: No, unfortunately not, but there will soon be for EPS-Sterna. Not exactly the same format, but close. I hope that the same code can be used for both, but using different yaml files perhaps

adybbroe:
@pytroll/core Do you what this is that cause the CI tests to fail now?
Doesn't look like anything specific to this PR?

pyproject.toml
```diff
@@ -64,6 +64,7 @@ hsaf_grib = ["pygrib"]
 remote_reading = ["fsspec"]
 insat_3d = ["xarray>=2024.10.0"]
 gms5-vissr_l1b = ["numba"]
+aws_l1b = ["xarray-datatree"]
```
pnuu, line 67: Just a check, but isn't this now a part of `xarray` and doesn't need a separate dependecy? So maybe make the reader dependent on a recent `xarray` release.

satpy/etc/composites/atms.yaml
```diff
@@ -14,5 +14,5 @@ composites:
     prerequisites:
     - name: '16'
     - name: '17'
-    - name: '22'
+    - name: '18'
```
pnuu, line 17: This isn't related to AWS, so shouldn't be part of this PR. Right?

satpy/etc/composites/aws.yaml
```diff
@@ -0,0 +1,46 @@
+sensor_name: aws
```
pnuu, line 1: Looking at https://space.oscar.wmo.int/satellites/view/aws_pfm , the instrument is named MWR, so I guess the sensor (and the file) should also be `mwr(.yaml)`

satpy/etc/enhancements/aws.yaml
```diff
@@ -0,0 +1,57 @@
+enhancements:
```
pnuu, line 1: Also this file should be named `mws.yaml` according to OSCAR.

satpy/etc/enhancements/generic.yaml
```diff
@@ -1289,3 +1289,18 @@ enhancements:
...
+    - name: stretch
+      method: !!python/name:satpy.enhancements.stretch
+      kwargs:
+        stretch: crude
+        min_stretch: [290, 290, 290]
+        max_stretch: [190, 190, 190]
+    - name: gamma
+      method: !!python/name:satpy.enhancements.gamma
+      kwargs:
+        gamma: [1.5, 1.2, 1.2]
```
pnuu, line 1306: No need for this, AWS/MWS enhancement is already defined. Oh, but why this definition is different to that one?

satpy/etc/readers/aws_l1b_nc.yaml
```diff
@@ -0,0 +1,535 @@
+reader:
+  name: aws_l1b_nc
```
pnuu, line 2: This should be `mws_l1b_nc` based on our naming convention using the sensor name and what OSCAR tells us it is.

satpy/etc/readers/aws_l1b_nc.yaml
pnuu: These also should be AWS/MWS (or just MWS).

satpy/etc/readers/aws_l1b_nc.yaml
```diff
@@ -0,0 +1,535 @@
...
+  short_name: AWS L1B RAD NetCDF4
+  long_name: AWS L1B Radiance (NetCDF4)
+  description: Reader for the ESA AWS (Arctic Weather Satellite) Sounder level-1b files in netCDF4.
+  reader: !!python/name:satpy.readers.yaml_reader.FileYAMLReader
+  sensors: [aws,]
```
pnuu, line 7:
```suggestion
  sensors: [mws,]
```

Not sure if it should be a list or just a string :thinking:

satpy/etc/readers/aws_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
+reader:
```
pnuu, line 1: Filename should be `mws_l1c_nc.yaml` and the metadata should also be updated accorindingly.

satpy/etc/readers/eps_sterna_l1b_nc.yaml
```diff
@@ -0,0 +1,532 @@
+reader:
```
pnuu, line 1: Hmm. Shouldn't there be only one reader named `mws_l1b_nc` and another for the level 1c data? The L1b reader for the MWS instrument should be the same for AWS-PFS and the future Sterna constellation.

satpy/readers/aws_l1b.py
```diff
@@ -0,0 +1,186 @@
+# Copyright (c) 2023, 2024 Pytroll Developers
```
pnuu, line 1: File naming: `mws_l1b.py`.

satpy/readers/aws_l1c.py
```diff
@@ -0,0 +1,122 @@
+# Copyright (c) 2023, 2024 Pytroll Developers
```
pnuu, line 1: File naming: `mws_l1c.py`. Haven't yet gone in-depth to see should there be a common base for the L1b and L1c readers.

satpy/readers/aws_l1c.py
```diff
@@ -0,0 +1,122 @@
...
+    This class implements the ESA Arctic Weather Satellite (AWS) Level-1b
+    NetCDF reader. It is designed to be used through the :class:`~satpy.Scene`
+    class using the :mod:`~satpy.Scene.load` method with the reader
+    ``"aws_l1c_nc"``.
+
+    """
+
+    def __init__(self, filename, filename_info, filetype_info, auto_maskandscale=True):
+        """Initialize the handler."""
+        super().__init__(filename, filename_info, filetype_info,
```
pnuu, line 45: Seems like the reader class is not tested.

satpy/tests/reader_tests/test_aws_l1b.py
```diff
@@ -0,0 +1,179 @@
+"""Tests for aws l1b filehandlers."""
```
pnuu, line 1: Naming: AWS -> MWS.

pnuu (changes requested):
The next round of initial comments inline. The major one is that I just realized the insturment is called MWS in OSCAR and to follow the Satpy naming convention pretty much all the `aws` and `AWS` strings in the filenames, metadata and docstrings should be changed to either `mws` or `MWS`. This would make the reader future-proof for the future Sterna constellation. If there are changes, then we ofc need to handle those somehow.

satpy/etc/composites/aws.yaml
adybbroe: Yes, right. Sensor name is "MWR"

satpy/etc/composites/atms.yaml
adybbroe: Yeah, you are right. I am revising the composite recipes now with AWS...

pyproject.toml
adybbroe: Yes, good catch!

adybbroe:
> The next round of initial comments inline. The major one is that I just realized the insturment is called MWS in OSCAR and to follow the Satpy naming convention pretty much all the `aws` and `AWS` strings in the filenames, metadata and docstrings should be changed to either `mws` or `MWS`. This would make the reader future-proof for the future Sterna constellation. If there are changes, then we ofc need to handle those somehow.

Yes, except that you mean "MWR"! Name is "MWR", right

satpy/tests/reader_tests/test_aws_l1b.py
adybbroe: MWS is another sensor onboard EPS-SG-A. MWR is the name, you are right!

satpy/readers/aws_l1c.py
adybbroe: Yes, there should probably, and a common base for the test functions.

satpy/etc/readers/eps_sterna_l1b_nc.yaml
adybbroe: No, unfortunately the formats are not the same! ESAs format is different. So one format for the AWS PFM, and another for the EPS-Sterna. So, we need two readers for the l1b.

satpy/etc/enhancements/generic.yaml
adybbroe: Yes, this is a new, and I think, better recipe for the MW humidity RGBs.

pnuu:
Three letters were evidently too many to get right yesterday :joy: Yeah, `MWR` is the correct instrument name.

satpy/readers/aws_l1c.py
adybbroe: Should be tested now.

adybbroe:
Example: 
```
from satpy import Scene
from glob import glob
import os

AREAID = 'arctic_south'
DATADIR = '/data/aws/l1b'
FILENAMES = glob(os.path.join(DATADIR, "W_XX-OHB*_G_D_2024091320*.nc"))
scn = Scene(filenames=FILENAMES, reader='aws1_mwr_l1b_nc')
scn.load(['mw183_humidity'])
local = scn.resample(AREAID, radius_of_influence=50000)
local.show('mw183_humidity')
```

aws_l1c_mw183_humidity_20240913_2048_arctic_south

satpy/etc/composites/mwr.yaml
```diff
@@ -0,0 +1,46 @@
+sensor_name: aws
```
pnuu, line 1:
```suggestion
sensor_name: mwr
```

satpy/etc/enhancements/aws.yaml
pnuu: The file is still `aws.yaml`

satpy/etc/enhancements/generic.yaml
pnuu: If this one is better, the one specified specifically for MWR should be deleted. It will be used instead of this one because it is defined in the `<instrument>.yaml` file.

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
```diff
@@ -0,0 +1,535 @@
+reader:
+  name: aws1_mwr_l1b_nc
```
pnuu, line 2: What is the `1` in `aws1`? I thought there'd ever be only one AWSat, and the next ones would be named "Sterna".

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
+reader:
+  name: aws1_mwr_l1c_nc
```
pnuu, line 2: Same here, what's the `1` in `aws1`?

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
...
+  short_name: AWS L1C RAD NetCDF4
+  long_name: AWS L1C Radiance (NetCDF4)
```
pnuu, line 4: Missing MWR.
```suggestion
  short_name: AWS/MWR L1C RAD NetCDF4
  long_name: AWS/MWR L1C Radiance (NetCDF4)
```

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
...
+        name: satellite_zenith
+        file_type: aws_l1c_nc
+        file_key: data/navigation/aws_satellite_zenith_angle
+        standard_name: satellite_zenith_angle
+        coordinates:
+            - longitude
+            - latitude
+
+file_types:
+    aws_l1c_nc:
```
pnuu, line 368:
```suggestion
    mwr_l1c_nc:
```

satpy/etc/readers/eps_sterna_mwr_l1b_nc.yaml
```diff
@@ -0,0 +1,532 @@
+reader:
+  name: eps_sterna_mwr_l1b_nc
+  short_name: AWS L1B RAD NetCDF4
+  long_name: AWS L1B Radiance (NetCDF4)
```
pnuu, line 4:
```suggestion
  short_name: EPS-Sterna/MWR L1B RAD NetCDF4
  long_name: EPS-Sterna/MWR L1B Radiance (NetCDF4)
```

satpy/etc/readers/eps_sterna_mwr_l1b_nc.yaml
```diff
@@ -0,0 +1,532 @@
...
+  description: Reader for the EUMETSAT EPS-Sterna Sounder level-1b files in netCDF4.
```
pnuu, line 5: I woudn't call MWR a sounder.
```suggestion
  description: Reader for the EUMETSAT EPS-Sterna radiometer level-1b files in netCDF4.
```

satpy/readers/mwr_l1b.py
```diff
@@ -0,0 +1,198 @@
...
+
+Sample EPS-Sterna l1b format AWS data from 16 orbits the 9th of November 2024.
+
+"""
+
+import xarray as xr
+
+from .netcdf_utils import NetCDF4FileHandler
+
+MWR_CHANNEL_NAMES = list(str(i) for i in range(1, 20))
```
pnuu, line 32:
```suggestion
MWR_CHANNEL_NAMES = [str(i) for i in range(1, 20)]
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+    dummy_array = (np.arange(0, geo_size) * maxval/geo_size).astype("int32")
+    fake_lon_data = xr.DataArray(dummy_array.reshape((10, 145)), dims=geo_dims)
+    maxval = 1800000
+    dummy_array = (np.arange(0, geo_size) * maxval/geo_size - maxval/2).astype("int32")
+    fake_lat_data = xr.DataArray(dummy_array.reshape((10, 145)), dims=geo_dims)
+    return (fake_lon_data, fake_lat_data)
+
+
+@pytest.fixture(scope="module")
+def eps_sterna_mwr_file(tmp_path_factory, fake_mwr_data_array):
```
pnuu, line 96: CodeScene is saying this and two other fixtures are pretty much the same. Maybe refactor the dataset creation to a function or fixture that is used in each.

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+    """Create an EPS-Sterna MWR l1b file."""
+    geo_dims = ["n_scans", "n_fovs", "n_feedhorns"]
+    geo_size = 10*145*4
```
pnuu, line 99: Has the pre-commit stuff been run? This should've been caught by the linter.
```suggestion
    geo_size = 10 * 145 * 4
```

satpy/readers/mwr_l1b.py
```diff
@@ -0,0 +1,198 @@
...
+                                    "solar_azimuth_horn3",
+                                    "solar_azimuth_horn4",
+                                    "solar_zenith_horn1",
+                                    "solar_zenith_horn2",
+                                    "solar_zenith_horn3",
+                                    "solar_zenith_horn4",
+                                    "satellite_azimuth_horn1",
+                                    "satellite_azimuth_horn2",
+                                    "satellite_azimuth_horn3",
+                                    "satellite_azimuth_horn4"]:
```
pnuu, line 145: The navigation dataset names could be moved to a `NAVIGATION_DATASET_NAMES` constant.

pnuu (changes requested):
Thanks for the work on the previous round of comments!

Some more requests in-line.

satpy/tests/reader_tests/conftest.py
adybbroe: Yes, I have pre-commit installed and activated!

satpy/readers/mwr_l1b.py
adybbroe: Good suggestion, done!

satpy/etc/enhancements/generic.yaml
pnuu: The reason most likely is that the enhancement file was still named as `aws.yaml` and not `mwr.yaml` so the name of the file and the sensor didn't match.

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
adybbroe: Yes, I am not really very happy about this. But those in charge claied you can't have a satellite with only three letters, thus AWS1, I have communicated this with EUMETSAT and ESA who has the responsibility towards WMO (Oscar). 
So, satellite name = "AWS1"
Instrument name = "MWR" as well as "EPS-Sterna Radiometer".
The latter came up while the radiomete name was still undecided at ESA/EUMETSAT side, and thus started being used. Think it is already in use in some BUFR tables, so we are stuck with two names for the radiometer! :-(

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
adybbroe: See above!

satpy/etc/composites/mwr.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - name: '16'
+    - name: '18'
+    - name: '19'
+    standard_name: mw325_humidity
+
+  ch1_tbs_colors:
+    compositor: !!python/name:satpy.composites.SingleBandCompositor
+    prerequisites:
+    - name: '1'
+    standard_name: tbs_colors
```
pnuu, line 40: I don't find a matching enhancement for this. Same in the `ch_tbs_colors` below.

satpy/etc/enhancements/mwr.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - name: inverse
+      method: !!python/name:satpy.enhancements.invert
+      args:
+      - [true, true, true]
+    - name: stretch
+      method: !!python/name:satpy.enhancements.stretch
+      kwargs: {stretch: linear}
+    - name: gamma
+      method: !!python/name:satpy.enhancements.gamma
+      kwargs: {gamma: 1.2}
```
pnuu, line 43: All of these three enhancements are identical, are they all needed? The corresponding composites could have the same `standard_name` so only one enhancement would be enough, `mw_humidty_surface` for example.

pnuu: If the enhancements are going to diverge in the future, that would be the time to add different enhancaments.

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
```diff
@@ -0,0 +1,535 @@
...
+  short_name: AWS1 MWR L1B RAD NetCDF4
+  long_name: AWS1 MWR L1B Radiance (NetCDF4)
+  description: Reader for the ESA AWS (Arctic Weather Satellite) Micorwave Radiometer (MWR) level-1b files in netCDF4.
```
pnuu, line 5:
```suggestion
  description: Reader for the ESA AWS (Arctic Weather Satellite) Microwave Radiometer (MWR) level-1b files in netCDF4.
```

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
pnuu: :facepalm:

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
```diff
@@ -0,0 +1,535 @@
...
+        horn: "4"
+        coordinates:
+            - longitude
+            - latitude
+
+file_types:
+    aws_l1b_nc:
+        # W_XX-OHB-Unknown,SAT,1-AWS-1B-RAD_C_OHB_20230707124607_G_D_20220621090100_20220621090618_T_B____.nc
+        # W_XX-OHB-Stockholm,SAT,AWS1-MWR-1B-RAD_C_OHB_20230823161321_G_D_20240115111111_20240115125434_T_B____.nc
+        # W_XX-OHB-Stockholm,SAT,AWS1-MWR-1B-RAD_C_OHB_20230816120142_G_D_20240115111111_20240115125434_T_B____radsim.nc
```
pnuu, line 530: These can be removed.

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
...
+        name: latitude
+        file_type: aws_l1c_nc
+        standard_name: latitude
+        units: degrees_north
+        file_key: data/navigation/aws_lat
+
+# --- Navigation data ---
+
+    solar_azimuth:
+        name: solar_azimuth
```
pnuu, line 332:
```suggestion
    solar_azimuth_angle:
        name: solar_azimuth_angle
```

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
...
+        file_type: aws_l1c_nc
+        file_key: data/navigation/aws_solar_azimuth_angle
+        standard_name: solar_azimuth_angle
+        coordinates:
+            - longitude
+            - latitude
+
+    solar_zenith:
+        name: solar_zenith
```
pnuu, line 341:
```suggestion
    solar_zenith_angle:
        name: solar_zenith_angle
```

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
...
+        file_type: aws_l1c_nc
+        file_key: data/navigation/aws_solar_zenith_angle
+        standard_name: solar_zenith_angle
+        coordinates:
+            - longitude
+            - latitude
+
+    satellite_azimuth:
+        name: satellite_azimuth
```
pnuu, line 350:
```suggestion
    satellite_azimuth_angle:
        name: satellite_azimuth_angle
```

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
...
+        file_type: aws_l1c_nc
+        file_key: data/navigation/aws_satellite_azimuth_angle
+        standard_name: satellite_azimuth_angle
+        coordinates:
+            - longitude
+            - latitude
+
+    satellite_zenith:
+        name: satellite_zenith
```
pnuu, line 359:
```suggestion
    satellite_zenith_angle:
        name: satellite_zenith_angle
```

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,375 @@
...
+        standard_name: satellite_zenith_angle
+        coordinates:
+            - longitude
+            - latitude
+
+file_types:
+    aws_l1c_nc:
+        # W_XX-OHB-Unknown,SAT,1-AWS-1B-RAD_C_OHB_20230707124607_G_D_20220621090100_20220621090618_T_B____.nc
+        # W_XX-OHB-Stockholm,SAT,AWS1-MWR-1B-RAD_C_OHB_20230823161321_G_D_20240115111111_20240115125434_T_B____.nc
+        # W_XX-OHB-Stockholm,SAT,AWS1-MWR-1B-RAD_C_OHB_20230816120142_G_D_20240115111111_20240115125434_T_B____radsim.nc
```
pnuu, line 371: These can be removed.

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
pnuu: Ignore this.

satpy/readers/mwr_l1c.py
```diff
@@ -0,0 +1,70 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Reader for the Arctic Weather Satellite (AWS) MWR level-1c data.
+
+MWR = Microwaver Radiometer, onboard AWS and EPS-Sterna
```
pnuu, line 17:
```suggestion
MWR = Microwave Radiometer, onboard AWS and EPS-Sterna
```

satpy/readers/mwr_l1c.py
```diff
@@ -0,0 +1,70 @@
...
+        """Get the sensor name."""
+        return "MWR"
+
+    def get_dataset(self, dataset_id, dataset_info):
+        """Get the data."""
+        if dataset_id["name"] in MWR_CHANNEL_NAMES:
+            data_array = self._get_channel_data(dataset_id, dataset_info)
+        elif (dataset_id["name"] in ["longitude", "latitude",
+                                     "solar_azimuth", "solar_zenith",
+                                     "satellite_zenith", "satellite_azimuth"]):
```
pnuu, line 50: My suggestions in the YAML portion adds `_angle` for clarity and consistency with other readers to these
```suggestion
        elif (dataset_id["name"] in ["longitude", "latitude",
                                     "solar_azimuth_angle", "solar_zenith_angle",
                                     "satellite_zenith_angle", "satellite_azimuth_angle"]):
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
 #
 # You should have received a copy of the GNU General Public License along with
 # satpy.  If not, see <http://www.gnu.org/licenses/>.
+
 """Setup and configuration for all reader tests."""
+
+import os
+from datetime import datetime, timedelta
```
pnuu, line 22: Please use `import datetime as dt` so that the imported `datetime` module does not make `datetime` package unaccesible.

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
```diff
@@ -0,0 +1,120 @@
+"""Tests for aws l1b filehandlers."""
+
+from datetime import datetime
```
pnuu, line 3: Please use `import datetime as dt` so that further imports are possible.

satpy/tests/reader_tests/test_aws1_mwr_l1c.py
```diff
@@ -0,0 +1,99 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Tests for ESA Arctic Weather Satellite (AWS) level-1c file reading."""
+
+
+
```
pnuu, line 22:
```suggestion


```

satpy/tests/reader_tests/test_aws1_mwr_l1c.py
```diff
@@ -0,0 +1,99 @@
...
+    assert "standard_name" in res.attrs
+    if id_name == "longitude":
+        assert res.max() <= 180
+
+
+@pytest.mark.parametrize(("id_name", "file_key", "fake_array"),
+                         [("solar_azimuth", "data/navigation/aws_solar_azimuth_angle", fake_sun_azi_data),
+                          ("solar_zenith", "data/navigation/aws_solar_zenith_angle", fake_sun_zen_data),
+                          ("satellite_azimuth", "data/navigation/aws_satellite_azimuth_angle", fake_sat_azi_data),
+                          ("satellite_zenith", "data/navigation/aws_satellite_zenith_angle", fake_sat_zen_data)])
```
pnuu, line 89: My consistency suggestion in the YAML and reader parts add the `_angle` ending
```suggestion
@pytest.mark.parametrize(("id_name", "file_key", "fake_array"),
                         [("solar_azimuth_angle", "data/navigation/aws_solar_azimuth_angle", fake_sun_azi_data),
                          ("solar_zenith_angle", "data/navigation/aws_solar_zenith_angle", fake_sun_zen_data),
                          ("satellite_azimuth_angle", "data/navigation/aws_satellite_azimuth_angle", fake_sat_azi_data),
                          ("satellite_zenith_angle", "data/navigation/aws_satellite_zenith_angle", fake_sat_zen_data)])
```

pnuu (changes requested):

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
adybbroe: Ah, yesh, food. Fixed.

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
adybbroe: Done

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
```diff
@@ -0,0 +1,535 @@
...
+        file_reader: !!python/name:satpy.readers.mwr_l1b.AWS_EPS_Sterna_MWR_L1BFile
+        file_patterns: [
+        'W_XX-OHB-Stockholm,SAT,{platform_name}-MWR-1B-RAD_C_OHB_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_{disposition_mode:1s}_{processing_mode:1s}____.nc',
+        'W_XX-OHB-Stockholm,SAT,{platform_name}-MWR-1B-RAD_C_OHB__{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_T_B____.nc',
+        'W_XX-OHB-Stockholm,SAT,{platform_name}-MWR-1B-RAD_C_OHB_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_T_B____radsim.nc']
```
pnuu, line 535: The patterns should have "tags" in place of `OHB` (two places) and `Stockholm.

satpy/readers/mwr_l1c.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    def __init__(self, filename, filename_info, filetype_info, auto_maskandscale=True):
+        """Initialize the handler."""
+        super().__init__(filename, filename_info, filetype_info, auto_maskandscale)
+        self.filename_info = filename_info
+
+    @property
+    def sensor(self):
+        """Get the sensor name."""
+        return "MWR"
```
pnuu, line 42: Seems like this should be lower case so that the enhancement file is found.
```suggestion
        return "mwr"
```

satpy/readers/mwr_l1b.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @property
+    def end_time(self):
+        """Get the end time."""
+        return self.filename_info["end_time"]
+
+    @property
+    def sensor(self):
+        """Get the sensor name."""
+        return self["/attr/instrument"]
```
pnuu, line 74: The sensor name in the test data is set as `"AWS"`, so it is wrong in both content and case (for Satpy, that is)
```suggestion
        return "mwr"
```

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
adybbroe: Ahh, yes, will fix

satpy/readers/mwr_l1c.py
adybbroe: Already fixed

satpy/readers/mwr_l1b.py
adybbroe: Fixed

satpy/etc/composites/mwr.yaml
adybbroe: I will add, have them off line...

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
```diff
@@ -0,0 +1,533 @@
...
+            - latitude
+
+file_types:
+    aws_l1b_nc:
+        # W_XX-OHB-Unknown,SAT,1-AWS-1B-RAD_C_OHB_20230707124607_G_D_20220621090100_20220621090618_T_B____.nc
+        # W_XX-OHB-Stockholm,SAT,AWS1-MWR-1B-RAD_C_OHB_20230823161321_G_D_20240115111111_20240115125434_T_B____.nc
+        file_reader: !!python/name:satpy.readers.mwr_l1b.AWS_EPS_Sterna_MWR_L1BFile
+        file_patterns: [
+        'W_{country:2s}-{organisation:s}-{location:s},SAT,{platform_name}-MWR-1B-RAD_C_OHB_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_{disposition_mode:1s}_{processing_mode:1s}____.nc',
+        'W_{country:2s}-{organisation:s}-{location:s},SAT,{platform_name}-MWR-1B-RAD_C_OHB__{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_{disposition_mode:1s}_{processing_mode:1s}____.nc']
```
pnuu, line 533: The second `OHB` needs to be changed to a Trollsift tag. Also, we should use US English.
```suggestion
        'W_{country:2s}-{organization:s}-{location:s},SAT,{platform_name}-MWR-1B-RAD_C_{organization2:s}_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_{disposition_mode:1s}_{processing_mode:1s}____.nc',
        'W_{country:2s}-{organization:s}-{location:s},SAT,{platform_name}-MWR-1B-RAD_C_{organization2:s}__{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_{disposition_mode:1s}_{processing_mode:1s}____.nc']
```

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
```diff
@@ -0,0 +1,372 @@
...
+        coordinates:
+            - longitude
+            - latitude
+
+file_types:
+    aws_l1c_nc:
+        file_reader: !!python/name:satpy.readers.mwr_l1c.AWS_MWR_L1CFile
+        file_patterns: [
+        'W_XX-OHB-Stockholm,SAT,{platform_name}-MWR-1C-RAD_C_OHB_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_T_B____.nc',
+        'W_XX-OHB-Stockholm,SAT,{platform_name}-MWR-1C-RAD_C_OHB__{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_T_B____.nc',]
```
pnuu, line 372: Patterns have OHB and Stockholm hardcoded.
```suggestion
        'W_XX-{organization:s}-{location:s},SAT,{platform_name}-MWR-1C-RAD_C_{organization2}_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_T_B____.nc',
        'W_XX-{organization:s}-{location:s},SAT,{platform_name}-MWR-1C-RAD_C_{organization2:s}__{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_T_B____.nc',]
```

satpy/etc/readers/eps_sterna_mwr_l1b_nc.yaml
```diff
@@ -0,0 +1,532 @@
...
+        coordinates:
+            - longitude
+            - latitude
+
+file_types:
+    eps_sterna_l1b_nc:
+        # W_XX-EUMETSAT-Darmstadt,SAT,AWS1-MWR-1B-RAD_C_EUMT_20241121085911_G_D_20241109234502_20241110004559_T_N____.nc
+        file_reader: !!python/name:satpy.readers.mwr_l1b.AWS_EPS_Sterna_MWR_L1BFile
+        file_patterns: [
+        'W_XX-EUMETSAT-Darmstadt,SAT,{platform_name}-MWR-1B-RAD_C_EUMT_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_{disposition_mode:1s}_{processing_mode:1s}____.nc',
```
pnuu, line 531: Hard-coded processing center etc.
```suggestion
        'W_XX-{organization:s}-{location:s},SAT,{platform_name}-MWR-1B-RAD_C_{organization2:s}_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_{disposition_mode:1s}_{processing_mode:1s}____.nc',
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
 #
 # You should have received a copy of the GNU General Public License along with
 # satpy.  If not, see <http://www.gnu.org/licenses/>.
+
 """Setup and configuration for all reader tests."""
+
+import os
+from datetime import datetime as dt
+from datetime import timedelta
```
pnuu, line 23:
```suggestion
import datetime as dt
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+
+
+rng = np.random.default_rng()
+
+def random_date(start, end):
+    """Create a random datetime between two datetimes."""
+    delta = end - start
+    int_delta = (delta.days * 24 * 60 * 60) + delta.seconds
+    random_second = randrange(int_delta)
+    return start + timedelta(seconds=random_second)
```
pnuu, line 49:
```suggestion
    return start + dt.timedelta(seconds=random_second)
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+        n_feedhorns="n_geo_groups"
+        prefix = "aws_"
+        longitude_attr = "aws_lon"
+        latitude_attr = "aws_lat"
+
+    geo_dims = ["n_scans", "n_fovs", n_feedhorns]
+    geo_size = 10 * 145 * 4
+
+    ds = DataTree()
+    start_time = dt(2024, 9, 1, 12, 0)
```
pnuu, line 109:
```suggestion
    start_time = dt.datetime(2024, 9, 1, 12, 0)
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+    ds.attrs["sensing_start_time_utc"] = start_time.strftime(DATETIME_FORMAT)
+    end_time = dt(2024, 9, 1, 12, 15)
```
pnuu, line 111:
```suggestion
    end_time = dt.datetime(2024, 9, 1, 12, 15)
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+@pytest.fixture(scope="module")
+def eps_sterna_mwr_file(tmp_path_factory, fake_mwr_data_array):
+    """Create an EPS-Sterna MWR l1b file."""
+    ds = aws_eps_sterna_mwr_l1bfile(fake_mwr_data_array, eps_sterna=True)
+
+    tmp_dir = tmp_path_factory.mktemp("eps_sterna_mwr_l1b_tests")
+    start_time = dt.fromisoformat(ds.attrs["sensing_start_time_utc"])
+    end_time = dt.fromisoformat(ds.attrs["sensing_end_time_utc"])
+    platform_name = "AWS1"
+    processing_time = random_date(dt(2024, 9, 1, 13), dt(2030, 6, 1))
```
pnuu, line 152:
```suggestion
    processing_time = random_date(dt.datetime(2024, 9, 1, 13), dt.datetime(2030, 6, 1))
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+@pytest.fixture(scope="module")
+def aws_mwr_file(tmp_path_factory, fake_mwr_data_array):
+    """Create an AWS MWR l1b file."""
+    ds = aws_eps_sterna_mwr_l1bfile(fake_mwr_data_array, eps_sterna=False)
+
+    tmp_dir = tmp_path_factory.mktemp("aws_l1b_tests")
+    start_time = dt.fromisoformat(ds.attrs["sensing_start_time_utc"])
+    end_time = dt.fromisoformat(ds.attrs["sensing_end_time_utc"])
+    platform_name = "AWS1"
+    processing_time = random_date(dt(2024, 9, 1, 13), dt(2030, 6, 1))
```
pnuu, line 173:
```suggestion
    processing_time = random_date(dt.datetime(2024, 9, 1, 13), dt.datetime(2030, 6, 1))
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+
+
+@pytest.fixture(scope="module")
+def aws_mwr_l1c_file(tmp_path_factory, fake_mwr_data_array):
+    """Create an AWS MWR l1c file."""
+    geo_dims = ["n_scans", "n_fovs"]
+    geo_size = 10 * 145
+
+    ds = DataTree()
+    start_time = dt(2024, 9, 1, 12, 0)
```
pnuu, line 193:
```suggestion
    start_time = dt.datetime(2024, 9, 1, 12, 0)
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+    ds.attrs["sensing_start_time_utc"] = start_time.strftime(DATETIME_FORMAT)
+    end_time = dt(2024, 9, 1, 12, 15)
```
pnuu, line 195:
```suggestion
    end_time = dt.datetime(2024, 9, 1, 12, 15)
```

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,243 @@
...
+    ds.attrs["sensing_end_time_utc"] = end_time.strftime(DATETIME_FORMAT)
+    processing_time = random_date(dt(2024, 6, 1), dt(2030, 6, 1))
```
pnuu, line 197:
```suggestion
    processing_time = random_date(dt.datetime(2024, 6, 1), dt.datetime(2030, 6, 1))
```

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
```diff
@@ -0,0 +1,120 @@
...
+fake_sun_zen_data = make_fake_angles(geo_size, geo_dims)
+fake_sat_azi_data = make_fake_angles(geo_size, geo_dims)
+fake_sat_zen_data = make_fake_angles(geo_size, geo_dims)
+
+
+
+def test_start_end_time(aws_mwr_handler):
+    """Test that start and end times are read correctly."""
+    assert aws_mwr_handler.start_time == dt(2024, 9, 1, 12, 0)
+    assert aws_mwr_handler.end_time == dt(2024, 9, 1, 12, 15)
```
pnuu, line 28:
```suggestion
    assert aws_mwr_handler.start_time == dt.datetime(2024, 9, 1, 12, 0)
    assert aws_mwr_handler.end_time == dt.datetime(2024, 9, 1, 12, 15)
```

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
```diff
@@ -0,0 +1,120 @@
+"""Tests for aws l1b filehandlers."""
+
+from datetime import datetime as dt
```
pnuu, line 3:
```suggestion
import datetime as dt
```

pnuu (changes requested):
In addition to the unresolved points, there was some kind of misunderstanding with the datetime imports, fixes suggested. Also the file patterns need some fixing.

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
adybbroe: Sorry, must have been to late, fixed!

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
adybbroe: Sorry, must have been to late, fixed!

satpy/tests/reader_tests/conftest.py
adybbroe: Sorry, must have been to late, fixed!

satpy/tests/reader_tests/conftest.py
adybbroe: Sorry, must have been to late, fixed!

satpy/tests/reader_tests/conftest.py
adybbroe: Sorry, must have been to late, fixed!

satpy/tests/reader_tests/conftest.py
adybbroe: Oh, yeah, fixed, thx!

satpy/etc/readers/aws1_mwr_l1b_nc.yaml
adybbroe: Yeah, right. Fixed now. I follow the British english spelling and in line with the EPS-Sterna L1b format description (attached above). And there it is "originator" for the second "OHB" you catched...

satpy/etc/readers/eps_sterna_mwr_l1b_nc.yaml
adybbroe: Fixed

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
adybbroe: Fixed

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
simonrp84: Do we need "netcdf4" in the short name? It means the "short" one is almost the same length as the "long" one!

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,242 @@
...
+                                                    originator="SMHI",
+                                                    start_time=start_time, end_time=end_time,
+                                                    processing_time=processing_time,
+                                                    platform_name=platform_name))
+    ds.to_netcdf(filename)
+    return filename
+
+
+@pytest.fixture(scope="module")
+def aws_mwr_l1c_file(tmp_path_factory, fake_mwr_data_array):
```
pnuu, line 186: A lot of this fixture is identical to `aws_eps_sterna_mwr_l1bfile()` function, so maybe that could be used to get the common parts and only update the parts that are different?

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,242 @@
...
+    filename = tmp_dir / compose(file_pattern, dict(country="SE",
+                                                    organisation="SMHI",
+                                                    location="Norrkoping",
+                                                    processing_level="1B",
+                                                    originator="SMHI",
+                                                    start_time=start_time, end_time=end_time,
+                                                    processing_time=processing_time,
+                                                    platform_name=platform_name))
+    ds.to_netcdf(filename)
+    return filename
```
pnuu, line 182: These two fixtures could be parametrized to a single one. The only things different are `eps_sterna=False/True` and the dictionary given to `compose()`.

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
```diff
@@ -0,0 +1,120 @@
...
+
+import datetime as dt
+from enum import Enum
+
+import numpy as np
+import pytest
+
+from satpy.tests.reader_tests.conftest import make_fake_angles, make_fake_mwr_lonlats
+
+platform_name = "AWS1"
```
pnuu, line 11:
```suggestion
PLATFORM_NAME = "AWS1"
```

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
```diff
@@ -0,0 +1,120 @@
...
+file_pattern = "W_XX-OHB-Stockholm,SAT,{platform_name}-MWR-1B-RAD_C_OHB_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_T_B____.nc"  # noqa
```
pnuu, line 12: This is not used and can be removed.

satpy/tests/reader_tests/test_aws1_mwr_l1b.py
```diff
@@ -0,0 +1,120 @@
...
+def test_orbit_number_start_end(aws_mwr_handler):
+    """Test that start and end orbit number is read correctly."""
+    assert aws_mwr_handler.orbit_start == 9991
+    assert aws_mwr_handler.orbit_end == 9992
+
+
+def test_metadata(aws_mwr_handler):
+    """Test that the metadata is read correctly."""
+    assert aws_mwr_handler.sensor == "mwr"
+    assert aws_mwr_handler.platform_name == platform_name
```
pnuu, line 40:
```suggestion
    assert aws_mwr_handler.platform_name == PLATFORM_NAME
```

satpy/tests/reader_tests/test_aws1_mwr_l1c.py
```diff
@@ -0,0 +1,98 @@
...
+
+"""Tests for ESA Arctic Weather Satellite (AWS) level-1c file reading."""
+
+
+import numpy as np
+import pytest
+
+from satpy.tests.reader_tests.conftest import make_fake_angles, make_fake_mwr_l1c_lonlats
+
+platform_name = "AWS1"
```
pnuu, line 27:
```suggestion
PLATFORM_NAME = "AWS1"
```

satpy/tests/reader_tests/test_aws1_mwr_l1c.py
```diff
@@ -0,0 +1,98 @@
...
+# W_XX-OHB-Stockholm,SAT,AWS1-MWR-1C-RAD_C_OHB__20241126183628_G_D_20240913222540_20240914000332_T_B____.nc
+file_pattern = "W_XX-OHB-Stockholm,SAT,{platform_name}-MWR-1C-RAD_C_OHB__{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_T_B____.nc"  # noqa
```
pnuu, line 29: Not used, can be deleted.

satpy/etc/readers/aws1_mwr_l1c_nc.yaml
adybbroe: Thanks, fixed now, will push soon.

satpy/tests/reader_tests/conftest.py
adybbroe: As the level-1c file format is not frozen and not documented (actually same can be said about the ESA AWS level-1b format). I would rather leave it as is for now.

satpy/tests/reader_tests/conftest.py
adybbroe: Well I have also made the tmp-directory path different. I can also change the EPS-Sterna platform name to "ST01" to make it visible that there may be more differences? :-)

satpy/etc/composites/mwr.yaml
adybbroe: Added now

adybbroe:
> In addition to the unresolved points, there was some kind of misunderstanding with the datetime imports, fixes suggested. Also the file patterns need some fixing.

Yeah, don't know what I did, sorry. Now fixed.

satpy/tests/reader_tests/conftest.py
adybbroe: I followed your advice here as well, thanks!

satpy/etc/composites/mwr.yaml
```diff
@@ -0,0 +1,46 @@
+sensor_name: mwr
```
djhoese, line 1: This is a microwave instrument right? Do we need to define a "base" sensor type for "microwave" making this `sensor_name: microwave/mwr`? Or does this work as-is and we won't worry about a shared set of "microwave" composites? If this isn't microwave, then this should probably be `sensor_name: visir/mwr`, right?

satpy/readers/mwr_l1b.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+
+class AWS_EPS_Sterna_MWR_L1BFile(AWS_EPS_Sterna_BaseFileHandler):
+    """Class implementing the AWS/EPS-Sterna MWR L1b Filehandler.
+
+    This class implements the ESA Arctic Weather Satellite (AWS) and EPS-Sterna
+    MWR Level-1b NetCDF reader. It is designed to be used through the
+    :class:`~satpy.Scene` class using the :mod:`~satpy.Scene.load` method with
+    the reader ``"mwr_l1b_nc"``.
```
djhoese, line 111: Is this needed?

satpy/readers/mwr_l1b.py
```diff
@@ -0,0 +1,201 @@
...
+
+    """
+    def __init__(self, filename, filename_info, filetype_info, auto_maskandscale=True):
+        """Initialize the handler."""
+        super().__init__(filename, filename_info, filetype_info, auto_maskandscale)
+
+        if filetype_info["file_type"].startswith("eps_sterna"):
+            self._feed_horn_group_name = "n_feedhorns"
+        else:
+            self._feed_horn_group_name = "n_geo_groups"
```
djhoese, line 121: This doesn't *need* to be changed, but would it make sense for this to go in the filetype info in the YAML and then accessed via `self.filetype_info`? Either directly for every use or reassign to the instance attribute as you've done here (ex. `self._feed_horn_group_name = filetype_info.get("feed_horn_group_name", "n_geo_groups")`).

satpy/readers/mwr_l1b.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+    @property
+    def sub_satellite_longitude_start(self):
+        """Get the longitude of sub-satellite point at start of the product."""
+        return self["status/satellite/subsat_longitude_start"].data.item()
```
djhoese, line 127: Does this compute a dask array/scalar?

satpy/readers/mwr_l1b.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def get_dataset(self, dataset_id, dataset_info):
+        """Get the data."""
+        if dataset_id["name"] in MWR_CHANNEL_NAMES:
+            data_array = self._get_channel_data(dataset_id, dataset_info)
+        elif dataset_id["name"] in NAVIGATION_DATASET_NAMES:
+            data_array = self._get_navigation_data(dataset_id, dataset_info)
+
+        elif dataset_id["name"] in ["longitude", "latitude"]:
+            data_array = self._get_navigation_data(dataset_id, dataset_info)
```
djhoese, line 152: Correct me if I'm wrong but this block could be removed if you did `in NAVIGATION_DATASET_NAMES + ["longitude", "latitude"]` in the above `elif`.

satpy/tests/reader_tests/conftest.py
```diff
@@ -15,4 +15,201 @@
...
+
+def random_date(start, end):
+    """Create a random datetime between two datetimes."""
+    delta = end - start
+    int_delta = (delta.days * 24 * 60 * 60) + delta.seconds
+    random_second = randrange(int_delta)
+    return start + dt.timedelta(seconds=random_second)
+
+
+@pytest.fixture(scope="session")
```
djhoese, line 51: Should this also be module scoped like the other fixtures?

satpy/tests/reader_tests/test_eps_sterna_mwr_l1b.py
```diff
@@ -0,0 +1,70 @@
...
+fake_lon_data, fake_lat_data = make_fake_mwr_lonlats(geo_size, geo_dims, shape)
+
+
+@pytest.mark.parametrize(("id_name", "file_key", "fake_array"),
+                         [("longitude", "data/navigation/longitude", fake_lon_data * 1e-4),
+                          ("latitude", "data/navigation/latitude", fake_lat_data),
+                          ])
+def test_get_navigation_data(eps_sterna_mwr_handler, id_name, file_key, fake_array):
+    """Test retrieving the geolocation (lon-lat) data."""
+    Horn = Enum("Horn", ["1", "2", "3", "4"])
```
djhoese, line 40: Should this be something that is importable from the reader module so users could do the same thing you're doing here and pass it to a DataQuery/DataID that is then given to `Scene.load`?

satpy/tests/reader_tests/conftest.py
adybbroe: Yes, I think so. Changed. Thanks.

satpy/readers/mwr_l1b.py
adybbroe: Good point. Done. The level-1c files does not "of course" have such a parameter. The level-1c processor takes the data on the four feedhorns and remap them on one feedhorn (typically the 183 GHz group).
So, what is the best way to handle this? Now I propose:
```
file_types:
    aws_l1c_nc:
        file_reader: !!python/name:satpy.readers.mwr_l1c.AWS_MWR_L1CFile
        file_patterns: [
        'W_{country:2s}-{organisation:s}-{location:s},SAT,{platform_name}-MWR-1C-RAD_C_{originator:4s}_{processing_time:%Y%m%d%H%M%S}_G_D_{start_time:%Y%m%d%H%M%S}_{end_time:%Y%m%d%H%M%S}_{disposition_mode:1s}_{processing_mode:1s}____.nc'
        ]
        feed_horn_group_name: none
```

adybbroe: This none is never used, it can be anything.

satpy/readers/mwr_l1b.py
adybbroe: I see what you mean, redundant. As this is nothing unique...
I have removed.

satpy/tests/reader_tests/test_eps_sterna_mwr_l1b.py
adybbroe: How do you mean? So, that you could get the lon,lat for a particular horn!?
Geolocation data are a 3-dimensional array, of nscans, nfovs, nhorns...

djhoese: I mean, as a user, what if I wanted to do:

```python
from satpy.readers.mrw_l1b import Horn

scn = Scene(...)
scn.load([DataQuery(name="1", horn=Horn("1"))])
```

But now that I look and think about it I notice a couple things:

1. I think the `Horn` class if done correctly (and satpy itself) the user should be able to pass `horn="1"` and it'll get cast to the Horn Enum automatically. Might be worth testing if not explicitly in a test then just manually as a sanity check. Except...
2. It looks like no dataset/variable has more than one horn? In that sense, I guess they are all unicorns.

adybbroe: Yeah, loading channel 1, should apply to horn number 1. There are four receivers/horns, and thus four channel groups.
Horn/Group 1: channel 1-8 (~55 GHz temperature channels)
Horn/Group 2: channel 9 (~89 GHz window channel)
Horn/Group 3: channel 10-15 (~183 GHz humidity channels)
Horn/Group 4: channel 16-19 (~325 GHz humidity channels)

djhoese: Ok so I guess the only real use case would be `scn.load([DataQuery(horn="1")])` but I'm not really sure we even want to support that or how that would work. I mean, I think that usage of a single DataQuery would only return a single channel anyway. Ok so let's ignore this comment/suggestion then.

PR #2567, opened 2023-09-06
Fix readthedocs creating a dirty git environment

djhoese:
Our version numbers are messed up in multiple ways, but this should get us closer.

https://docs.readthedocs.io/en/latest/build-customization.html#avoid-having-a-dirty-git-index

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,5 +13,8 @@ build:
   os: "ubuntu-20.04"
   tools:
     python: "mambaforge-4.10"
+  jobs:
+    pre_install:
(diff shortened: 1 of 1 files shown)


mraspaud (approved):
Lgtm

djhoese:
I'm still waiting on more results from https://github.com/readthedocs/readthedocs.org/issues/10715 before merging this to make sure I don't need to do anything else. There definitely seems to be a change in RTD that is stopping our old/existing workflow from working properly.

mraspaud (approved):
LGTM

PR #2574, opened 2023-09-13
Rename ABI "night_microphysics_abi" composite to "night_microphysics"

djhoese:
A simple change with a big difference. The default `night_microphysics` recipe results in the ABI C14 channel being used, but NASA/NOAA/CIRA recommend using C13. This changes the final result a decent amount. The enhancement crude stretch was already changed a long time ago to match the CIRA Quick Guide:

https://rammb.cira.colostate.edu/training/visit/quick_guides/QuickGuide_GOESR_NtMicroRGB_Final_20191206.pdf

But I didn't realize that asking for `night_microphysics` would give you the C14 version of the composite. This C13 version was only available via the name `night_microphysics_abi`. This PR changes the name of that composite so it is the main and only night_microphysics loaded for ABI data.

CC @kathys @graemely @scottlindstrom for any thoughts or disagreement with this being the new default in our Geo2Grid project.

Looking at the differences it is kind of like you add a shadow effect to the red regions of the image.

### Old Night Micro

image

### New Night Micro

image

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -555,6 +555,22 @@ composites:
       - name: C13
     standard_name: night_microphysics
 
+  night_microphysics_eum:
+    description: >
+      Nighttime Microphysics RGB following the EUMETSAT recipe
+    compositor: !!python/name:satpy.composites.GenericCompositor
+    prerequisites:
+      - compositor: !!python/name:satpy.composites.DifferenceCompositor
+        prerequisites:
+          - name: C15
+          - name: C14
+      - compositor: !!python/name:satpy.composites.DifferenceCompositor
+        prerequisites:
+          - name: C14
+          - name: C07
+      - name: C14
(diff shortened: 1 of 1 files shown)


pnuu (approved):
LGTM.

The "original" recipe from EUMETSAT uses the SEVIRI bands at 3.9, 10.8 and 12.0 um. The ABI C14 channel (barely) meets the 10.8 specification (`[10.80, 11.20, 11.60]`) so I guess that's why it's been used. I think I'll also adapt my own configs to use C13 (`[10.10, 10.35, 10.60]`) instead of the current C14.

simonrp84:
Yeah, this looks good to me. A EUM version might also be useful to add.

djhoese:
I did add an `_eum` alias for the old recipe. I think this is good to go now. Merging...

PR #2575, opened 2023-09-20
Remove use of deprecated setuptools_scm_git_archive build package

djhoese:
In my Polar2Grid project I install Satpy from source. I noticed today that CI was failing because Satpy is still using the deprecated setuptools_scm_git_archive utility package. The functionality of this package has been merged into setuptools_scm and is no longer needed. It now fails because setuptools_scm (the newest version) no longer includes modules that that package was importing.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/.github/workflows/deploy-sdist.yaml
+++ b/.github/workflows/deploy-sdist.yaml
@@ -14,8 +14,6 @@ jobs:
     steps:
       - name: Checkout source
         uses: actions/checkout@v4
-        with:
-          fetch-tags: true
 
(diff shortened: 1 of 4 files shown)


djhoese:
I also think I can simplify the setup.py by removing a hack we had for setuptools_scm that I *think* is now not needed.

mraspaud (approved):
LGTM, do you think this is ready to merge?

PR #2584, opened 2023-10-02
Convert AHI HSD dask chunking to be based on band resolution

djhoese:
As discussed in #2562, performance (memory usage for sure) is better when chunking is done based on resolution. Especially when native resampling is done. This means the 500m band's chunks are twice the size of the 1km band's chunks and 4 times the size of the 2km band's chunks. This should mean dask doesn't have to do any rechunking.

This PR is a continuation of #2052 and should be rebased/remerged after that is merged first. This PR also includes a fix for sunz correction where the dtype for 32-bit floats was being bumped up to 64-bit floats inside dask computations causing more memory usage than we thought.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese:
Oh I still need to add tests for the reader chunking and dtype fixes there.

mraspaud (approved):
Lgtm! Feel free to merge when the rebating is done and the tests pass.

djhoese:
Reran my profiling and got the same results as before all this rebasing and refactoring so I think we're good here.

PR #2608, opened 2023-10-23
Fix ABI L2 to only convert reflectances to percentages

djhoese:
This is an add-on to #2607 where I realized after merging that the `if` statement (copied from the L1b reader) would be applied to any non-count data with units of `"1"`. For L1b this is really only reflectances I think, but for L2 this could be many many other products. This PR changes this to be specifically `calibration == "reflectance"` which I think is safe enough for now.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


mraspaud (approved):
LGTM, thank for the quick fix.

Issue #1955, opened 2021-12-27
Add OSI-SAF reader


gerritholl:
## Feature Request

**Is your feature request related to a problem? Please describe.**

We are currently using an unmaintained custom reader for OSI-SAF ice maps to produce NinJoTIFF images for our users.

**Describe the solution you'd like**

We would like to integrate this reader into Satpy.

**Describe any changes to existing user workflow**

No changes to existing user workflow foreseen.

Developers would have one more reader to adapt the next times Satpy fundamentals change.

**Additional context**

The alternative would be to keep the status quo (custom reader in DWD only).  Advantage: less maintenance for Satpy developers.  Disadvantage: more maintenance for DWD (+ partners) and lack of opportunity for others to use the reader.

simonrp84:
I was actually considering the need for an OSI-SAF reader and had it filed under "whenever I have time" on my to-do list. Happy to lend a hand if you do decide to implement this in Satpy.

gerritholl:
There is some starting point at https://github.com/ch-k/satpy/blob/feature-osisaf-iceconc-nc-reader/satpy/readers/osisaf_conc_eumetcast_nc.py (I believe the author there is a copy-paste error, to be clarified), but lacking unit tests and in need of some cleanup.

ch-k:
The file osisaf_conc_eumetcast_nc.py  is based on ghrsst_l3c_sst.py. It's copy-paste with some adjustments. It was enough to supply the Polarstern ship on its mission to the north in 2019. But for sure, there is some cwork to do.

PR #2632, opened 2023-11-08
Add reader for OSI SAF L3 products

simonrp84:
This PR adds a reader for the OSI SAD level 3 products on an EASE or polar stereographic grid in netCDF format. It's been tested with the following products:
 - Ice concentration
 - Ice type
 - Ice edge
 - Ice emissivity

 
 - [x] Tests added


satpy/readers/osisaf_l3_nc.py
```diff
@@ -0,0 +1,153 @@
...
+# Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# satpy is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
+# type: ignore
```
djhoese, line 18: @simonrp84 Why is this line here? This is why mypy is failing.

djhoese:
```suggestion
```

simonrp84: That line was there because I copy/pasted the GHRSST reader as the base for my code...and that reader has the line. Should it be removed there too?

djhoese: Lol yeah, in the long run we should be type checking everything...or at least letting mypy choose what it does. That likely needs its own PR since that ignore is probably there for a reason.

satpy/readers/osisaf_l3_nc.py
```diff
@@ -0,0 +1,241 @@
...
+        else:
+            return self._get_ftype_grid()
+
+
+    def _get_ds_attr(self, a_name):
+        """Get a dataset attribute and check it's valid."""
+        try:
+            return self[a_name]
+        except KeyError:
+            return None
```
gerritholl, line 142: Why swallow the `KeyError`s? Wouldn't that hide bugs elsewhere?

satpy/tests/reader_tests/test_osisaf_l3.py
```diff
@@ -0,0 +1,364 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
gerritholl, line 2: Those lines are redundant and can be removed.

satpy/readers/osisaf_l3_nc.py
```diff
@@ -0,0 +1,241 @@
...
+            except KeyError:
+                return "unknown_sensor"
+
+    def _get_platname(self):
+        """Get platform name."""
+        try:
+            return self["/attr/platform_name"]
+        except KeyError:
+            return self["/attr/platform"]
+
```
djhoese, line 219:
```suggestion
```

satpy/readers/osisaf_l3_nc.py
```diff
@@ -0,0 +1,241 @@
...
+        file_units = self._get_ds_units(ds_info, var_path)
+
+        # Try to get the valid limits for the data.
+        # Not all datasets have these, so fall back on assuming no limits.
+        valid_min = self._get_ds_attr(var_path + "/attr/valid_min")
+        valid_max = self._get_ds_attr(var_path + "/attr/valid_max")
+        if valid_min is not None and valid_max is not None:
+            data = data.where(data >= valid_min, np.nan)
+            data = data.where(data <= valid_max, np.nan)
```
djhoese, line 173: If the DataArray are floats then removing the `np.nan` should make xarray smart enough to pick the NaN with the right dtype. Otherwise you have to do `data.dtype.type(np.nan)`. Note I say this because of a bug currently in xarray when combined with numpy 2.x which is not released yet. This is a preventative suggestion.

Without this change if xarray doesn't fix their usage, then when numpy 2.0 comes out your dtype would suddenly get upcast to 64-bit floats. If the variable arrays are integers then you may also have to force the casting to 32-bit floats before this call.

```suggestion
            data = data.where(data >= valid_min)
            data = data.where(data <= valid_max)
```

satpy/readers/osisaf_l3_nc.py
```diff
@@ -0,0 +1,241 @@
...
+
+        # Try to get the fill value for the data.
+        # If there isn't one, assume all remaining pixels are valid.
+        fill_value = self._get_ds_attr(var_path + "/attr/_FillValue")
+        if fill_value is not None:
+            data = data.where(data != fill_value, np.nan)
```
djhoese, line 179: Same here.

```suggestion
            data = data.where(data != fill_value)
```

satpy/readers/osisaf_l3_nc.py
```diff
@@ -0,0 +1,241 @@
...
+    def get_dataset(self, dataset_id, ds_info):
+        """Load a dataset."""
+        logger.debug(f"Reading {dataset_id['name']} from {self.filename}")
+        var_path = ds_info.get("file_key", f"{dataset_id['name']}")
+
+        shape = self[var_path + "/shape"]
+        data = self[var_path]
+        if shape[0] == 1:
+            # Remove the time dimension from dataset
+            data = data[0]
```
djhoese, line 163: Is the time dimension always size 1?

gerritholl (approved):
Looks good.  Some minor comments.

It can't read the files we get from EUMETCAST, which have the format `S-OSI_-DMI_-MULT-GL_NH_CONCn__-202304251200Z.nc`.  But then again, nor can we (our internal reader gives us an HDF error since a while).

satpy/readers/osisaf_l3_nc.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @staticmethod
+    def _parse_datetime(datestr):
+        try:
+            return datetime.strptime(datestr, "%Y-%m-%d %H:%M:%S")
+        except ValueError:
+            try:
+                return datetime.strptime(datestr, "%Y%m%dT%H%M%SZ")
+            except ValueError:
+                return datetime.strptime(datestr, "%Y-%m-%dT%H:%M:%SZ")
```
djhoese, line 41: Can this be moved closer to where it is used?

You could also do:

```python
for dt_format in ("%Y...", ...):
    try:
        return datetime.strptime(datestr, dt_fmt)
    except ValueError:
        continue
raise ValueError(f"Unsupported date format: {datestr}")

**Review by djhoese — CHANGES_REQUESTED** [review]

File: satpy/readers/osisaf_l3_nc.py [hunk 9]
```diff
@@ -0,0 +1,241 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
simonrp84: Fixed!

satpy/readers/osisaf_l3_nc.py
simonrp84: Fixed!

satpy/readers/osisaf_l3_nc.py
simonrp84: Not all files have a time dimension, but if it is there it'll always be the first dim and always have length=1. Hence this check (copied from the GRHSST reader).

satpy/tests/reader_tests/test_osisaf_l3.py
simonrp84: Fixed!

satpy/readers/osisaf_l3_nc.py
djhoese: If the main way this is used is to try multiple possible attribute/variable names, then maybe a different version of this with a for loop would be better?

```python
attr_val = self._get_first_attr(("/attr/X", "/attr/Y", "/attr/Z"))


def _get_first_attr(self, possible_names):
    for attr_name in possible_names:
        try:
            return self[attr_name]
        except KeyError:
            continue
    raise KeyError("...")

```

satpy/readers/osisaf_l3_nc.py
```diff
@@ -0,0 +1,241 @@
...
+        """Get grid in case of filetype info being used."""
+        if self.filetype_info["file_type"] == "osi_radflux_grid":
+            self.area_def = self._get_geographic_grid()
+            return self.area_def
+        elif self.filetype_info["file_type"] == "osi_sst":
+            self.area_def = self._get_polar_stereographic_grid()
+            return self.area_def
+
+    def get_area_def(self, area_id):
+        """Override abstract baseclass method"""
```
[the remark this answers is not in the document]
simonrp84: Updated to be slightly less pointless.

satpy/readers/osisaf_l3_nc.py
simonrp84: Code seems to work with your suggestion in place, so have found and removed all instances of `np.nan` in the `where` statements.

satpy/readers/osisaf_l3_nc.py
simonrp84: Done.

satpy/readers/osisaf_l3_nc.py
simonrp84: Think I inadvertently fixed this while sorting out another suggestion.

simonrp84:
I think I've addressed all your comments, thanks for the reviews!

gerritholl:
Thanks! It works with our data.

gerritholl:
[x] Closes https://github.com/pytroll/satpy/issues/1955 

(does that work automatically in a comment?)

satpy/readers/osisaf_l3_nc.py
simonrp84: Hopefully my newest change is suitable. If not, I'll come back to it some time in the new year.

PR #2634, opened 2023-11-10
Remove duplicate entries of required netcdf variables in FCI reader

pnuu:
While testing things to speed-up FCI reader, I noticed there were some duplicate/triplicate entries in the FCI reader YAML listing the required netcdf variables. Each of these in themselves caused extra file access, so this also speeds up the reader a tiny bit.


mraspaud (approved):
LGTM

pnuu:
There could be some others that are read twice even if not listed, like `data/mtg_geos_projection` bringing in all its attributes which are listed also separately. I'll see if this is true and clean also these.

pnuu:
Yep, all the separately listed attributes of `data/mtg_geos_projection` were already read when collecting the parent.

ameraner (approved):
LGTM, well spotted!

PR #2640, opened 2023-11-15
Keep original dtype in DayNightCompositor

pnuu:
This PR is a start to keep the `dtype` of the input data until saving. To have the tests pass, another PR is needed for Trollimage so that the enhancements don't upscale the data.

The tests require https://github.com/pytroll/trollimage/pull/150 to be merged (done) and a new Trollimage release including it. This is a side-effect of the Satpy feature applying a linear stretch by default, and in the current release this up-casts `float32` to `float64`.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/modifiers/angles.py
+++ b/satpy/modifiers/angles.py
@@ -403,8 +403,9 @@ def get_cos_sza(data_arr: xr.DataArray) -> xr.DataArray:
     """
     chunks = _geo_chunks_from_data_arr(data_arr)
     lons, lats = _get_valid_lonlats(data_arr.attrs["area"], chunks)
-    lons = lons.astype(data_arr.dtype)
-    lats = lats.astype(data_arr.dtype)
+    if lons.dtype != data_arr.dtype and np.issubdtype(data_arr.dtype, np.floating):
+        lons = lons.astype(data_arr.dtype)
+        lats = lats.astype(data_arr.dtype)
     cos_sza = _get_cos_sza(data_arr.attrs["start_time"], lons, lats)
     return _geo_dask_to_data_array(cos_sza)
 
--- a/satpy/modifiers/angles.py
+++ b/satpy/modifiers/angles.py
@@ -395,6 +395,8 @@ def get_cos_sza(data_arr: xr.DataArray) -> xr.DataArray:
     """
     chunks = _geo_chunks_from_data_arr(data_arr)
     lons, lats = _get_valid_lonlats(data_arr.attrs["area"], chunks)
+    lons = lons.astype(data_arr.dtype)
+    lats = lats.astype(data_arr.dtype)
     cos_sza = _get_cos_sza(data_arr.attrs["start_time"], lons, lats)
     return _geo_dask_to_data_array(cos_sza)
 
--- a/satpy/tests/test_composites.py
+++ b/satpy/tests/test_composites.py
@@ -490,7 +490,6 @@ class TestDayNightCompositor(unittest.TestCase):
             res = comp((self.data_b,))
             res = res.compute()
         expected_l_channel = np.array([[np.nan, 0.], [0.5, 1.]], dtype=np.float32)
-        # FIXME: with the current changes the np.nan becomes 0.0 instead, why?!
         expected_alpha = np.array([[np.nan, 0.], [0., 0.]], dtype=np.float32)
         assert res.dtype == np.float32
         np.testing.assert_allclose(res.values[0], expected_l_channel)
--- a/satpy/tests/test_composites.py
+++ b/satpy/tests/test_composites.py
@@ -437,7 +437,7 @@ class TestDayNightCompositor(unittest.TestCase):
             comp = DayNightCompositor(name="dn_test", day_night="day_night")
             res = comp((self.data_a, self.data_b, self.sza))
             res = res.compute()
-        expected = np.array([[0., 0.22122352], [0.5, 1.]], dtype=np.float32)
+        expected = np.array([[0., 0.22122374], [0.5, 1.]], dtype=np.float32)
         assert res.dtype == np.float32
         np.testing.assert_allclose(res.values[0], expected, rtol=1e-6)
 
@@ -462,7 +462,7 @@ class TestDayNightCompositor(unittest.TestCase):
             comp = DayNightCompositor(name="dn_test", day_night="night_only", include_alpha=True)
             res = comp((self.data_b, self.sza))
             res = res.compute()
-        expected_red_channel = np.array([[np.nan, 0.], [0.49999994, 1.]], dtype=np.float32)
+        expected_red_channel = np.array([[np.nan, 0.], [0.5, 1.]], dtype=np.float32)
         expected_alpha = np.array([[0., 0.3329599], [1., 1.]], dtype=np.float32)
         assert res.dtype == np.float32
         np.testing.assert_allclose(res.values[0], expected_red_channel)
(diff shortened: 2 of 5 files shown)


satpy/tests/test_composites.py
```diff
@@ -485,8 +489,10 @@ def test_night_only_area_with_alpha(self):
             comp = DayNightCompositor(name="dn_test", day_night="night_only", include_alpha=True)
             res = comp((self.data_b,))
             res = res.compute()
-        expected_l_channel = np.array([[np.nan, 0.], [0.5, 1.]])
-        expected_alpha = np.array([[np.nan, 0.], [0., 0.]])
+        expected_l_channel = np.array([[np.nan, 0.], [0.49999994, 1.]], dtype=np.float32)
+        # FIXME: with the current changes the np.nan becomes 0.0 instead, why?!
+        expected_alpha = np.array([[np.nan, 0.], [0., 0.]], dtype=np.float32)
```
pnuu, line 494: The `np.nan` will be changed to `0.0` with the PR. No idea why, so didn't update the test.

pnuu:
The tests will require https://github.com/pytroll/trollimage/pull/150

satpy/modifiers/angles.py
```diff
@@ -395,6 +395,8 @@ def get_cos_sza(data_arr: xr.DataArray) -> xr.DataArray:
     """
     chunks = _geo_chunks_from_data_arr(data_arr)
     lons, lats = _get_valid_lonlats(data_arr.attrs["area"], chunks)
+    lons = lons.astype(data_arr.dtype)
+    lats = lats.astype(data_arr.dtype)
```
gerritholl, line 399: Should this simply be hardcoded `float32` (or even `float16`)? If those are only used for SZA corrections, I don't see why we would ever need a higher precision, or why the precision should depend on the dtype of the input data.  I don't know if `data_arr.dtype` could ever be integer here, but if it is, that would impact the quality of the correction.

djhoese: I could have sworn I fixed this inside the `_get_cos_sza` call or something. Let me track it down. I don't think I agree with casting lon/lats to the data array dtype here.

djhoese: https://github.com/pytroll/satpy/commit/e6e5c6e5f0085c8c8216951b80a46a468b0f75a2

pnuu: The `sunzen_corr_cos()` isn't used because, well, we're not correcting for cosine angle, just using it to weight the overlap. The casting is needed to keep the data at proper dtype. I wanted to pass the dtype to `area.get_lonlats()`, but the decorated zarr-caching is out of my understanding and just passing the dtype as another argument didn't work.

satpy/tests/test_composites.py
pnuu: Ok, just needed a simple update to getting the mask after changing the "missing" data from integer 0 to `data.dtype`.

satpy/modifiers/angles.py
djhoese: Ah ok.

I don't think the cached functions should be correcting for dtype in this sense. By doing that you'd end up with possibly two cached versions of the data with the only different being 32-bit and 64-bit floating dtype. What about casting the `cos_sza` after the `_get_cos_sza` call? Or is that not good enough?

The `_get_cos_sza` function is also used in the below `_get_sun_angles` function too so we may want this type of casting done there too. Also, isn't this casting dangerous if the input array is an integer type?

Side note: The `get_valid_lonlats` should maybe make sure that `np.nan` is the same dtype as the lon/lats when masking to avoid upcasting there. I think we usually use 64-bit floats for all x/y coordinates so the lon/lats are usually 64-bit too so probably not a huge problem.

pnuu: Casting as it is in this PR, processing time 49.1 s:
Screenshot 2023-11-20 at 11-04-43 Bokeh Plot

Casting the result of call to `cos_sza()`, processing time 51.6 s:
Screenshot 2023-11-20 at 11-06-11 Bokeh Plot

The latter uses ~1 GB more memory.

pnuu: Oh, and no visible difference for 8-bit imagery.

pnuu: The lons/lats casting is made more strict in https://github.com/pytroll/satpy/pull/2640/commits/e98eb3ee54831dd2449f7fc73a72c5bdff6b4324 . The casting is made only if the dtypes of lons/lats and the data don't match, and data are floats.

pnuu:
The composite tests will require also https://github.com/pytroll/trollimage/pull/151 and a new release of Trollimage.

satpy/modifiers/angles.py
djhoese: Ah this seems much safer. Thanks.

pnuu:
The changes in Trollimage might have affected the "thruth" values in also other tests, I'll have a look.

pnuu:
https://github.com/pytroll/satpy/pull/2640/commits/2ff0b2acad2becd2a86b070e16dfdd73e2432b78 should fix three of the tests that were affected by floating point inaccuracy. There's a remaining Ninjo geotiff failure and three CF failures, I'd guess they are again because of a new XArray release without the compression fix.

pnuu:
Oh, now there's only the single Ninjo tiff failure. I'll compare the results with previous Trollimage version with the current one tomorrow and see what could be the problem. I don't know what the `NinJoTagGenerator` does, @gerritholl do you have time to have a look?

gerritholl:
Some background info to the failing test:

The `NinjoTagGenerator` class derives NinJo tags from data and metadata.  In the failing test, it calculates the minimum grey value for an RGB image that was originally with data ranged from 0 to 255, but that should become 1 to 255 because we used `fill_value=0`, such that the rest of the data have to be stretched from 1 to 255.  I haven't run the test with this branch yet to dig deeper, but it looks like it became 255 rather than 1, which is not good.

I have no clue what that has to do with dtype preservation in the `DayNightCompositor` :O

pnuu:
> I have no clue what that has to do with dtype preservation in the DayNightCompositor :O

Nothing, the change is due to my updates to Trollimage (most likely https://github.com/pytroll/trollimage/pull/150) which were released today and needed for the DNC tests to pass in this PR.

pnuu:
Ok.

```python
#!/usr/bin/env python

import numpy as np
import xarray as xr

from trollimage import xrimage

arr = np.arange(75, dtype=np.uint8).reshape(5, 5, 3)
arr[4, 4, :] = 255
data = xr.DataArray(arr.copy(), dims=['y', 'x', 'bands'],
                    coords={'bands': ['R', 'G', 'B']})
img = xrimage.XRImage(data)
print("img", img.data.values.min(), img.data.values.max())
img.stretch_linear()
print("stretched img", img.data.values.min(), img.data.values.max())
```

With tag Trollimage `v1.21.0` this prints
```
$ python test_xr_uint8.py
img 0 255
stretched img -0.0015614156835530892 1.0960743801652901
```
and with `main`
```
$ python test_xr_uint8.py
img 0 255
stretched img 0.0 2.2130435
```

I'll create an issue on Trollimage.

pnuu:
https://github.com/pytroll/trollimage/issues/152

pnuu:
The integer stretching is fixed in https://github.com/pytroll/trollimage/pull/153 . So we need a patch release of Trollimage to resolve this.

mraspaud:
> The integer stretching is fixed in [pytroll/trollimage#153](https://github.com/pytroll/trollimage/pull/153) . So we need a patch release of Trollimage to resolve this.

We will do a patch release as soon as the PR is merged.

PR #2652, opened 2023-11-28
Add file pattern for CRRPh of NWC SAF GEO v2021

pnuu:
The 2021 version NWC SAF GEO software producess CRRPh ("convective rain rate from physical somethingorother") using file names like `S_NWC_CRRPh_MSG3_MSG-N-VISIR_20231128T051500Z.nc`. The previous version (v2018) has the files named with a dash: `S_NWC_CRR-Ph_MSG3_MSG-N-VISIR_20231128T051500Z.nc`. This PR adds the new pattern.

There's a similar change to `PC-Ph` -> `PCPh` but we don't have that product currently in the config, and I'm not sure its yet supported.

In addition one unrelated failing test assertion is adjusted for floating point handling.


mraspaud (approved):
LGTM

PR #2669, opened 2023-12-08
Fix RealisticColors compositor upcasting data to float64

pnuu:
The `RealisticColors` compositor was upcasting the data to `float64` (via `np.where()` usage). This PR fixes the issue, and adds the first test for the compositor.

 - [x] Tests added

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -992,17 +992,17 @@ class RealisticColors(GenericCompositor):
         hrv = projectables[2]
 
         try:
-            ch3 = 3 * hrv - vis06 - vis08
+            ch3 = 3.0 * hrv - vis06 - vis08
             ch3.attrs = hrv.attrs
         except ValueError:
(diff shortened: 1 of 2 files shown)


pnuu:
Creating the `realistic_colors` composite with `main`:
Screenshot 2023-12-08 at 09-47-55 Bokeh Plot

The same with this PR:
Screenshot 2023-12-08 at 09-49-23 Bokeh Plot

There's no change in memory usage, but the PR'd version is 17 seconds faster (20.0 s vs. 37 s).

djhoese (approved):
How did this never have tests before? Looks good to me. Thanks for fixing it.

One confusion: how does avoiding upcasting change the execution time so much?

pnuu:
> One confusion: how does avoiding upcasting change the execution time so much?

My guess is that the tasks now interleave better with each other. The same can be observer with most of the other `dtype` and premature computation fixes I've made so far.

PR #2671, opened 2023-12-08
Workaround AWIPS bug not handling integers properly in "awips_tiled" writer

djhoese:
Discovered by @spruceboy and the rest of the GINA team in Alaska. It seems AWIPS is fine displaying data that is unsigned 16-bit integers stored in a 16-bit *signed* integer with the `_Unsigned: true` attribute. However, AWIPS does like this data when using special "derivedParameters" functionality where it treats the data as signed data before passing it to additional processing. In an attempt workaround this issue @spruceboy tried saving the data as signed integers (no `_Unsigned` attribute) and that fixed the addition processing, but broke the visualization. The simplest fix for now (seeing as AWIPS development will take 6-12 months minimum once the fix is implemented) is in this PR which is to only use the positive side of the signed variable.

Note: I had to update one of the tests because the precision of the saved NetCDF data compared to the original input data was reduced, *BUT* this PR is not fixing a Satpy bug. This PR is working around an AWIPS bug.

CC @kathys

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/writer_tests/test_awips_tiled.py
+++ b/satpy/tests/writer_tests/test_awips_tiled.py
@@ -198,7 +198,7 @@ class TestAWIPSTiledWriter:
             check_required_properties(unmasked_ds, output_ds)
             scale_factor = output_ds["data"].encoding["scale_factor"]
             np.testing.assert_allclose(input_data_arr.values, output_ds["data"].data,
-                                       atol=scale_factor / 2)
+                                       atol=scale_factor * 0.75)
 
     def test_units_length_warning(self, tmp_path):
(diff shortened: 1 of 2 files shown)


spruceboy:
Thanks for your help with this @djhoese!  Just to be clear, this appears to be a bug in edex, the part of AWIPS the pre-processes data for display in the AWIPS visualization tool CAVE.  There are probibly ways to work around this issue inside EDEX, but we can't seem to find anyone who can explain what is going on.  Thanks again!

PR #2674, opened 2023-12-11
Update xarray version in CF writer tests for compression kwarg

pnuu:
Yet another XArray release without the compression kwarg fix, so lets bump the version up a notch.


sfinkens (approved):
Thanks!

PR #2675, opened 2023-12-11
Make CF encoding of dataset attributes public

sfinkens:
In some of my projects I have been using `satpy.writer.encode_attrs_nc`, which has been made private in Satpy 0.45. This is an attempt to get it back :slightly_smiling_face: 

~~Also, xarray still doesn't support the `compression` kwarg, so I bumped the anticipated version number by two months.~~ Solved by Panu

 - [X] Tests added 
 - [X] Fully documented

### DIFF
--- a/satpy/cf/attrs.py
+++ b/satpy/cf/attrs.py
@@ -128,8 +128,8 @@ def _encode_to_cf(obj):
         return _encode_python_objects(obj)
 
 
-def _encode_nc_attrs(attrs):
-    """Encode dataset attributes in a netcdf compatible datatype.
+def encode_attrs_to_cf(attrs):
+    """Encode dataset attributes as a netcdf compatible datatype.
 
     Args:
         attrs (dict):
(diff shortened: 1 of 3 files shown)


djhoese (approved):
I'm fine merging this, but just so we're sure we're not missing some feature or other use case here, @sfinkens could you provide more information about your use case? Are you using xarray? What makes you need this lower-level function versus the higher-level ones that call it?

sfinkens:
Sure! I have an algorithm that returns an xarray dataset, which I want to save to disk. The dataset's `to_netcdf` method is almost fine, I just need to make the attributes CF compatible.

```
ds = my_algo()
ds = encode_attrs_nc(ds)
ds.to_netcdf()
```

The higher-level method would be `satpy.cf.attrs.preprocess_attrs`, which does more than just encoding (adding long name for example).

djhoese:
Good point on the preprocess function. Makes sense.

PR #2676, opened 2023-12-11
Convert times in SEVIRI readers to nanosecond precision to silence warnings

pnuu:
The SEVIRI readers (at least for HRIT) issues a ton of warnings about not having nanosecond precision for the times passed to DataArray. This PR converts the times to ns precision to silence the warnings, and separates the three associated tests.

 - [x] Tests added

### DIFF
--- a/satpy/tests/reader_tests/test_seviri_base.py
+++ b/satpy/tests/reader_tests/test_seviri_base.py
@@ -95,7 +95,7 @@ class SeviriBaseTest(unittest.TestCase):
         expected = np.datetime64("2016-03-03 12:00:00.000")
         res = get_cds_time(days=days, msecs=msecs)
         np.testing.assert_equal(res, expected)
-        assert ".000000000" in res.__repr__()
+        assert res.dtype == np.dtype("datetime64[ns]")
 
     def test_pad_data_horizontally_bad_shape(self):
         """Test the error handling for the horizontal hrv padding."""
--- a/satpy/tests/reader_tests/test_seviri_base.py
+++ b/satpy/tests/reader_tests/test_seviri_base.py
@@ -74,23 +74,28 @@ class SeviriBaseTest(unittest.TestCase):
         exp = chebyshev4(coefs, time, domain)
         np.testing.assert_allclose(res, exp)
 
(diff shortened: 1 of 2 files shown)


djhoese (approved):
Awesome! Thank you!

satpy/tests/reader_tests/test_seviri_base.py
```diff
@@ -74,23 +74,28 @@ def test_chebyshev(self):
...
+    def test_get_cds_time_nanoseconds(self):
+        """Test the get_cds_time function for having nanosecond precision."""
         days = 21246
-        msecs = 12*3600*1000
+        msecs = 12 * 3600 * 1000
         expected = np.datetime64("2016-03-03 12:00:00.000")
-        np.testing.assert_equal(get_cds_time(days=days, msecs=msecs), expected)
+        res = get_cds_time(days=days, msecs=msecs)
+        np.testing.assert_equal(res, expected)
+        assert ".000000000" in res.__repr__()
```
sfinkens, line 98: Does this test the dtype? Can this be tested directly, instead of indirectly via string representation?

pnuu: Hmm, the dtype is `dtype('<M8[ns]')`. There is also the `ns` part, but I need to dig a bit more (hopefully tomorrow) to find the best way.

sfinkens: Interesting, I thought the precision was more easily accessible. What about
```
>>> a.dtype.name
'datetime64[ns]'
```

sfinkens: Or simply
```
a.dtype == np.dtype("datetime64[ns]")
```

pnuu: Adjusted in https://github.com/pytroll/satpy/pull/2676/commits/df26a586eff71ed269e449dd8c59ef12a4b129a6

sfinkens (approved):
LGTM, thanks for fixing this!

mraspaud (approved):
LGTM!

PR #2681, opened 2023-12-12
Get rid of warnings in resampler tests

pnuu:
This PR adjusts the resamplers so that the tests do not issue any warnings.

Also warning from import gradient search is handled. This is not visible in the tests, but I've seen this in real usage.

 - [x] Tests added

### DIFF
--- a/satpy/resample.py
+++ b/satpy/resample.py
@@ -148,10 +148,8 @@ from weakref import WeakValueDictionary
 
 import dask.array as da
 import numpy as np
-import pyresample
 import xarray as xr
 import zarr
-from packaging import version
 from pyresample.ewa import DaskEWAResampler, LegacyDaskEWAResampler
 from pyresample.geometry import SwathDefinition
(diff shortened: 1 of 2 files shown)


satpy/resample.py
```diff
@@ -148,13 +148,11 @@
...
 import dask.array as da
 import numpy as np
-import pyresample
 import xarray as xr
 import zarr
-from packaging import version
 from pyresample.ewa import DaskEWAResampler, LegacyDaskEWAResampler
 from pyresample.geometry import SwathDefinition
-from pyresample.gradient import GradientSearchResampler
+from pyresample.gradient import create_gradient_search_resampler
```
djhoese, line 155: The module docstring still uses the class. If it shouldn't, could you update it to point to this function too?

djhoese (approved):
I made one comment about the docstring. Besides that this looks good to go. Thanks!

PR #2687, opened 2023-12-13
add support of masking olci l1b products by using quality flags

### LINKED ISSUE #1209: OLCI reader can't process quality flags
## Feature Request

The OLCI L1 reader does not recognise the quality flag products that are shipped with the data. These are in the `qualityFlags.nc` file.

For bright scenes, OLCI frequently saturates (see image) and it would be good to have the quality flags available to mask these saturated pixels.
baddata

1) Add quality flags to the reader YAML
2) Update the reader functions to understand and use the quality flags, masking bad pixels

**Additional context**
This issue is mainly a reminder for me, so I can come back to it and implement a fix when I have some time.

yufeizhu600:
This PR added support of using the quality flags from the qualityFlags.nc file to mask the OCLI L1b products, similar as using wqsf file masking OLCI L2 products.

 - [x] Closes #1209 
 - [x] Tests added 
 - [ ] Fully documented


satpy/readers/olci_nc.py
```diff
@@ -184,24 +196,40 @@ def _get_solar_flux(self, band):
...
                              solar_flux=solar_flux, dtype=solar_flux.dtype)
 
+    def getbitmask(self, quality_flags, items=None):
+        """Get the quality flags bitmask."""
+        if items is None:
+            items = DEFAULT_L1B_MASK_ITEMS
+
+        bflags = BitFlags(quality_flags, flag_list=L1B_QUALITY_FLAGS)
+
+        return reduce(np.logical_or, [bflags[item] for item in items])
```
djhoese, line 206: Is this the same?

```suggestion
        return np.array([bflags[item] for item in items]).any()
```

satpy/readers/olci_nc.py
```diff
@@ -184,24 +196,40 @@ def _get_solar_flux(self, band):
...
+
     def get_dataset(self, key, info):
         """Load a dataset."""
-        if self.channel != key["name"]:
+        if self.channel is not None and self.channel != key["name"]:
```
djhoese, line 210: When is `self.channel` `None`?

satpy/tests/reader_tests/test_olci_nc.py
```diff
@@ -241,7 +296,7 @@ def test_chl_nn(self, mocked_dataset):
         assert res.values[-1, -1] == 1e29
 
 
-class TestBitFlags(unittest.TestCase):
+class TestL2BitFlags(unittest.TestCase):
```
djhoese, line 299: This class is only a single function and has no setup. Feel free to extract the single method into a function and remove the class. This is also not using anything specific to unittest so it doesn't need to be a TestCase unittest anymore (but is being removed regardless).

satpy/tests/reader_tests/test_olci_nc.py
```diff
@@ -274,3 +329,45 @@ def test_bitflags(self):
                              False, False,  True,  True, False, False, True,
                              False])
         assert all(mask == expected)
+
+
+class TestL1bBitFlags(unittest.TestCase):
```
djhoese, line 334: Same here. Remove the class and make the method a regular function.

satpy/readers/olci_nc.py
```diff
@@ -166,10 +176,12 @@ def __init__(self, filename, filename_info, filetype_info, engine=None):
 class NCOLCI1B(NCOLCIChannelBase):
     """File handler for OLCI l1b."""
 
-    def __init__(self, filename, filename_info, filetype_info, cal, engine=None):
+    def __init__(self, filename, filename_info, filetype_info, cal=None, engine=None, mask_items=None):
         """Init the file handler."""
         super().__init__(filename, filename_info, filetype_info, engine)
-        self.cal = cal.nc
+        if cal is not None:
```
djhoese, line 182: In what conditions is `cal` None? Should this be required (but keep it as a kwarg for consistency) and raise an error if it is None?

djhoese (changes requested):
I won't pretend to understand everything that's going on here, but I had a couple questions and suggestions to start with.

satpy/readers/olci_nc.py
yufeizhu600: No, the reduce() function return one 2D Boolean array from multiply arrays, and the result from np.array() call is a single Boolean value.

satpy/tests/reader_tests/test_olci_nc.py
yufeizhu600: Sure, I will simplify the test from a class to a single function.

satpy/readers/olci_nc.py
djhoese: Ok. I suppose you could do a `np.stack` then an or or any along a certain axis/dimension, but probably not worth it.

satpy/readers/olci_nc.py
yufeizhu600: Thanks for comments. 
When loading `mask` instead of an actual channel, the self.channel=None;  similar as the get_dataset() in the L2 Class at https://github.com/pytroll/satpy/blob/main/satpy/readers/olci_nc.py#L218

satpy/readers/olci_nc.py
yufeizhu600: The `cal` will be `None` when loading `mask` into the scene. In that case, that is alright since we only care about the `mask_items`.

satpy/tests/reader_tests/test_olci_nc.py
yufeizhu600: Sure, I will modify it.

satpy/readers/olci_nc.py
djhoese: Ok. It makes me feel like this should be done differently, but this reader code is too big to try to change it now for such a "simple" addition/change. Looks good.

djhoese (approved):
I don't know enough about this reader to have the final say, I think @mraspaud should do that. This looks good enough to me though.

satpy/tests/reader_tests/test_olci_nc.py
```diff
@@ -137,6 +137,61 @@ def test_get_mask_with_alternative_items(self, mocked_dataset):
...
+
+        from satpy.readers.olci_nc import NCOLCI1B
+        from satpy.tests.utils import make_dataid
+        mocked_dataset.return_value = xr.Dataset({"quality_flags": (["rows", "columns"],
+                                                           np.array([1 << (x % 32) for x in range(35)]).reshape(5, 7))},
+                                                 coords={"rows": np.arange(5),
+                                                         "columns": np.arange(7)})
+        ds_id = make_dataid(name="mask")
+        filename_info = {"mission_id": "S3A", "dataset_name": "mask", "start_time": 0, "end_time": 0}
+        test = NCOLCI1B("somedir/somefile.nc", filename_info, "c")
```
djhoese, line 155: Depending on how bored you are, you could change this to not using mocking but instead create a real file on disk. I realize you're just copying what was here before which is why I won't push it.

mraspaud (changes requested):

satpy/tests/reader_tests/test_olci_nc.py
```diff
@@ -241,36 +296,27 @@ def test_chl_nn(self, mocked_dataset):
...
+    import numpy as np
 
-        import numpy as np
+    from satpy.readers.olci_nc import DEFAULT_L1B_MASK_ITEMS, L1B_QUALITY_FLAGS, BitFlags
+
+    bits = np.array([1 << x for x in range(len(L1B_QUALITY_FLAGS))])
+
+    bflags = BitFlags(bits, flag_list=L1B_QUALITY_FLAGS)
+
+    mask = reduce(np.logical_or, [bflags[item] for item in DEFAULT_L1B_MASK_ITEMS])
```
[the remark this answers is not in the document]
yufeizhu600: I originally `added` testing bitFlags for l1b products into this test, then found it is sort of duplicated to use two sets of mask to test one function and decided to remove one. If you prefer, I can add back the WQSF part.

mraspaud (approved):
@yufeizhu600 I made an attempt at merging your changes with main, let's see how the tests go. If they pass I'm happy and I'll merge.

PR #2610, opened 2023-10-23
Remove legacy resampler code

djhoese:
Satpy depends on a new enough version of pyresample that all of theses imports are guaranteed.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 
 - [ ] Add your name to `AUTHORS.md` if not there already

PR #2690, opened 2023-12-14
Fix composites failing on non-aligned geolocation coordinates

### LINKED ISSUE #2668: FCI HRFI true_color unavailable even after native resampling if upper_right_corner is used
**Describe the bug**

When loading FCI FDHSI and HRFI data and loading the `true_color` composite, this composite is unavailable even after native resampling.

**To Reproduce**

```python
import hdf5plugin
from satpy import Scene
from glob import glob
fci_files = glob(f"/media/nas/x23352/MTG/FCI/2023/12/03/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-*-FD--CHK-BODY--DIS-NC4E_C_EUMT_20231203*_IDPFI_OPE_20231203*_20231203*_N_JLS_C_0070_*.nc")
sc.load(["true_color"], upper_right_corner="NE")
ls = sc.resample(resampler="native")
ls.save_datasets()
```

**Expected behavior**

I expect to get the true color composite.

**Actual results**

Output without debug messages:

```
[INFO: 2023-12-07 17:28:15 : satpy.readers.yaml_reader] Flipping Dataset vis_06 upsidedown.
[INFO: 2023-12-07 17:28:16 : satpy.readers.yaml_reader] Flipping Dataset vis_05 upsidedown.
[INFO: 2023-12-07 17:28:16 : satpy.readers.yaml_reader] Flipping Dataset vis_04 upsidedown.
[INFO: 2023-12-07 17:28:17 : satpy.readers.yaml_reader] Flipping Dataset vis_08 upsidedown.
[INFO: 2023-12-07 17:28:17 : satpy.readers.yaml_reader] Flipping Dataset unknown_name upsidedown.
[INFO: 2023-12-07 17:28:17 : satpy.readers.yaml_reader] Flipping Dataset unknown_name upsidedown.
[INFO: 2023-12-07 17:28:17 : satpy.readers.yaml_reader] Flipping Dataset unknown_name upsidedown.
[INFO: 2023-12-07 17:28:17 : satpy.readers.yaml_reader] Flipping Dataset unknown_name upsidedown.
[INFO: 2023-12-07 17:28:17 : satpy.modifiers.atmosphere] Removing Rayleigh scattering with atmosphere 'us-standard' and aerosol type 'rayleigh_only' for 'vis_06'
[INFO: 2023-12-07 17:28:17 : pyspectral.rayleigh] Atmosphere chosen: us-standard
[WARNING: 2023-12-07 17:28:17 : satpy.scene] The following datasets were not created and may require resampling to be generated: DataID(name='true_color')
[INFO: 2023-12-07 17:28:17 : satpy.modifiers.atmosphere] Removing Rayleigh scattering with atmosphere 'us-standard' and aerosol type 'rayleigh_only' for 'vis_05'
[INFO: 2023-12-07 17:28:17 : pyspectral.rayleigh] Atmosphere chosen: us-standard
[INFO: 2023-12-07 17:28:17 : satpy.composites.spectral] Applying NDVI-weighted hybrid-green correction with limits [0.15, 0.05] and strength 3.0.
[INFO: 2023-12-07 17:28:17 : satpy.modifiers.atmosphere] Removing Rayleigh scattering with atmosphere 'us-standard' and aerosol type 'rayleigh_only' for 'vis_04'
[INFO: 2023-12-07 17:28:17 : pyspectral.rayleigh] Atmosphere chosen: us-standard
[WARNING: 2023-12-07 17:28:17 : satpy.scene] The following datasets were not created and may require resampling to be generated: DataID(name='true_color')
  File "/data/gholl/checkouts/protocode/mwe/fci-true-color-native-resamplying.py", line 9, in <module>
    ls.save_datasets()
  File "/data/gholl/checkouts/satpy/satpy/scene.py", line 1271, in save_datasets
    raise RuntimeError("None of the requested datasets have been "
RuntimeError: None of the requested datasets have been generated or could not be loaded. Requested composite inputs may need to have matching dimensions (eg. through resampling).
```

The complete debug output is too long for a GitHub issue.

**Environment Info:**
 - OS: openSUSE 15.3
 - Satpy Version: main (v0.45.0-36-gfbe017c55)

djhoese:
See #2668 for detailed discussion. Basically #2610 switched from the internal satpy base resampler to the pyresample base resampler class. This overall isn't a problem except for the pyresample base resample has a shortcut where it won't process an input if it detects that the target geometry is the same as the source geometry. In those cases it returns the original DataArray with no additional modification. This logically/theoretically shouldn't be a problem, but it reveals a larger issue.

When y and x coordinates are generated from different sources they may have very small differences due to floating point precision issues. So example an x and y that are a subset/slice of a larger pair of x/y arrays may be different than an x/y generated from a sliced/subset AreaDefinition of the same original area definition (again, look at my comments in the referenced issue).

This PR updates one part of Satpy to make these operations work as overall the resampler change revealed an existing issue that was previously being hidden. The composite base class now performs a call to `xarray.align` to make sure that input arrays are compatible, but for `y` and `x` coordinates it will re-assign them to match the first input. The docstring in the new method explains any concerns about this.

 - [x] Closes #2668 
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -157,7 +157,7 @@ class CompositeBase:
                 elif o.get(k) is not None:
                     d[k] = o[k]
 
-    def match_data_arrays(self, data_arrays):
+    def match_data_arrays(self, data_arrays: Sequence[xr.DataArray]) -> list[xr.DataArray]:
         """Match data arrays so that they can be used together in a composite.
 
         For the purpose of this method, "can be used together" means:
@@ -189,46 +189,7 @@ class CompositeBase:
         new_arrays = list(unify_chunks(*new_arrays))
         return new_arrays
 
-    def drop_coordinates(self, data_arrays):
-        """Drop negligible non-dimensional coordinates.
-
-        Drops negligible coordinates if they do not correspond to any
-        dimension.  Negligible coordinates are defined in the
-        :attr:`NEGLIGIBLE_COORDS` module attribute.
-
-        Args:
-            data_arrays (List[arrays]): Arrays to be checked
-        """
-        new_arrays = []
-        for ds in data_arrays:
-            drop = [coord for coord in ds.coords
-                    if coord not in ds.dims and
-                    any([neglible in coord for neglible in NEGLIGIBLE_COORDS])]
-            if drop:
-                new_arrays.append(ds.drop_vars(drop))
-            else:
-                new_arrays.append(ds)
-
-        return new_arrays
-
-    def align_geo_coordinates(self, data_arrays: Sequence[xr.DataArray]) -> list[xr.DataArray]:
-        """Align DataArrays along geolocation coordinates.
-
-        See :func:`~xarray.align` for more information. This function uses
-        the "override" join method to essentially ignore differences between
-        coordinates. The :meth:`check_geolocation` should be called before
-        this to ensure that geolocation coordinates and "area" are compatible.
-        The :meth:`drop_coordinates` method should be called before this to
-        ensure that coordinates that are considered "negligible" when computing
-        composites do not affect alignment.
-
-        """
-        non_geo_coords = tuple(
-            coord_name for data_arr in data_arrays
-            for coord_name in data_arr.coords if coord_name not in ("x", "y"))
-        return xr.align(*data_arrays, join="override", exclude=non_geo_coords)
-
-    def check_geolocation(self, data_arrays):
+    def check_geolocation(self, data_arrays: Sequence[xr.DataArray]) -> None:
         """Check that the geolocations of the *data_arrays* are compatible.
 
         For the purpose of this method, "compatible" means:
@@ -238,7 +199,7 @@ class CompositeBase:
         - If all have an area, the areas should be all the same.
 
         Args:
-            data_arrays (List[arrays]): Arrays to be checked
+            data_arrays: Arrays to be checked
 
         Raises:
             :class:`IncompatibleAreas`:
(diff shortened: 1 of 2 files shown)


simonrp84:
Adding a note here to say that this fixes the problem I have been experiencing with cropping Himawari data:
```
scn = Scene(ifiles_l15, reader='ahi_hsd')
scn.load(['true_color_reproduction'])
scn2 = scn.crop(xy_bbox = bbox)
scn3 = scn2.resample(scn2.coarsest_area(), resampler='native', reduce_data=False)
scn3.save_datasets(base_dir=tod, writer='simple_image')
```

The above code fails on satpy's `main` but runs correctly with this PR.

@yufeizhu600 Maybe you can try this PR with the problems you were having with @strandgren's geocolor PR - guess the two PRs can be used together...somehow?

yufeizhu600:
> @yufeizhu600 Maybe you can try this PR with the problems you were having with @strandgren's geocolor PR - guess the two PRs can be used together...somehow?
@simonrp84 , you tagged wrong person. :). I think you meant to tag @yukaribbba.

yukaribbba:
@yufeizhu600 @simonrp84 Thanks. I'll have a try.

simonrp84:
Sorry for tagging the wrong person! Typed 'y' and went with githubs first suggestion 🙃 that's why you shouldn't code when tired,folks.

mraspaud (approved):
LGTM! I think it's good to merge as it is, I'm just wondering if we should have an new method called eg `unify_geo_coordinates` that groups `check_geolocation` and `align_geo_coordinates` into one place?

ameraner (changes requested):
Indeed the fix works, thanks for implementing a nice solution so quickly! 

Unfortunately however, the NDVI-corrected True Color still fails; I took a look and it's because the `NDVIHybridGreen` calls `match_data_arrays` only on a subset of projectables, and then mixes these with the other unmatched projectables. This is fixed by 

```diff
diff --git a/satpy/composites/spectral.py b/satpy/composites/spectral.py
index 448d7cb..ba417ac 100644
--- a/satpy/composites/spectral.py
+++ b/satpy/composites/spectral.py
@@ -160,9 +160,9 @@ class NDVIHybridGreen(SpectralBlender):
         LOG.info(f"Applying NDVI-weighted hybrid-green correction with limits [{self.limits[0]}, "
                  f"{self.limits[1]}] and strength {self.strength}.")

-        ndvi_input = self.match_data_arrays([projectables[1], projectables[2]])
+        ndvi_input = self.match_data_arrays(projectables)

-        ndvi = (ndvi_input[1] - ndvi_input[0]) / (ndvi_input[1] + ndvi_input[0])
+        ndvi = (ndvi_input[2] - ndvi_input[1]) / (ndvi_input[2] + ndvi_input[1])

```
if you want you can include this in this PR, or I can make a quick separate one.

djhoese:
> I'm just wondering if we should have an new method called eg unify_geo_coordinates that groups check_geolocation and align_geo_coordinates into one place?

@mraspaud I had this same thought, but with drop_coordinates and align_geo_coordinates. So...maybe that's a sign that they should stay the way they are?

PR #2557, opened 2023-08-24
Add baseline for GeoColor composite including FCI, AHI and ABI recipes

strandgren:
This PR adds the baseline for the GeoColor composite blend developed by [Miller et al. 2020](https://journals.ametsoc.org/view/journals/atot/37/3/JTECH-D-19-0134.1.xml). The GeoColor implementation in this PR follows the general approach described in Miller et al., but uses equivalent or similar Satpy functionalities, methods and composites where available. Some notable differences are:

- The green correction used for the true color layer is based on the hybrid green method used in Miller et al., but including the pixel-level scaling by NDVI available in Satpy.
- The true color layer for ABI is currently using the already available  `true_color` composite based in the green channel simulation in Satpy, with no use of the green channel simulation approach and hybrid green correction used in Miller et al. (2020).
- We do not account for cloud height information in the IR channel when doing the Rayleigh correction. Currently, there is also no blend with an uncorrected image to improve the imagery at high sunz angles (does not appear to be needed though). 
- The night-time layer used here is simply the NASA Black Marble geotiff image. In Miller et al. the use an elevation-enhanced nightscape layer with lights data from VIIRS. They also use an orange color for the city light to better replicate the appearance sodium lamps commonly used in the U.S.

The PR implements the missing compositors/layers needed for the GeoColor composite as well as recipes for FCI, AHI and ABI. Based on this, it should be easy to add recipes for other sensors like AMI, AGRI or even polar orbiting satellites in the future.

Below two early examples for FCI and AHI:

**FCI**
image

**AHI**
image

 - [x] Add land-sea mask to online repository and implement automatic download and caching
 - [x] Improve or decide on way forward for improved low-level cloud detection
 - [x] Merge main into this branch once #2554 and #2556 have been merged and adapt recipes accordingly.
 - [x] Make sure that new compositors are optimized and do not trigger any early unnecessary dask computations.
 - [ ] Add a description of GeoColor to https://github.com/pytroll/satpy/wiki/Built-in-composites
 - [x] Tests added 
 - [x] Fully documented

PR #2696, opened 2023-12-16
Add double alpha channel support and improve metadata behaviours for BackgroundCompositor

### LINKED ISSUE #2695: Improvements for BackgroundCompositor
## Feature Request

1. Make background fill-in an option, not a necessity
2. Add support for double alpha channel

I'll start a PR for this.

yukaribbba:
Current BackgroundCompositor assumes that only foreground has an alpha band. What if both of them do? Like the newly added #2557 , both LowCloudCompositor and HighCloudCompositor generate images with alpha channels. 

Here's an example of a `high_cloud` stacked on a `low_cloud`:
This is the result of current compositor. **Apparently it can't handle the background alpha correctly.**
image

Below is by this PR. **It recognizes both alpha channels, and builds a new alpha.**
image

 - [x] Closes #2695  
 - [x] Tests added 
 - [x] Fully documented


satpy/composites/__init__.py
```diff
@@ -1677,11 +1677,18 @@ def __call__(self, *args, **kwargs):
...
 
 class BackgroundCompositor(GenericCompositor):
-    """A compositor that overlays one composite on top of another."""
+    """A compositor that overlays one composite on top of another.
 
-    def __call__(self, projectables, *args, **kwargs):
+    Args:
+        bg_fill_in (bool): True means the compositor will fill the area where
+                           foreground is Nan with background.
+                           False means it will just leave the area blank.
```
djhoese, line 1685: "blank" -> "transparent"?

yukaribbba: > "blank" -> "transparent"?

Ok. I'm still working on this thing.

djhoese:
Could you describe the use case here? If the foreground has invalid pixels that you don't want to be filled in with background, why are you using the background compositor? Additionally, I believe there are "Filler" compositors that will fill in invalid data with a specific value. This could be done before calling the BackgroundCompositor.

satpy/composites/__init__.py
djhoese: Understood. I'm just trying to understand in case there are other ways of doing this that already exist.

yukaribbba:
Well this is a little more complicated than I expected. I have to organize my thoughts. I think maybe I need a third channel used for masking area, just like there's a "high_resolution_band" in SharpenedCompositor

yukaribbba:
pre-commit.ci autofix

yukaribbba:
@djhoese Ready for review :}

djhoese:
In your updated description, what is currently produced in `main` if both inputs have alphas? What is this PR choosing to do (sorry, it isn't clear to me from the picture)?

yukaribbba:
Here's an example of a high_cloud stacked on a low_cloud:
This is the result of current compositor. Apparently it can't handle the background alpha correctly.
image

Below is by this PR. It recognizes both alpha channels, and generates a new alpha channel.
image

pnuu:
Just a thought, but could the first use-case be implemented by combining the existing `Background` and `Masking` compositors?

yukaribbba:
> Just a thought, but could the first use-case be implemented by combining the existing `Background` and `Masking` compositors?

Well maybe, but that looks a little complicated and I want to do it in one step....

satpy/composites/__init__.py
```diff
@@ -1677,23 +1677,74 @@ def __call__(self, *args, **kwargs):
...
+    def __init__(self, name, mask_value=None, **kwargs):  # noqa: D417
+        """Collect custom configuration values.
+
+        Args:
+            mask_value (float / None): All the pixels on the stacked image where the values
+                                       of the third dataset that are equal to this will be
+                                       masked out.
+                                       If not set by the user, it will be ``np.nan``.
+                                       This argument could be helpful when you try to use a
+                                       local image(StaticImageCompositor) for masking.
```
djhoese, line 1712: Reworded this, but also types should be separated by I think a vertical bar `|` or maybe a comma in google style docstrings. Let's see what the readthedocs job says.

```suggestion
            mask_value (float | None): Value of the third dataset used to generate
                the mask for the stacked image. Defaults to ``np.nan``. This is
                useful when wanting additional masking for a static image
                (e.g. :class:`StaticImageCompositor`).
```

satpy/composites/__init__.py
```diff
@@ -1677,23 +1677,74 @@ def __call__(self, *args, **kwargs):
...
-        projectables = self.match_data_arrays(projectables)
+        optional_datasets = [] if optional_datasets is None else optional_datasets
+        projectables = self.match_data_arrays(projectables + optional_datasets)
+
         # Get enhanced datasets
         foreground = enhance2dataset(projectables[0], convert_p=True)
         background = enhance2dataset(projectables[1], convert_p=True)
+        before_bg_mode = background.attrs["mode"]
+
+        mask_dataset = projectables[2] if not optional_datasets == [] else None
```
djhoese, line 1729: Check your condition here. `not optional_datasets == []` is confusing. You could just do `mask_dataset = projectables[2] if len(projecables) >= 3 else None`.

djhoese:
So is another way of summarizing this PR that use case 1 is a new feature and use case 2 (the two alphas) is a bug fix?

I think overall I agree with @pnuu. There is something to be said about individual components having a single purpose (the unix philosophy). Especially when that purpose (masking) seems to be covered by an existing component. Even if it requires using two or more compositors this could be done with "inline" compositors in a single definition if necessary.

yukaribbba:
Understood. So I decide to remove the first part, keeping it for myself. If someone wants that in the future I'll re-add it. I'll still commit one with fixes you mentioned as a backup here.

yukaribbba:
The commit 707334f is the last one with masking function.

yukaribbba:
There's a problem that happens both on `main` and this PR. If we stack L/L or RGB/RGB together(well that's boring), we'll get an error.

My test script:
```
files = find_files_and_readers(base_dir='C:/Users/45107/Downloads/Sat/Geo/gk2a_20231215_1600_FLDK', reader='ami_l1b')

area = AreaDefinition(area_id='world', description='world latlon', proj_id='latlon',
                      projection='+proj=latlon +datum=WGS84 +ellps=WGS84',
                      width=4096, height=2048, area_extent=(-180, -90, 180, 90))

scn4 = Scene(filenames=files)
product = 'blackmarble_with_blackmarble' # RGB/RGB
scn4.load([product])
scn4 = scn4.resample(area, resampler='nearest') # Whether resampled or not, error raises
scn4.show(product)
```
```
  static_night:
    filename: C:/Users/45107/Downloads/Sat/Geo/satpy_night.tif
    standard_name: static_image_spec

  blackmarble_with_blackmarble:
    compositor: !!python/name:satpy.composites.BackgroundCompositor
    prerequisites:
      - name: static_night
      - name: static_night
```

DEBUG
```
[DEBUG: 2023-12-20 10:54:54 : satpy.readers.yaml_reader] Reading ('C:\\Users\\45107\\miniforge3\\Lib\\site-packages\\satpy\\etc\\readers\\ami_l1b.yaml',)
[DEBUG: 2023-12-20 10:54:54 : satpy.readers.yaml_reader] Reading ('C:\\Users\\45107\\miniforge3\\Lib\\site-packages\\satpy\\etc\\readers\\ami_l1b.yaml',)
[DEBUG: 2023-12-20 10:54:54 : satpy.readers.yaml_reader] Assigning to ami_l1b: ['C:/Users/45107/Downloads/Sat/Geo/gk2a_20231215_1600_FLDK\\gk2a_ami_le1b_ir105_fd020ge_202312151600.nc', 'C:/Users/45107/Downloads/Sat/Geo/gk2a_20231215_1600_FLDK\\gk2a_ami_le1b_sw038_fd020ge_202312151600.nc']
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] PROJ data files are available at built-in paths.
C:\Users\45107\miniforge3\Lib\site-packages\xarray\core\dataset.py:278: UserWarning: The specified chunks separate the stored chunks along dimension "dim_image_y" starting at index 4096. This could degrade performance. Instead, consider rechunking after loading.
  warnings.warn(
C:\Users\45107\miniforge3\Lib\site-packages\xarray\core\dataset.py:278: UserWarning: The specified chunks separate the stored chunks along dimension "dim_image_x" starting at index 4096. This could degrade performance. Instead, consider rechunking after loading.
  warnings.warn(
[DEBUG: 2023-12-20 10:54:54 : satpy.composites.config_loader] Looking for composites config file ami.yaml
[DEBUG: 2023-12-20 10:54:54 : satpy.composites.config_loader] Looking for composites config file visir.yaml
[DEBUG: 2023-12-20 10:54:54 : pyorbital.tlefile] Path to the Pyorbital configuration (where e.g. platforms.txt is found): C:\Users\45107\miniforge3\Lib\site-packages\pyorbital\etc
[DEBUG: 2023-12-20 10:54:54 : satpy.readers.yaml_reader] Reading ('C:\\Users\\45107\\miniforge3\\Lib\\site-packages\\satpy\\etc\\readers\\generic_image.yaml',)
[DEBUG: 2023-12-20 10:54:54 : satpy.readers.yaml_reader] Assigning to generic_image: [<FSFile "C:/Users/45107/Downloads/Sat/Geo/satpy_night.tif">]
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Entering env context: <rasterio.env.Env object at 0x000001E29B666390>
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Starting outermost env
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] No GDAL environment exists
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] New GDAL environment <rasterio._env.GDALEnv object at 0x000001E29D5F69E0> created
[DEBUG: 2023-12-20 10:54:54 : rasterio._filepath] Installing FilePath filesystem handler plugin...
[DEBUG: 2023-12-20 10:54:54 : rasterio._env] GDAL_DATA found in environment.
[DEBUG: 2023-12-20 10:54:54 : rasterio._env] PROJ data files are available at built-in paths.
[DEBUG: 2023-12-20 10:54:54 : rasterio._env] Started GDALEnv: self=<rasterio._env.GDALEnv object at 0x000001E29D5F69E0>.
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Entered env context: <rasterio.env.Env object at 0x000001E29B666390>
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Sharing flag: 0
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Nodata success: 0, Nodata value: 0.000000
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Nodata success: 0, Nodata value: 0.000000
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Nodata success: 0, Nodata value: 0.000000
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Dataset <open DatasetReader name='C:/Users/45107/Downloads/Sat/Geo/satpy_night.tif' mode='r'> is started.
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Exiting env context: <rasterio.env.Env object at 0x000001E29B666390>
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Cleared existing <rasterio._env.GDALEnv object at 0x000001E29D5F69E0> options
[DEBUG: 2023-12-20 10:54:54 : rasterio._env] Stopped GDALEnv <rasterio._env.GDALEnv object at 0x000001E29D5F69E0>.
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Exiting outermost env
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Exited env context: <rasterio.env.Env object at 0x000001E29B666390>
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Entering env context: <rasterio.env.Env object at 0x000001E29D1C8B90>
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Starting outermost env
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] No GDAL environment exists
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] New GDAL environment <rasterio._env.GDALEnv object at 0x000001E29D5F7B20> created
[DEBUG: 2023-12-20 10:54:54 : rasterio._env] GDAL_DATA found in environment.
[DEBUG: 2023-12-20 10:54:54 : rasterio._env] PROJ data files are available at built-in paths.
[DEBUG: 2023-12-20 10:54:54 : rasterio._env] Started GDALEnv: self=<rasterio._env.GDALEnv object at 0x000001E29D5F7B20>.
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Entered env context: <rasterio.env.Env object at 0x000001E29D1C8B90>
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Sharing flag: 0
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Nodata success: 0, Nodata value: 0.000000
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Nodata success: 0, Nodata value: 0.000000
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Nodata success: 0, Nodata value: 0.000000
[DEBUG: 2023-12-20 10:54:54 : rasterio._base] Dataset <open DatasetReader name='C:/Users/45107/Downloads/Sat/Geo/satpy_night.tif' mode='r'> is started.
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Exiting env context: <rasterio.env.Env object at 0x000001E29D1C8B90>
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Cleared existing <rasterio._env.GDALEnv object at 0x000001E29D5F7B20> options
[DEBUG: 2023-12-20 10:54:54 : rasterio._env] Stopped GDALEnv <rasterio._env.GDALEnv object at 0x000001E29D5F7B20>.
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Exiting outermost env
[DEBUG: 2023-12-20 10:54:54 : rasterio.env] Exited env context: <rasterio.env.Env object at 0x000001E29D1C8B90>
[DEBUG: 2023-12-20 10:54:54 : satpy.composites.config_loader] Looking for composites config file images.yaml
[DEBUG: 2023-12-20 10:54:54 : satpy.composites.config_loader] No composite config found called images.yaml
[DEBUG: 2023-12-20 10:54:54 : satpy.readers.generic_image] Reading 'image.'
[WARNING: 2023-12-20 10:54:54 : satpy.readers.generic_image] cannot convert float NaN to integer
[DEBUG: 2023-12-20 10:54:54 : satpy.writers] Adding enhancement configuration from file: C:\Users\45107\miniforge3\Lib\site-packages\satpy\etc\enhancements\generic.yaml
[DEBUG: 2023-12-20 10:54:55 : satpy.writers] Adding enhancement configuration from file: D:\satpy_config\enhancements\generic.yaml
[DEBUG: 2023-12-20 10:54:55 : satpy.writers] Data for DataID(name='static_night') will be enhanced with options:
	[{'name': 'stretch', 'method': <function stretch at 0x000001E29D632200>, 'kwargs': {'stretch': 'crude', 'min_stretch': [0, 0, 0], 'max_stretch': [255, 255, 255]}}]
[DEBUG: 2023-12-20 10:54:55 : trollimage.xrimage] Applying stretch crude with parameters {'min_stretch': [0, 0, 0], 'max_stretch': [255, 255, 255]}
[DEBUG: 2023-12-20 10:54:55 : satpy.writers] Adding enhancement configuration from file: C:\Users\45107\miniforge3\Lib\site-packages\satpy\etc\enhancements\generic.yaml
[DEBUG: 2023-12-20 10:54:55 : satpy.writers] Adding enhancement configuration from file: D:\satpy_config\enhancements\generic.yaml
[DEBUG: 2023-12-20 10:54:55 : satpy.writers] Data for DataID(name='static_night') will be enhanced with options:
	[{'name': 'stretch', 'method': <function stretch at 0x000001E29D632200>, 'kwargs': {'stretch': 'crude', 'min_stretch': [0, 0, 0], 'max_stretch': [255, 255, 255]}}]
[DEBUG: 2023-12-20 10:54:55 : trollimage.xrimage] Applying stretch crude with parameters {'min_stretch': [0, 0, 0], 'max_stretch': [255, 255, 255]}
[DEBUG: 2023-12-20 10:54:55 : satpy.scene] Resampling DataID(name='static_night')
[INFO: 2023-12-20 10:54:55 : satpy.scene] Not reducing data before resampling.
[DEBUG: 2023-12-20 10:54:55 : satpy.resample] Computing kd-tree parameters
[DEBUG: 2023-12-20 10:54:55 : satpy.resample] Resampling where-0c2efa9774b75cb3d825d9cc0911c8b2
  File "C:\Users\45107\miniforge3\Lib\site-packages\satpy\dataset\data_dict.py", line 169, in __getitem__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'blackmarble_with_blackmarble'

  File "C:\Users\45107\PycharmProjects\Satellite\bcdef.py", line 16, in <module>
    scn4.show(product)
  File "C:\Users\45107\miniforge3\Lib\site-packages\satpy\scene.py", line 1010, in show
                             ~~~~^^^^^^^^^^^^
  File "C:\Users\45107\miniforge3\Lib\site-packages\satpy\scene.py", line 825, in __getitem__
    return self._datasets[key]
           ~~~~~~~~~~~~~~^^^^^
  File "C:\Users\45107\miniforge3\Lib\site-packages\satpy\dataset\data_dict.py", line 171, in __getitem__
    key = self.get_key(item)
          ^^^^^^^^^^^^^^^^^^
  File "C:\Users\45107\miniforge3\Lib\site-packages\satpy\dataset\data_dict.py", line 158, in get_key
    return get_key(match_key, self.keys(), num_results=num_results,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\45107\miniforge3\Lib\site-packages\satpy\dataset\data_dict.py", line 107, in get_key
KeyError: "No dataset matching 'DataQuery(name='blackmarble_with_blackmarble')' found"
```

Do you think it's related to #2668 and #2690 ?

djhoese:
> scn4 = scn4.resample

First, don't reassign the Scene variable. Instead use a new name to avoid Python garbage collecting the initial `scn4`.

The error is a KeyError because the composite never successfully generated a result. From the log output it isn't clear to me what happened. Some exception is being silenced/hidden.

yukaribbba:
Ok but this also happens when there's no resampling.

djhoese:
Right, but the compositor is called in both cases. I was just pointing out that it is not a good idea to reassign the Scene. I realize this isn't the primary issue, but just a heads up to avoid it.

My only other guess is that the dependency tree is getting confused because you're depending on a single dataset twice in your composite but I don't *think* that should be a problem.

You could put a huge try/except around everything after `match_data_arrays` in the compositor and see what error(s) you get.

yukaribbba:
Didn't get anything from try/except. But you're right, this may not be a problem since we rarely do that...

djhoese:
If you put print statements in the compositor, does it get called? Does it get all the way to returning a DataArray?

yukaribbba:
print `data`
```
[<xarray.DataArray (y: 2048, x: 4096)>
dask.array<where, shape=(2048, 4096), dtype=float32, chunksize=(2048, 4096), chunktype=numpy.ndarray>
Coordinates:
  * x            (x) float64 -180.0 -179.9 -179.8 -179.7 ... 179.8 179.9 180.0
  * y            (y) float64 89.96 89.87 89.78 89.69 ... -89.78 -89.87 -89.96
    spatial_ref  int32 0
    crs          object GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",E...
    bands        <U1 'A', <xarray.DataArray (y: 2048, x: 4096)>
dask.array<where, shape=(2048, 4096), dtype=float32, chunksize=(2048, 4096), chunktype=numpy.ndarray>
Coordinates:
  * x            (x) float64 -180.0 -179.9 -179.8 -179.7 ... 179.8 179.9 180.0
  * y            (y) float64 89.96 89.87 89.78 89.69 ... -89.78 -89.87 -89.96
    spatial_ref  int32 0
    crs          object GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",E...
    bands        <U1 'A', <xarray.DataArray (y: 2048, x: 4096)>
dask.array<where, shape=(2048, 4096), dtype=float32, chunksize=(2048, 4096), chunktype=numpy.ndarray>
Coordinates:
  * x            (x) float64 -180.0 -179.9 -179.8 -179.7 ... 179.8 179.9 180.0
  * y            (y) float64 89.96 89.87 89.78 89.69 ... -89.78 -89.87 -89.96
    spatial_ref  int32 0
    crs          object GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",E...
    bands        <U1 'A']
```

print `res`
```
<xarray.DataArray 'where-7cb3c772602cf5919c594884bafe9c7a' (bands: 3, y: 2048,
                                                            x: 4096)>
dask.array<where, shape=(3, 2048, 4096), dtype=float32, chunksize=(1, 2048, 4096), chunktype=numpy.ndarray>
Coordinates:
  * x            (x) float64 -180.0 -179.9 -179.8 -179.7 ... 179.8 179.9 180.0
  * y            (y) float64 89.96 89.87 89.78 89.69 ... -89.78 -89.87 -89.96
    spatial_ref  int32 0
    crs          object GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",E...
  * bands        (bands) <U1 'R' 'G' 'B'
Attributes:
    wavelength:              None
    name:                    blackmarble_with_blackmarble
    _satpy_id:               DataID(name='blackmarble_with_blackmarble')
    prerequisites:           [DataQuery(name='static_night'), DataQuery(name=...
    optional_prerequisites:  []
    sensor:                  None
    mode:                    RGB
```

yukaribbba:
Looks like the compositor did its job correctly. The error is somewhere else.

djhoese:
Does this fail to load if you use `main` instead of this branch?

yukaribbba:
Yep.

print `data`
```
[<xarray.DataArray (y: 2048, x: 4096)>
dask.array<getitem, shape=(2048, 4096), dtype=float32, chunksize=(2048, 4096), chunktype=numpy.ndarray>
Coordinates:
  * x            (x) float64 -180.0 -179.9 -179.8 -179.7 ... 179.8 179.9 180.0
  * y            (y) float64 89.96 89.87 89.78 89.69 ... -89.78 -89.87 -89.96
    spatial_ref  int32 0
    bands        <U1 'R'
    crs          object GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",E..., <xarray.DataArray (y: 2048, x: 4096)>
dask.array<getitem, shape=(2048, 4096), dtype=float32, chunksize=(2048, 4096), chunktype=numpy.ndarray>
Coordinates:
  * x            (x) float64 -180.0 -179.9 -179.8 -179.7 ... 179.8 179.9 180.0
  * y            (y) float64 89.96 89.87 89.78 89.69 ... -89.78 -89.87 -89.96
    spatial_ref  int32 0
    bands        <U1 'G'
    crs          object GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",E..., <xarray.DataArray (y: 2048, x: 4096)>
dask.array<getitem, shape=(2048, 4096), dtype=float32, chunksize=(2048, 4096), chunktype=numpy.ndarray>
Coordinates:
  * x            (x) float64 -180.0 -179.9 -179.8 -179.7 ... 179.8 179.9 180.0
  * y            (y) float64 89.96 89.87 89.78 89.69 ... -89.78 -89.87 -89.96
    spatial_ref  int32 0
    bands        <U1 'B'
    crs          object GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",E...]
```

print `res`
```
<xarray.DataArray 'where-a7023d79eee34bda86d5737ee7a3d36e' (bands: 3, y: 2048,
                                                            x: 4096)>
dask.array<where, shape=(3, 2048, 4096), dtype=float32, chunksize=(1, 2048, 4096), chunktype=numpy.ndarray>
Coordinates:
  * x            (x) float64 -180.0 -179.9 -179.8 -179.7 ... 179.8 179.9 180.0
  * y            (y) float64 89.96 89.87 89.78 89.69 ... -89.78 -89.87 -89.96
    spatial_ref  int32 0
    crs          object GEOGCRS["WGS 84",DATUM["World Geodetic System 1984",E...
  * bands        (bands) <U1 'R' 'G' 'B'
Attributes:
    wavelength:              None
    name:                    blackmarble_with_blackmarble
    _satpy_id:               DataID(name='blackmarble_with_blackmarble')
    prerequisites:           [DataQuery(name='static_night'), DataQuery(name=...
    optional_prerequisites:  []
    sensor:                  None
    mode:                    RGB
```

yukaribbba:
Still keyerror

yukaribbba:
> My only other guess is that the dependency tree is getting confused because you're depending on a single dataset twice in your composite

But we can do this without any problems.

```
  IR105_RGB:
    prerequisites:
      - name: IR105
      - name: IR105
      - name: IR105
```

djhoese:
In the compositor code can you take the result and set a value for `.attrs["sensor"] = "viirs"` and `.attrs["start_time"] = datetime.utcnow()` and see if it behaves better? Oh you could also try adding a `standard_name` to the composite YAML for the background compositor. Technically for an actual output image you'd want a corresponding enhancement for that standard_name, but if you're not concerned with what the final output looks like then that doesn't matter.

yukaribbba:
Sorry it's late at night in my time and I have to go to bed. I'll try what you said tomorrow.

yukaribbba:
> In the compositor code can you take the result and set a value for `.attrs["sensor"] = "viirs"` and `.attrs["start_time"] = datetime.utcnow()` and see if it behaves better?

Nothing changed...
```
        attrs = self._combine_metadata_with_mode_and_sensor(foreground, background)
        data = self._get_merged_image_data(foreground, background)
        res = super(BackgroundCompositor, self).__call__(data, **kwargs)
        res.attrs.update(attrs)
        res.attrs["sensor"] = "viirs"
        res.attrs["start_time"] = datetime.datetime.utcnow()
        print(res.attrs)
        return res
```

```
{'wavelength': None, 'name': 'static_night', '_satpy_id': DataID(name='static_night'), 'prerequisites': [], 'optional_prerequisites': [], 'sensor': 'viirs', 'mode': 'RGB', 'standard_name': 'static_image_spec', 'area': Area ID: None
Description: None
Projection ID: WGS 84
Projection: {'datum': 'WGS84', 'no_defs': 'None', 'proj': 'longlat', 'type': 'crs'}
Number of columns: 4096
Number of rows: 2048
Area extent: (-180.0, -90.0, 180.0, 90.0), 'ancillary_variables': [], 'add_offset': 0.0, 'file_type': 'graphic', 'end_time': datetime.datetime(2023, 12, 21, 2, 42, 55, 850157), 'AREA_OR_POINT': 'Area', 'reader': 'generic_image', 'scale_factor': 1.0, 'start_time': datetime.datetime(2023, 12, 21, 2, 42, 56, 411302), 'nodatavals': (nan, nan, nan)}
```

So the attrs has been successfully set but KeyError still...Just like I thought before this could be something outside the the whole compositor base.

yukaribbba:
I tried all these just to know if there're any serious problems behind it that could affect other things.

djhoese:
Obviously the fact that you said it happens in `main` suggests something is wrong beyond your changes in this PR. My suggestions for changes in start_time/sensor and adding a standard_name were meant to hopefully reveal (to me at least) where the problem might lie. I might have to try this myself to further debug it, but I'm not sure I'll have time for the next couple weeks. We really need to track down what happens to the DataArray after it is returned by the compositor. Oh...I wonder if the Scene is deleting the composite because it thinks it isn't needed. You could try passing `unload=False` to the `Scene.load` call and see if that changes anything.

yukaribbba:
Unfortunately `unload=False` doesn't help...

yukaribbba:
In the Scene.show
```
    def show(self, dataset_id, overlay=None):
        """Show the *dataset* on screen as an image.

        Show dataset on screen as an image, possibly with an overlay.

        Args:
            dataset_id (DataID, DataQuery or str):
                Either a DataID, a DataQuery or a string, that refers to a data
                array that has been previously loaded using Scene.load.
            overlay (dict, optional):
                Add an overlay before showing the image.  The keys/values for
                this dictionary are as the arguments for
                :meth:`~satpy.writers.add_overlay`.  The dictionary should
                contain at least the key ``"coast_dir"``, which should refer
                to a top-level directory containing shapefiles.  See the
                pycoast_ package documentation for coastline shapefile
                installation instructions.

        .. _pycoast: https://pycoast.readthedocs.io/

        """
        from satpy.utils import in_ipynb
        try:
        except Exception as e:
            print(e)
        if not in_ipynb():
            img.show()
        return img
```

`'NoneType' object has no attribute 'squeeze'`

djhoese:
You said IR105 didn't have a problem even when it was configured in the YAML, right? It was only when the static compositor was involved I thought?

yukaribbba:
I'll summarize.

**Succeed:**
```
  IR105_with_SW038:
    compositor: !!python/name:satpy.composites.BackgroundCompositor
    prerequisites:
      - name: IR105
      - name: SW038
```

**Fail:**
```
  IR105_with_IR105:
    compositor: !!python/name:satpy.composites.BackgroundCompositor
    prerequisites:
      - name: IR105
      - name: IR105
```

```
  black_marble_with_black_marble:
    compositor: !!python/name:satpy.composites.BackgroundCompositor
    prerequisites:
      - name: static_night
      - name: static_night
```

Note they're all in YAML mode, not the test way above.

yukaribbba:
Perhaps one of the dataset got filtered/dropped in Scene, cause they have the same name?...

djhoese:
Ok I was able to reproduce this with ABI. So the problem is that when all the "prerequisites" are combined (in my case C01 and C01) the metadata is also combined. Since the inputs are exactly the same all the metadata is preserved. If they had been different then anything not equal between them would have been removed. This includes the special `_satpy_id` attribute and all of its subcomponents (name, wavelength, etc). On `main` at least the code in the compositor is structured in a way that the merged/combined metadata overwrites anything from the generated data or the compositor which is not great. Put another way, the resulting composite takes over the identity of the input DataArray for a bit (if you're curious I can point you to the Scene code where this is breaking). So in my case my final Scene had C01 and not the composite (`"test"`) that I was creating. I'm not sure if the C01 in the Scene was the composite or the original dataset from the reader. Let me see if there is a quick fix in the compositor, but I think overall this is an unsupported use case.

yukaribbba:
Well that's a little bit complicated for me...I'll try to understand if you can tell me the broken part.
Anyway your fix also works on this PR. We can finally move on! Thanks!

yukaribbba:
BTW, in `test_multiple_sensors`, `assert res.attrs["sensor"] == {"abi", "glm"}` is now replaced by `assert res.attrs["sensor"] is None`. Do you think this is acceptable / what we want?

yukaribbba:
BTW, in `test_multiple_sensors`, `assert res.attrs["sensor"] == {"abi", "glm"}` is now replaced by `assert res.attrs["sensor"] is None`. Do you think this is acceptable / what we want?

djhoese:
Oh nope. That's not good. Hhhmmm why would `sensor` be in the compositor definition in a way that would overwrite the combined metadata sensors?

yukaribbba:
I have no idea...Do you think we really need sensor in the metadata? Usually BackgroundCompositor in the last few steps of a composite. Before it's called, we may already finish atmosphere correction/nir reflectance or such jobs which requires sensor name.

djhoese:
Enhancements need to know the sensor to find the right config and there is no telling what users want to do with the metadata after. I know that for my own applications I use the sensor names in the output filename.

yukaribbba:
Whoops 😬

yukaribbba:
**Okay sensor is back** ✅
`{'area': 'foo', 'sensor': {'abi', 'glm'}, 'wavelength': None, 'name': 'name', 'prerequisites': [], 'optional_prerequisites': [], 'mode': 'L'}`
**IR105_with_SW038** ✅
**IR105_with_IR105** ✅
**HighCloud_with_LowCloud** ✅
**HighLowCloud_with_blackmarble** ✅

How can I describe you fix?

djhoese:
Sorry, I'm not sure what you mean by describing the fix?

yukaribbba:
Yeah I don't know if I need a description for the metadata fix at the beginning. My part is already there. If you think it's just a minor fix maybe we don't need to mention it.

yukaribbba:
For the first floor of this post……

satpy/composites/__init__.py
```diff
@@ -1714,24 +1726,51 @@ def _combine_metadata_with_mode_and_sensor(self,
...
+
+        alpha_fore = _get_alpha(foreground)
+        alpha_back = _get_alpha(background)
+        new_alpha = alpha_fore + alpha_back * (1 - alpha_fore)
+
+        data = []
+
+        # Unless background has an initial alpha band, there will be no alpha band in the output image
+        # Let the writer decide
+        output_mode = background.mode if initial_bg_alpha else background.mode.replace("A", "")
```
djhoese, line 1756: Is this `background.attrs["mode"]`? Please access DataArray attributes from `.attrs`.

satpy/composites/__init__.py
```diff
@@ -1682,20 +1682,32 @@ class BackgroundCompositor(GenericCompositor):
...
+
         # Get enhanced datasets
         foreground = enhance2dataset(projectables[0], convert_p=True)
         background = enhance2dataset(projectables[1], convert_p=True)
+        before_bg_mode = background.attrs["mode"]
+
         # Adjust bands so that they match
         # L/RGB -> RGB/RGB
         # LA/RGB -> RGBA/RGBA
         # RGB/RGBA -> RGBA/RGBA
```
djhoese, line 1694: Are these cases still the same after your changes? Your tests suggest that `LA/RGB` will remain `RGB` instead of what is specified here (the second line) of `RGBA`.

satpy/composites/__init__.py
```diff
@@ -1714,24 +1726,51 @@ def _combine_metadata_with_mode_and_sensor(self,
...
+            # If the dataset contains an alpha channel, just use it
+            # If not, we still need one. So build it and fill it with 1
+            if "A" in dataset.attrs["mode"]:
+                alpha = dataset.sel(bands="A")
+            else:
+                first_band = dataset.isel(bands=0)
+                alpha = xr.full_like(first_band, 1)
+                alpha["bands"] = "A"
+
+            # There could be Nans in the alpha, especially through 'add_bands'
```
djhoese, line 1742: In what cases does `add_bands` add an alpha with NaNs as described in this comment? Skimming the `add_bands` code I only see a `da.ones` call to create a fully opaque alpha channel.

satpy/composites/__init__.py
```diff
@@ -1714,24 +1726,51 @@ def _combine_metadata_with_mode_and_sensor(self,
...
+        # The best way is through the new alpha
+        new_alpha_nan = xr.where(alpha_fore + alpha_back == 0, np.nan, new_alpha) if "A" not in output_mode \
+            else new_alpha
+
+        for band in output_mode:
+            fg_band = foreground.sel(bands=band) if band != "A" else new_alpha
+            bg_band = background.sel(bands=band) if band != "A" else new_alpha
+
+            chan = (fg_band * alpha_fore +
+                    bg_band * alpha_back * (1 - alpha_fore)) / new_alpha_nan if band != "A" else new_alpha
```
djhoese, line 1768: What is the `/ new_alpha_nan` doing here?

satpy/composites/__init__.py
```diff
@@ -1714,24 +1726,51 @@ def _combine_metadata_with_mode_and_sensor(self,
...
+        data = []
+
+        # Unless background has an initial alpha band, there will be no alpha band in the output image
+        # Let the writer decide
+        output_mode = background.attrs["mode"] if initial_bg_alpha else background.attrs["mode"].replace("A", "")
+
+        # If we let the writer decide alpha band, we must fill the transparent areas in the image with np.nan first
+        # The best way is through the new alpha
+        new_alpha_nan = xr.where(alpha_fore + alpha_back == 0, np.nan, new_alpha) if "A" not in output_mode \
+            else new_alpha
```
djhoese, line 1761: Could some of this logic and maybe most of `_get_alpha` be removed if we were more clever with the `bands` argument that is passed to `add_bands` earlier in the compositor? For example, if instead of passing `backgrounds["bands"]` to `add_bands` we figured out what the final `output_mode` was going to be and passed that to `add_bands` instead, then we wouldn't have to do some of this `xr.where` call to create a new alpha band...I hope.

What I'm really saying is that it is really hard for me to follow what this method is doing. That may just be me or it may be the complicated logic being implemented here or it might be that there is room for improvement in the code.

djhoese (commented):
Thanks for trying to tackle this. I made some inline comments. I'm really not sure I understand all of the code so let me know in my comments below what I'm missing.

satpy/composites/__init__.py
```diff
@@ -1682,20 +1682,32 @@ class BackgroundCompositor(GenericCompositor):
...
         # LA/RGB -> RGBA/RGBA
         # RGB/RGBA -> RGBA/RGBA
         foreground = add_bands(foreground, background["bands"])
         background = add_bands(background, foreground["bands"])
+        after_bg_mode = background.attrs["mode"]
+
+        # It's important to judge whether the alpha band of background is initially generated, e.g. by CloudCompositor
+        # Or it's just added through 'add_bands'
+        # The result will be used to decide the output image mode
+        initial_bg_alpha = True if "A" in before_bg_mode and "A" in after_bg_mode else False
```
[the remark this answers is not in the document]
yukaribbba: Right, no need for `after_bg_mode`.

satpy/composites/__init__.py
yukaribbba: I'm not sure whether these comments are about the whole compositor or just `add_bands` it self...

satpy/composites/__init__.py
yukaribbba: If this line got disabled I'll get a weird result.
image

satpy/composites/__init__.py
yukaribbba: Like the comments say, when there's no initial alpha band in the background then we won't pass the alpha band to the writer. Instead we leave the writer for decision. Before that, we must fill all the transparent areas with nan. The best way to do this is letting the `new_alpha` get involved in `chan = (fg_band * alpha_fore + bg_band * alpha_back * (1 - alpha_fore)) / new_alpha`  ([https://en.wikipedia.org/wiki/Alpha_compositing](url)) since `anything / nan = nan`. However, I don't want to change the `new_alpha` so I make a duplicate here `new_alpha_nan`

satpy/composites/__init__.py
yukaribbba: I think `add_bands` requires both the band data and the mode attribute. I don't know how to do it by just passing mode.

satpy/composites/__init__.py
djhoese: Ok. I would think other alpha-based composites (like some of the cloud ones you've talked about in this PR) are producing NaNs in the alpha. `add_bands` doesn't seem to be doing that.

satpy/composites/__init__.py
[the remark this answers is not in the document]
djhoese: Ok good. We can resolve this conversation when the code is changed.

satpy/composites/__init__.py
djhoese: I do think it would be possible to move some of this up, but given how much I see this logic mirrors the wikipedia article then maybe that isn't for the best.

satpy/composites/__init__.py
djhoese: Thanks for the wikipedia article. I was starting to wonder if this was equivalent to the pre-multiplied alpha, but the article says it is a straight alpha so OK.

Would it be possible to separate out the alpha channel handling (the `band == "A"` case) to be done after the loop? Because in that case `chan` just becomes `new_alpha`, right? I'm not sure I understand what the below `xr.where(chan.isnull(), ...` does in that case though.

Next question, how much extra work is this compositor now doing in the case where no alphas are involved? In the old code there was the `else:` block that did a simple `xr.where`. Is it worth maybe have a special branch of `if "A" not in foreground_mode and "A" not in background_mode: return self._simple_overlay()` or something?

satpy/composites/__init__.py
yukaribbba: I'll change the comment there.

satpy/composites/__init__.py
yukaribbba: 1. Agree. Leaving the alpha outside the loop will make it more tidy.
2. `chan = xr.where(chan.isnull(), bg_band * alpha_back, chan)` is to fill the areas where `foreground` is completely transparent with `background`. Without it you may get an image like this (HighLowCloud_with_background):
image
Actually, it is:
image
3. Well I use this loop just to avoid `if/else` block and to make things more general. But if you want to simplify the workflow for no alpha cases I can bring it back(could be a little huge) or make another method for this just like you said.

djhoese: For 2, what about the data in this case is making it NaN? Why aren't other areas over land NaN in the cloud image? I guess another way for me to think about this is that the alpha band would have to be set to 0 (for foreground and background) where ever the band data is NaN, but also the band data would need to be set to 0 where ever the alpha is 0. Ok ok, too many cases to work around.

For 3, although the code got longer I do think overall it simplifies the understanding.

I think my biggest confusion with this code overall is that the NaN handling seems relatively magical. I mean, it isn't explicit and comes from some dividing and 0 or NaN filling at various parts. I might be the only one that is having trouble with this though so we should wait to see what @mraspaud and @pnuu say when they get back from their holidays.

yukaribbba: > but also the band data would need to be set to 0 where ever the alpha is 0

It could be a little risky for float32 output...So in my understanding, the band data should to be set to NaN. And that is exactly what `new_alpha_nan` does

yukaribbba: I understand your concern. It's kinda tricky here so we may wait for their thoughts.

djhoese: @mraspaud @pnuu I'd still like your comments here if you have time.

satpy/composites/__init__.py
```diff
@@ -1707,24 +1720,68 @@ def _combine_metadata_with_mode_and_sensor(self,
...
+            if "A" in dataset.attrs["mode"]:
+                alpha = dataset.sel(bands="A")
+            else:
+                first_band = dataset.isel(bands=0)
+                alpha = xr.full_like(first_band, 1)
+                alpha["bands"] = "A"
+
+            # There could be Nans in the alpha
+            # Replace them with 0 to prevent cases like 1 + nan = nan, so they won't affect new_alpha
+            alpha = xr.where(alpha.isnull(), 0, alpha)
```
mraspaud, line 1738: This should be move to the first case, as the second case i just ones, right?

satpy/composites/__init__.py
```diff
@@ -1707,24 +1720,68 @@ def _combine_metadata_with_mode_and_sensor(self,
...
+
+            return alpha
```
mraspaud, line 1740: is it possible to have this function outside of the method and class?

mraspaud (commented):
just some comments

satpy/composites/__init__.py
mraspaud: Ok, I've been looking at the code here, and although I understand what the code is doing, I understand @djhoese 's concern regarding the nan magic. I think though it can be made clearer if these lines:
```
            chan = (fg_band * alpha_fore +
                    bg_band * alpha_back * (1 - alpha_fore)) / new_alpha_nan
```
are split in two steps: first compute the new value for the channels, and then apply the nans.

As this function is quite technical, maybe refactoring into even smaller function with relevant names could help clarify things?

satpy/composites/__init__.py
yukaribbba: agree.

satpy/composites/__init__.py
yukaribbba: done

satpy/composites/__init__.py
yukaribbba: After some tests I suddenly realised that the whole NaN thing is completely unnecesary. By dropping it the logic is now almost the same as the old one and is easy to understand.

mraspaud: That's great @yukaribbba , thanks a lot for looking into this again :) Looking forwards to your new commits

mraspaud (approved):
LGTM! thanks for the simplifications

mraspaud:
@djhoese rtd seems to fail, but I can't see any problem in the logs, any idea where I should look next?

mraspaud:
> Right in the middle of the log:
> 
> ```
> 
> /home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/2696/lib/python3.10/site-packages/satpy/composites/__init__.py:docstring of satpy.composites.BackgroundCompositor:5: ERROR: Unexpected indentation.
> ```

duh, I was grep for warning...  thanks
@yukaribbba the identation of the docstring doesn't seem to agree with sphinx...

yukaribbba:
> > Right in the middle of the log:
> > ```
> > 
> > /home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/2696/lib/python3.10/site-packages/satpy/composites/__init__.py:docstring of satpy.composites.BackgroundCompositor:5: ERROR: Unexpected indentation.
> > ```
> 
> duh, I was grep for warning... thanks @yukaribbba the identation of the docstring doesn't seem to agree with sphinx...

Dang...This should be fine now.

mraspaud:
@yukaribbba shinx build now passes, but I'm not sure it looks like you intended, see here https://satpy--2696.org.readthedocs.build/en/2696/api/satpy.composites.html#satpy.composites.BackgroundCompositor

yukaribbba:
damn, how can I do that? Is there something special like `\n` that rtd can recogonize?

mraspaud:
See here to find something you can use: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html
Maybe a bullet list?

yukaribbba:
Maybe a talbe would be better. Let's see how it renders.

yukaribbba:
Heading to bed. I'll see this tomorrow.

yukaribbba:
Done. [https://satpy--2696.org.readthedocs.build/en/2696/api/satpy.composites.html#satpy.composites.BackgroundCompositor](url)

PR #2699, opened 2023-12-18
Move Scene.to_hvplot internals to _scene_converters

bornagain1981:
As requested in PR #2106, the to_hvplot and to_geoviews methods are  moved under _scene_converters.py

- [x] Test added
- [x] Fully documented

PR #2700, opened 2023-12-18
Fix eps_l1b reader Delayed usage causing docs failures 

djhoese:
This was discussed on slack and first seen failing in #2699. Readthedocs builds were failing with a new version of dask. Sphinx apidocs were trying to analyze if a method was a staticmethod and failing with:

```
  File "/home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/2699/lib/python3.10/site-packages/sphinx/ext/autodoc/__init__.py", line 2159, in import_object
    inspect.isstaticmethod(obj, cls=self.parent, name=self.object_name)):
  File "/home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/2699/lib/python3.10/site-packages/sphinx/util/inspect.py", line 214, in isstaticmethod
    if meth:
  File "/home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/2699/lib/python3.10/site-packages/dask/delayed.py", line 645, in __bool__
    raise TypeError("Truth of Delayed objects is not supported")
TypeError: Truth of Delayed objects is not supported
```

I was able to reproduce it locally with the newest dask and discovered that it was the `eps_l1b.py` reader. This reader declared a regular class method as a Delayed object. This is generally considered bad practice as it makes the Delayed object non-serializable in distributed environments/workflows. Technically making it a staticmethod should have fixed it but when I tried it locally I couldn't make it work. I instead decided to make it a global function as there was no real reason to tie it to the class. This seems to fix it.

I then checked trollimage which does use delayed staticmethods and it doesn't fail, but that's because the staticmethod isn't made delayed until it is called. For example `res = dask.delayed(self._some_static_method)(a, b, c)`. This type of calling seems fine.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -278,7 +278,7 @@ latex_documents = [
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     "dask": ("https://docs.dask.org/en/latest", None),
-    "geoviews": ("http://geoviews.org", None),
+    "geoviews": ("https://geoviews.org", None),
     "jobqueue": ("https://jobqueue.dask.org/en/latest", None),
     "numpy": ("https://numpy.org/doc/stable", None),
     "pydecorate": ("https://pydecorate.readthedocs.io/en/stable", None),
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM, thanks for the fix!

PR #2707, opened 2023-12-21
Fix 'viirs_edr' renaming two sets of dimensions to the same names

djhoese:
So this wasn't actually a problem until xarray 2023.9.0 where something changed to be more strict about dimensions and coordinates. This also wasn't failing in tests because my tests assume the ideal case of the file metadata. The real files however seem to have incorrect coordinates for some variables:

```
        short \750m\ Surface\ Reflectance\ Band\ M10(Along_Track_750m, Along_Scan_750m) ;
                \750m\ Surface\ Reflectance\ Band\ M10:scale_factor = 0.0001f ;
                \750m\ Surface\ Reflectance\ Band\ M10:add_offset = 0.f ;
                \750m\ Surface\ Reflectance\ Band\ M10:coordinates = "Latitude_at_375m_resolution, Longitude_at_375m_resolution" ;
```

Obviously the 750m data should not be pointing to the 375m lon/lat variables as coordinates. This was made worse by my renaming in `__init__` from both sets of dimensions (375m and 750m) to the same name "y" and "x". This PR switches that renaming to happen per DataArray. Note I am not testing this particular case with bad coordinate names because I believe the algorithm developers have been told and the file coordinates should get fixed...hopefully.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/viirs_edr.py
+++ b/satpy/readers/viirs_edr.py
@@ -62,6 +62,7 @@ from __future__ import annotations
 import logging
 from typing import Iterable
 
+import dask.array as da
 import xarray as xr
 
 from satpy import DataID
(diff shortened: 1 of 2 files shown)


djhoese:
I need these recent PRs I've made to continue my work and I'm not sure I can wait any longer. If someone wants to do a post-merge review I'll address any issues then.

PR #2711, opened 2024-01-03
Add support for NOAA-21 in MiRS limb correction

djhoese:
In the near future we hope to have limb correction for NOAA-21 from the algorithm developers. In the mean time we (the CSPP Polar2Grid team - @kathys) will use the NOAA-20 coefficients as if they were meant for NOAA-20. This PR fixes the MiRS reader so it can detect coefficient files for other NOAA satellites, not just NOAA-20 as it did previously.

This PR also includes a lot of refactoring and fixes for preserving 32-bit floats instead of accidentally upcasting to 64-bit floats.

Note to @kathys: I'm not sure we knew this but Sfc_type (surface type) is never treated as a mask variable from what I can tell in the unit tests here. Due to the way these files are structured (and that some satellite's files have less metadata than others) we use the global "missing_value" attribute which triggers the reader code to say "let's check for fills and replace with NaN". So this converts 16-bit integers to 32-bit floats. I could fix this, but it would risk breaking existing behavior and I'm not sure we have the time/funding/desire to go down that road so I'm leaving the behavior as is. Let me know if you think otherwise. In the end it probably doesn't change anything for us in Polar2Grid either way (for surface_type that is).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/mirs.py
+++ b/satpy/readers/mirs.py
@@ -336,19 +336,21 @@ class MiRSL2ncHandler(BaseFileHandler):
             return np.timedelta64("NaT")
         if np.issubdtype(data_arr_dtype, np.datetime64):
             return np.datetime64("NaT")
-        return np.nan
+        return np.float32(np.nan)
 
     @staticmethod
     def _scale_data(data_arr, scale_factor, add_offset):
         """Scale data, if needed."""
         scaling_needed = not (scale_factor == 1 and add_offset == 0)
         if scaling_needed:
-            data_arr = data_arr * scale_factor + add_offset
+            data_arr = data_arr * np.float32(scale_factor) + np.float32(add_offset)
         return data_arr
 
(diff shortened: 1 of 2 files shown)


kathys:
@djhoese I read your comments about the missing data and 16 bit integers versus 32 bit floats, and I agree with you, that it is not worth the risk of breaking the existing behavior.  I do not know of anyone who is using the `Sfc_type` parameter right now from these MiRS files.

djhoese:
Note: If NOAA-21 data is provided to this reader as-is it will fail to process because it will try to load/retrieve coefficient files that are not configured in the reader's YAML. We're currently doing this in our own custom configuration in Polar2Grid by replicating the NOAA-20 files for NOAA-21.

djhoese:
I need this PR to continue the work I'm doing that has a pretty tight deadline. I'm going to merge this now and if anyone does a post-merge review I can address issues then.

PR #2713, opened 2024-01-05
Add kwargs config option to turn off mitiff corner correction

### LINKED ISSUE #2712: mitiff writer add config option to add or not to add the size of a pixel in the proj string
## Feature Request

When plotting geotiff and mitiff images from the same satellite, different products, I see the geo location differs.

image
mitiff with the current correction
image
geotiff

Looking in the code I see I add the size of a pixel to the lower left corner of the proj4 string fort DIANA to plot the data correctly.

For some reasons I don't understand now, this apparently is not correct any more.

I want to add a config option to turn off this correction.

I don't think there are any other users of this writer, but to be sure I want to keep the current correction the default.

TAlonglong:
- [X] Closes #2712 
 - [X] Tests added 
 - [X] Fully documented 

### DIFF
--- a/satpy/writers/mitiff.py
+++ b/satpy/writers/mitiff.py
@@ -245,6 +245,20 @@ class MITIFFWriter(ImageWriter):
         # FUTURE: Use pyproj 2.0+ to convert EPSG to PROJ4 if possible
         proj4_string, x_0 = self._convert_epsg_to_proj(proj4_string, x_0)
 
+        proj4_string = self._special_correction_of_proj_string(proj4_string)
+
+        if isinstance(datasets, list):
+            _dataset = first_dataset
+        else:
+            _dataset = datasets
+        mitiff_pixel_adjustment = kwargs.get("mitiff_pixel_adjustment", False)
+        proj4_string = self._append_projection_center(proj4_string, _dataset, x_0, y_0, mitiff_pixel_adjustment)
+        LOG.debug("proj4_string: %s", proj4_string)
+        proj4_string += "\n"
+
+        return proj4_string
+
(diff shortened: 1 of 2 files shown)


satpy/writers/mitiff.py
```diff
@@ -259,31 +259,39 @@ def _add_proj4_string(self, datasets, first_dataset):
...
         else:
             dataset = datasets
+        corner_correction_x = dataset.attrs["area"].pixel_size_x
+        corner_correction_y = dataset.attrs["area"].pixel_size_y
+        try:
+            if kwargs['mitiff_pixel_adjustment'] is False:
+                corner_correction_x = 0
+                corner_correction_y = 0
+        except KeyError:
+            pass
```
djhoese, line 280:
```suggestion
        if kwargs.get("mitiff_pixel_adjustment", False):
            corner_correction_x = 0
            corner_correction_y = 0
```

satpy/writers/mitiff.py
```diff
@@ -563,7 +571,7 @@ def _make_image_description(self, datasets, **kwargs):
 
         _image_description += " Map projection: Stereographic\n"
 
-        _image_description += self._add_proj4_string(datasets, first_dataset)
+        _image_description += self._add_proj4_string(datasets, first_dataset, **kwargs)
```
djhoese, line 574: You only expect a single keyword argument here, right? Could you pass only the specific one?

TAlonglong: Man, you are so fast.

Yeah your are correct.

mraspaud (approved):
LGTM

PR #2726, opened 2024-01-24
Fix AGRI L1 C07 having a valid LUT value for its fill value

djhoese:
Closes https://github.com/ssec/polar2grid/issues/565

CC @simonrp84 

This was first encountered in Geo2Grid development by @kathys. Some debugging showed that C07 files for AGRI L1 have a LUT with 65536 elements in it even though 65535 is a fill value. The end result is fill values being replaced by real BT values.

This PR implements one possible solution. It says "if the LUT already includes the fill value, force that LUT value to NaN". An alternative could be slicing the LUT array to the input data's `valid_range` and then getting NaN for anything beyond the LUT. The ending implementation ends up looking pretty similar I think.

Note: I haven't added tests for this yet because I wanted to give @simonrp84 time to review the solution as soon as possible. I also wanted to see if this would work without having to (likely) rewrite a ton of the tests to fit the situation...which is basically a bug in the input data.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/fy4_base.py
+++ b/satpy/readers/fy4_base.py
@@ -28,6 +28,7 @@ from datetime import datetime
 
 import dask.array as da
 import numpy as np
+import numpy.typing as npt
 import xarray as xr
 
 from satpy._compat import cached_property
@@ -86,7 +87,7 @@ class FY4Base(HDF5FileHandler):
 
         return ref
 
-    def apply_lut(self, data, lut):
+    def _apply_lut(self, data: xr.DataArray, lut: npt.NDArray[np.float32]) -> xr.DataArray:
         """Calibrate digital number (DN) by applying a LUT.
 
         Args:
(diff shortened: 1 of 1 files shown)


djhoese:
Ok this is ready for review @mraspaud and @simonrp84 if you have the time. The existing test code was...rough. The expected values and the coefficients used were based on the index of the channel in the file, not the actual channel number and it made it very confusing. Additionally there was a lot of "group all this stuff into a method to make linters happy about less complexity", but a lot of it didn't need to be that way. Or rather, it should have been done even further to make the code even simpler.

satpy/tests/reader_tests/test_agri_l1.py
```diff
@@ -390,6 +370,7 @@ def test_agri_for_one_resolution(self, resolution_to_test, satname):
         available_datasets = reader.available_dataset_ids
         band_names = CHANNELS_BY_RESOLUTION[resolution_to_test]
         self._assert_which_channels_are_loaded(available_datasets, band_names, resolution_to_test)
+        # band_names = ["C07"]
```
mraspaud, line 373: Is this a reminder of the contents of band_names or a forgotten comment?

mraspaud (approved):
Looks good to me! Thanks for refactoring the tests on the way there.

simonrp84:
Unless I'm doing somethign wrong, this doesn't seem to solve the issue. With the data linked to earlier in this issue I still get a value of 200K for deep space pixels.

satpy/tests/reader_tests/test_agri_l1.py
djhoese: Forgotten comment. I'll get rid of it.

djhoese:
image

Created with:

```python
In [1]: from satpy import Scene; from glob import glob
scn
In [2]: scn = Scene(reader="agri_fy4a_l1", filenames=glob("/data/satellite/agri_l1/20220120/FY4A-_AGRI--_N_DISK_1047E_L1-_*"))

In [3]: scn.load(["C07"])

In [4]: scn["C07"].plot.imshow().figure.show()

```

@simonrp84 possibly incorrect installation on your end?

simonrp84:
Aah,I was using the FY4B reader for FY4A. Maybe that is the problem. Will test later.

djhoese:
I'm going to consider this good to go. Merging...

PR #2731, opened 2024-01-29
Remove slstr_l2 reader in favor of ghrsst_l2

### LINKED ISSUE #2319: slstr_l2.yaml points to deleted slstr_l2.py
**Describe the bug**
The PR https://github.com/pytroll/satpy/pull/1410 (@adybbroe) removed `slstr_l2.py`, but did not update `slstr_l2.yaml` accordingly, which still points to now non-existing classes (see https://github.com/pytroll/satpy/blob/a109584faf2e8efcc1cc4c524f181655f77a9646/satpy/etc/readers/slstr_l2.yaml#L14). 

The PR comments discuss the similarity between the `ghrsst_l2` and `slstr_l2` readers, maybe the `slstr_l2.yaml` reader should be updated to point to the `ghrsst_l2.py` classes? Or, maybe even better, `slstr_l2` shall be added to the list of renamed readers, see here https://github.com/pytroll/satpy/blob/a109584faf2e8efcc1cc4c524f181655f77a9646/satpy/readers/__init__.py#L40

djhoese:
This fixes a long time broken reader. As discussed by @ameraner in #2319, @adybbroe made a PR renaming/refactoring that included moving the `slstr_l2` reader python code to the `ghrsst_l2` reader, but that PR forgot to remove/rename/deprecate the YAML file. This PR finally removes that and adds the reader name to the dictionary of deprecated readers.

I'm not sure I agree with the "equivalence" of the two readers since the variables were renamed to be, for example, `sea_surface_temperature_slstr`. I'm not familiar with these readers and data files, but wouldn't it be more user friendly to share the Python code, but have names without the instrument suffix on the variables?

CC @simonrp84 who originally brought this to my attention.

 - [ ] Closes #2319 
 - [ ] Tests added 
 - [ ] Fully documented 


simonrp84:
Seems fine to me.

For context, I am writing an actual L2 reader for SLSTR (in support of fire products) so I won't make my PR for that until this one is merged.

djhoese:
If there are other L2 products that can be supported doesn't that mean that I shouldn't merge this as-is but instead we fix the reference to the python code in this YAML file?

ameraner:
Looks to me like the "formally cleanest" way would be to merge this PR as-is, in order to remove the broken `slstr_l2` reader, also considering that fixing the python reference in the `slstr_l2.yaml` file would only lead to a "SLSTR L2" reader that only accepts one specific L2 product, which is not optimal. After that, Simon can add a proper `slstr_l2` reader, possibly by re-using/merging the python parts depending on the formats.

djhoese:
Sorry, I need some clarification about what all of this stuff is. SLSTR is the instrument, right? GHRSST is a algorithm/SST group, but is it also a processing/algorithm software package? Are the SST files generated and previously supported by slstr_l2 an output of this GHRSST package? For the other products that @simonrp84 wants to support, are these "official" products? Or are they all outputs of different processing packages and from different algorithm groups and none of them officially supported/recommended by the organization in charge of SLSTR?

It is not uncommon for an L2 reader to exist with support for only some datasets. I don't have a huge problem with that although it is misleading a bit to new users. Even if it takes @simonrp84 6-12 months to add support for other datasets, I'm not that concerned that there will be that many upset users who are surprised by the limited offering of the current version of the reader. Especially since it existed like that for many years before this and was only just discovered to not be functional.

djhoese:
As discussed in the monthly status meeting today, the GHRSST files will likely not be similar to any files that @simonrp84 wants to add support for. So it would make sense for the SST files to only be supported in the `ghrsst_l2` reader. We also agreed that merging this PR to deprecate/remove the old reader is the best move for now. In the future if/when @simonrp84 re-adds the slstr_l2 reader he can do that, but will need to remember to remove the deprecation in `satpy/readers/__init__.py`.

ameraner:
@simonrp84 you mentioned here that you were writing an actual `slstr_l2` reader.. any update/progress on that?

simon-sat:
I got about 80% through writing something, but then moved to ESA.

My colleagues have the code but I don't have high hopes that they completed it...will check! In the worst case maybe they can give the code back to me so I can finish it. ;-)

ameraner:
Would be nice, thanks! I think there will be an increasing demand for it in the near future.

PR #2740, opened 2024-02-06
Add VIIRS L2 Reader

### LINKED ISSUE #2729: Add Viirs L2 Reader + Enhancments
## Feature Request

I would like to add a reader for level 2 nc files that use the VIIRS instrument and have a layer in NASA Worldview. There is currently no reader available for a lot of the products that we deliver.

I would like to add a reader for CLDPROP, Clear Sky Confidence, DB Aerosol Optical Depth and DB Angstrom Exponent L2 files and then show/save the images with the same colormaps that NASA Worldview uses.

This feature should be fully backwards compatible. The only changes to existing code that I have found necessary is to make some very minor changes to the yaml_reader. I will be sure to point these out when I believe the changes are ready for a PR.

**Additional context**
This feature would likely be a first step which could be expanded upon in the future to support many more of the VIIRS L2 worldview layers once I am more familiar with the Satpy project and workflow.

wjsharpe:
This PR adds the code for the (initial) functionality for the VIIRS Level 2 reader. There is currently no reader available for many SNPP and NOAA20 products we deliver, this should help with that.

Components of PR:
1. VIIRS_L2 reader and corresponding yaml file for four VIIRS_L2 products and other VIIRS_L2 file types. This could be expanded in the future to support all VIIRS_L2  products in NASA Worldview
2. ~Added four colormaps/palettes to generic.yaml in enhancements (could change these to filename reference if there is a good spot to put the txt files in Satpy project)~
3. ~Added composites to viirs.yaml to correctly mask Cloud Top Height NaN values~

 - [ ] Closes #2729 
 - [ ] Tests added
 - [ ] Fully documented 


satpy/tests/reader_tests/test_viirs_l2.py
```diff
@@ -0,0 +1,152 @@
...
+        )
+        r.create_filehandlers(loadables)
+        datasets = r.load(
+            ["Aerosol_Optical_Thickness_550_Land_Ocean_Best_Estimate", "Angstrom_Exponent_Land_Ocean_Best_Estimate"]
+        )
+        assert len(datasets) == 2
+        for d in datasets.values():
+            assert d.shape == DEFAULT_FILE_SHAPE
+            assert d.dims == ("y", "x")
+            assert d.attrs["sensor"] == "viirs"
```
djhoese, line 124: The last 3 tests all have the same body. Could you parametrize them? I think it would be `("filename", "variables")`. You could also add checks for dtype which is a relatively new thing we've been striving for. Something like:

```
for d in datasets.values():
    d_np = d.compute()
    assert d.dtype == d_np.dtype
    assert d.dtype == np.float32
```

There are cases where some operations (scaling the data, masking invalid values, etc) where xarray or numpy will cast data to 64-bit floats. For most satellite data 32-bit floats is good enough. Or in other cases like category products (ex. cloud type) we want things to remain integer types. The dtype could be another parameter in the parametrize if needed.

djhoese (changes requested):
Nice job! I had a couple requests and questions that I've made inline. My only other question is if there is a reason the enhancements in `generic.yaml` couldn't be put in `viirs.yaml`? Sorry if I missed/forgot this in a past discussion.

satpy/etc/readers/viirs_l2.yaml
```diff
@@ -0,0 +1,126 @@
...
+    file_key: Angstrom_Exponent_Land_Ocean_Best_Estimate
+    file_type: [aerdb_l2_viirs, aerdb_l2_viirs_nrt]
+
+  Aerosol_Optical_Thickness_550_Land_Ocean:
+    name: Aerosol_Optical_Thickness_550_Land_Ocean_Best_Estimate
+    long_name: Deep Blue/SOAR aerosol optical thickness at 550 nm over land and ocean
+    units: None
+    coordinates: [aerdb_lon,aerdb_lat]
+    file_key: Aerosol_Optical_Thickness_550_Land_Ocean_Best_Estimate
+    file_type: [aerdb_l2_viirs, aerdb_l2_viirs_nrt]
```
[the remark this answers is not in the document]
wjsharpe: Long_name does match what is in the file. Everything is now "1" instead of None. For the AERDB products it was "1" in the file so that was an oversight on my part. For Clear_Sky_Confidence it is None in the file but "1" should be good for the CF compatibility you mentioned.

satpy/readers/viirs_l2.py
```diff
@@ -0,0 +1,186 @@
...
+        """Get sensor name."""
+        return self["/attr/instrument"].lower()
+
+    def _get_dataset_file_units(self, ds_info, var_path):
+        file_units = ds_info.get("units")
+        if file_units is None:
+            file_units = self.get(var_path + "/attr/units")
+        if file_units == "none" or file_units == "None":
+            file_units = "1"
+        return file_units
```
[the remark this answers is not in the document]
wjsharpe: I just changed the YAML units from None to "1", so this shouldn't come up.

satpy/tests/reader_tests/test_viirs_l2.py
wjsharpe: Everything is parameterized with dtype check now, all the datasets are of type 32-bit float so that was a simple addition.

satpy/readers/viirs_l2.py
```diff
@@ -0,0 +1,186 @@
...
+
+        """
+        for is_avail, ds_info in configured_datasets or []:
+            if is_avail is not None:
+                yield is_avail, ds_info
+                continue
+            ft_matches = self.file_type_matches(ds_info["file_type"])
+            var_path = ds_info.get("file_key", ds_info["name"])
+            is_in_file = var_path in self
+            yield ft_matches and is_in_file, ds_info
```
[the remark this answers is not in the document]
wjsharpe: I added a conditional making `is_in_file=None` if not ft_matches and checked locally to verify that all datasets show up when multiple file types are provided.

satpy/etc/readers/viirs_l2.yaml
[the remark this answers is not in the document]
djhoese: In general I think my preference is for the YAML to only specify metadata that we know needs to be overwritten. For example when "units" are missing or are incorrect (not CF compliant, typos, etc). When you say the units for `Clear_Sky_Confidence` is None, do you mean the string "None", "none", it doesn't exist, or something else?

satpy/etc/composites/viirs.yaml
```diff
@@ -685,3 +685,23 @@ composites:
...
+    description: >
+      Cloud Top Height composite from NOAA/SNPP
+    compositor: !!python/name:satpy.composites.MaskingCompositor
+    prerequisites:
+      - Cloud_Top_Height
+      - cth
+    conditions:
+      - method: isnan
+        transparency: 100
+    standard_name: viirs_cloud_top_height
```
djhoese, line 707: Am I reading this correctly that both `Cloud_Top_Height` and `cth` come from the `viirs_l2` reader? What's the difference between them again?

I think this solution is reasonable, but we've also made it pretty common to include a boolean keyword argument in the file handler to enable or disable masking for certain products. This simplifies things in that users see the dataset as being part of the reader rather than being confused why something that comes from the file is being considered a composite. I think I'd prefer this, but I also don't know anything about the VIIRS L2 files and this/these products so I'm curious what you think?

satpy/etc/readers/viirs_l2.yaml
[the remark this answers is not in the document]
wjsharpe: In the file the units for Clear Sky Confidence are equal to the string "none". I'm not really sure of the distinction or best practices of what to use between the different nones.

satpy/etc/enhancements/generic.yaml
```diff
@@ -1230,3 +1230,255 @@ enhancements:
          stretch: crude
          min_stretch: [0,0,0]
          max_stretch: [1,1,1]
+
+  viirs_cloud_top_height:
+    name: viirs_cloud_top_height
+    operations:
+      - name: palettize
```
djhoese, line 1237: Given my other comment about potentially moving the filtering/masking to the reader, I think this enhancement specifically should go in `enhancements/viirs.yaml` given the `viirs_` prefix. Additionally, the other enhancements are being applied/set based on the specific `name: ` of the product. Are these names the same for the MODIS data that you were looking at supporting? Or do they differ? Should we maybe have the enhancements controlled by standard_name instead of the specific name? However, I then fear we'd be changing long held default behavior for all instruments/readers that produce these types of products. Essentially making decisions for all satellite instruments/readers based on one NASA Worldview likes. Not that these are bad colormaps (I haven't look at them), but up until now we've been pretty relaxed as far as defaults for enhancements on products. Usually users, even core maintainers of satpy, end up having their own custom `etc/` directory with custom enhancements.

satpy/etc/readers/viirs_l2.yaml
```diff
@@ -0,0 +1,126 @@
+reader:
+  name: viirs_l2
+  short_name: VIIRS L2
+  long_name: SNPP VIIRS Level 2 data in netCDF4 format
+  description: Generic NASA VIIRS L2 Reader
+  status: Alpha
+  supports_fsspec: false
+  reader: !!python/name:satpy.readers.yaml_reader.GEOFlippableFileYAMLReader
+  sensors: [viirs]
+  default_datasets:
```
djhoese, line 10: This isn't actually used and can be removed. It was something we thought we might use, but never did.

```suggestion
```

satpy/etc/readers/viirs_l2.yaml
```diff
@@ -0,0 +1,126 @@
...
+    file_patterns:
+      - 'CLDMSK_L2_VIIRS_{spacecraft_name:s}.A{start_time:%Y%j.%H%M}.{collection:03d}.{production_time:%Y%j%H%M%S}.nc'
+  aerdb_l2_viirs:
+    file_reader: !!python/name:satpy.readers.viirs_l2.VIIRSL2FileHandler
+    file_patterns:
+      - 'AERDB_L2_VIIRS_{spacecraft_name:s}.A{start_time:%Y%j.%H%M}.{collection:03d}.{production_time:%Y%j%H%M%S}.nc'
+  aerdb_l2_viirs_nrt:
+    file_reader: !!python/name:satpy.readers.viirs_l2.VIIRSL2FileHandler
+    file_patterns:
+      - 'AERDB_L2_VIIRS_{spacecraft_name:s}.A{start_time:%Y%j.%H%M}.{collection:03d}.nrt.nc'
```
djhoese, line 28: Unless you expect people to provide the AERDB production/archive files and the NRT files at the same time, these file types could be combined by moving the NRT file pattern to the `file_patterns` of the non-NRT file type.

satpy/etc/readers/viirs_l2.yaml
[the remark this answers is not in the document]
djhoese: I think CF-wise the unit "1" literally means unitless which I think (at least in English) is equivalent to what is intended with "none".

satpy/etc/composites/viirs.yaml
wjsharpe: `cth` is just `Cloud_Top_Height` with the colormap applied, and then that is masked so the NaN values show up correctly with the applied colormap. The suggestion you made seems reasonable and if that's the best practice I can change it to have the boolean keyword.

satpy/readers/viirs_l2.py
```diff
@@ -0,0 +1,188 @@
...
+
+        if file_units == "1" and output_units == "%":
+            LOG.debug(
+                "Adjusting scaling factors to convert '%s' to '%s'",
+                file_units,
+                output_units,
+            )
+            factors[::2] = np.where(factors[::2] != -999, factors[::2] * 100.0, -999)
+            factors[1::2] = np.where(factors[1::2] != -999, factors[1::2] * 100.0, -999)
+            return factors
```
djhoese, line 133: If this isn't used, this could be removed. Are there any level 2 products that are the same as the Level 1B datasets? I know for ABI there are L2 products that are the same as the calibrated L1 band data.

satpy/readers/viirs_l2.py
```diff
@@ -0,0 +1,188 @@
...
+            if is_avail is not None:
+                yield is_avail, ds_info
+                continue
+            ft_matches = self.file_type_matches(ds_info["file_type"])
+            if not ft_matches:
+                is_in_file = None
+            else:
+                var_path = ds_info.get("file_key", ds_info["name"])
+                is_in_file = var_path in self
+            yield ft_matches and is_in_file, ds_info
```
djhoese, line 155: This is mostly nitpicking, but I think this version is easier to read.

```suggestion
            if ft_matches is None:
                yield None, ds_info
                continue
            var_path = ds_info.get("file_key", ds_info["name"])
            yield var_path in self, ds_info
```

satpy/etc/enhancements/generic.yaml
wjsharpe: That viirs_cloud_top_height is just poorly names by me, it could also be applied to MODIS data. I can change the name of it to reflect that. The names are the same for the MODIS data that I am looking to support in the future. As far as potential downstream behavior, would there be a better way for me to specify how to apply the enhancement? I was hoping to keep it in the project as some of my coworkers will theoretically also be using these colormaps.

satpy/tests/reader_tests/test_viirs_l2.py
```diff
@@ -0,0 +1,133 @@
...
+    @pytest.mark.parametrize(
+        ("filename", "datasets"),
+        [
+            pytest.param("CLDPROP_L2_VIIRS_SNPP.A2023364.2230.011.2023365115856.nc",["Cloud_Top_Height"],id="CLDPROP"),
+            pytest.param("CLDMSK_L2_VIIRS_SNPP.A2023364.2230.001.2023365105952.nc",["Clear_Sky_Confidence"],id="CLDMSK"),
+            pytest.param("AERDB_L2_VIIRS_SNPP.A2023364.2230.011.2023365113427.nc",
+                         ["Aerosol_Optical_Thickness_550_Land_Ocean_Best_Estimate",
+                          "Angstrom_Exponent_Land_Ocean_Best_Estimate"],id="AERDB"),
+        ],
+    )
```
djhoese, line 119: Our current linter doesn't seem to be complaining about whitespace properly, but these should really have spaces after commas.

```suggestion
            pytest.param("CLDPROP_L2_VIIRS_SNPP.A2023364.2230.011.2023365115856.nc", ["Cloud_Top_Height"], id="CLDPROP"),
            pytest.param("CLDMSK_L2_VIIRS_SNPP.A2023364.2230.001.2023365105952.nc", ["Clear_Sky_Confidence"], id="CLDMSK"),
            pytest.param("AERDB_L2_VIIRS_SNPP.A2023364.2230.011.2023365113427.nc",
                         ["Aerosol_Optical_Thickness_550_Land_Ocean_Best_Estimate",
                          "Angstrom_Exponent_Land_Ocean_Best_Estimate"], id="AERDB"),
        ],
    )
```

satpy/etc/composites/viirs.yaml
djhoese: I don't know about "best" practice, but it is what I've been doing most of the time. What is the correct output you expect for NaN values going into a colormap? I guess I'm surprised that NaN in the dataset isn't coming out as fully transparent when saved as a geotiff.

You said that `cth` is the cloud top height with the colormap applied? How so? `cth` doesn't have an enhancement applied from what I see. Otherwise, if you meant that `Cloud_Top_Height` is the one with the colormap applied, that doesn't happen until after this compositing step unless I'm missing something. That is, MaskingCompositor and SingleBandCompositor don't apply enhancements internally. So the intention with this composite definition is to take a single band (L) array and add an Alpha (LA) where it is transparent where the data has NaN values. Isn't that already the default behavior?

wjsharpe: That is what I thought the default behavior would be. For whatever reason NaN values were being shown as the first color in my palette and were not transparent, this composite fixed that problem for me.

satpy/etc/enhancements/generic.yaml
djhoese: I think there are two possible solutions, either of which is fine and have their own pros and cons. In either case I think having the colormaps as separate files in `etc/colormaps/` is fine. In both of the below cases I think you could include the colormap files in Satpy if they are any type of "standard" colormap that people are used to seeing and using, even if they are never used in Satpy itself...although now that I say that maybe that is a little weird to have them and not use them. Anyway...

1. Configure all of these composites and enhancements in Satpy as you have now. This assumes that none of these new definitions conflict with existing definitions and won't cause too much headache down the line.
2. Distribute your own collection of `etc/` YAML and colormap files that users include as part of their environment or local working directory to get access to these custom enhancements. There is a plugin system in Satpy that would make this decently easy as a separate python package.

Obviously the best part of the first option is no extra maintenance on your part as far as user installations go. The second option is great because it means you have full control over what your users/colleagues are using, that Satpy's future changes don't modify your configuration, and that we avoid any conflicts with other products/instruments. But there is a maintenance aspect to it. If there are conflicts with existing definitions or other readers/instruments then we're stretching Satpy's historically lazy/relaxed policy on this kind of stuff and proving that we need to add more features for things like "user profiles" to control which set of configurations users are using even when they are all provided by Satpy.

satpy/etc/composites/viirs.yaml
djhoese: Before I look into this more, do you *need* the palette to be preserved in your output or would an RGB/A as output be OK with you?

Basically, there is the colorized and the palettized variants of applying a colormap to something. I forget what the default behavior of the geotiff writer is, but there is an option to say whether the final dataset is saved with a geotiff ColorTable or converted to an RGB/A first. I think by default, especially since your colormap doesn't have an Alpha channel, the palettize enhancement will try to assign every incoming value to some color in that colormap. If you use the colorize enhancement instead I think it provides some freedom to the enhancement step to say "this is NaN, that's invalid, don't assign it a color in the colormap".

wjsharpe: The reason I chose the palettize instead of colorize in this case is because (if I'm not mistaken) colorize will interpolate between the color values given. E.g if I have a 900m and 950m color and a value of 925m at a given point it will interpolate between the two, when it should just be the 900m color. I am hoping to avoid that interpolation. Colorizing instead of using a palette did mask the NaN by default for me (hence why I don't have composites for the other datasets), like the comment you linked suggested.

djhoese: I have to get some other work done, but do you have a screenshot of what the CTH colormap is supposed to look like? What we've done in the past for interpolation issues is replicated the colors so you have one block of color from 0 to 5 and then the next color is 5.0001 to 10.0, then the next is 10.0001 to 15.0, and so on. Not great, but it allows us to cheat in most cases without having to have special cases for discrete colormaps.

Skimming the colors in the CTH colormap, I'm curious how you generated it. In the first block of colors I see the G and the B hover between 0 and 2, is that difference even perceptible by humans?

wjsharpe: As you can see in the screenshot it looks like there are only ~15 colors but in the json worldview uses there are many similar colors within each box. I reflected that json in my colormap, but I could just simplify it by using one color per "box" in the image and not use a palette to avoid headaches.
Screenshot 2024-02-08 at 10 17 54 AM

djhoese: My current recommendation then is to do what I suggested above. You would use colorize and you would define your colormap as color1 -> color1 with values (control points) A to B, then color2 -> color2 with values B + 0.0001 + C, and so on. If you only defined the colors with a single point then you'd get interpolation between them. This is what I've ended up doing and I'm not sure there are many (any) other options within Satpy at the moment for a discrete colormap like this.

wjsharpe: This should be updated now. I removed the enhancements and switched from palette to colormap with the values and it produced the same result.

satpy/etc/enhancements/generic.yaml
wjsharpe: I updated this similar to option 1, but to hopefully mitigate potential headaches I changed my colormaps from using name to standard_name and made the standard_name formatted like filetype+name (e.g. cldprop_cloud_top_height). This naming scheme would still allow reuse for the MODIS data which could come later, but shouldn't conflict with anything else.

satpy/tests/reader_tests/test_viirs_l2.py
wjsharpe: I updated this, I did notice the linter not picking up on some little things when I ran it in the terminal. I started to just run Black first which seems to pick up on everything and then your linter afterwards to reconcile any difference between the two.

satpy/etc/enhancements/generic.yaml
```diff
@@ -1230,3 +1230,60 @@ enhancements:
...
+          - {colors: [[255, 255, 0], [255, 255, 0]], min_value: 4800.0001, max_value: 5600}
+          - {colors: [[0, 220, 0], [0, 220, 0]], min_value: 5600.0001, max_value: 6400}
+          - {colors: [[0, 136, 0], [0, 136, 0]], min_value: 6400.0001, max_value: 7200}
+          - {colors: [[0, 80, 0], [0, 80, 0]], min_value: 7200.0001, max_value: 8000}
+          - {colors: [[0, 136, 238], [0, 136, 238]], min_value: 8000.0001, max_value: 8800}
+          - {colors: [[0, 0, 255], [0, 0, 255]], min_value: 8800.0001, max_value: 9600}
+          - {colors: [[0, 0, 170], [0, 0, 170]], min_value: 9600.0001, max_value: 10400}
+          - {colors: [[0, 0, 100], [0, 0, 100]], min_value: 10400.0001, max_value: 11200}
+          - {colors: [[183,15,141], [183, 15, 141]], min_value: 11200.0001, max_value: 12000}
+          - {colors: [[102, 0, 119], [102, 0, 119]], min_value: 12000.0001, max_value: 18000}
```
djhoese, line 1256: You should be able to define this as two parameters: colors and values. This should be much faster for Satpy to load this information in. As you have it Satpy/trollimage is considering each of these as a separate colormap.

satpy/etc/enhancements/generic.yaml
```diff
@@ -1230,3 +1230,60 @@ enhancements:
...
+
+  Clear_Sky_Confidence:
+    standard_name: cldmsk_clear_sky_confidence
+    operations:
+      - name: colorize
+        method: !!python/name:satpy.enhancements.colorize
+        kwargs:
+          palettes:
+          - {
+             filename: clear_sky_confidence.txt
```
djhoese, line 1266: Does Satpy find this colormap in `etc/colormaps/`?

djhoese (commented):
Thanks for doing all this work. I had a few more inline comments. Otherwise, could you rename the colormap files to `.csv` and remove the whitespace?

satpy/etc/enhancements/generic.yaml
wjsharpe: I set the config path to the etc/colormaps/ directory locally before running any of this

satpy/etc/enhancements/generic.yaml
wjsharpe: Why I did it like this is because when I just had values and colors it interpolated between the colors like we were talking about before.  The only way I could get it to not interpolate was to have satpy see it as separate colormaps with those defined values.

wjsharpe: I reverted back to just having the two values. The images attached show the issue
Screenshot 2024-02-09 at 1 41 31 PM
Screenshot 2024-02-09 at 1 43 26 PM

djhoese: Unless github mobile is showing it wrong, you still need to double every color and have the values as you had them. I'll check more closely when I'm at my computer later.

wjsharpe: No, I think you're right. I can make those changes and then put it in a csv in a few minutes

wjsharpe: should be all fixed now

satpy/etc/enhancements/generic.yaml
djhoese: Which config path?

satpy/etc/enhancements/generic.yaml
djhoese: I'm not sure I'm a fan of the filetype appearing in the standard_name. After all we'd like the standard_names to be CF-ish, but I understand why you did it. Let's let others review this PR and we'll get back to this after others have had a chance to think about conflicts.

pnuu (commented):
Couple of comments:
* What if we add other cloud top height colormaps "native" to other such products? Should the naming show which producer these colormaps belong to?
* We have the generic colormaps in Trollimage. Should these be included there? Probably not, as they are data specific. Just a thought.
* All the other enhancement names are lower case.
* Should the colormaps have some headers to explain their source and where they are expected to be used?

satpy/etc/enhancements/generic.yaml
wjsharpe: I set SATPY_CONFIG_PATH using `satpy.config.set(config_path=['xxx/etc/colormaps/'])`

djhoese: Could you try not modifying the config path and specify these in the YAML as `colormaps/clear_sky_confidenct.csv` and so on?

djhoese:
> What if we add other cloud top height colormaps "native" to other such products? Should the naming show which producer these colormaps belong to?

Good point. If something has a name, it should be meaningful. And in this shared context they should try to follow some kind of defined standard naming. Otherwise we end up renaming everything in the future like we did with the reader names because everyone followed a different but kind of similar scheme.

> We have the generic colormaps in Trollimage. Should these be included there? Probably not, as they are data specific. Just a thought.

Another good point, but I think you're right. You're saying should these (and all?) colormaps be put in trollimage. I feel pretty strongly that the answer is no. Trollimage provides generic functionality, but is for the most part not specific to any task/field.

> Should the colormaps have some headers to explain their source and where they are expected to be used?

Maybe a README.txt in `etc/colormaps/` that explains each colormap? There may come a day where people are adding numpy binary files for colormaps into this directory in which case they can't add a header so an external description is probably a good place to start and avoids wasted time processing the header every time.

I'm mostly just really uncertain on where to draw the line with this as a user-specific config thing versus builtin functionality. The biggest issue is that some of us (the satpy maintainers) have gotten away with being lazy for the most part with sharing configuration of composites and enhancements and just getting lucky that things mostly don't conflict. In the case of @pnuu and a few others, you've just always had your own custom configs from the start and not depended on Satpy's defaults/builtins. In my Polar2Grid/Geo2Grid projects I depend on Satpy's defaults, but also overwrite a few things the way we want them in our project.

These colormaps and enhancements feel like a user-specific configuration kind of thing...but I/we probably wouldn't feel that way if they didn't involve new colormaps. By that I mean if the enhancement that @wjsharpe wanted was a black and white gamma 0.5 (or 2 or whatever to get square root) for one or more of these products we'd say "yeah, ok, that's not too jarring", but that is still a user-specific configuration that could conflict in the future. So I don't know what the right approach is here.

satpy/etc/enhancements/generic.yaml
wjsharpe: That works as well, just updated.

wjsharpe:
@djhoese So I think in the interest of trying to get at least the reader + enhancements merged there are a few things I could do.  If you think it is too user specific I can remove the /etc/colormaps/ directory and just put them in a shared location for my team and point the config path there. Then after you and the other maintainers decide on some of the specifics discussed above I could be the guinea pig for those changes in a separate merge, if you decide you want them in Satpy at all. For the naming of my enhancements I think how I have it now with using standard_name should avoid any of those conflicts we were talking about last week and would still remain generic enough to be correct for the future Aqua/MODIS version of these products. As for CF compliance of those standard_names, standard_name as far as I know just has the purpose of definitively defining a physical quantity. In this context where there are multiple similar datasets adding an additional modified seems reasonable to accomplish that purpose to me, especially since there is already a set of standard name modifiers that are accepted in the CF standard. Those modifiers are at the end, so I could change to have the product/file type string at the end instead of the beginning if you would prefer (e.g. cloud_top_height_cloudprop instead of cldprop_cloud_top_height).

djhoese:
>  As for CF compliance of those standard_names, standard_name as far as I know just has the purpose of definitively defining a physical quantity. In this context where there are multiple similar datasets adding an additional modified seems reasonable to accomplish that purpose to me, especially since there is already a set of standard name modifiers that are accepted in the CF standard. Those modifiers are at the end, so I could change to have the product/file type string at the end instead of the beginning if you would prefer (e.g. cloud_top_height_cloudprop instead of cldprop_cloud_top_height).

Could you point me to the documentation in CF about modifiers on standard names? I didn't think they went as far as user customizations, but were more about defining the specific state of a particular variable. I'd really like to find a way to keep the standard names and if there needs to be non-conflicting enhancements then I think specifying `name: ` and `reader: ` for each product's enhancement is the clearest without complicating/corrupting the standard_name. This may mean duplicate enhancements between VIIRS and MODIS and other sensors but I don't think that's the worst thing in this case.

> So I think in the interest of trying to get at least the reader + enhancements merged there are a few things I could do. If you think it is too user specific I can remove the /etc/colormaps/ directory and just put them in a shared location for my team and point the config path there. Then after you and the other maintainers decide on some of the specifics discussed above I could be the guinea pig for those changes in a separate merge, if you decide you want them in Satpy at all.

If you're OK with this, then separating the colormaps out would really clarify what this PR is. Especially if you go as far as separating out the `enhancements/` YAML changes. Note that in addition to saving them in a shared location you could make a plugin (https://satpy.readthedocs.io/en/stable/dev_guide/plugins.html) package that could be installed by your colleagues.

Unless @mraspaud (who is on a business trip right now) has a decently positive reaction to including these colormaps in Satpy and definitely wants them in this PR, I think my preference is leaning towards separating them out for now.

djhoese:
@wjsharpe Just for the record, are these colormaps used by any other organizations/tools? For example AWIPS (VIIRS or otherwise) or maybe a non-US based tool?

mraspaud:
Thanks for asking for my opinion here. I would actually prefer to keep the colourmaps outside if possible. If you feel the colourmaps would still be of use to other users, I can recommend making and distributing a satpy plugin for it.
Other than that, this PR looks quite good to me.

wjsharpe:
@djhoese I just made the change separating out the colormaps as well as identifying the enhancements by name and reader instead of standard_name. One thing I would prefer is keeping the enhancement yaml changes. There won't be any conflicts with the convention of specifying the name and reader discussed above, and removing them would necessitate the creation of a plugin like you mentioned. This seems like additional development work for myself and anyone using the reader. In my opinion not having the enhancements would somewhat render the remaining changes in this PR not particularly useful without the plugin, at least for what is currently planned. If you feel strongly about having those changes removed I can do so, as my main priority is resolving this PR as quickly as possible to move onto downstream tasks outside of the Satpy project.

As far as other usages of the colormaps, I'm not positive of their usages outside of worldview. I haven't seen them anywhere else, but I am also not familiar with AWIPS.

djhoese:
I'm going on parental leave in a little bit, but while waiting in the hospital I had time to read this. If the enhancements that use the colormaps stay in Satpy then the colormaps do too. Right? Or am I missing something with the way you've done the change. If you're distributing the colormaps in an etc directory to colleagues then you can include the enhancement YAML alongside. With or without a full plugin package that should work if the config path includes the location of this extra etc directory.

mraspaud:
Well, without the csv files, someone wanting to visualise the product in the current state will just get a crash. If you take out the enhancements on the other hand, we would at least get a gray-scale image to look at...
As @djhoese if you provide the enhancement config alongside the csv files, you're good.

wjsharpe:
Enhancements are now removed

djhoese (approved):
This looks good to me. Thanks for all the back and forth even if it ultimately meant removing the enhancements and colormaps. I've updated the title and original description to not give the wrong impression to people stumbling on the PR.

It is a little late for me here and I don't fully trust myself to merge this tonight. I'll merge this tomorrow after one more look, but other maintainers can feel free to merge this before if it looks good to them.

PR #2741, opened 2024-02-06
Drop Python 3.9 support

djhoese:
Following the guidance of:

https://scientific-python.org/specs/spec-0000/

This PR drops Python 3.9 support and adjusts CI, pre-commit, and readthedocs to use newer versions of Python. This means the currently supported versions of Python will be 3.10, 3.11, and 3.12.

CC @pytroll/satpy-core I've asked for reviews from some of you, but I think as long as there are no complaints this should be a simple agreement and merge.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ["windows-latest", "ubuntu-latest", "macos-latest"]
-        python-version: ["3.9", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12"]
         experimental: [false]
         include:
           - python-version: "3.12"
--- a/.github/workflows/deploy-sdist.yaml
+++ b/.github/workflows/deploy-sdist.yaml
@@ -19,7 +19,7 @@ jobs:
         shell: bash -l {0}
         run: |
           python -m pip install -q build
-          python -m build -s
+          python -m build
 
       - name: Publish package to PyPI
         if: github.event.action == 'published'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
           - types-pkg-resources
           - types-PyYAML
           - types-requests
-        args: ["--python-version", "3.9", "--ignore-missing-imports"]
+        args: ["--python-version", "3.10", "--ignore-missing-imports"]
   - repo: https://github.com/pycqa/isort
     rev: 5.13.2
     hooks:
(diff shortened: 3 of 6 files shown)


djhoese:
The remaining unstable environment breakage has been reported here: https://github.com/Unidata/cftime/issues/318

sfinkens (approved):
LGTM, thanks! Also for sharing that nice table.

ameraner:
I understand that the fact that EUMETSAT's `fcidecomp` package (https://anaconda.org/Eumetsat/fcidecomp, and https://gitlab.eumetsat.int/open-source/data-tailor-plugins/fcidecomp/-/tree/2.0.1?ref_type=heads, see also threads on slack) only currently supports only up to py3.9 is going to be a problem after this PR. 

Considering that FCI will go pre-operational soon, and satpy will get a lot of attention in the next months because of this, is it an option to wait a bit more for dropping 3.9? EUM "promised" an update in May, but maybe with a little more pressure (e.g. at the OPSWG in March) this could be accelerated. 

I fully understand that it's highly annoying to wait for updates for a package that is not maintained properly, but considering this special phase with an update of a major met satellite, maybe it's worth considering.

If this is not possible/too painful, maybe we should consider writing a docs page for possible workarounds (manual installation from Pytroll's fcidecomp fork, or `hdf5plugin`)?

simonrp84:
I agree with @ameraner, if possible it'd be good to delay the dropping of 3.9 until FCI can be easily supported.

djhoese:
I don't have a problem waiting a little longer, but don't all the readers use hdf5plugin right now? I think @gerritholl had shown that fcidecomp doesn't even work with modern numpy versions.

pnuu:
> I don't have a problem waiting a little longer, but don't all the readers use hdf5plugin right now?

We don't import `hdf5plugins` in Satpy. It was decided that we shouldn't force the user to one solution, so either `import hdf5lugins` before importing satpy/xarray/netcdf4/h5netcdf, or installing `fcidecomp`, by the user is always required.

djhoese:
Ah ok. So not as strict/severe of a case then. Still, if fcidecomp doesn't work with modern numpy then users won't be able to create a new python environment that works with fcidecomp anyway. How do we anticipate people wanting to process FCI data will create their python environments? If they don't specify a specific Python version then they're going to get Python 3.11 or 3.12 as it is. If they are told to install Python 3.9 (or less) then they'll get an older version of Satpy that is compatible with that. So maybe the question is if the newest FCI changes (composites, L2, etc) are requested...eh OK we'll wait.

gerritholl:
> I think @gerritholl had shown that fcidecomp doesn't even work with modern numpy versions.

That problem was fixed by EUMETSAT.  fcidecomp now works with modern numpy.

PR #2742, opened 2024-02-12
Fix nominal end time in AHI HSD

sfinkens:
I noticed that for AHI HSD, the attributes `start_time/end_time` and `nominal_start_time/nominal_end_time` are identical.

```python
import satpy
import glob

filenames = glob.glob("*.DAT")
scene = satpy.Scene(filenames, reader="ahi_hsd")
scene.load(["B13"])

print(scene["B13"].attrs["start_time"])
print(scene["B13"].attrs["end_time"])
print(scene["B13"].attrs["time_parameters"]["nominal_start_time"])
print(scene["B13"].attrs["time_parameters"]["nominal_end_time"])
```

Output:

```
2020-01-01 12:00:00
2020-01-01 12:00:00
2020-01-01 12:00:00
2020-01-01 12:00:00
```

This has two reasons:

1. The `end_time` property returns `nominal_start_time`. See https://github.com/pytroll/satpy/blob/main/satpy/readers/ahi_hsd.py#L417
2. The `observation_end_time` (e.g. 03:10:12.34567 for a full disk scan) is rounded to the "observation timeline" (03:00:00 or "0300" in the header). See https://github.com/pytroll/satpy/blob/main/satpy/readers/ahi_hsd.py#L473. Here, the offset `dt` only accounts for the start of the scan, not the total scan duration.

To fix this, I modified the `nominal_end_time` property to add the scan duration to the nominal start time.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/ahi_hsd.py
+++ b/satpy/readers/ahi_hsd.py
@@ -429,12 +429,13 @@ class AHIHSDFileHandler(BaseFileHandler):
     @property
     def nominal_start_time(self):
         """Time this band was nominally to be recorded."""
-        return self._modify_observation_time_for_nominal(self.observation_start_time, "start")
+        return self._modify_observation_time_for_nominal(self.observation_start_time)
 
     @property
     def nominal_end_time(self):
         """Get the nominal end time."""
-        return self._modify_observation_time_for_nominal(self.observation_end_time, "end")
+        freq = self._observation_frequency
+        return self.nominal_start_time + timedelta(minutes=freq // 60, seconds=freq % 60)
 
     @staticmethod
     def _is_valid_timeline(timeline):
@@ -443,7 +444,16 @@ class AHIHSDFileHandler(BaseFileHandler):
             return False
         return True
 
-    def _modify_observation_time_for_nominal(self, observation_time, start_or_end_time="start"):
+    @property
+    def _observation_frequency(self):
+        frequencies = {"FLDK": 600, "JP": 150, "R3": 150, "R4": 30, "R5": 30}
+        area = self.observation_area
+        if area != "FLDK":
+            # e.g. JP01, JP02 etc
+            area = area[:2]
+        return frequencies[area]
+
+    def _modify_observation_time_for_nominal(self, observation_time):
         """Round observation time to a nominal time based on known observation frequency.
 
         AHI observations are split into different sectors including Full Disk
@@ -454,7 +464,6 @@ class AHIHSDFileHandler(BaseFileHandler):
         sector. So if the observation time is 13:32:48 for the "JP02" sector
         which is the second Japan observation where every Japan observation is
         2.5 minutes apart, then the result should be 13:32:30.
-
         """
         timeline = "{:04d}".format(self.basic_info["observation_timeline"][0])
         if not self._is_valid_timeline(timeline):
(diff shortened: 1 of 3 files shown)


djhoese:
> The observation_end_time (e.g. 03:10:12.34567 for a full disk scan) is rounded to the "observation timeline" (03:00:00 or "0300" in the header). See [main/satpy/readers/ahi_hsd.py#L473](https://github.com/pytroll/satpy/blob/main/satpy/readers/ahi_hsd.py?rgh-link-date=2024-02-12T10%3A55%3A34Z#L473). Here, the offset dt only accounts for the start of the scan, not the total scan duration.

I'm having trouble understanding this sentence, sorry. When you say "observation_end_time" are you referring to `self.observation_end_time` or the `self.basic_info["observation_end_time"]` that comes from the file?

djhoese:
The new code I think I do understand and looks good to me. One fear is that if the observation times (start or end) coming from the file are ever before the theoretical nominal time then the rounding will give the wrong answer. This isn't something new introduced in this PR, but correct me if I'm wrong:

FLDK theoretical start at 12:00:00. Time in file is 11:59:59.999. Replacing the start time hours/minutes/seconds would put this at 11:50:00, wouldn't it? Or does the `timeline` variable fix this? Or do we assume that this is impossible? Meaning, the instrument isn't triggered/scheduled to make an observation until 12:00:00 so it could never have a time before this?

sfinkens:
> I'm having trouble understanding this sentence, sorry. When you say "observation_end_time" are you referring to `self.observation_end_time` or the `self.basic_info["observation_end_time"]` that comes from the file?

I meant `self.observation_end_time`, sorry for not being clear. I'll try with more words :) To compute the nominal end time, the reader currently takes `self.observation_end_time` and replaces hours/minutes/seconds with the "timeline". Then adds an offset `dt` to get from the "timeline" to the actual start of the scan.


> FLDK theoretical start at 12:00:00. Time in file is 11:59:59.999. Replacing the start time hours/minutes/seconds would put this at 11:50:00, wouldn't it? Or does the `timeline` variable fix this? Or do we assume that this is impossible? Meaning, the instrument isn't triggered/scheduled to make an observation until 12:00:00 so it could never have a time before this?

Oh, nice catch! I think the 11:59 case would still work because hours/minutes/seconds are snapped to the timeline variable. But at midnight it doesn't work anymore, for example

```
timeline = 0000
observation end time = 2023-01-01 23:59
=> nominal time = 2023-01-01 00:00 (instead of 2023-01-02)
```

I'll try to take that into account

sfinkens:
>  Or do we assume that this is impossible? Meaning, the instrument isn't triggered/scheduled to make an observation until 12:00:00 so it could never have a time before this?

I wouldn't rely on that

mraspaud (approved):
LGTM!

satpy/readers/ahi_hsd.py
```diff
@@ -775,3 +743,96 @@ def _ir_calibrate(self, data):
         c2_ = self._header["calibration"]["c2_rad2tb_conversion"][0]
 
         return (c0_ + c1_ * Te_ + c2_ * Te_ ** 2).clip(0)
+
+
+class NominalTimeCalculator:
```
djhoese, line 748: Can we make this private so we don't have to worry about someone trying to use it outside this module?

```suggestion
class _NominalTimeCalculator:
```

sfinkens: Good idea, done

djhoese:
Awesome job. It is surprising and almost sad how much code had to go into get a start and end time, but it is what it is. Thanks for putting all this work into it.

sfinkens:
> Awesome job. It is surprising and almost sad how much code had to go into get a start and end time, but it is what it is. Thanks for putting all this work into it.

Haha, True. But I enjoyed it somehow :smiley:

PR #2754, opened 2024-03-08
Bugfix vgac reader

ninahakansson:
Fix a bug I intruded when writing unit test for the time var in the vgac reader. The time_proj0 is a scalar in the files, and now also in the unittest case.

 - [x] Tests added


mraspaud (approved):
LGTM

PR #2760, opened 2024-03-13
Fix instrument name reading from attributes for VIIRS L2 data

pnuu:
The some of the attributes in the CSPP/ASCI files have been renamed between v3r0 and v3r2. The only on I've seen that affects Satpy is the instrument name. The [original reader version](https://github.com/pytroll/satpy/blob/v0.43.0.post0/satpy/readers/viirs_l2.py#L58) was based on older data and works at least until v3r0 files, but the current one seems to handle only the newer files. With this PR both can be read.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/viirs_l2.py
+++ b/satpy/readers/viirs_l2.py
@@ -70,7 +70,13 @@ class VIIRSL2FileHandler(NetCDF4FileHandler):
     @property
     def sensor_name(self):
         """Get sensor name."""
-        return self["/attr/instrument"].lower()
+        try:
+            # Until v3r0 (or v3r1)
+            instrument = self["/attr/instrument_name"].lower()
+        except KeyError:
+            # Changed in ASCI v3r1 or v3r2
+            instrument = self["/attr/instrument"].lower()
+        return instrument
 
     def _get_dataset_file_units(self, ds_info, var_path):
         file_units = ds_info.get("units")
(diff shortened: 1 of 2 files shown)


pnuu:
Any ideas what would be the best way to adapt the tests to check for both attribute versions? Dupplicating the fake filehandler class and the test class doesn't seem very attractive...

pnuu:
Ok, duplicated the minimal amount of test code I thought was necessary.

djhoese:
CC @wjsharpe, Does this look good to you?

djhoese:
Posting the realization from slack here for the record: @pnuu was using the old "viirs_l2.py" reader that he had created a while ago. A couple months ago I reworked all the VIIRS EDR (note EDR, not L2) to be one `viirs_edr` reader. The new VIIRS L2 reader being modified here is from #2740 by @wjsharpe and is for the NASA VIIRS L2 products *not* the NOAA EDR products.

pnuu:
I should've remembered that `viirs_l2` -> `viirs_edr` change. I'll test the v3r2 and v3r0 files tomorrow with the current Satpy and close this if both already work.

pnuu:
Yeah, this PR was a mixup with the old `viirs_l2.py` and the current `viirs_edr.py` for the ASCI files. Closing, `viirs_edr` already works with both of the versions I have.

PR #2776, opened 2024-04-07
Add option to choose start time to MSI SAFE reader

simonrp84:
Currently, the `msi_safe` reader uses the `filename_info` to get the `start_time` for a Scene. This can be problematic as multiple tiles can have the same `start_time` and will hence overwrite each other:
```
for files in [[tile1_files], [tile2_files]]:
   scn = Scene(files, reader='msi_safe')
   scn.load(['true_color'])
   scn.save_dataset('true_color')
```
The user would expect that two images are produced from this, one from each set of files. However, in many cases the second tile will overwrite the image saved by the first.

This PR attempts to fix the problem by allowing an optional `reader_kwarg` that allows the user to specify that the `start_time` should be read from the tile metadata rather than the filename:
```
for files in [[tile1_files], [tile2_files]]:
   scn = Scene(files, reader='msi_safe', reader_kwargs={'use_tile_time': True})
   scn.load(['true_color'])
   scn.save_dataset('true_color')
```
This will generate two output images as expected.

 - [x] Tests added
 - [x] Fully documented


mraspaud (approved):
LGTM, thanks for adding this functionality.

mraspaud (approved):
LGTM!

PR #2779, opened 2024-04-13
Convert Sentinel-2 MSI sensor name to lowercase in the reader YAML config file and add support for "counts" calibration

yukaribbba:
This PR has two purposes:
1. The current sensor name in YAML is "MSI", which triggers a minor warning `Inconsistent sensor/satellite input` in pyspectral since the latter one stores sensor name in lowercase. So it's replace by lowercase.
2. Add "counts" calibration just like other sensors.

### DIFF
--- a/satpy/etc/readers/msi_safe.yaml
+++ b/satpy/etc/readers/msi_safe.yaml
@@ -38,7 +38,7 @@ datasets:
         units: W m-2 um-1 sr-1
       counts:
         standard_name: counts
-        units: "1"        
+        units: "1"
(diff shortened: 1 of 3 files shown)


djhoese (approved):
I'm surprised the sensor is defined per dataset, but that's not your doing so don't worry about it. This looks good to me, but I'd like someone else who has more experience with this data hit the merge button.

@mraspaud ?

mraspaud (approved):
LGTM, thanks for making the counts  available and making tests for it.

PR #2784, opened 2024-04-19
Fix various issues in unstable CI

djhoese:
cftime's master branch should support numpy 2 now, but from what I could tell our CI still wasn't working. My best guess is the conda-installed version is getting used before the pip one. This PR first removes the conda installed one in unstable CI before installing from source.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese:
@mraspaud Ok I've got CI working in unstable as far as numpy 2 goes and what we have control over. The remaining things are:

1. The xarray dtype/scalar handling that I have a PR for in xarray.
2. Things based on numba fail since numba does not support numpy 2 yet.

I think this is ready for merge.

mraspaud (approved):
LGTM

Issue #2752, opened 2024-03-06
Replacement of deprecated `datetime.datetime.utcnow()`


pnuu:
## Feature Request

As of Python 3.12 the usage of `datetime.datetime.utcnow()` and `.utcfromtimestamp()` are deprecated and will be removed in a later version. We need to address this in Satpy (and in all the other Pytroll packages).

**Describe the solution you'd like**

There are some options I can think of:

1. use naive datetimes and leave it to the user to handle problems with `.now()` returing time their computer is set to
2. use naive datetimes and force UTC times for the process (`os.environ["TZ"]; time.tzset()`)
3. start using timezone-aware datetimes and use `dt.datetime.now(dt.datetime.timezone.utc)`

The first option is problematic in many ways if the computer is not in UTC, one of which is the unpredictability of the tests. Forcing local time to UTC might cause weird problems if the user wants to use real local time for some part of the processing.

So we have the third option and make all the datetimes in Satpy (and Pytroll in general) timezone-aware. This was the option we kind of agreed in the last Pytroll monthly meeting, but the attendance was pretty limited. It'll require some (lots of?) work to make sure all the datetime stuff still work as intended.

**Describe any changes to existing user workflow**
Depends on the option. With the third option should only affect developers.

**Additional context**
Some links that discuss the deprecation:
* https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated
* https://github.com/python/cpython/issues/103857 (see also the linked https://blog.ganssle.io/articles/2019/11/utcnow.html)

Ping: @djhoese @mraspaud @gerritholl @simonrp84 @ameraner @sfinkens

simonrp84:
The third option seems like the most sensible, if work-intensive, option. I can imagine the first two options causing potential unexpected issues for users with stange setups etc.

gerritholl:
Agree that option three is the best alternative here.  I remember now why I used option three in the ninjogeotiff writer: I was blissfully unaware of `utcnow` :smile: 

https://github.com/pytroll/satpy/blob/2c27b35081c59e27fbc03d2f9e3df14fd8ffa215/satpy/writers/ninjogeotiff.py#L401

pnuu:
> It could affect (monitoring) tools that interpret logfiles.

The log formatting can be tuned by the user if necessary. I this mostly affects Trollflow2 and the collectors.

> It could also change the format in which the CF writer writes timestamps.

[...]

PR #2787, opened 2024-04-22
Fix datetime imports

pnuu:
In many places, we have used `from datetime import datetime`. This is bad, since it masks the module for further imports:

```python
from datetime import datetime
from datetime.timezone import utc
```

Sorting these the other way around will cause complaints from flake8/ruff.

This PR changes all of these import to `import datetime as dt`, so we need on the single import to access other items:

```python
import datetime as dt

dt.datetime(1980, 3, 19)
dt.timedelta(minutes=42)
now = dt.datetime.now(dt.timezone.utc)
```

This is a change that makes implementing the timezone aware datetimes of https://github.com/pytroll/satpy/issues/2752 easier.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


satpy/tests/reader_tests/test_seviri_l1b_icare.py
```diff
@@ -133,7 +133,7 @@ def test_load_dataset_vis(self):
         datasets = r.load(["VIS008"])
         assert len(datasets) == 1
         for v in datasets.values():
-            dt = datetime(2004, 12, 29, 12, 27, 44)
+            dt = dt.datetime(2004, 12, 29, 12, 27, 44)
```
sfinkens, line 136: This overrides the module with a timestamp

sfinkens (changes requested):
Oh, nasty... Thanks for fixing this!

Variables previously named `dt` are now called `date`. I think this might be misleading in some cases. Sometimes it's a date, sometimes a timestamp with more precision, sometimes an offset or a timedelta. Obviously the previous name wasn't any better, so feel free to ignore this if takes too much time :)

satpy/tests/reader_tests/test_seviri_l1b_icare.py
pnuu: That only fixed the one in the snippet, I'll have a look at what each of the `dt` -> `date` actually mean and update accordinly.

pnuu: And the somewhat better naming in https://github.com/pytroll/satpy/pull/2787/commits/37301177ea7bded84500ca596c82e621a1b1ef4d

sfinkens (approved):
Thanks for all the hard work!

PR #2788, opened 2024-04-22
Add an optional kwarg to CloudCompositor for RGBA output

yukaribbba:
Usually `CloudCompositor` produces an LA mode image. But sometimes you may want RGBA result. So this PR will add "mode" kwarg just like `MaskingCompositor`.

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -1033,8 +1033,10 @@ class RealisticColors(GenericCompositor):
 class CloudCompositor(GenericCompositor):
     """Detect clouds based on thresholding and use it as a mask for compositing."""
 
+    _supported_modes = ["LA", "RGBA"]
+
     def __init__(self, name, transition_min=258.15, transition_max=298.15,  # noqa: D417
-                 transition_gamma=3.0, invert_alpha=False, **kwargs):
+                 transition_gamma=3.0, invert_alpha=False, mode="LA", **kwargs):
         """Collect custom configuration values.
 
         Args:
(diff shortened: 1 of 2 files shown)


djhoese:
Could you explain more why/when this is needed? Normally writers or other compositors are supposed to handle these cases for you.

yukaribbba:
My fault. Didn't realize that. Closing now

djhoese:
If you have a use case where this isn't true let me know. There is very likely a case where things won't work, I just didn't know what that situation would be.

yukaribbba:
So `GenericCompositor` isn't helpful since `CloudCompositor` will add a "band" dimension which is not present in the first two. Therefore triggers the "incompatible areas". So I turn to MaskingCompositor and BackgroundCompositor.

```
  masking:
    compositor: !!python/name:satpy.composites.MaskingCompositor
    prerequisites:
      - name: C14
      - name: C14
    conditions:
      - method: less_equal
        value: 0
        transparency: 100
      - method: greater
        value: 0
        transparency: 100
      - method: isnan
        transparency: 100
    mode: RGBA
    standard_name: no_enhancement_spec

  ir_cloud_day_default_2000:
    compositor: !!python/name:satpy.composites.BackgroundCompositor
    prerequisites:
      - name: masking
      - compositor: !!python/name:satpy.composites.CloudCompositor
        prerequisites:
          - name: C14
    standard_name: ir_cloud_day_default
```

So the logic is: lay a completely transparent RGBA image upon the LA image(day cloud) then you get the RGBA result. It's kinda magical but anyway I got what I want and no changes needed.

djhoese:
Why is C14 being used here? Is the final result you want just a geotiff with the cloud compositor output as RGBA? And you need that so more image viewers will show the image correctly?

yukaribbba:
Yes I want a geotiff in RGBA mode so that my image viewer can show it properly. C14 is used just for convenience since cloudcompositor will also use that band. Actually it can be any bands since all I need from that is just a transparent image . I'm not sure if other image viewers have such problems but in my Windows pc that is certain. Could just be a rare case, who knows...

djhoese:
I think this is a good feature, but should maybe not be added to just the CloudCompositor. I wonder if there is a way to do this with enhancements. @pnuu @simonrp84 or @mraspaud any ideas how to take the output of a composite that is LA and produce an RGBA without modifying Satpy or trollimage?

pnuu:
The first option that comes to mind is to create a composite that has C14 on all 3 channels and then use that with `CloudCompositor`.

yukaribbba:
@pnuu Unfortunately `CloudCompositor` can only accept a single band otherwise it'll end it up like this:
```
<xarray.DataArray 'bands' (bands: 6)> Size: 24B
array(['R', 'G', 'B', 'R', 'G', 'B'], dtype='<U1')
Coordinates:
    crs      object 8B PROJCRS["unknown",BASEGEOGCRS["unknown",DATUM["Unknown...
  * bands    (bands) <U1 24B 'R' 'G' 'B' 'R' 'G' 'B'
```

PR #2789, opened 2024-04-22
Activate LI L2 accumulated products gridding by default

ameraner:
This PR changes the default behaviour of the LI L2 reader, such that the accumulated products are delivered automatically as 2-d arrays.

**Background and current state**:
The LI L2 products consist of two types of products: 
- point products (LE, LEF, LFL, LGR): "classic" lightning products consisting of values with a lat-lon 
- accumulated (and gridded) products (AF, AFR, AFA): these products are the result of temporal accumulation (e.g. 30 seconds), and have been regridded to the FCI 2km grid by the ground processor, in order to facilitate the synergistic usage together with FCI.

Now, both products are actually stored in the NetCDF files as 1-d arrays. In the case of the accumulated products, each entry in the 1-d array corresponds to a specific pixel in the FCI grid (i.e. the corresponding column and row number are given in the file). 

The mapping of these points onto the FCI grid, to obtain a 2-d array with an `AreaDefinition`, is already implemented in the reader, but needs to be activated by passing the reader kwarg `with_area_definition=True`.  **The default value of this is currently set to `False`.** This is because, during the first implementation of the reader, it was decided to follow the rationale of "the reader shall return data as it's stored in the file and not modify it".

**Why do we want to change this (set `with_area_definition=True` as default)?**
The reasons are multiple:
- the accumulated products are designed and computed to be used in the grid. It is more intuitive for a user if the reader returns an accumulated product already in the 2-d grid, ready to be used, rather than a 1-d array that needs remapping in a specific way.
- activating the automatic regridding by default avoids the risk of users resampling the 1-d points to a 2-d grid with methods that can introduce artifacts (e.g. nearest or bilinear that introduce "pixel bleeding" effects)
- the usage of `reader_kwargs` through the `Scene` object is a rather advanced operation, and requires the user to fully find and understand the reader documentation, which is not always trivial. It also introduces an extra step that is otherwise not required in the standard Satpy API usage.
- It simplifies the usage and combination with FCI, e.g. by allowing the generation of multi-sensor composites out-of-the box with the same handling of the FCI data.
- Advanced users that would like to keep the data as 1-d arrays for further customised processing, can of course still do so via the `Scene` `reader_kwargs` interface.

For these reasons, we think that this is a good exception to the "reader doesn't modify data" guideline, since it applies an operation to the data that is anyway expected to be performed for the product to be usable, improving the user-friendliness.

Note that LI L2 accumulated products have not been widely distributed so far, and the most attention for these will start with the LI pre-operational dissemination (hopefully in a couple of months time), so we think it's a good time to introduce this now.

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/li_l2_nc.py
+++ b/satpy/readers/li_l2_nc.py
@@ -46,7 +46,7 @@ CHUNK_SIZE = get_legacy_chunk_size()
 class LIL2NCFileHandler(LINCFileHandler):
     """Implementation class for the unified LI L2 satpy reader."""
 
-    def __init__(self, filename, filename_info, filetype_info, with_area_definition=False):
+    def __init__(self, filename, filename_info, filetype_info, with_area_definition=True):
         """Initialize LIL2NCFileHandler."""
         super(LIL2NCFileHandler, self).__init__(filename, filename_info, filetype_info)
 
--- a/satpy/readers/li_l2_nc.py
+++ b/satpy/readers/li_l2_nc.py
@@ -13,34 +13,50 @@
 # You should have received a copy of the GNU General Public License
 # along with satpy.  If not, see <http://www.gnu.org/licenses/>.
 
-"""MTG Lightning Imager (LI) L2 unified reader.
+"""MTG Lightning Imager (LI) Level-2 (L2) unified reader.
 
 This reader supports reading all the products from the LI L2
 processing level:
 
+Point products:
   * L2-LE Lightning Events
   * L2-LEF Lightning Events Filtered
   * L2-LFL Lightning Flashes
   * L2-LGR Lightning Groups
+Accumulated products:
   * L2-AF Accumulated Flashes
   * L2-AFA Accumulated Flash Area
   * L2-AFR Accumulated Flash Radiance
 
-Point-based products (LE, LEF, LFL, LGR) are provided as 1-D arrays, with a ``pyresample.geometry.SwathDefinition`` area
+Per default, the unified LI L2 reader returns the data either as an 1-D array
+or as a 2-D array depending on the product type.
+
+Point-based products (LE, LEF, LFL, LGR) are "classic" lightning products
+consisting of values with attached latitude and longitude coordinates.
+Hence, these products are provided by the reader as 1-D arrays,
+with a ``pyresample.geometry.SwathDefinition`` area
 attribute containing the points lat-lon coordinates.
 
-Accumulated products (AF, AFA, AFR) are provided as 2-D arrays in the FCI 2km grid as per intended usage,
-with a ``pyresample.geometry.AreaDefinition`` area attribute containing the grid geolocation information.
+Accumulated products (AF, AFA, AFR) are the result of temporal accumulation
+of events (e.g. over 30 seconds), and are gridded in the FCI 2km geostationary
+projection grid, in order to facilitate the synergistic usage together with FCI.
+Compared to the point products, the gridded products also give information
+about the spatial extent of the lightning activity.
+Hence, these products are provided by the reader as 2-D arrays in the FCI 2km
+grid as per intended usage, with a ``pyresample.geometry.AreaDefinition`` area
+attribute containing the grid geolocation information.
 In this way, the products can directly be overlaid to FCI data.
-If needed, the products can still be accessed as 1-d array by setting the reader kwarg ``with_area_definition=False``,
-eg::
+If needed, the accumulated products can also be accessed as 1-d array by
+setting the reader kwarg ``with_area_definition=False``,
+e.g.::
 
   scn = Scene(filenames=filenames, reader="li_l2_nc", reader_kwargs={'with_area_definition': False})
 
-The lat-lon coordinates of the points/grid pixels can be accessed using e.g.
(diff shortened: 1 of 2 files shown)


satpy/readers/li_l2_nc.py
```diff
@@ -13,18 +13,36 @@
...
-  * L2-AF
-  * L2-AFR
+  * L2-LE Lightning Events
+  * L2-LEF Lightning Events Filtered
+  * L2-LFL Lightning Flashes
+  * L2-LGR Lightning Groups
+  * L2-AF Accumulated Flashes
+  * L2-AFA Accumulated Flash Area
+  * L2-AFR Accumulated Flash Radiance
+
```
sjoro, line 28: i'd like to have a bit more "meat" and documentation in here for the reader... something like:
"The unified LI L2 reader returns the data either as an 1-D array or as a 2-D array depenging on the product type."

satpy/readers/li_l2_nc.py
```diff
@@ -13,18 +13,36 @@
...
+Point-based products (LE, LEF, LFL, LGR) are provided as 1-D arrays, with a ``pyresample.geometry.SwathDefinition`` area
+attribute containing the points lat-lon coordinates.
+
+Accumulated products (AF, AFA, AFR) are provided as 2-D arrays in the FCI 2km grid as per intended usage,
+with a ``pyresample.geometry.AreaDefinition`` area attribute containing the grid geolocation information.
+In this way, the products can directly be overlaid to FCI data.
+If needed, the products can still be accessed as 1-d array by setting the reader kwarg ``with_area_definition=False``,
```
sjoro, line 35: a small clarification suggestion: "If needed, the accumulated products can also be..."

sjoro (changes requested):
thank you @ameraner for the PR! tiny suggested editorials for the docstrings, otherwise LGTM.

sjoro:
i agree with @ameraner and have discussed this change with him earlier, so i'm good with this change.

pnuu (approved):
LGTM.

gerritholl:
I fully support that this should be true by default.  We are already modifying the data in most readers, because by default, we use `calibration=radiance` or `calibration=brightness_temperature`, not `calibration=counts`, which would be the data as they are in the file.

But I'm not sure if this should be a reader kwarg, or a DataID key (like for calibration or polarisation).  Probably I'm rather late with this thought because the gridding feature already exists, just isn't activated by default.

pnuu:
> activating the automatic regridding by default avoids the risk of users resampling the 1-d points to a 2-d grid with methods that can introduce artifacts (e.g. nearest or bilinear that introduce "pixel bleeding" effects)

How big of an issue do you think the pixel bleeding will be when resampling the 2D array to a local area definition? Or should that be ideally done from the non-gridded data (LFL)?

satpy/readers/li_l2_nc.py
ameraner: Done

satpy/readers/li_l2_nc.py
ameraner: Done

ameraner:
> How big of an issue do you think the pixel bleeding will be when resampling the 2D array to a local area definition? Or should that be ideally done from the non-gridded data (LFL)?

That is indeed a good point, and it's a bit tricky... My understanding:

The accumulated products show the extent of the measured lightning events in the FCI grid. Therefore, ideally, one would stay in the native geostationary projection (and possibly use the `native` resampler to up/downscale the grid if needed) to avoid resampling artefacts. If the projection needs to be changed, any of the existing resampling algorithms would end up changing the area covered by the lightnings eventually (since we map pixels between grids using only the pixel center position, and not the pixel footprint area - correct me if I'm wrong here). The bucket resampler would do the best job, in that at least it avoids the pixel bleeding effect, which could artificially increase the area covered by the lightning, possibly causing a wrong interpretation of the scene.

For the point products: the groups and flashes products represent spatial and temporal clusters of events; however, we only get one lat-lon coordinate for each group or flash (i.e. the average location of the group/flash) hence we anyway lose the spatial extent information. For this reason, I think that also in this case the bucket resampler does the best job, since it maps one group/flash to exactly one pixel in the target grid, so-to-say maintaining the "lat-lon marker" functionality. 

I hope this is understandable/makes some sense. Otherwise I'm happy to discuss this further with examples and in front of a flipchart, e.g. at the user days :) What is sure is that there's still a lot to be experimented with using this data!

pnuu:
Thanks @ameraner for the insight! I just today got my MTG/LI processing chain completely functional (needed the new Satpy version for S3 support), and I'm using `resampler='bucket_count'` to aggregate the LFL data to a 4 km grid. I'm putting the resulting images to WMS so they can be used as an overlay with any data.

I'll include my LI setup in my PUD presentation in June.

sjoro:
@pnuu it can be pretty significant, we see this when visualising GII BUFR data, which is essentially "point data" and needs to be put in an MSG grid before viewing the result. i did the first experiments with "nearest neighbor" resampling and it all depends on the search radius... too big and we get round blobs of data bleeding to neighboring pixels, too small and we get gaps in the data. bucket resampling does the best job, or simply converting the given lonlats first to rows and cols and use those to index the data points to correct location.

PR #2796, opened 2024-05-03
Add support for GK-2B/GOCI2 data at L1b/c

simon-sat:
This PR adds support for level 1 data produced by the GOCI-II ocean color instrument flying aboard GeoKompsat-2A.

Data is produced by the sensor in a series of sections that, together, cover either an area around Korea or the full disk. Each section is sensed at a different time. There is also a data product joining these sections together to create a full-area image at L1c.

Right now, this PR is a draft as:
1) I haven't added any tests
2) The full disk data is not read successfully.
3) I need to add a proper area definition, which should be possible from the L1 attributes (the data is in GEOS projection). Currently I cannibalised code from the L2 GOCI reader that generates a swathdefinition.

 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/etc/composites/goci2.yaml
+++ b/satpy/etc/composites/goci2.yaml
@@ -1,0 +1,148 @@
+sensor_name: visir/goci2
+
+
+modifiers:
+
+  rayleigh_corrected:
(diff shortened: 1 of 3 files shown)


simonrp84:
pre-commit.ci autofix

PR #2799, opened 2024-05-11
Add netcdf4 to goci2 optional dependency in `pyproject.toml`

Isotr0py:
- This PR adds `goci2` optional dependency with `netCDF4 >= 1.1.8` since we used `NetCDF4FileHandler` for GOCI2 reader.


mraspaud (approved):
LGTM

PR #2802, opened 2024-05-17
Adapt reader mviri_l1b_fiduceo_nc

bkremmli:
This PR fixes the mviri_l1b_fiduceo_nc reader when being used with a new xarray version (2024.3.0). When using the original reader, a ValueError about not being able to decode the times is thrown. The file is now opened without decoding. The decoding is now done in DatasetWrapper()._decode_cf(). The time is decoded separatly from the other data values. FillValues for the time are recognized to replace time values with NaT and time is decoded using the offset values included within the attributes.

Also, opening the dataset using chunks is deactivated because the input files contains dimensions of the same name which cannot be processed by xarray at the moment. The chunking as well as renaming the dimensions is also performed in DatsetWrapper().

 - [x ] Tests added 
 - [ ] Fully documented --> this change should not be visible to users
 - [x ] Add your name to `AUTHORS.md` if not there already


satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -513,11 +514,19 @@ def get_time(self):
...
+            time = self["time_ir_wv"]
         except KeyError:
-            return self["time"]
+            time = self["time"]
+
+        timeOffset = time.attrs["add_offset"]
+        condition = time == time.attrs["_FillValue"]
+
+        time = xr.where(condition, numpy.nan, time + timeOffset)
+        time = (time*1e9).astype("datetime64[ns]")
```
mraspaud, line 528: how about
```suggestion
        time = time.astype("datetime64[s]").astype("datetime64[ns]")
```
this way you don't risk integer overflow

mraspaud (changes requested):
Thanks for adding this feature! I'm no expert here, so problably @sfinkens needs to have a look, but I wrote some comments already :)

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -555,11 +564,14 @@ def __init__(self, filename, filename_info, filetype_info,  # noqa: D417
...
         nc_raw = xr.open_dataset(
             filename,
-            chunks={"x": CHUNK_SIZE,
-                    "y": CHUNK_SIZE,
-                    "x_ir_wv": CHUNK_SIZE,
-                    "y_ir_wv": CHUNK_SIZE}
+            # chunks={"x": CHUNK_SIZE,
+            #         "y": CHUNK_SIZE,
+            #         "x_ir_wv": CHUNK_SIZE,
+            #         "y_ir_wv": CHUNK_SIZE},
```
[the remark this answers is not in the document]
bkremmli: Thanks for the first comments!
Concerning the chunks, it throws a ValueError at the moment: "ValueError: This function cannot handle duplicate dimensions, but dimensions {'srf_size'} appear more than once on this object's dims: ('srf_size', 'srf_size')"

It could be related to this post: https://github.com/pydata/xarray/issues/8579

Any other suggestions than disabling it are very welcome :)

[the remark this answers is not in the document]
sfinkens: In that discussion they propose a workaround:
```
ds.variables["covariance..."].dims = ("srf_size_1", "srf_size_2")
ds.chunk(mychunks)
```
I tried that here: https://github.com/sfinkens/satpy/commit/82eb8cd67c6d7119b1cabeb4b34683b1c1f8b47e Let me know what you think!

I also updated the tests to trigger that situation:
1) Add a new variable with identical dimension names
2) Write a fake file to disk
3) Read it back in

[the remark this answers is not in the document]
sfinkens: There's still a bit of work to be done, because the tests now fail due to some extra attributes and coordinates.

sfinkens (commented):
Nice work @bkremmli! Thanks for updating the tests, so that the problem is triggered.

sfinkens:
Also I noticed that space pixels now have some finite values (instead of NaN), because `decode_cf=False`. You can use `decode_cf=True` together with `time.encoding["add_offset/_FillValue"]`, see https://github.com/sfinkens/satpy/commit/7045a87cc5e8066712df153ba4e9998c06df6265

satpy/readers/mviri_l1b_fiduceo_nc.py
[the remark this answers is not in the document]
bkremmli: Thank you for your feedback! As already discussed, I will include your proposed changes. Concerning the tests, I am currently working on it.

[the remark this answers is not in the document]
bkremmli: @sfinkens I submitted some changes and the tests are mostly ok now. The test "test_reassign_coords(self)" still fails - as I am not familiar with the Mock Object Library, it would be great if you could have a look on it - thank you!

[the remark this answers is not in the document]
sfinkens: Ok, I'll check after lunch!

sfinkens (commented):
Nice work, almost there! 

I think in the tests you need to write the dataset to disk and read it back in. Otherwise the dask error with duplicate dimensions is not triggered.

satpy/readers/mviri_l1b_fiduceo_nc.py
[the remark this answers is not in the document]
bkremmli: sfinkens and I did some pair programming: Calling time.astype("datetime64[s]").astype("datetime64[ns]") is not properly working because a float input is needed. We propose to open the dataset with decode_cs=False and afterwards decode time and other variables separatly to properly take care of time FillValues and offsets.

sfinkens added a test for the Interpolator.

The test "test_reassign_coords() is still failing because the new functionalities in DatasetWrapper() are not yet considered (mocking should not be of importance here). Maybe we should call the assign_coords() method directly instead of DatasetWrapper().

Also, a test for DatasetWrapper should be included to test the separate encoding.

satpy/readers/mviri_l1b_fiduceo_nc.py
bkremmli: see conversation above

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -459,6 +457,29 @@ def __init__(self, nc):
         """Wrap the given dataset."""
         self.nc = nc
 
+        # remove time before decoding and add again.
+        raw_time = nc["time_ir_wv"]
+        self.nc = self.nc.drop_vars(["time_ir_wv"])
+        self.nc = xr.decode_cf(self.nc)
+        self.nc["time_ir_wv"] = raw_time
```
sfinkens, line 464: Can you please extract that into a separate method, for example `DatasetWrapper._decode_cf()`

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -459,6 +457,29 @@ def __init__(self, nc):
...
+
+    def _chunk(self, nc):
+
+        (chunk_size, chunk_size) = nc.variables["quality_pixel_bitmask"].encoding["chunksizes"]
+        chunks = {
+            "x": chunk_size,
+            "y": chunk_size,
+            "x_ir_wv": chunk_size,
+            "y_ir_wv": chunk_size
+        }
```
sfinkens, line 480: I'd prefer separate x/y chunksizes

```suggestion
        (chunk_size_y, chunk_size_x) = nc.variables["quality_pixel_bitmask"].encoding["chunksizes"]
        chunks = {
            "x": chunk_size_x,
            "y": chunk_size_y,
            "x_ir_wv": chunk_size_x,
            "y_ir_wv": chunk_size_y
        }
```

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -513,11 +534,17 @@ def get_time(self):
...
         """
         try:
-            return self["time_ir_wv"]
+            time = self["time_ir_wv"]
         except KeyError:
-            return self["time"]
+            time = self["time"]
+
+        time = xr.where(time == time.attrs["_FillValue"], np.datetime64("NaT"),
+                        (time + time.attrs["add_offset"]).astype("datetime64[s]").astype("datetime64[ns]"))
```
sfinkens, line 545: Is it possible to do this once after decoding?

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
```diff
@@ -403,6 +416,7 @@ def test_get_dataset(self, file_handler, name, calibration, resolution,
             xr.testing.assert_allclose(ds, expected)
             assert ds.dtype == expected.dtype
             assert ds.attrs == expected.attrs
+            assert True
```
sfinkens, line 419: Debugging leftover?

satpy/readers/mviri_l1b_fiduceo_nc.py
sfinkens: BTW, I just noticed that in `get_time` there's a note:

> Variable is sometimes named "time" and sometimes "time_ir_wv"

I guess this would have to be handled here, too.

sfinkens: Maybe just renaming to a common time variable name is sufficient

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
```diff
@@ -28,6 +28,7 @@
 import xarray as xr
 from pyproj import CRS
 from pyresample.geometry import AreaDefinition
+from pytest_lazyfixture import lazy_fixture
```
sfinkens, line 31: This should fix the tests

```suggestion
from pytest_lazy_fixtures import lf as lazy_fixture
```

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
bkremmli: True - sorry for that, will be fixed shortly.

satpy/readers/mviri_l1b_fiduceo_nc.py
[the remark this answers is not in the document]
bkremmli: Test is working again, DatasetWrapper is implicitly tested within the other tests, so no separate test is needed.

satpy/readers/mviri_l1b_fiduceo_nc.py
bkremmli: changed according to the results of our discussion last week.

bkremmli:
The test for the duplicate dimensions still needs to be added.

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -455,10 +452,40 @@ def is_high_resol(resolution):
...
+        self.nc = xr.decode_cf(self.nc)
+        self.nc[time.name] = (time_dims, time.values)
+
+    def _fix_duplicate_dimensions(self, nc):
+        nc.variables["covariance_spectral_response_function_vis"].dims = ("srf_size_1", "srf_size_2")
+        self.nc = nc.drop_dims("srf_size")
+
+    def _chunk(self, nc):
+
+        (chunk_size_y, chunk_size_x) = nc.variables["quality_pixel_bitmask"].encoding["chunksizes"]
```
sfinkens, line 480: Can you please add a comment here, that accessing the encoding doesn't load the data into memory? (Martin just raised the concern that this loads the bitmask array into memory. We tested and it doesn't)

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -558,13 +586,24 @@ def __init__(self, filename, filename_info, filetype_info,  # noqa: D417
...
 
         # Projection longitude is not provided in the file, read it from the
         # filename.
-        self.projection_longitude = float(filename_info["projection_longitude"])
+        if "." in str(filename_info["projection_longitude"]):
+            self.projection_longitude = float(filename_info["projection_longitude"])
+        else:
+            self.projection_longitude = (
+                float(filename_info["projection_longitude"][:2] + "." + filename_info["projection_longitude"][2:])
+            )
```
sfinkens, line 606: This could be moved to a separate method `get_projection_longitude`.

sfinkens: In the second case, I think you could do 

```python
float(filename_info["projection_longitude"]) / 100
```

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -558,13 +586,24 @@ def __init__(self, filename, filename_info, filetype_info,  # noqa: D417
...
-                    "y_ir_wv": CHUNK_SIZE}
+                    "y_ir_wv": CHUNK_SIZE},
+            decode_cf=False,
+            decode_times=False,
+            mask_and_scale=False,
         )
+
         self.nc = DatasetWrapper(nc_raw)
+        # decode times and fix duplicate dimensions
+        self.nc.prepare_input()
```
sfinkens, line 597: I'd prefer this to be done in `DatasetWrapper.__init__`

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
```diff
@@ -547,17 +560,44 @@ def test_file_pattern(self, reader):
...
         files = reader.select_files_from_pathnames(filenames)
         # only 3 out of 4 above should match
-        assert len(files) == 3
+        assert len(files) == 6
 
 
 class TestDatasetWrapper:
     """Unit tests for DatasetWrapper class."""
 
+    def test_fix_duplicate_dimensions(self):
```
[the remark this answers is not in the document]
sfinkens: If `prepare_input` is called in `DatasetWrapper.__init__` (see my other comment) I think the test is fine.

satpy/readers/mviri_l1b_fiduceo_nc.py
bkremmli: sounds good

satpy/readers/mviri_l1b_fiduceo_nc.py
bkremmli: Thanks for reviewing!
I moved it back into DatasetWrapper.__init__ and adapted test_fix_duplicate_dimensions().
However, test_reassign_coords() now fails as it expects data to be time decoded - this was the original reason why I put it into a separate function. It should be possible to adapt the test, but the test case would get much bigger and test everything in DatasetWrapper.__init__ instead of only the dimension reassignment.
Any better idea is welcome ;)

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -459,6 +461,12 @@ def __init__(self, nc):
         """Wrap the given dataset."""
         self.nc = nc
 
+        # decode data
```
sfinkens, line 464:
```suggestion

```

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -558,13 +588,16 @@ def __init__(self, filename, filename_info, filetype_info,  # noqa: D417
             chunks={"x": CHUNK_SIZE,
                     "y": CHUNK_SIZE,
                     "x_ir_wv": CHUNK_SIZE,
-                    "y_ir_wv": CHUNK_SIZE}
+                    "y_ir_wv": CHUNK_SIZE},
+            decode_cf=False,
```
sfinkens, line 592:
```suggestion
            # See DatasetWrapper for why this is disabled
            decode_cf=False,
```

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
```diff
@@ -36,13 +37,16 @@
...
     FiduceoMviriFullFcdrFileHandler,
+    Interpolator,
 )
 from satpy.tests.utils import make_dataid
 
 # NOTE:
 # The following fixtures are not defined in this file, but are used and injected by Pytest:
 # - request
 
+fill_val = int("4294967295")
```
sfinkens, line 48: Why is it defined as string and then converted to int?

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
```diff
@@ -547,17 +568,51 @@ def test_file_pattern(self, reader):
...
             "FIDUCEO_FCDR_L15_MVIRI_MET7-57.0_201701201000_201701201030_EASY_v2.6_fv3.1.nc",
             "FIDUCEO_FCDR_L15_MVIRI_MET7-00.0_201701201000_201701201030_EASY_v2.6_fv3.1.nc",
+            "MVIRI_FCDR-EASY_L15_MET7-E0000_200607060600_200607060630_0200.nc",
+            "MVIRI_FCDR-EASY_L15_MET7-E5700_200607060600_200607060630_0200.nc",
+            "MVIRI_FCDR-FULL_L15_MET7-E0000_200607060600_200607060630_0200.nc",
             "abcde",
         ]
 
         files = reader.select_files_from_pathnames(filenames)
         # only 3 out of 4 above should match
```
sfinkens, line 578:
```suggestion
```

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -574,6 +607,13 @@ def __init__(self, filename, filename_info, filetype_info,  # noqa: D417
             self._get_acq_time_uncached
         )
 
+    def _get_projection_longitude(self, filename_info):
+        """Read projection longitude from filename as it is not provided in the file."""
+        if "." in str(filename_info["projection_longitude"]):
```
sfinkens, line 612: I just learned from an AI that you can use `filename_info["projection_longitude"].is_integer()` to check if a number has zero digits

sfinkens (commented):
Looks very good, thanks! I'm just nitpicking with minor things now 🙂

satpy/readers/mviri_l1b_fiduceo_nc.py
bkremmli: resolved because the function does not exist anymore

satpy/readers/mviri_l1b_fiduceo_nc.py
```diff
@@ -509,6 +517,28 @@ def _cleanup_attrs(self, ds):
         # satpy warnings.
         ds.attrs.pop("ancillary_variables", None)
 
+    def _decode_cf(self):
+        # remove time before decoding and add again.
```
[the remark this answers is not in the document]
bkremmli: Partly: once, the add_offset is given in float which seems to be an issue for decode_cf(); secondly, (if first converting the attribute to int), decode_cf is raising an error that the conversion from sec. to nsec. gives an overflow..

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
bkremmli: this was done as a temporary bugfix but can be changed back (done)

satpy/readers/mviri_l1b_fiduceo_nc.py
bkremmli: nice function! However, it does not serve the purpose of the code above which needs to distinguish e.g. 57.0 from 5700 in mviri filenames version 1 and 2

sfinkens (approved):
LGTM!

sfinkens:
Let's see if we can make CodeScene happy

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
bkremmli: I now remember more clearly- after pushing to Github, a unit test fails for windows 3.9 (overflow error), from web search I found out that this minor change resolves the issue.

sfinkens: Ok, can you please add a comment :)

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
```diff
@@ -249,6 +256,10 @@
     height=2
 )
 
+time_fake_dataset = np.arange(4) * 60 * 60
+time_fake_dataset[0] = fill_val
+time_fake_dataset[1] = fill_val
+time_fake_dataset = time_fake_dataset.reshape(2, 2)
```
sfinkens, line 262: I think it would be nice to have those lines together in a method

sfinkens (commented):
Good job!

satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
sfinkens: Thanks!

sfinkens (approved):

sfinkens:
@mraspaud Do you want to have a final look? In the next PR we'll split "DatasetWrapper" into preprocessor and accessor

mraspaud:
I just had a quick look and this looks good to merge. In a future PR, you might want to ensure the sensor data is in float32 to increase performance.

sfinkens:
> you might want to ensure the sensor data is in float32 to increase performance.

That's checked in the tests

https://github.com/pytroll/satpy/blob/main/satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py#L433
https://github.com/pytroll/satpy/blob/main/satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py#L77

PR #2804, opened 2024-05-22
Fix LI L2 accumulated products `'with_area_definition': False` 1-d coordinates computation

ameraner:
This PR fixes a bug in the `li_l2_nc` reader when requesting accumulated products as 1-d arrays (by using `reader_kwargs={'with_area_definition': False}`).  *Note that it is anyway recommended to request accumulated products as 2-d arrays using `reader_kwargs={'with_area_definition': True}` (as discussed in https://github.com/pytroll/satpy/pull/2789).*

The bug was due to a missing `*= -1` on the azimuth geos angle coordinate (used inside the 1-d coordinates calculation), needed to account for a convention difference between MTG data and Proj. Without this fix, the coordinates (particularly the longitude) are computed with fully wrong values.

This PR fixes this issue and adds a test to check the consistency between the 1-d computed coordinates and the according coordinates extracted from the 2-d arrays with `AreaDefinition`. Before this fix this test would have failed.
It also updates the test filecontents for more realistic tests.

 - [x] Tests added 

Checking the reader on a real-life case, comparing the resampled 1-d arrays and the 2-d arrays:

Before this PR, the lightning data is at fully wrong locations (left: resampled 1-d arrays, right: 2-d):
image

After this PR, the results are comparable (left: resampled 1-d arrays, right: 2-d)::
image

According code:
```python
from satpy import Scene
from glob import glob
import matplotlib.pyplot as plt
import os
import numpy as np
from satpy.utils import debug_on;

debug_on()
import matplotlib

matplotlib.use("TkAgg")

ll_bbox = [33, 13, 60, 27]

# # initialise SEVIRI Scene
sevscn = Scene(
    filenames=['/tcenas/fbf/MSG/in/MSG3/OPE3/SEVI-MSG15/MSG3-SEVI-MSG15-0100-NA-20240501121242.333000000Z-NA.nat'],
    reader='seviri_l1b_native')
sevscn.load(['natural_color'])
sevscn_r = sevscn.resample('mtg_fci_fdss_2km', resampler='nearest')
sevscn_r = sevscn_r.crop(ll_bbox=ll_bbox)
seviri_img = np.moveaxis(get_enhanced_image(sevscn_r['natural_color']).data.values, 0, -1)

# initialise LI L2 acc product scene as 1-d array output
path_to_testdata = '/tcenas/scratch/andream/lil2a/'
scn = Scene(filenames=glob(os.path.join(path_to_testdata, '*AF*---*T_0073_0001.nc')), reader='li_l2_nc',
            reader_kwargs={'with_area_definition': False})
scn.load(['flash_accumulation'])
# resample 1-d array onto grid
scn_r = scn.resample('mtg_fci_fdss_2km', resampler='bucket_avg')
scn_r_c = scn_r.crop(ll_bbox=ll_bbox)

# plot resampled 1-d arrays on top of SEVIRI
crs = scn_r_c["flash_accumulation"].attrs['area'].to_cartopy_crs()
fig, axs = plt.subplots(1, 2, subplot_kw={'projection': crs})
ax = axs[0]
ax.coastlines(resolution='10m')

ax.imshow(seviri_img, transform=crs, extent=crs.bounds, origin='upper',
          interpolation='none')
ax.imshow(scn_r_c["flash_accumulation"], transform=crs, extent=crs.bounds, origin='upper',
          interpolation='none')


# initialise LI L2 acc product scene as 2-d array output (recommended, does not require resampling)
scn_adef = Scene(filenames=glob(os.path.join(path_to_testdata, '*AF*---*T_0073_0001.nc')), reader='li_l2_nc',
            reader_kwargs={'with_area_definition': True})
scn_adef.load(['flash_accumulation'], upper_right_corner='NE')
scn_adef_c = scn_adef.crop(ll_bbox=ll_bbox)

# plot 2d-array on top of SEVIRI
crs = scn_adef_c["flash_accumulation"].attrs['area'].to_cartopy_crs()
ax = axs[1]
ax.coastlines(resolution='10m')

ax.imshow(seviri_img, transform=crs, extent=crs.bounds, origin='upper',
          interpolation='none')
ax.imshow(scn_adef_c["flash_accumulation"], transform=crs, extent=crs.bounds, origin='upper',
          interpolation='none')

plt.show(block=False)
```

### DIFF
--- a/satpy/readers/li_base_nc.py
+++ b/satpy/readers/li_base_nc.py
@@ -371,6 +371,9 @@ class LINCFileHandler(NetCDF4FsspecFileHandler):
         azimuth = azimuth.values * point_height
         elevation = elevation.values * point_height
 
+        # In the MTG world, azimuth is defined as positive towards west, while proj expects it positive towards east
+        azimuth *= -1
+
         lon, lat = projection(azimuth, elevation, inverse=True)
 
         return np.stack([lon.astype(azimuth.dtype), lat.astype(elevation.dtype)])
--- a/satpy/readers/li_l2_nc.py
+++ b/satpy/readers/li_l2_nc.py
@@ -73,7 +73,7 @@ class LIL2NCFileHandler(LINCFileHandler):
         if var_with_swath_coord and self.with_area_def:
             return get_area_def("mtg_fci_fdss_2km")
 
-        raise NotImplementedError("Area definition is not supported for accumulated products.")
+        raise NotImplementedError("Area definition is not supported for non-accumulated products.")
 
     def is_var_with_swath_coord(self, dsid):
         """Check if the variable corresponding to this dataset is listed as variable with swath coordinates."""
--- a/satpy/tests/reader_tests/_li_test_utils.py
+++ b/satpy/tests/reader_tests/_li_test_utils.py
@@ -521,9 +521,13 @@ def accumulation_dimensions(nacc, nobs):
 
 def fci_grid_definition(axis, nobs):
     """FCI grid definition on X or Y axis."""
+    scale_factor = 5.58871526031607e-5
+    add_offset = -0.15561777642350116
     if axis == "X":
         long_name = "azimuth angle encoded as column"
         standard_name = "projection_x_coordinate"
+        scale_factor *= -1
+        add_offset *= -1
     else:
         long_name = "zenith angle encoded as row"
         standard_name = "projection_y_coordinate"
@@ -531,10 +535,10 @@ def fci_grid_definition(axis, nobs):
     return {
         "format": "i2",
         "shape": ("pixels",),
-        "add_offset": -0.155619516,
+        "add_offset": add_offset,
         "axis": axis,
         "long_name": long_name,
-        "scale_factor": 5.58878e-5,
+        "scale_factor": scale_factor,
         "standard_name": standard_name,
         "units": "radian",
         "valid_range": np.asarray([1, 5568]),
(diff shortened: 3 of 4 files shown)


satpy/readers/li_base_nc.py
```diff
@@ -371,6 +371,9 @@ def inverse_projection(self, azimuth, elevation, proj_dict):
         azimuth = azimuth.values * point_height
         elevation = elevation.values * point_height
 
+        # In the MTG world, azimuth is defined as positive towards west, while proj expects it positive towards east
+        azimuth *= -1
```
pnuu, line 375: I was going to comment that in-place multiplication with Dask arrays is dangerous, but it seems the array is already computed (`azimuth.values` is accessed). Is that necessary for some further computations?

pnuu (approved):
LGTM, just one question.

satpy/readers/li_base_nc.py
ameraner: That `self.inverse_projection` method is called inside a `da.map_blocks`: https://github.com/pytroll/satpy/blob/d29268cd98a175a686898e10ae31123aa837189e/satpy/readers/li_base_nc.py#L339-L344
so believe it should remain daskified. IIRC, we needed to get the `.values` inside there as the following `Proj` call accepts only computed values.

pnuu: I would have thought `da.map_blocks()` would pass Numpy arrays to the method being used :thinking:  Haven't used it in a while so might remember incorrectly. And if it works, then it must be correct because Numpy arrays wouldn't have `.values` attribute causing a crash :sweat_smile:

PR #2806, opened 2024-05-28
Add ability to clip AMI negative radiances

simonrp84:
This PR adds the ability to clip negative radiances from AMI datasets. It uses the same method already in place for GOES/ABI and, like ABI, this clipping is disabled by default.

 - [x] Tests added 
 - [x] Fully documented


satpy/readers/ami_l1b.py
```diff
@@ -201,11 +207,12 @@ def get_dataset(self, dataset_id, ds_info):
         data = data.where(qf == 0)
 
         # Calibration values from file, fall back to built-in if unavailable
-        gain = self.nc.attrs["DN_to_Radiance_Gain"]
-        offset = self.nc.attrs["DN_to_Radiance_Offset"]
+        self.gain = self.nc.attrs["DN_to_Radiance_Gain"]
+        self.offset = self.nc.attrs["DN_to_Radiance_Offset"]
```
djhoese, line 211: Why `self.` versus just passing them to `_clip_negative_radiance`?

Additionally, it is considered bad practice to create instance attributes that are not defined at creation time (`__init__`). It isn't obvious to me though why these should be instance attributes though.

satpy/tests/reader_tests/test_ami_l1b.py
```diff
@@ -58,14 +58,25 @@ class TestAMIL1bNetCDFBase(unittest.TestCase):
     """Common setup for NC_ABI_L1B tests."""
 
     @mock.patch("satpy.readers.ami_l1b.xr")
-    def setUp(self, xr_, counts=None):
+    def setUp(self, xr_, counts=None, irtest=False):
```
djhoese, line 61: I didn't notice that `counts` was done this way too, but these kwargs to the setup function should be changed. I'd say either:

1. (Preferred) Separate out the data creation steps as a function that has to be called from within the tests (I haven't looked at how many tests depend on this) and pass the kwargs from there. You could also at that point switch this class to a pytest test (remove the unittest base class).
2. Same as 1, but make the data creation pytest fixtures. Not sure that provides much benefit given the keyword arguments.
3. Set `counts` and `irtest` as a class attribute and use `self.counts` and `self.irtest` inside the setup function.

djhoese (changes requested):
Thanks for tackling this. I just have some concerns about how information is stored/kept/used.

satpy/readers/ami_l1b.py
simonrp84: Fixed!

satpy/tests/reader_tests/test_ami_l1b.py
simonrp84: I have tried to update this, but some of the tests are now failing. Unfortunately I don't have more time to debug what's going on - I barely understand this testing stuff and it takes up too much time. So we'll have to leave it at this point. Am happy to revert to the old way (following what was in the tests before I made this PR).

satpy/tests/reader_tests/test_ami_l1b.py
```diff
@@ -57,42 +57,59 @@ def close(self):
...
-                    "data_quality_flag_meaning":
-                        "0:good_pixel, 1:conditionally_usable_pixel, 2:out_of_scan_area_pixel, 3:error_pixel",
-                    "ground_sample_distance_ew": 1.4e-05,
-                    "ground_sample_distance_ns": 1.4e-05,
-                }
-            )
+            if self.irtest:
+                # If testing IR clipping, set one pixel to negative radiance
+                rad_data[0, 0] = 16364
+                self.counts = xr.DataArray(
```
djhoese, line 86: I'm rewriting these tests, but I'm realizing that this block for `self.counts = ` is probably indented one level too deep so `self.counts` never gets set. I have no idea how your tests were working.

djhoese: Oh `self.counts` is set in the `setUp` :man_facepalming: nevermind then

djhoese:
Ok @simonrp84 I've refactored the tests and hopefully they are easier to follow now *BUT* while doing this I noticed that the IR clipping tests actually use the fake visible data including the visible metadata. Any issue with me changing this to use the fake IR data?

simonrp84:
OK, thanks for your help on this Dave - really appreciate it. I tested your changes and they work nicely for me so I'm happy that it can be merged.

In the process of testing, though, I discovered that the AMI reader is broken with `numpy>=2`. I have added an additional commit that fixes this bug and enables the reader to work again.

mraspaud (approved):
LGTM!

djhoese:
I did a little more cleanup, but now I'll merge this when CI passes.

djhoese (approved):

PR #2807, opened 2024-05-31
Update the vii_l1b-reader, for new testdata format of VII

shornqui:
EUMETSAT has updated the EPS-SG testdata for VII (METimage), to version 2.1. There is one variable in the netCDF-file that has changed names. Because of that the VII-level1b-reader needs to be update.
(I have also checked with EUMETSAT, this change is meant to stay -the old variable name was a bug.)

There were no need to add new tests, but there was an updated test, due to this change.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


mraspaud (approved):
LGTM

PR #2813, opened 2024-06-12
Add missing hatchling dependency to unstable CI

djhoese:
Unstable is failing because hatchling isn't being installed. This should fix that.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -93,7 +93,6 @@ jobs:
           git+https://github.com/Unidata/netcdf4-python \
           git+https://github.com/dask/dask \
           git+https://github.com/dask/distributed \
-          git+https://github.com/zarr-developers/zarr \
           git+https://github.com/Unidata/cftime \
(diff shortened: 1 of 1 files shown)


mraspaud (approved):
Thanks a lot for fixing this!

djhoese:
@mraspaud @pnuu this is ready for review. The unstable environment now has "proper" failing tests. There are failures for numba which doesn't support numpy 2 yet, but also some unexpected failures and warnings. Mainly some h5netcdf `close` context manager failures or something and a lot of unexpected output dtypes in the DayNightCompositor tests. The numpy 2 dtype PR in xarray has been merged so I expected these to go away. We should look into these failures and make sure that it isn't a missed use case or something we should/can fix in Satpy. That said, I think this PR should be merged as-is so that other PRs can implement the fixes.

djhoese:
Haha, my open tab didn't refresh to show me that this was already merged many hours ago. Nevermind the review, thanks for merging!

PR #2822, opened 2024-06-14
Make NetCDF file cache handling compatible with dask distributed

### LINKED ISSUE #2815: LI and FCI readers do not work with dask distributed scheduler
**Describe the bug**

When using the dask distributed `LocalCluster`, computing LI or FCI data gives corrupted data.  Attempting to save the datasets to disk fails with several exceptions, with the root cause that a `_netCDF4.Variable` is not picklable.

**To Reproduce**
For FCI:

```python
from dask.distributed import Client, LocalCluster
from glob import glob
from satpy import Scene
fci_files = glob("W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001*_IDPFI_VAL_20231001*_20231001*_N__C_0054_003[456].nc")

def main():
    cluster = LocalCluster()
    client = Client(cluster)
    sc.load(["ir_105"], calibration="counts", pad_data=False)
    print(sc["ir_105"][100, 1000].compute().item())
    sc.save_datasets("simple_image")

if __name__ == "__main__":
    main()
```

And for LI accumulated products:

```python
from dask.distributed import Client, LocalCluster
from glob import glob
from satpy import Scene
li_files = glob("W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+LI-2-AF*--FD--CHK-BODY---NC4E_C_EUMT_20240613045114_L2PF_OPE_20240613045030_20240613045100_N__T_0030_0002.nc")

def main():
    cluster = LocalCluster()
    client = Client(cluster)
    sc = Scene(filenames={"li_l2_nc": li_files})
    sc.load(["accumulated_flash_area"])
    print(sc["accumulated_flash_area"][2500, 3080].compute())
    sc.save_datasets("simple_image")

if __name__ == "__main__":
    main()
```

**Expected behavior**

I expect that the print statements result in the correct value, namely the same value as when I comment out the cluster usage.  Furthermore, I expect both scripts to produce a simple image.

**Actual results**

With the distributed scheduler for FCI:

```
[DEBUG: 2024-06-13 13:11:24 : asyncio] Using selector: EpollSelector
[DEBUG: 2024-06-13 13:11:25 : asyncio] Using selector: EpollSelector
[DEBUG: 2024-06-13 13:11:25 : asyncio] Using selector: EpollSelector
[DEBUG: 2024-06-13 13:11:25 : asyncio] Using selector: EpollSelector
[DEBUG: 2024-06-13 13:11:25 : asyncio] Using selector: EpollSelector
[DEBUG: 2024-06-13 13:11:25 : satpy.readers.yaml_reader] Reading ('/data/gholl/checkouts/satpy/satpy/etc/readers/fci_l1c_nc.yaml',)
[DEBUG: 2024-06-13 13:11:25 : satpy.readers.yaml_reader] Assigning to fci_l1c_nc: ['/media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090004_IDPFI_VAL_20231001085755_20231001085835_N__C_0054_0034.nc', '/media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090012_IDPFI_VAL_20231001085811_20231001085849_N__C_0054_0035.nc', '/media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090018_IDPFI_VAL_20231001085827_20231001085856_N__C_0054_0036.nc']
[DEBUG: 2024-06-13 13:11:25 : satpy.readers.fci_l1c_nc] Reading: /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090004_IDPFI_VAL_20231001085755_20231001085835_N__C_0054_0034.nc
[DEBUG: 2024-06-13 13:11:25 : satpy.readers.fci_l1c_nc] Start: 2023-10-01 08:50:00
[DEBUG: 2024-06-13 13:11:25 : satpy.readers.fci_l1c_nc] End: 2023-10-01 09:00:00
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Reading: /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090012_IDPFI_VAL_20231001085811_20231001085849_N__C_0054_0035.nc
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Start: 2023-10-01 08:50:00
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] End: 2023-10-01 09:00:00
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Reading: /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090018_IDPFI_VAL_20231001085827_20231001085856_N__C_0054_0036.nc
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Start: 2023-10-01 08:50:00
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] End: 2023-10-01 09:00:00
[DEBUG: 2024-06-13 13:11:26 : satpy.composites.config_loader] Looking for composites config file fci.yaml
[DEBUG: 2024-06-13 13:11:26 : pyorbital.tlefile] Path to the Pyorbital configuration (where e.g. platforms.txt is found): /data/gholl/mambaforge/envs/py312/lib/python3.12/site-packages/pyorbital/etc
[DEBUG: 2024-06-13 13:11:26 : satpy.composites.config_loader] Looking for composites config file visir.yaml
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Reading ir_105 from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090004_IDPFI_VAL_20231001085755_20231001085835_N__C_0054_0034.nc
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Reading ir_105 from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090012_IDPFI_VAL_20231001085811_20231001085849_N__C_0054_0035.nc
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Reading ir_105 from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090018_IDPFI_VAL_20231001085827_20231001085856_N__C_0054_0036.nc
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Channel ir_105 resolution: 5568
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Row/Cols: 139 / 5568
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Calculated area extent: (-5567999.994203018, 3895999.995943777, 5567999.994203017, 3617999.996233209)
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Channel ir_105 resolution: 5568
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Row/Cols: 140 / 5568
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Calculated area extent: (-5567999.994203018, 4175999.995652262, 5567999.994203017, 3895999.995943776)
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Channel ir_105 resolution: 5568
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Row/Cols: 139 / 5568
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Calculated area extent: (-5567999.994203018, 4453999.9953628285, 5567999.994203017, 4175999.995652261)
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.yaml_reader] Requested orientation for Dataset ir_105 is 'native' (default). No flipping is applied.
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Reading ir_105_pixel_quality from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090004_IDPFI_VAL_20231001085755_20231001085835_N__C_0054_0034.nc
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Reading ir_105_pixel_quality from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090012_IDPFI_VAL_20231001085811_20231001085849_N__C_0054_0035.nc
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.fci_l1c_nc] Reading ir_105_pixel_quality from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090018_IDPFI_VAL_20231001085827_20231001085856_N__C_0054_0036.nc
[DEBUG: 2024-06-13 13:11:26 : satpy.readers.yaml_reader] Requested orientation for Dataset None is 'native' (default). No flipping is applied.
[DEBUG: 2024-06-13 13:11:26 : satpy.writers] Reading ['/data/gholl/checkouts/satpy/satpy/etc/writers/simple_image.yaml']
[DEBUG: 2024-06-13 13:11:26 : satpy.writers] No enhancement being applied to dataset
[DEBUG: 2024-06-13 13:11:26 : satpy.writers.simple_image] Saving to image: ir_105_20231001_085000.png
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing BlpImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing BmpImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing BufrStubImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing CurImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing DcxImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing DdsImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing EpsImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing FitsImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing FliImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing FpxImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Image: failed to import FpxImagePlugin: No module named 'olefile'
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing FtexImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing GbrImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing GifImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing GribStubImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing Hdf5StubImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing IcnsImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing IcoImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing ImImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing ImtImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing IptcImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing JpegImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing Jpeg2KImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing McIdasImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing MicImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Image: failed to import MicImagePlugin: No module named 'olefile'
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing MpegImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing MpoImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing MspImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing PalmImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing PcdImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing PcxImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing PdfImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing PixarImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing PngImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing PpmImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing PsdImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing QoiImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing SgiImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing SpiderImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing SunImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing TgaImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing TiffImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing WebPImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing WmfImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing XbmImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing XpmImagePlugin
[DEBUG: 2024-06-13 13:11:26 : PIL.Image] Importing XVThumbImagePlugin
[INFO: 2024-06-13 13:11:26 : satpy.writers] Computing and writing results...
2024-06-13 13:11:26,944 - distributed.protocol.pickle - ERROR - Failed to serialize <ToPickle: HighLevelGraph with 44 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7f8a6c437110>
 0. full_like-696a7234367a96400427da4ddc3c09fb
 1. invert-c86f30564af6929deacc155ead13f4d1
 2. original-array-e0f4e2816622aec99d8d202747d85b9f
 3. array-e0f4e2816622aec99d8d202747d85b9f
 4. le-d7d994d4bb2dacedd4efe49a96fcdbd7
 5. ge-6525ae48396405fdd51cac3616bc1b5f
 6. and_-b70dd6e75897c6c84093f7a1fa507819
 7. where-55dd48b805817a5efab42d71c5bd8cc2
 8. original-array-7e05d901070c5ceae9e85d9ee69991f6
 9. array-7e05d901070c5ceae9e85d9ee69991f6
 10. le-0cfb672d92259c0a4f66414a0f34cf0e
 11. ge-00cfb616d89fc90f4efafd7a287420f1
 12. and_-cf161ec58d41ada77980ff85dfbf4182
 13. where-910dce532cab82a93d00f64c04fb404d
 14. original-array-92330af6cf7bfca79e5a6116279bac9e
 15. array-92330af6cf7bfca79e5a6116279bac9e
 16. le-ce6a025d03a2f725f94de6d96eb0c140
 17. ge-37880d5c8c615ab2f4a67380e8830bc1
 18. and_-be8954ad7264317b4b0e82597182f16b
 19. where-d73b65825da59584be331d7d8feade6a
 20. concatenate-eb0ea4f18c208f8db3d37fe422cdc388
 21. broadcast_to-7f5b963d05ba33d499efbf900da3354b
 22. getitem-6bde8934bd55ad8109bbabc9fb0a6aa7
 23. ne-b133c489c45f840b5eee637617834f8e
 24. any-15bb9bf9193a46778437db9e43b0bc31
 25. any-aggregate-06570645fc6883d9786108644ab05894
 26. broadcast_to-a07bbc298b7fcd3892f638759b336b74
 27. clip-cf26d7576b4382959563d04f809f545c
 28. mul-d576f52b2a38443f0c9e1c8ab4fb5b7a
 29. add-cee50f65a8dd904280ef521a4e73f325
 30. round-febda8ca0b95edda8d5aa435f049db50
 31. where-3996b77335e9b452415d985190e60c1b
 32. astype-03fdd3b253bf9da74dcd5953d469e252
 33. clip-514b533a12d10b0cfb427d0c5c427bfb
 34. round-adc4823fe757e220b0d282d51cbbda7e
 35. full_like-9ad15347397c7233476b6973843cf60f
 36. invert-5bb4b1de69ff7f5779267d39403501a6
 37. where-ace75b00e9abe1e2db982f759b457b77
 38. astype-5951719beae63373cac280e05293e64c
 39. concatenate-9da4581706091d03c5ea32c5ee24bf7e
 40. transpose-c5b49f55d6ed82f09054cbb10bf22eee
 41. finalize-4c7205b4-bc9d-4530-8ad0-8eae37f87e51
 42. fromarray-2721f204-3bc5-4122-9899-40fc78b569f8
 43. delayed_pil_save-c558dd2e0cd6be31c80780ab4ddf2d0b
>.
  File "/data/gholl/mambaforge/envs/py312/lib/python3.12/site-packages/distributed/protocol/pickle.py", line 63, in dumps
    result = pickle.dumps(x, **dump_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_pickle.PicklingError: Can't pickle <function delayed_pil_save at 0x7f8ab81436a0>: it's not the same object as trollimage.xrimage.delayed_pil_save

  File "/data/gholl/mambaforge/envs/py312/lib/python3.12/site-packages/distributed/protocol/pickle.py", line 68, in dumps
    pickler.dump(x)
_pickle.PicklingError: Can't pickle <function delayed_pil_save at 0x7f8ab81436a0>: it's not the same object as trollimage.xrimage.delayed_pil_save

  File "/data/gholl/mambaforge/envs/py312/lib/python3.12/site-packages/distributed/protocol/pickle.py", line 81, in dumps
    result = cloudpickle.dumps(x, **dump_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py312/lib/python3.12/site-packages/cloudpickle/cloudpickle.py", line 1479, in dumps
    cp.dump(obj)
  File "/data/gholl/mambaforge/envs/py312/lib/python3.12/site-packages/cloudpickle/cloudpickle.py", line 1245, in dump
    return super().dump(obj)
           ^^^^^^^^^^^^^^^^^
  File "src/netCDF4/_netCDF4.pyx", line 6061, in netCDF4._netCDF4.Variable.__reduce__
NotImplementedError: Variable is not picklable
43.085777282714844
  File "/data/gholl/checkouts/protocode/test-dask-distributed-fci-2.py", line 27, in <module>
    main()
  File "/data/gholl/checkouts/protocode/test-dask-distributed-fci-2.py", line 23, in main
    sc.save_datasets("simple_image", enhance=False)
  File "/data/gholl/checkouts/satpy/satpy/scene.py", line 1293, in save_datasets
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/checkouts/satpy/satpy/writers/__init__.py", line 756, in save_datasets
    return compute_writer_results([results])
  File "/data/gholl/checkouts/satpy/satpy/writers/__init__.py", line 594, in compute_writer_results
    da.compute(delayeds)
  File "/data/gholl/mambaforge/envs/py312/lib/python3.12/site-packages/dask/base.py", line 661, in compute
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/mambaforge/envs/py312/lib/python3.12/site-packages/distributed/protocol/serialize.py", line 389, in serialize
    raise TypeError(msg, str_x) from exc
TypeError: ('Could not serialize object of type HighLevelGraph', '<ToPickle: HighLevelGraph with 44 layers.\n<dask.highlevelgraph.HighLevelGraph object at 0x7f8a6c437110>\n 0. full_like-696a7234367a96400427da4ddc3c09fb\n 1. invert-c86f30564af6929deacc155ead13f4d1\n 2. original-array-e0f4e2816622aec99d8d202747d85b9f\n 3. array-e0f4e2816622aec99d8d202747d85b9f\n 4. le-d7d994d4bb2dacedd4efe49a96fcdbd7\n 5. ge-6525ae48396405fdd51cac3616bc1b5f\n 6. and_-b70dd6e75897c6c84093f7a1fa507819\n 7. where-55dd48b805817a5efab42d71c5bd8cc2\n 8. original-array-7e05d901070c5ceae9e85d9ee69991f6\n 9. array-7e05d901070c5ceae9e85d9ee69991f6\n 10. le-0cfb672d92259c0a4f66414a0f34cf0e\n 11. ge-00cfb616d89fc90f4efafd7a287420f1\n 12. and_-cf161ec58d41ada77980ff85dfbf4182\n 13. where-910dce532cab82a93d00f64c04fb404d\n 14. original-array-92330af6cf7bfca79e5a6116279bac9e\n 15. array-92330af6cf7bfca79e5a6116279bac9e\n 16. le-ce6a025d03a2f725f94de6d96eb0c140\n 17. ge-37880d5c8c615ab2f4a67380e8830bc1\n 18. and_-be8954ad7264317b4b0e82597182f16b\n 19. where-d73b65825da59584be331d7d8feade6a\n 20. concatenate-eb0ea4f18c208f8db3d37fe422cdc388\n 21. broadcast_to-7f5b963d05ba33d499efbf900da3354b\n 22. getitem-6bde8934bd55ad8109bbabc9fb0a6aa7\n 23. ne-b133c489c45f840b5eee637617834f8e\n 24. any-15bb9bf9193a46778437db9e43b0bc31\n 25. any-aggregate-06570645fc6883d9786108644ab05894\n 26. broadcast_to-a07bbc298b7fcd3892f638759b336b74\n 27. clip-cf26d7576b4382959563d04f809f545c\n 28. mul-d576f52b2a38443f0c9e1c8ab4fb5b7a\n 29. add-cee50f65a8dd904280ef521a4e73f325\n 30. round-febda8ca0b95edda8d5aa435f049db50\n 31. where-3996b77335e9b452415d985190e60c1b\n 32. astype-03fdd3b253bf9da74dcd5953d469e252\n 33. clip-514b533a12d10b0cfb427d0c5c427bfb\n 34. round-adc4823fe757e220b0d282d51cbbda7e\n 35. full_like-9ad15347397c7233476b6973843cf60f\n 36. invert-5bb4b1de69ff7f5779267d39403501a6\n 37. where-ace75b00e9abe1e2db982f759b457b77\n 38. astype-5951719beae63373cac280e05293e64c\n 39. concatenate-9da4581706091d03c5ea32c5ee24bf7e\n 40. transpose-c5b49f55d6ed82f09054cbb10bf22eee\n 41. finalize-4c7205b4-bc9d-4530-8ad0-8eae37f87e51\n 42. fromarray-2721f204-3bc5-4122-9899-40fc78b569f8\n 43. delayed_pil_save-c558dd2e0cd6be31c80780ab4ddf2d0b\n>')
```

If I comment out the `save_datasets` call, the code completes with:

```
[DEBUG: 2024-06-13 13:13:34 : satpy.readers.fci_l1c_nc] Reading ir_105_pixel_quality from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090004_IDPFI_VAL_20231001085755_20231001085835_N__C_0054_0034.nc
[DEBUG: 2024-06-13 13:13:34 : satpy.readers.fci_l1c_nc] Reading ir_105_pixel_quality from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090012_IDPFI_VAL_20231001085811_20231001085849_N__C_0054_0035.nc
[DEBUG: 2024-06-13 13:13:34 : satpy.readers.fci_l1c_nc] Reading ir_105_pixel_quality from /media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY---NC4E_C_EUMT_20231001090018_IDPFI_VAL_20231001085827_20231001085856_N__C_0054_0036.nc
[DEBUG: 2024-06-13 13:13:34 : satpy.readers.yaml_reader] Requested orientation for Dataset None is 'native' (default). No flipping is applied.
43.085777282714844
```

The value is wrong (it shouldn't even be a float).  For reference, when I use the regular scheduler, I get the (probably correct) value of 1025 (integer)

If I load calibrated data, all data are NaN.

I get similar problems with LI, but not with IASI L2 CDR.  All three use file handlers deriving from `NetCDF4FsspecFileHandler`, but only FCI and LI use `cache_handle=True`, which seems to trigger the problem.  I did not try MWS (which also uses `cache_handle=True`).

**Environment Info:**
 - OS: openSUSE Leap 15.3
 - Satpy Version: latest main (v0.49.0-55-g367016e29)

**Additional context**

I ran into this problem when working on #2686.

It would seem that references to the file handler object end up in the dask graph, which should probably be avoided.  They fail to be pickled, because they contain references to NetCDF4 objects.

In #1546, a similar problem was solved for VIIRS Compact.

I don't know yet if the hypothesis is correct and if so, how those references end up in the dask graph.

gerritholl:
This PR makes file cache handling in the NetCDF4FileHandler compatible with dask distributed.  It adds a utility function in `satpy.readers.utils` called `get_distributed_friendly_dask_array`, which can be used to produce a dask.array from a netCDF4 variable that can be used in an xarray, but dask graphs remain picklable and thus computable when including this one.  This utility function is now used in NetCDF4FileHandler, which replaces homegrown file handle caching by caching using xarray.backends.CachingFileManager, which is needed to implement the aforementioned utility function.

 - [x] Closes #2815 
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -26,7 +26,7 @@ import xarray as xr
 
 from satpy.readers.core.file_handlers import BaseFileHandler
 from satpy.readers.core.remote import open_file_or_filename
-from satpy.readers.core.utils import get_distributed_friendly_dask_array, np2str
+from satpy.readers.core.utils import get_serialisable_dask_array, np2str
 from satpy.utils import get_legacy_chunk_size
 
 LOG = logging.getLogger(__name__)
@@ -352,13 +352,13 @@ class NetCDF4FileHandler(BaseFileHandler):
             else:
                 v = ds[key]
         if group is None:
-            dv = get_distributed_friendly_dask_array(
+            dv = get_serialisable_dask_array(
                     self.manager, key,
                     chunks=v.shape, dtype=v.dtype,
                     auto_maskandscale=self._auto_maskandscale)
         else:
-            dv = get_distributed_friendly_dask_array(
-                    self.manager, key, group=group,
+            dv = get_serialisable_dask_array(
+                    self.manager, "/".join([group, key]),
                     chunks=v.shape, dtype=v.dtype,
                     auto_maskandscale=self._auto_maskandscale)
         attrs = self._get_object_attrs(v)
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -17,6 +17,7 @@
 # satpy.  If not, see <http://www.gnu.org/licenses/>.
 """Helpers for reading netcdf-based files."""
 
+import functools
 import logging
 import warnings
 
@@ -122,7 +123,9 @@ class NetCDF4FileHandler(BaseFileHandler):
 
         if cache_handle:
             self.manager = xr.backends.CachingFileManager(
-                    netCDF4.Dataset, self.filename, mode="r")
+                    functools.partial(_NCDatasetWrapper,
+                                      auto_maskandscale=auto_maskandscale),
+                    self.filename, mode="r")
         else:
             file_handle.close()
 
@@ -477,3 +480,24 @@ class NetCDF4FsspecFileHandler(NetCDF4FileHandler):
         if self._use_h5netcdf:
             return obj.attrs[key]
         return super()._get_attr(obj, key)
+
+class _NCDatasetWrapper(netCDF4.Dataset):
+    """Wrap netcdf4.Dataset setting auto_maskandscale globally.
+
+    Helper class that wraps netcdf4.Dataset while setting extra parameters.
+    By encapsulating this in a helper class, we can
+    pass it to CachingFileManager directly.  Currently sets
+    auto_maskandscale globally (for all variables).
+    """
+
+    def __init__(self, *args, auto_maskandscale=False, **kwargs):
+        """Initialise object."""
+        super().__init__(*args, **kwargs)
+        self._set_extra_settings(auto_maskandscale=auto_maskandscale)
+
+    def _set_extra_settings(self, auto_maskandscale):
+        """Set our own custom settings.
+
+        Currently only applies set_auto_maskandscale.
+        """
+        self.set_auto_maskandscale(auto_maskandscale)
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -18,6 +18,7 @@
 """Helpers for reading netcdf-based files."""
 
 import logging
+import warnings
 
 import netCDF4
 import numpy as np
@@ -126,6 +127,16 @@ class NetCDF4FileHandler(BaseFileHandler):
 
     def _get_file_handle(self):
         return netCDF4.Dataset(self.filename, "r")
+
+    @property
+    def file_handle(self):
+        """Backward-compatible way for file handle caching."""
+        warnings.warn(
+                "attribute .file_handle is deprecated, use .manager instead",
+                DeprecationWarning)
+        if self.manager is None:
+            return None
+        return self.manager.acquire()
 
     @staticmethod
     def _set_file_handle_auto_maskandscale(file_handle, auto_maskandscale):
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -105,7 +105,7 @@ class NetCDF4FileHandler(BaseFileHandler):
         self._auto_maskandscale = auto_maskandscale
         if cache_handle:
             self.manager = xr.backends.CachingFileManager(
-                    functools.partial(_NCDatasetWrapper,
+                    functools.partial(_nc_dataset_wrapper,
                                       auto_maskandscale=auto_maskandscale),
                     self.filename, mode="r")
             file_handle = self.manager.acquire()
@@ -481,23 +481,14 @@ class NetCDF4FsspecFileHandler(NetCDF4FileHandler):
             return obj.attrs[key]
         return super()._get_attr(obj, key)
 
-class _NCDatasetWrapper(netCDF4.Dataset):
+def _nc_dataset_wrapper(*args, auto_maskandscale, **kwargs):
     """Wrap netcdf4.Dataset setting auto_maskandscale globally.
 
-    Helper class that wraps netcdf4.Dataset while setting extra parameters.
-    By encapsulating this in a helper class, we can
+    Helper function that wraps netcdf4.Dataset while setting extra parameters.
+    By encapsulating this in a helper function, we can
     pass it to CachingFileManager directly.  Currently sets
     auto_maskandscale globally (for all variables).
     """
-
-    def __init__(self, *args, auto_maskandscale=False, **kwargs):
-        """Initialise object."""
-        super().__init__(*args, **kwargs)
-        self._set_extra_settings(auto_maskandscale=auto_maskandscale)
-
-    def _set_extra_settings(self, auto_maskandscale):
-        """Set our own custom settings.
-
-        Currently only applies set_auto_maskandscale.
-        """
-        self.set_auto_maskandscale(auto_maskandscale)
+    nc = netCDF4.Dataset(*args, **kwargs)
+    nc.set_auto_maskandscale(auto_maskandscale)
+    return nc
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -211,10 +211,7 @@ class NetCDF4FileHandler(BaseFileHandler):
     def __del__(self):
         """Delete the file handler."""
         if self.manager is not None:
-            try:
-                self.manager.close()
-            except RuntimeError:  # presumably closed already
-                pass
+            self.manager.close()
 
     def _collect_global_attrs(self, obj):
         """Collect all the global attributes for the provided file object."""
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -19,14 +19,13 @@
 
 import logging
 
-import dask.array as da
 import netCDF4
 import numpy as np
 import xarray as xr
 
 from satpy.readers.core.file_handlers import BaseFileHandler
 from satpy.readers.core.remote import open_file_or_filename
-from satpy.readers.core.utils import np2str
+from satpy.readers.core.utils import get_distributed_friendly_dask_array, np2str
 from satpy.utils import get_legacy_chunk_size
 
 LOG = logging.getLogger(__name__)
@@ -85,10 +84,12 @@ class NetCDF4FileHandler(BaseFileHandler):
         xarray_kwargs (dict): Addition arguments to `xarray.open_dataset`
         cache_var_size (int): Cache variables smaller than this size.
         cache_handle (bool): Keep files open for lifetime of filehandler.
+            Uses xarray.backends.CachingFileManager, which uses a least
+            recently used cache.
 
     """
 
-    file_handle = None
+    manager = None
 
     def __init__(self, filename, filename_info, filetype_info,
                  auto_maskandscale=False, xarray_kwargs=None,
@@ -118,7 +119,8 @@ class NetCDF4FileHandler(BaseFileHandler):
         self.collect_cache_vars(cache_var_size)
 
         if cache_handle:
-            self.file_handle = file_handle
+            self.manager = xr.backends.CachingFileManager(
+                    netCDF4.Dataset, self.filename, mode="r")
         else:
             file_handle.close()
 
@@ -196,9 +198,9 @@ class NetCDF4FileHandler(BaseFileHandler):
 
     def __del__(self):
         """Delete the file handler."""
-        if self.file_handle is not None:
+        if self.manager is not None:
             try:
-                self.file_handle.close()
+                self.manager.close()
             except RuntimeError:  # presumably closed already
                 pass
 
@@ -301,8 +303,8 @@ class NetCDF4FileHandler(BaseFileHandler):
             group, key = parts
         else:
             group = None
-        if self.file_handle is not None:
-            val = self._get_var_from_filehandle(group, key)
+        if self.manager is not None:
+            val = self._get_var_from_manager(group, key)
         else:
             val = self._get_var_from_xr(group, key)
         return val
@@ -331,18 +333,26 @@ class NetCDF4FileHandler(BaseFileHandler):
                 val.load()
         return val
 
-    def _get_var_from_filehandle(self, group, key):
+    def _get_var_from_manager(self, group, key):
         # Not getting coordinates as this is more work, therefore more
         # overhead, and those are not used downstream.
+        with self.manager.acquire_context() as ds:
+            if group is not None:
+                v = ds[group][key]
+            else:
+                v = ds[key]
         if group is None:
-            g = self.file_handle
+            dv = get_distributed_friendly_dask_array(
+                    self.manager, key,
+                    chunks=v.shape, dtype=v.dtype)
         else:
-            g = self.file_handle[group]
-        v = g[key]
+            dv = get_distributed_friendly_dask_array(
+                    self.manager, key, group=group,
+                    chunks=v.shape, dtype=v.dtype)
         attrs = self._get_object_attrs(v)
         x = xr.DataArray(
-                da.from_array(v), dims=v.dimensions, attrs=attrs,
-                name=v.name)
+                dv,
+                dims=v.dimensions, attrs=attrs, name=v.name)
         return x
 
     def __contains__(self, item):
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -118,7 +118,7 @@ class NetCDF4FileHandler(BaseFileHandler):
                 raise
 
             self._set_file_handle_auto_maskandscale(file_handle, auto_maskandscale)
-            self._set_xarray_kwargs(xarray_kwargs, auto_maskandscale)
+        self._set_xarray_kwargs(xarray_kwargs, auto_maskandscale)
 
         listed_variables = filetype_info.get("required_netcdf_variables")
         if listed_variables:
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -103,15 +103,22 @@ class NetCDF4FileHandler(BaseFileHandler):
         self.cached_file_content = {}
         self._use_h5netcdf = False
         self._auto_maskandscale = auto_maskandscale
-        try:
-            file_handle = self._get_file_handle()
-        except IOError:
-            LOG.exception(
-                "Failed reading file %s. Possibly corrupted file", self.filename)
-            raise
+        if cache_handle:
+            self.manager = xr.backends.CachingFileManager(
+                    functools.partial(_NCDatasetWrapper,
+                                      auto_maskandscale=auto_maskandscale),
+                    self.filename, mode="r")
+            file_handle = self.manager.acquire()
+        else:
+            try:
+                file_handle = self._get_file_handle()
+            except IOError:
+                LOG.exception(
+                    "Failed reading file %s. Possibly corrupted file", self.filename)
+                raise
 
-        self._set_file_handle_auto_maskandscale(file_handle, auto_maskandscale)
-        self._set_xarray_kwargs(xarray_kwargs, auto_maskandscale)
+            self._set_file_handle_auto_maskandscale(file_handle, auto_maskandscale)
+            self._set_xarray_kwargs(xarray_kwargs, auto_maskandscale)
 
         listed_variables = filetype_info.get("required_netcdf_variables")
         if listed_variables:
@@ -121,12 +128,7 @@ class NetCDF4FileHandler(BaseFileHandler):
             self.collect_dimensions("", file_handle)
         self.collect_cache_vars(cache_var_size)
 
-        if cache_handle:
-            self.manager = xr.backends.CachingFileManager(
-                    functools.partial(_NCDatasetWrapper,
-                                      auto_maskandscale=auto_maskandscale),
-                    self.filename, mode="r")
-        else:
+        if not cache_handle:
             file_handle.close()
 
     def _get_file_handle(self):
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -101,6 +101,7 @@ class NetCDF4FileHandler(BaseFileHandler):
         self.file_content = {}
         self.cached_file_content = {}
         self._use_h5netcdf = False
+        self._auto_maskandscale = auto_maskandscale
         try:
             file_handle = self._get_file_handle()
         except IOError:
@@ -347,6 +348,7 @@ class NetCDF4FileHandler(BaseFileHandler):
     def _get_var_from_manager(self, group, key):
         # Not getting coordinates as this is more work, therefore more
         # overhead, and those are not used downstream.
+
         with self.manager.acquire_context() as ds:
             if group is not None:
                 v = ds[group][key]
@@ -355,11 +357,13 @@ class NetCDF4FileHandler(BaseFileHandler):
         if group is None:
             dv = get_distributed_friendly_dask_array(
                     self.manager, key,
-                    chunks=v.shape, dtype=v.dtype)
+                    chunks=v.shape, dtype=v.dtype,
+                    auto_maskandscale=self._auto_maskandscale)
         else:
             dv = get_distributed_friendly_dask_array(
                     self.manager, key, group=group,
-                    chunks=v.shape, dtype=v.dtype)
+                    chunks=v.shape, dtype=v.dtype,
+                    auto_maskandscale=self._auto_maskandscale)
         attrs = self._get_object_attrs(v)
         x = xr.DataArray(
                 dv,
--- a/satpy/readers/core/netcdf.py
+++ b/satpy/readers/core/netcdf.py
@@ -27,7 +27,7 @@ import xarray as xr
 
 from satpy.readers.core.file_handlers import BaseFileHandler
 from satpy.readers.core.remote import open_file_or_filename
-from satpy.readers.core.utils import get_serialisable_dask_array, np2str
+from satpy.readers.core.utils import get_serializable_dask_array, np2str
 from satpy.utils import get_legacy_chunk_size
 
 LOG = logging.getLogger(__name__)
@@ -355,15 +355,13 @@ class NetCDF4FileHandler(BaseFileHandler):
             else:
                 v = ds[key]
         if group is None:
-            dv = get_serialisable_dask_array(
+            dv = get_serializable_dask_array(
                     self.manager, key,
-                    chunks=v.shape, dtype=v.dtype,
-                    auto_maskandscale=self._auto_maskandscale)
+                    chunks=v.shape, dtype=v.dtype)
         else:
-            dv = get_serialisable_dask_array(
+            dv = get_serializable_dask_array(
                     self.manager, "/".join([group, key]),
-                    chunks=v.shape, dtype=v.dtype,
-                    auto_maskandscale=self._auto_maskandscale)
+                    chunks=v.shape, dtype=v.dtype)
         attrs = self._get_object_attrs(v)
         x = xr.DataArray(
                 dv,
--- a/satpy/readers/core/utils.py
+++ b/satpy/readers/core/utils.py
@@ -28,6 +28,7 @@ from io import BytesIO
 from shutil import which
 from subprocess import PIPE, Popen  # nosec
 
+import dask.array as da
 import numpy as np
 import pyproj
 import xarray as xr
@@ -699,3 +700,32 @@ class CalibrationCoefficientPicker:
 
 def _make_coefs(coefs, mode):
     return {"coefs": coefs, "mode": mode}
+
+
+def get_distributed_friendly_dask_array(manager, varname):
+    """Construct a dask array from a variable for dask distributed.
+
+    When we construct a dask array using da.array and use that to create an
+    xarray dataarray, the result is not serialisable and dask graphs using
+    this dataarray cannot be computed when the dask distributed scheduler
+    is in use.  To circumvent this problem, xarray provides the
+    CachingFileManager.  See GH#2815 for more information.
+
+    Args:
+        manager (xarray.backends.CachingFileManager):
+            Instance of xarray.backends.CachingFileManager encapsulating the
+            dataset to be read.
+        varname (str):
+            Name of the variable.
+    """
+    def get_chunk(block_info):
+        with manager.acquire_context() as nc:
+            loc = block_info[None]["array-location"][0]
+            rv = nc[varname][loc[0]:loc[1]]
+        return rv
+
+    return da.map_blocks(
+            get_chunk,
+            chunks=(10,),
+            meta=np.array([]),
+            dtype="i4")
--- a/satpy/readers/core/utils.py
+++ b/satpy/readers/core/utils.py
@@ -703,7 +703,7 @@ def _make_coefs(coefs, mode):
 
 
 def get_distributed_friendly_dask_array(manager, varname, chunks, dtype,
-                                        group="/"):
+                                        group="/", auto_maskandscale=None):
     """Construct a dask array from a variable for dask distributed.
 
     When we construct a dask array using da.array and use that to create an
@@ -733,9 +733,16 @@ def get_distributed_friendly_dask_array(manager, varname, chunks, dtype,
             What dtype to use.
         group (str):
             What group to read the variable from.
+        auto_maskandscale (bool, optional):
+            Apply automatic masking and scaling.  This will only
+            work if CachingFileManager.acquire returns a handler with a
+            method set_auto_maskandscale, such as is the case for
+            NetCDF4.Dataset.
     """
     def get_chunk():
         with manager.acquire_context() as nc:
+            if auto_maskandscale is not None:
+                nc.set_auto_maskandscale(auto_maskandscale)
             return nc["/".join([group, varname])][:]
 
     return da.map_blocks(
--- a/satpy/readers/core/utils.py
+++ b/satpy/readers/core/utils.py
@@ -702,7 +702,7 @@ def _make_coefs(coefs, mode):
     return {"coefs": coefs, "mode": mode}
 
 
-def get_distributed_friendly_dask_array(manager, varname, chunks=None):
+def get_distributed_friendly_dask_array(manager, varname, chunks, dtype):
     """Construct a dask array from a variable for dask distributed.
 
     When we construct a dask array using da.array and use that to create an
@@ -726,8 +726,10 @@ def get_distributed_friendly_dask_array(manager, varname, chunks=None):
             dataset to be read.
         varname (str):
             Name of the variable.
-        chunks (tuple or None, optional):
+        chunks (tuple):
             Chunks to use when creating the dask array.
+        dtype (dtype):
+            What dtype to use.
     """
     def get_chunk():
         with manager.acquire_context() as nc:
@@ -736,4 +738,5 @@ def get_distributed_friendly_dask_array(manager, varname, chunks=None):
     return da.map_blocks(
             get_chunk,
             chunks=chunks,
+            dtype=dtype,
             meta=np.array([]))
--- a/satpy/readers/core/utils.py
+++ b/satpy/readers/core/utils.py
@@ -722,7 +722,7 @@ def get_serializable_dask_array(manager, varname, chunks, dtype):
 
     Args:
         manager (xarray.backends.CachingFileManager):
-            Instance of xarray.backends.CachingFileManager encapsulating the
+            Instance of :class:`~xarray.backends.CachingFileManager` encapsulating the
             dataset to be read.
         varname (str):
             Name of the variable (possibly including a group path).
@@ -741,4 +741,4 @@ def get_serializable_dask_array(manager, varname, chunks, dtype):
             get_chunk,
             chunks=chunks,
             dtype=dtype,
-            meta=np.array([]))
+            meta=np.array([], dtype=dtype))
--- a/satpy/readers/core/utils.py
+++ b/satpy/readers/core/utils.py
@@ -702,9 +702,9 @@ def _make_coefs(coefs, mode):
     return {"coefs": coefs, "mode": mode}
 
 
-def get_distributed_friendly_dask_array(manager, varname, chunks, dtype,
-                                        group="/", auto_maskandscale=None):
-    """Construct a dask array from a variable for dask distributed.
+def get_serialisable_dask_array(manager, varname, chunks, dtype,
+                                auto_maskandscale=None):
+    """Construct a serialisable dask array from a variable.
 
     When we construct a dask array using da.array and use that to create an
     xarray dataarray, the result is not serialisable and dask graphs using
@@ -726,13 +726,11 @@ def get_distributed_friendly_dask_array(manager, varname, chunks, dtype,
             Instance of xarray.backends.CachingFileManager encapsulating the
             dataset to be read.
         varname (str):
-            Name of the variable.
+            Name of the variable (possibly including a group path).
         chunks (tuple):
             Chunks to use when creating the dask array.
         dtype (dtype):
             What dtype to use.
-        group (str):
-            What group to read the variable from.
         auto_maskandscale (bool, optional):
             Apply automatic masking and scaling.  This will only
             work if CachingFileManager.acquire returns a handler with a
@@ -744,7 +742,7 @@ def get_distributed_friendly_dask_array(manager, varname, chunks, dtype,
         with manager.acquire_context() as nc:
             if auto_maskandscale is not None:
                 nc.set_auto_maskandscale(auto_maskandscale)
-            var = nc["/".join([group, varname])]
+            var = nc[varname] #"/".join([group, varname])]
             return var[tuple(slice(*x) for x in arrloc)]
 
     return da.map_blocks(
--- a/satpy/readers/core/utils.py
+++ b/satpy/readers/core/utils.py
@@ -705,11 +705,11 @@ def _make_coefs(coefs, mode):
 def get_serializable_dask_array(manager, varname, chunks, dtype):
     """Construct a serializable dask array from a variable.
 
-    When we construct a dask array using da.array and use that to create an
-    xarray dataarray, the result is not serializable and dask graphs using
-    this dataarray cannot be computed when the dask distributed scheduler
-    is in use.  To circumvent this problem, xarray provides the
-    CachingFileManager.  See GH#2815 for more information.
+    When we construct a dask array using da.array from a file, and use
+    that to create an xarray dataarray, the result is not serializable
+    and dask graphs using this dataarray cannot be computed when the dask
+    distributed scheduler is in use.  To circumvent this problem, xarray
+    provides the CachingFileManager.  See GH#2815 for more information.
 
     Should have at least one dimension.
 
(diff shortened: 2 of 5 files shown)


satpy/readers/netcdf_utils.py
```diff
@@ -196,11 +210,8 @@ def _get_required_variable_names(listed_variables, variable_name_replacements):
 
     def __del__(self):
         """Delete the file handler."""
-        if self.file_handle is not None:
-            try:
-                self.file_handle.close()
-            except RuntimeError:  # presumably closed already
-                pass
+        if self.manager is not None:
+            self.manager.close()
```
djhoese, line 214: Is it worth it to wrap this with a try/except?

satpy/readers/utils.py
```diff
@@ -474,3 +475,53 @@ def remove_earthsun_distance_correction(reflectance, utc_date=None):
...
+    CachingFileManager.  See GH#2815 for more information.
+
+    Should have at least one dimension.
+
+    Example::
+
+        >>> import NetCDF4
+        >>> from xarray.backends import CachingFileManager
+        >>> cfm = CachingFileManager(NetCDF4.Dataset, fn, mode="r")
+        >>> arr = get_distributed_friendly_dask_array(cfm, "my_var")
```
djhoese, line 497:
```suggestion
        >>> import netCDF4
        >>> from xarray.backends import CachingFileManager
        >>> cfm = CachingFileManager(netCDF4.Dataset, fn, mode="r")
        >>> arr = get_distributed_friendly_dask_array(cfm, "my_var")
```

satpy/readers/utils.py
```diff
@@ -474,3 +475,53 @@ def remove_earthsun_distance_correction(reflectance, utc_date=None):
...
+
+    Args:
+        manager (xarray.backends.CachingFileManager):
+            Instance of xarray.backends.CachingFileManager encapsulating the
+            dataset to be read.
```
djhoese, line 502: We should check how the docs render this. If the argument type isn't "clickable" to go directly to the xarray docs for the CFM then we could wrap the mention of it in the description with:

```
:class:`xarray.backends.CachingFileManager`
```

satpy/readers/utils.py
```diff
@@ -474,3 +475,53 @@ def remove_earthsun_distance_correction(reflectance, utc_date=None):
     with xr.set_options(keep_attrs=True):
         reflectance = reflectance / reflectance.dtype.type(sun_earth_dist * sun_earth_dist)
     return reflectance
+
+
+def get_distributed_friendly_dask_array(manager, varname, chunks, dtype,
+                                        group="/", auto_maskandscale=None):
```
djhoese, line 481: I'm not sure how I feel about this function name. Obviously it makes sense in this PR because it solves this specific problem, but it feels like there is a (shorter) more generic name that gets the point across. Another thing is that `distributed_friendly` is mentioned here, but that friendliness is a side effect of the "serializable" nature of the way you're accessing the data here, right? `get_serializable_dask_array`?

I don't feel super strongly about this, but the name was distracting to me so I thought I'd say something.

satpy/readers/utils.py
```diff
@@ -474,3 +475,53 @@ def remove_earthsun_distance_correction(reflectance, utc_date=None):
...
+            Apply automatic masking and scaling.  This will only
+            work if CachingFileManager.acquire returns a handler with a
+            method set_auto_maskandscale, such as is the case for
+            NetCDF4.Dataset.
+    """
+    def get_chunk():
+        with manager.acquire_context() as nc:
+            if auto_maskandscale is not None:
+                nc.set_auto_maskandscale(auto_maskandscale)
+            return nc["/".join([group, varname])][:]
```
djhoese, line 521: This code is specific to `netCDF4`. While I understand that is the use case here, it suggests that either the function name and purpose needs to change or the `set_auto_maskandscale` functionality should be changed. Along with that, if I'm understanding the CFM properly, the `nc` object here should be the same between accesses. I mean, the exact same object if accessed in the same process. Although unlikely in the current usage setting this global attribute on the `netCDF4.Dataset` object could cause race conditions or unexpected behavior...although very unlikely.

To resolve both of these concerns, could this function and the current usage be changed to:

1. Use a wrapper (or `functools.partial` is possible) around `netCDF4.Dataset` when creating the CFM to set the auto_maskandscale option when `netCDF4.Dataset` is called inside the CFM. Although more hacky, there could also be a callback/init_func or maybe getitem function (see next 2 points) that handles setting of these per-variable things there (auto mask and scale can be set on the file object or the variable object).
2. The `group` and `varname` being joined is NetCDF specific and as far as I can tell isn't needed if the user is just expected to do it themselves.
3. This function assumes that `__getitem__` works for whatever file-like object is accessed and that `[:]` is needed. I don't think `__getitem__/[]` access for a variable is a bad assumption, but if there is a way to make this generic maybe that's what we need. If this function ended up being used with another library for reading files then the `[:]` may make an unnecessary copy and may not be noticeable to someone using the function.

satpy/readers/utils.py
```diff
@@ -474,3 +475,53 @@ def remove_earthsun_distance_correction(reflectance, utc_date=None):
...
+            What dtype to use.
+        group (str):
+            What group to read the variable from.
+        auto_maskandscale (bool, optional):
+            Apply automatic masking and scaling.  This will only
+            work if CachingFileManager.acquire returns a handler with a
+            method set_auto_maskandscale, such as is the case for
+            NetCDF4.Dataset.
+    """
+    def get_chunk():
```
djhoese, line 517: The `chunks` is never used here. The current calling from the file handler is accessing the full shape of the variable so this is fine, but only for now. I mean that `map_blocks` will only ever call this function once. *However*, if you added a `block_info` kwarg to the function signature or whatever the `map_blocks` special keyword argument is, then you could change `[:]` to access a specific sub-set of the NetCDF file variable and only do a partial load. This should improve performance a lot (I think :crossed_fingers:) if it was actually used in the file handler.

satpy/readers/utils.py
```diff
@@ -474,3 +475,53 @@ def remove_earthsun_distance_correction(reflectance, utc_date=None):
...
+        with manager.acquire_context() as nc:
+            if auto_maskandscale is not None:
+                nc.set_auto_maskandscale(auto_maskandscale)
+            return nc["/".join([group, varname])][:]
+
+    return da.map_blocks(
+            get_chunk,
+            chunks=chunks,
+            dtype=dtype,
+            meta=np.array([]))
```
djhoese, line 527: I think `dtype=dtype` should be added to the `np.array` call too. Otherwise it looks like numpy defaults to `[]` being a `np.float64` array.

satpy/readers/netcdf_utils.py
```diff
@@ -85,10 +85,12 @@ class NetCDF4FileHandler(BaseFileHandler):
         xarray_kwargs (dict): Addition arguments to `xarray.open_dataset`
         cache_var_size (int): Cache variables smaller than this size.
         cache_handle (bool): Keep files open for lifetime of filehandler.
+            Uses xarray.backends.CachingFileManager, which uses a least
+            recently used cache.
```
djhoese, line 89: I've lost track of the caching functionality of this helper class. What are the main differences now between the xarray implementation and the "cached" version specific to Satpy? Is the main difference the support for NetCDF groups and loading per-variable?

djhoese (changes requested):
Awesome job deciphering how to use the CachingFileManager and wrapping things in a map_blocks task. I think this is really close to being done, but I had some concerns about the helper function.

satpy/readers/netcdf_utils.py
gerritholl: `CachingFileManager` from xarray uses (by default) xarrays global LRU file cache, which by default keeps at most 128 files open (configurable).  Satpy kept everything open indefinitely. This could have an impact on performance, but is probably a good thing overall. 

By using `CachingFileManager`, the resulting `xarray.DataArray` is serialisable and therefore dask-distributed friendly.

Not sure about your point on groups — groups were already supported?

satpy/readers/netcdf_utils.py
gerritholl: Maybe?  I figured the `RuntimeError` could not happen because the `CachingFileManager` does not fail when close is called multiple times.  That leaves other exceptions, but I'm not sure what we would do in the `except:`-block, nor what actually happens if an exception is uncaught in `__del__`.

satpy/readers/netcdf_utils.py
djhoese: I didn't think xarray's default `xr.open_dataset` supported groups so we had this helper class and added on caching functionality later. I think the idea I was going for when I made the above comment originally was: why can't we just use `xr.open_dataset` instead of this class using `CachingFileManager`? My assumption was that we're supporting NetCDF groups here while xarray doesn't.

satpy/readers/utils.py
djhoese: Yes I think that's what I'm saying. I think the result of `get_chunk()` right now is broken for any chunk size other than the full shape of the array because you never do any slicing of the NetCDF variable inside `get_chunk()`. So, if you had a full array of 100x100 and a chunk size of 50x50, then map_blocks would call this function 4 times ((0-50, 0-50), (0-50, 50-100), (50-100, 0-50), (50-100, 50-100)). BUT each call would return the full variable 100x100. So I *think* this would be a case where the dask array would say "yeah, I have shape 100x100", but then once you computed it you'd get a 200x200 array back.

satpy/readers/netcdf_utils.py
djhoese: Yeah I'm not sure either. Let's see what happens I guess.

satpy/readers/netcdf_utils.py
gerritholl: Oh, do you mean the entire `NetCDF4FileHandler` class is redundant if we use `CachingFileManager` and `xr.open_dataset`s group ability?  It would seem to be a bit more, such as providing a convenient indexing mechanism.  But maybe I still misunderstand.

satpy/readers/utils.py
gerritholl: I've added a wrapper along with `functools.partial` and moved the `set_auto_maskandscale` there.  I've only set this globally; I find per variable too complicated and we set it globally so far already, so there is no change there.  I've also changed the group handling and moved this responsibility to the user.

As for the `__getitem__` assumption, yes, there is such an assumption.  I'm not sure how to make this generic without making it harder to use.  To make it generic, the user would need to pass information in some way on how the variable is obtained, via a callable or a wrapper class.  Do you have an idea on what would be a good way to implement this?

satpy/readers/utils.py
gerritholl: Renamed `get_serializable_dask_array`.

satpy/readers/utils.py
gerritholl: The argument type was already clickable, but in the description it was not.  I have now made it clickable in both cases (screenshot from local doc production):
 
Bildschirmfoto_2024-07-25_11-10-06

satpy/readers/utils.py
gerritholl: Fixed it now, I think.

satpy/readers/netcdf_utils.py
djhoese: I guess that's what I'm asking, not stating. You're right there is the `__getitem__` logic and the group handling is important enough. I guess I was mostly just curious what the difference was overall.

satpy/readers/utils.py
djhoese: I'm ok with the `__getitem__` assumption and even the `[:]` if it just seems annoying to make it more generic. One option could be to do something like `if not isinstance(x, np.ndarray)` (or some other "is not numpy-like or is netcdf Variable-like object" type of check) then you could do the `[:]` inside that `if` block.

gerritholl: I replaced `[:]` by taking information from `block_info`:

```
    def get_chunk(block_info=None):
        arrloc = block_info[None]["array-location"]
        with manager.acquire_context() as nc:
            var = nc[varname]
            return var[tuple(slice(*x) for x in arrloc)]
```

It's less obvious than `[:]` (at least to me) but I think it's correct :D

If something doesn't support such indexing, then it will fail with an exception and the user will notice that it doesn't work.

djhoese: Sounds good.

satpy/readers/utils.py
```diff
@@ -476,6 +477,48 @@ def remove_earthsun_distance_correction(reflectance, utc_date=None):
     return reflectance
 
 
+def get_serializable_dask_array(manager, varname, chunks, dtype):
+    """Construct a serializable dask array from a variable.
+
+    When we construct a dask array using da.array and use that to create an
```
mraspaud, line 483: Should we specify that this happens when a dask array is constructed from a file?

mraspaud (approved):
LGTM, just one comment inline. How is this affecting the performance of the reader? Do you consider this ready to be merged?

satpy/readers/utils.py
gerritholl: Yes.  Fixed in 5b137e8e7.

gerritholl:
> How is this affecting the performance of the reader?

I had not tested that yet.  I have now.  Sadly, it gets worse :(

A simple test script that reads FCI, loads some channels, resamples, and writes them again, not specifying any dask scheduler.

With Satpy main:

time 0:45.93, RAM 9.27 GB

With this PR:

time 0:55.0, RAM 10.0 GB

Additionally, upon exiting, there is the repeated error message:

```
 Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:
```

> Do you consider this ready to be merged?

Sadly no, considering the problems described above.  I will dig into this.

gerritholl:
With Satpy main, three runs, times in seconds:

Scene creation: 13.9, 11.9, 11.5

Loading: 1.5, 1.3, 1.0

Computing: 6.5, 6.9, 6.5

With this PR:

Scene creation: 13.2,  12.0, 12.0

Loading: 3.5, 3.5, 4.4

Computing: 5.5, 5.8, 5.6

So it's in particular the loading that gets slower.

gerritholl:
Profiling reveals that there are 160 calls to `acquire_context`, but only 40 to `_NCDatasetWrapper.__init__`, so the caching appears to be doing its job.  But the load call is not reusing the opening that happened upon Scene creation, which is why performance is impacted.  I will see what I can improve.

gerritholl:
With c2b1533 loading is much faster, although with more variability than for satpy main.  Scene creation is a little slower.

Scene creation: 13.4, 14.5, 12.8, 12.6, 12.8, 13.0

Loading: 1.9, 1.0, 1.6, 1.4, 1.3, 1.0

gerritholl:
I can't reliably reproduce the performance differences.  Running it again with the main branch gives:

Scene creation, main branch: 14.5, 14.3, 13.2

Scene creation, this PR: 13.6, 12.7, 13.8

And with cProfile, it's always faster with my PR.

Considering those uncertainties, I will declare performance is the same within the measurement uncertainty.

gerritholl:
Fixed the problem with the strange exception/error messages upon exit in 9fce5a7.

mraspaud:
I'm happy with this. @djhoese can you just confirm that you are good with this being merged? (and feel free to merge it if that's the case)

satpy/readers/netcdf_utils.py
```diff
@@ -117,14 +128,22 @@ def __init__(self, filename, filename_info, filetype_info,
...
+        warnings.warn(
+                "attribute .file_handle is deprecated, use .manager instead",
+                DeprecationWarning)
+        if self.manager is None:
+            return None
+        return self.manager.acquire()
+
     @staticmethod
     def _set_file_handle_auto_maskandscale(file_handle, auto_maskandscale):
         if hasattr(file_handle, "set_auto_maskandscale"):
```
djhoese, line 149: Not that this has to be handled in your PR, but if I remember correctly this `Dataset`-level `set_auto_maskandscale` was added to netcdf4-python quite a while ago. It seems error prone and confusing to silently call the method only if it exists and to not log/inform the user that it wasn't used when it was expected. Maybe we should remove this method on the file handler class and always call `file_handle.set_auto_maskandscale` no matter what. Your wrapper does it already.

satpy/tests/reader_tests/test_utils.py
```diff
@@ -514,6 +514,64 @@ def test_generic_open_binary(tmp_path, data, filename, mode):
...
+    slow.
+    """
+
+    @pytest.fixture(scope="class")
+    def dask_dist_client(self):
+        """Set up and close a dask distributed client."""
+        from dask.distributed import Client
+        cl = Client()
+        yield cl
+        cl.close()
```
djhoese, line 531: It looks like dask developer's recommend using their test utilities for writing distributed-based tests:

https://distributed.dask.org/en/latest/develop.html#writing-tests

Would it be possible to use their tools instead of this?

djhoese (changes requested):

satpy/tests/reader_tests/test_utils.py
djhoese: Looks like you might be able to import this fixture and use it:

https://github.com/dask/distributed/blob/386e5fea1cde4aefaf821e319405188266b41832/distributed/utils_test.py#L521-L525

gerritholl: Fixed in 222648afe

gerritholl:
`ruff check --fix` and `isort` are reverting each others changes to the list of imports.  I don't know how to satisfy both.

satpy/readers/netcdf_utils.py
gerritholl: If it doesn't need to be handled in this PR then I'd rather defer it.

gerritholl:
I think CodeScene is not fair in complaining about the excess number of arguments:

```def test_get_serializable_dask_array(self, tmp_path, client, shape, dtype, grp):```

Two of those are fixtures, so there is no way I can get rid of them.  I don't think combining shape, dtype, grp is going to improve this method in any way.  I tried to suppress the finding, but it doesn't work:

> Error occurred while suppressing! Please try again.

**Comment by djhoese** [comment]
I agree with you that CodeScene is being too picky here. You could remove the surrounding class as it isn't doing anything. That would get rid of the `self` argument too.

**Comment by gerritholl** [comment]
> You could remove the surrounding class as it isn't doing anything. That would get rid of the `self` argument too.

Good point, done.  But CodeScene already discounts `self`, so it doesn't resolve the failure.

File: satpy/readers/core/netcdf.py [hunk 12]
```diff
@@ -196,11 +215,8 @@ def _get_required_variable_names(listed_variables, variable_name_replacements):
 
     def __del__(self):
         """Delete the file handler."""
-        if self.file_handle is not None:
-            try:
-                self.file_handle.close()
-            except RuntimeError:  # presumably closed already
-                pass
+        if self.manager is not None:
+            self.manager.close()
```
djhoese, line 219: Is there any chance of `self.manager.close()` raising an exception? Should we try/except here? Or, should `self.manager` be set to `None` after it is closed?

gerritholl: I don't know, but this is `__del__` — exceptions are swallowed anyway?  And what good would it do to change the value of `self.manager` if the object is under destruction?

satpy/readers/core/utils.py
```diff
@@ -699,3 +700,45 @@ def get_coefs(self, channel):
...
+        chunks (tuple):
+            Chunks to use when creating the dask array.
+        dtype (dtype):
+            What dtype to use.
+    """
+    def get_chunk(block_info=None):
+        arrloc = block_info[None]["array-location"]
+        with manager.acquire_context() as nc:
+            var = nc[varname]
+            return var[tuple(slice(*x) for x in arrloc)]
```
djhoese, line 738: I'm not sure where our discussion ended on this, but I still think `manager` and `varname` could/should be keyword arguments to this function. Oh actually, for full distributed compatibility I think this `get_chunk` function needs to be in the global/module namespace otherwise distributed can't serialize it easily or at all.

satpy/readers/core/netcdf.py
djhoese: I think I was thinking of race conditions in a multi-threaded environment. I know we've seen users get errors where `__del__` is called on a file handler that is already closed or deleted and they get NetCDF errors or something. Anyway...I guess leave it as is and we'll see what happens.

satpy/readers/core/utils.py
gerritholl: I think the discussion got lost when I resolved merge conflicts.  I will check implementing your suggestion, I need to retrieve understanding how it works.

mraspaud:
Should we suggest this change to xarray, or is this too specific?

PR #2824, opened 2024-06-14
add import rioxarray where readers actually need them

ludwigVonKoopa:
some readers use `rasterio` engine from xarray, but it actually use rioxarray in the backend, without stating it in the dependencies (see https://github.com/pydata/xarray/issues/7831).

It should be a xarray problems, and in the PR they tend to agree to udpate the message without stating explicitly which exact dependency is problematic (which I understand).

So importing `rioxarray` in the reader file seems the best compromise : It complain with an `ModuleNotFoundError`, which is easier to understand why it doesn't work and how to correct that.


satpy/readers/generic_image.py
```diff
@@ -31,6 +31,7 @@
 import dask.array as da
 import numpy as np
 import rasterio
+import rioxarray  # noqa: F401, need by xarray with the engine rasterio
```
djhoese, line 34: This comment uses a `,` to separate the rule from the description, but the other one uses a `:`. What is the recommended way of doing this (I assume it is supported)?

pyproject.toml
```diff
@@ -66,7 +67,7 @@ gms5-vissr_l1b = ["numba"]
 # Writers:
 cf = ["h5netcdf >= 0.7.3"]
 awips_tiled = ["netCDF4 >= 1.1.8"]
-geotiff = ["rasterio", "trollimage[geotiff]"]
+geotiff = ["rasterio", "rioxarray", "trollimage[geotiff]"]
```
djhoese, line 70: Correct me if I'm wrong, but I'm 99% sure the geotiff *writer* does not use `rioxarray`.

satpy/readers/generic_image.py
ludwigVonKoopa: Right, the `,` can be used when multiple rules should be ignored. i.e. : `x=5  # noqa: E731,E123`  So If I want to be fully correct I should stick with some dot `.` or simply a space.

Moreover, as [stated in the documentation](https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors) : 

> Contents before and after the `# noqa: ...` portion are ignored so multiple comments may appear on one line. Here are several examples:
> ```python
># mypy requires `# type: ignore` to appear first
>x = 5  # type: ignore  # noqa: ABC123
># can use to add useful user information to a noqa comment
>y = 6  # noqa: ABC456  # TODO: will fix this later
>```

pyproject.toml
ludwigVonKoopa: Exact, I saw `rasterio` so I had an urge to add `rioxarray` too

djhoese (approved):

djhoese:
Thanks @ludwigVonKoopa!

PR #2825, opened 2024-06-14
Update docs for Developer's Guide, add dev-key for dependency install…

### LINKED ISSUE #2823: Documentation under "Developer's Guide" wrong for "Development installation"
**Describe the bug**
Documentation under "Developer's Guide" is wrong for "Development installation" and "Documentation"asking the use to run
`pip install -e ".[all]"`

When following the instructions and installing a new `conda`-environment quick eye in the terminal outputs can reveal the following warning

`WARNING: satpy 0.49.1.dev83+g0f8bf8714.d20240614 does not provide the extra 'all'`

**Expected behavior**
Update the instructions and/or define a new key in `pyproject.toml`, e.g. `dev`, which installs the desired/required dependencies.

**Additional context**
This change/bug may have been introduced when switching from `setup.py` to `pyproject.toml`?

sjoro:
This PR fixes documentation under "Developer's Guide" by introducing new dependency group `dev`in `pyproject.toml`-file and updating the docs accordingly.

Old documentation instructed the user to run `pip install -e ".[all]"`, which only results to a warning as dependency key `all` does not exist in `pyproject.toml`-file.

 - [x] Closes #2823  
 - [x] Fully documented

### DIFF
--- a/doc/source/dev_guide/index.rst
+++ b/doc/source/dev_guide/index.rst
@@ -72,7 +72,7 @@ All the required dependencies for a full development environment, i.e. running t
 tests and building the documentation, can be installed with::
 
     conda install eccodes
-    pip install -e ".[all]"
+    pip install -e ".[dev]"
 
 Running tests
 =============
@@ -125,7 +125,7 @@ Satpy's documentation is built using Sphinx. All documentation lives in the
 ``doc/`` directory of the project repository. For building the documentation,
 additional packages are needed. These can be installed with ::
 
-    pip install -e ".[all]"
+    pip install -e ".[doc]"
 
 After editing the source files there the documentation can be generated locally::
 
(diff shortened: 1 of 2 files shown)


djhoese:
Did you come up with this idea of listing `satpy` as a dependency of itself or is this recommended by some documentation?

sjoro:
@djhoese not sure what you refer to? this one `dev = ["satpy[doc,tests]"]`?

djhoese:
Yes that. I've never seen a project reference itself as a dependency.

sjoro:
it should not be a `satpy`-dependency (maybe i have misunderstood something then), but this notation in toml, according to what i found online (stackoverflow), is to bind two dependency groups into to one so that `pip install -e ".[dev]"` installs the dependecies for `doc` and `test`.
https://stackoverflow.com/questions/75490231/pip-pyproject-toml-can-optional-dependency-groups-require-other-optional-depend

sjoro:
what i would like to achieve here is to have a straight forward way to install a `dev` dependencies, which to me essentially are the `docs`and `test` dependency groups. if the notation is wrong, i can just update the docs and instruct to install `doc` and `test` dependencies and leave the `toml`-file untouched.

djhoese:
Ok it looks like this is supported in modern versions of pip. I found this thread:

https://discuss.python.org/t/pyproject-toml-optional-dependencies-redundancy-aka-dry-extras/8428/5

And the last 3 comments talk about recursion/inconsistency issues but reference a filed github issue:

https://github.com/pypa/pip/issues/10393

Where it is pointed out that this syntax is supported. That is good enough for me.

mraspaud (approved):
LGTM

PR #2686, opened 2023-12-12
Preload (FCI) filehandlers for eager processing

gerritholl:
Add functionality to preload filehandlers for files that have not yet arrived on disk.  This works by passing a single file (the first chunk) and then it will generate glob patterns for all remaining chunks.

It seems to work for loading only FDHSI data.

- [x] How should the user configure or pass the number of attempts and the waiting time between attempts?
 - [x] Closes https://github.com/pytroll/trollflow2/issues/160 
 - [x] Tests added 
 - [x] Fully documented
 - [x] Implement secede/rejoin with dask distributed scheduler
 - [x] Find a solution for https://github.com/pytroll/satpy/issues/2815 so that the former can be used.
 - [x] Make sure no computes are happening in a `Scene.load` call, including for preloaded segments.

Deferred: support for mixing FDHSI and HRFI.  This PR is ambitious enough as it is, and I would like to postpone that complication to a later PR.

As of 2024-06-20, it works with some limitations.

The default scheduler offers no control of the order of tasks.  The result is that dask tasks waiting for segments 33–40 might be scheduled first, while segments 2–32 are coming in but the corresponding tasks being later in the queue.  With the dask.distributed scheduler we can avoid this problem, but the dask distributed scheduler has limited support in Satpy.  PR #2822 makes the FCI reader work with dask.distributed, but nearest neighbour resampling or the GeoTIFF writer still fail (see #1762).  Therefore, to use eager processing, the user either:

* needs to accept suboptimal ordering of tasks, or
* use many workers for many chunks, or
* use dask.distributed and accept its limitations.

PR #2828, opened 2024-06-14
Share FCI L1c metadata between segments

pnuu:
Some of the metadata are identical in every FCI L1c segment, so reading those only once is possible. This will save a lot of time in `Scene` creation when the data are in S3 storage:
* `main` - 37.0 s
* PR - 23.0 s

There will be conflicts with https://github.com/pytroll/satpy/pull/2686, but maybe adding the pickle would benefit also this feature.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 


pnuu:
I'll see what I can do for testing. It's not easy, because the file handler is never used in the tests.

pnuu:
I've now added two tests that hopefully show that the storing and reusing of common items between the FCI L1c segments work.

pnuu:
Anything in the list shown [here](https://github.com/pytroll/satpy/blob/main/satpy/etc/readers/fci_l1c_nc.yaml#L23) that don't end in the strings listed in `NONSHAREABLE_VARIABLE_ENDINGS`. Note that the `{chan_name}` is replaced with the 16 channels FCI has, so the list is quite long.

gerritholl:
> Interesting! In the code I can see some metadata listed as non-shareable. Do you have a list of the shareable metadata? i.e: Which ones don't change between segments and are hence affected by this PR?

In
https://github.com/pytroll/satpy/pull/2686/files#diff-2170f8edf16088150763d5f3a6cbd69d62600d238c0ba80a41afcb4832fb7b5d
I explicitly list what metadata can be shared between segments.

satpy/readers/fci_l1c_nc.py
```diff
@@ -700,6 +709,31 @@ def calibrate_rad_to_refl(self, radiance, key):
         res = 100 * radiance * np.float32(np.pi) * np.float32(sun_earth_distance) ** np.float32(2) / cesi
         return res
 
+    def _collect_listed_variables(self, file_handle, listed_variables, filetype_info):
+        listed_variables = self._recycle_shared_info(listed_variables, filetype_info)
+        super()._collect_listed_variables(file_handle, listed_variables, filetype_info)
+        self._store_shared_info(filetype_info)
+
+    def _recycle_shared_info(self, listed_variables, filetype_info):
```
gerritholl, line 717: When I see the word `recycle`, I think of garbage collection, but I don't think that's what you mean here.  Maybe something like `collect_from_inter_segment_cache`, if I understand correctly what your intention is?

satpy/readers/fci_l1c_nc.py
```diff
@@ -700,6 +709,31 @@ def calibrate_rad_to_refl(self, radiance, key):
...
+                except ValueError:
+                    pass
+        return listed_variables
+
+    def _store_shared_info(self, filetype_info):
+        if "shared_info" not in filetype_info:
+            shared_info = {}
+            for key in self.file_content:
+                if any(key.endswith(k) for k in NONSHAREABLE_VARIABLE_ENDINGS):
+                    continue
```
gerritholl, line 733: By default you are sharing (in case of unknown variable names).  Would it be safer to default to not-sharing, i.e. use a whitelist rather than a blacklist?

satpy/readers/fci_l1c_nc.py
```diff
@@ -700,6 +709,31 @@ def calibrate_rad_to_refl(self, radiance, key):
...
+                shared_info[key] = self.file_content[key]
+            filetype_info["shared_info"] = shared_info
```
gerritholl, line 735: Is `filetype_info` really an appropriate place to put this cache?  I fear this might be confusing.

gerritholl (changes requested):
Thank you for the work!  I left some comments/questions where I think things might be implemented differently.  In addition, could more implementation be in `netcdf_utils` so it can be used by other readers where relevant?

satpy/readers/fci_l1c_nc.py
pnuu: I had the same thought earlier today, that I'm not recycling but reusing the data. I'll think of a better name for this.

satpy/readers/fci_l1c_nc.py
pnuu: I think your dict of sharing stuff would make this obsolete in any case as it explicitly says what can be shared and how.

satpy/readers/fci_l1c_nc.py
```diff
@@ -158,6 +158,15 @@
...
 
+NONSHAREABLE_VARIABLE_ENDINGS = [
+    "index",
+    "time",
+    "measured/effective_radiance",
+    "measured/y",
+    "position_row",
+    "index_map",
+    "pixel_quality"]
+
```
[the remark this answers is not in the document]
pnuu: Yup. I kinda made this on the basis "lets see if this makes things faster" so the approach was the simplest I saw. I think yours is more general and I'll update this PR to match yours when yours is ready.

Could be that `swath_number` can't be shared, but apparently sharing it didn't break anything when creating imagery :sweat_smile:

pnuu:
> In addition, could more implementation be in `netcdf_utils` so it can be used by other readers where relevant?

I think so. As I said in another comment above this snowballed from a quick test at PCW so I touched as little other parts of the code as possible. The FCI L1c data format is the most demanding in this regard, and most important to me, so started here.

I'll see about generalizing things later and converting this as a draft for the summer period.

satpy/readers/fci_l1c_nc.py
pnuu: We talked about this with @ameraner at PCW and this was already used in somewhere. LI L2, perhaps? It was also the path of least resistance as it was already in place and the same dict is passed between different filehandlers by the YAML reader.

If there are other backwards compatible ways to pass the info between the file handlers then let me know.

PR #2832, opened 2024-06-19
Add a read-only FSFile.fs property

pnuu:
This PR exposes the private `FSFile._fs` via a read-only property which would be nicer than accessing the private attribute. I'll need this for using `fsspec` caching I'm adding to Trollflow2 in https://github.com/pytroll/trollflow2/pull/205 

 - [x] Tests added 
 - [ ] Fully documented 


mraspaud (approved):
LGTM

PR #2833, opened 2024-06-19
Fix conversion of valid_range metadata to tuple in viirs_edr reader

djhoese:
While testing `viirs_edr` with our upcoming release of Polar2Grid 3.1, @kathys noticed that the AWIPS tiled writer fails to handle `valid_range` metadata that is a numpy array. It assumes it is a tuple or something else boolean-able. I can't find any documentation in the Satpy documentation on this, but I'm fairly certain that we tend to recommend either removing metadata fields like `valid_range` or to convert them to tuples. This PR converts the `viirs_edr` reader's metadata to a tuple for `valid_range` and now writing to AWIPS tiled files works as expected.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/viirs_edr.py
+++ b/satpy/readers/viirs_edr.py
@@ -128,18 +128,13 @@ class VIIRSJRRFileHandler(BaseFileHandler):
         """Get the dataset."""
         data_arr = self.nc[info["file_key"]]
         data_arr = self._mask_invalid(data_arr, info)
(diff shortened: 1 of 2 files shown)


djhoese:
I need this pull request for my work. Since everyone is going on vacation, I'm going to flex my github permission muscles and push this through. The overall fix is pretty simple, but we should probably have a larger discussion about the "standard practice" being utilized here.

PR #2847, opened 2024-06-30
Add "performance_test" to the benchmarks

yukaribbba:
To have a better understanding of the performance topics in FAQ, several tests are taken. This PR will give brief on them.

 - [x] Fully documented


djhoese (changes requested):
Very interesting. Thanks for putting this together. I have some concerns though.

1. I don't see the benefit of the performance table of contents section having a `maxdepth` of 2. It could just go along with the release notes or at least be brought down to a maxdepth of 1.
2. The testing platform and environments being on the main performance page suggests that this platform will be used for all series of tests and it won't be (unless you're volunteering your time and your machine for the next N years). I think it'd be better to have that information maybe per-test document *OR* run these tests on a specific cloud-based platforms. We already have benchmarks running on the European Weather Cloud.
3. I'm not sure the different resampling strategies need to be tested and documented in Satpy. This information seems like it would be better for pyresample.
4. We will need the code that generated all of this to be added to the repository and easily runnable. Preferably it would generate these tables automatically.
5. I feel like these types of results are highly dependent on the CPU and available memory of a system (assuming this is the only thing the machine is doing at the time). So I'm not sure how useful it is to show this to a user and have them expect similar numbers.
6. I think a graph, maybe even an embedded javascript one, would be easier to digest. I wouldn't have recommended this normally, but since the main argument for this being added is making it easier to understand how worker and chunk size work together I think it is important.
7. Lastly, I'm not sure how long lasting and useful this will be. These results could vary wildly with CPU and memory on the system (as I mentioned), but also with version changed of Satpy and satpy's dependencies. There was discussion somewhere else (github or slack) about having a script to show how a users specific machine performs with these different values. That may be better in the long term to make available to users and to make the results "personal" and actionable since it is their system they're seeing the results for.

I understand this is a draft and I am not the only satpy maintainer, but I just wanted to give my thoughts on this before you spend too much time on it if it ends up not being merged.

yukaribbba:
I agree. Actually I'm also not sure how useful my results could be for other platforms (That's why I only post one of them). I do have the test scripts that can finally output a csv file just like the tables in this PR. I'm making some improvements for public use. Maybe we can just let the user run it and decide what's the best settings like you said. But where shall we put it?

yukaribbba:
pre-commit.ci autofix

yukaribbba:
pre-commit.ci autofix

yukaribbba:
@djhoese CI failed again. Anyway my part is done temporarily. Here's an example of the test report.
FireShot Capture 005 - Satpy Performance Test Report for abi_l1b -

benchmarks/performance_test.py
```diff
@@ -0,0 +1,505 @@
...
+        resampler = "native" if diff_res else "none"
+        generate = not diff_res
+        total_rounds = len(self.chunk_size_opts) * len(self.worker_opts)
+
+        print(f"{self.reader_name} test started. Composite is \"{self.composite}\".\n")  # noqa
+        i = 0
+        for chunk_size in self.chunk_size_opts:
+            for num_worker in self.worker_opts:
+                self.single_loop((chunk_size, num_worker, resampler), generate=generate)
+                i = i + 1
```
djhoese, line 204:
```suggestion
for chunk_size, num_worker in itertools.product(self.chunk_size_opts, self.worker_opts):
```

I wonder if the final report could be put after the for loop and instead find a way to only do the sleep (maybe split the completed message and the 1-min rest message into two prints) if we aren't on the last round.

benchmarks/performance_test.py
```diff
@@ -0,0 +1,505 @@
...
+        import dask.config
+        self.running = True
+
+        chunk_size, num_worker, resampler = conditions
+
+        dask.config.set({"array.chunk-size": f"{chunk_size}MiB"})
+        dask.config.set(num_workers=num_worker)
+
+        try:
+            num_thread = os.environ["OMP_NUM_THREADS"]
```
djhoese, line 147: FYI we almost never see a good improvement if this is over 1 or 2.

doc/source/performance_test.rst
```diff
@@ -0,0 +1,149 @@
...
+    2024/06/29  09:06   <DIR>   G16_s20241691700214_e20241691709522_FLDK
+    2024/06/29  09:06   <DIR>   G16_s20241701700215_e20241701709523_FLDK
+    2024/06/29  09:06   <DIR>   G16_s20241711700217_e20241711709525_FLDK
+    2024/06/29  09:06   <DIR>   G16_s20241721700219_e20241721709527_FLDK
+    2024/06/29  09:06   <DIR>   G16_s20241731700220_e20241731709528_FLDK
+
+
+4. Do I have enough swap memory?
+--------------------------------
+Some conditions or resamplers may consume a hell of physical memory and then swap. When both are at their limits,
```
djhoese, line 60: We generally try to avoid language like this.

```suggestion
Some conditions or resamplers may consume a lot of physical memory and even use swap. When both are at their limits,
```

djhoese (commented):
I really didn't think you were going to put more work into this until we got more feedback from other maintainers. It looks like you've done a lot. It does look more useful this way, but it is still a lot. I guess I was hoping for a python script that could be called and provided command line options for all of these things. But please don't put more work toward this until others have weighed in with their opinions.

I see why the module was added to the docs and their are API docs since users are expected to run it themselves, but I don't like the extra dependencies that it requires in the docs building (especially matplotlib). Note that pandas shouldn't be an extra dependency since xarray should depend on it.

Lastly, our documentation is usually written with linux-based examples. You're running on Windows so it does give a different "look" to the documentation you've added.

yukaribbba:
I did think about command line options but eventually I found it looks nasty to let users put all these options in command. Beside, there're some kwargs hard to input that way. But ok, I'll be waiting for others' thoughts.

yukaribbba:
For dependencies, we can drop `py-cpuinfo` but `psutil` is necessary. Otherwise how can we record cpu/memory usage?

djhoese:
> For dependencies, we can drop py-cpuinfo but psutil is necessary. Otherwise how can we record cpu/memory usage?

Basically I was thinking that the script/module wouldn't be added to the docs (no API docs) and therefore we wouldn't need any of its dependencies because sphinx would never import it to be documented. Alternatively, we could tell sphinx to mock its dependencies in `conf.py`.

yukaribbba:
Well for me reading the docs are always better than reading the description inside the codes. Thats why I embed it. But the doc thing is just secondary. First we gonna hear what others say.

mraspaud:
Sorry for being so late on this.
First I'd like to thank you @yukaribbba for working on this! As a user, I can imaging the script being really useful when you want to know what parameters to tweak for your particular data.

A couple of comments/questions (I haven't checked the code in detail yet, so forgive me if I overlooked something):
- for read the docs, you don't actually need matplotlib I think, if you use `autodoc_mock_imports`.
- Have you considered including dask graphs also?
- For other satpy benchmarks, we are using [airspeed velocity](https://asv.readthedocs.io/en/v0.6.1/) and showing the results here: https://mraspaud.github.io/satpy/ (hasn't been updated for a year though :scream: , I need to fix that). Would this be something we could use ? It uses a VM on the European weather cloud to do the benchmarking.

PR #2850, opened 2024-07-05
Add parallel test execution to CI with pytest-xdist library

ClementLaplace:
This PR is related to the parralelisation of the tests as mentioned in this [issue](https://github.com/pytroll/satpy/issues/2826)

### DIFF
--- a/satpy/tests/test_writers.py
+++ b/satpy/tests/test_writers.py
@@ -18,11 +18,14 @@
 
 from __future__ import annotations
 
+import contextlib
 import datetime as dt
 import os
+import pathlib
 import shutil
 import unittest
 import warnings
+from typing import Iterator
 from unittest import mock
 
 import dask.array as da
@@ -30,6 +33,8 @@ import numpy as np
 import pytest
 import xarray as xr
 from trollimage.colormap import greys
+
+from satpy.writers import ImageWriter
 
 
 class TestWritersModule(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


djhoese:
Very interesting failures. I've restarted the ubuntu 3.12 test just to make sure it isn't macos-specific. As a last resort I think I remember there being a way to tell pytest that some tests can't be run in parallel, but I'm also surprised that these are failing.

djhoese:
I've been doing some testing of this locally and I think I've discovered one of the main issues although I'm not 100% sure it is the only issue. In `satpy/tests/test_writers.py` I wrote a ton of stuff that on `setup_class` and `teardown_class` will create enhancement YAML files. It currently puts these in the current working directory which messes up other portions of Satpy that are scanning for configuration files and look for YAML files in the current directory or based in the current directory (ex. `test_sensor1.yaml`, `enhancements/test_sensor1.yaml`, etc). I'll see if I can clean them up and push them to this branch and see how things behave.

djhoese:
@pytroll/satpy-core and @ClementLaplace any reason not to use `-n auto` on CI? If they give us the cores why not use them, right?

djhoese:
I'm not sure I'm happy with my choice of fixing my enhancement tests by using a fixture, but I didn't really like any the solutions I came up with either.

satpy/tests/test_writers.py
```diff
@@ -118,41 +118,41 @@ def test_init_nonexistent_enh_file(self):
...
+        The root temporary directory is changed to and returned.
 
-        class CustomImageWriter(ImageWriter):
-            def __init__(self, **kwargs):
-                super(CustomImageWriter, self).__init__(name="test", config_files=[], **kwargs)
-                self.img = None
+        """
+        prev_cwd = pathlib.Path.cwd()
+        tmp_path = tmp_path_factory.mktemp("config")
+        os.chdir(tmp_path)
```
sfinkens, line 143: What is the `chdir` needed for?

sfinkens (approved):
Looks good to me, just one question.

sfinkens:
> @pytroll/satpy-core and @ClementLaplace any reason not to use `-n auto` on CI? If they give us the cores why not use them, right?

I think it's a great idea! Thanks for bringing this up @ClementLaplace

satpy/tests/test_writers.py
djhoese: The tests are testing that configs in the current working directory are discovered and used. However, now that I say that that might not be important to all the tests. I'll try to review the tests and see if this chdir can be limited to only a few tests.

djhoese:
Ah github actions only gives use 2 cores except on macos it gives us 3 it seems. I guess "auto" is still a good thing to do for now.

satpy/tests/test_writers.py
djhoese: So there is at least one class of tests that I think it is important that the configs come from the current working directory. Or at least it is nice that they test it. The other tests would be fine if the config path was instead added using the `satpy.config` interface(s) for adding user custom configs. Either way, this current directory versus satpy.config directory stuff should be handled by other `satpy.config`-specific tests and is not necessary to test here. The tests look about the same whether `satpy.config` is used versus `chdir`, but I think the `chdir` ends up being a little cleaner but this is more of a personal preference and if I'm being honest some level of laziness that I don't want to change it.

Unless someone feels strongly I think I'd like this merged as-is and if more cleanup is needed in the future I can do that.

mraspaud (approved):
looks good to me, thanks @ClementLaplace @djhoese !

PR #2851, opened 2024-07-05
Fix start/end time properties of `hrit_jma` reader

sfinkens:
Fix `start/end_time` properties of `hrit_jma` reader to return `datetime` objects (currently integers).

Reason: `datetime64.astype(dt.datetime)` doesn't always return a `datetime` object. (see https://github.com/numpy/numpy/issues/20351).

```python
>>> np.datetime64("2000-01-01T00:00:00.000001").astype(dt.datetime)  # microsecond precision
datetime.datetime(2000, 1, 1, 0, 0, 0, 1)
>>> np.datetime64("2000-01-01T00:00:00.000000001").astype(dt.datetime)  # nanosecond precision
946684800000000001
```

I've updated usages of `datetime64.astype(dt.datetime)` in other readers, too. Although it doesn't seem to be a problem there, because they use micro/millisecond precision.

 - [X] Tests added

### DIFF
--- a/satpy/readers/viirs_vgac_l1c_nc.py
+++ b/satpy/readers/viirs_vgac_l1c_nc.py
@@ -22,7 +22,7 @@ import numpy as np
 import xarray as xr
 
 from satpy.readers.file_handlers import BaseFileHandler
-from satpy.utils import datetime64_to_pydatetime, get_legacy_chunk_size
+from satpy.utils import get_legacy_chunk_size
 
 CHUNK_SIZE = get_legacy_chunk_size()
 logger = logging.getLogger(__name__)
@@ -72,10 +72,6 @@ class VGACFileHandler(BaseFileHandler):
             data.attrs["end_time"] = dt.datetime.strptime(data.attrs["EndTime"], "%Y-%m-%dT%H:%M:%S")
             self._end_time = data.attrs["end_time"]
             self._start_time = data.attrs["start_time"]
-
-    def dt64_to_datetime(self, dt64):
-        """Conversion of numpy.datetime64 to datetime objects."""
-        return datetime64_to_pydatetime(dt64)
 
     def extract_time_data(self, data, nc):
         """Decode time data."""
--- a/satpy/readers/viirs_vgac_l1c_nc.py
+++ b/satpy/readers/viirs_vgac_l1c_nc.py
@@ -22,7 +22,7 @@ import numpy as np
 import xarray as xr
 
 from satpy.readers.file_handlers import BaseFileHandler
-from satpy.utils import get_legacy_chunk_size
+from satpy.utils import datetime64_to_pydatetime, get_legacy_chunk_size
 
 CHUNK_SIZE = get_legacy_chunk_size()
 logger = logging.getLogger(__name__)
@@ -75,9 +75,7 @@ class VGACFileHandler(BaseFileHandler):
 
     def dt64_to_datetime(self, dt64):
         """Conversion of numpy.datetime64 to datetime objects."""
-        if isinstance(dt64, np.datetime64):
-            return dt64.astype(dt.datetime)
-        return dt64
+        return datetime64_to_pydatetime(dt64)
 
     def extract_time_data(self, data, nc):
         """Decode time data."""
--- a/satpy/utils.py
+++ b/satpy/utils.py
@@ -31,7 +31,6 @@ from urllib.parse import urlparse
 
 import dask.utils
 import numpy as np
-import pandas as pd
 import xarray as xr
 import yaml
 from yaml import BaseLoader, UnsafeLoader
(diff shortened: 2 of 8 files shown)


satpy/readers/viirs_vgac_l1c_nc.py
```diff
@@ -75,9 +75,7 @@ def set_time_attrs(self, data):
 
     def dt64_to_datetime(self, dt64):
         """Conversion of numpy.datetime64 to datetime objects."""
-        if isinstance(dt64, np.datetime64):
-            return dt64.astype(dt.datetime)
-        return dt64
+        return datetime64_to_pydatetime(dt64)
```
djhoese, line 78: Is it worth it to keep this method? Or can we just replace it with the new utility function?

satpy/utils.py
```diff
@@ -841,3 +842,17 @@ def find_in_ancillary(data, dataset):
...
+
+    Discards nanosecond precision, because Python datetime only has microsecond
+    precision.
+
+    Args:
+        dt64 (np.datetime64): Timestamp to be converted
+    Returns (dt.datetime):
+        Converted timestamp
+    """
+    return pd.Timestamp(dt64).to_pydatetime(warn=False)
```
djhoese, line 858: What are the cases where this warns that are being ignored? Is there a way to do this without pandas? I know when I've looked this up in the past I think the pandas solution was also the best option, but I'm just curious.

sfinkens: Pandas warns when the conversion to datetime loses precision:

```python
In [4]: t = np.datetime64("2000-01-01T00:00:00.000000001")
In [5]: pd.Timestamp(t).to_pydatetime()
<ipython-input-5-1572263aba35>:1: UserWarning: Discarding nonzero nanoseconds in conversion.
  pd.Timestamp(t).to_pydatetime()
Out[5]: datetime.datetime(2000, 1, 1, 0, 0)
```

An alternative solution could be to explicitly cast to datetime precision (microseconds) and then convert to datetime:

```python
In [6]: t.astype("datetime64[us]").astype(dt.datetime)
Out[6]: datetime.datetime(2000, 1, 1, 0, 0)
```

Actually I think that's better, because it's more transparent

satpy/readers/viirs_vgac_l1c_nc.py
sfinkens: I was hesitating because it's a public method, but we can probably replace it. It's only used once in the tests.

satpy/readers/viirs_vgac_l1c_nc.py
```diff
@@ -73,12 +73,6 @@ def set_time_attrs(self, data):
             self._end_time = data.attrs["end_time"]
             self._start_time = data.attrs["start_time"]
 
-    def dt64_to_datetime(self, dt64):
-        """Conversion of numpy.datetime64 to datetime objects."""
-        if isinstance(dt64, np.datetime64):
-            return dt64.astype(dt.datetime)
-        return dt64
```
djhoese, line 80: How are you able to straight remove this with no other changes in this file? Was nothing else using this?

djhoese (approved):
I had one question for something that was confusing me, otherwise looks good from what I can tell.

Are there any warnings generated when using numpy 2 and timezone aware python datetime objects?

```python
now = dt.datetime.now(dt.timezone.utc)
np.datetime64(now)
```

^ produces a warning.

satpy/readers/viirs_vgac_l1c_nc.py
sfinkens: Yes, it was only used once in the tests

sfinkens:
> Are there any warnings generated when using numpy 2 and timezone aware python datetime objects?

No I don't see any. In this case I'm going the other way round from datetime64 to datetime. Since datetime64 has no timezone support I wouldn't expect warnings. There are other timezone related warnings, though. For example:

```
/home/runner/work/satpy/satpy/satpy/cf/attrs.py:212: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    _history_create = "Created by pytroll/satpy on {}".format(datetime.datetime.utcnow())
```

djhoese:
Yeah in another PR we should fix those. I thought Panu had started but maybe that was just the "dt" import change. Anyway...should you maybe add a UTC timezone to these datetimes? Or maybe we'll remember to do that in the future utcnow PR? Probably not. 🤷‍♂️

pnuu:
Oh, and as a reference the `utcnow()` issue: https://github.com/pytroll/satpy/issues/2752

sfinkens:
> Oh, and as a reference the `utcnow()` issue: #2752

Oh, I forgot about that, thanks!

sfinkens:
> should you maybe add a UTC timezone to these datetimes? 

Maybe it makes more sense to change them all at once when we decided on the way to go?

pnuu:
I also think the datetime aware stuff should happen in "one go" so that nothing breaks in any of the comparisons etc.

mraspaud (approved):
LGTM

PR #2861, opened 2024-07-26
Enhance support for FY-3F/MERSI-3

MengXin001:
This PR enhanced support for FY-3F/MERSI-3 data. Includes fixed geolocation dataset name in 250m data, support get coefficients, etc.
Actually, I'm not sure whether the test config is correct. theoretically, this test is basically the same as the FY-3D.

 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/readers/mersi3_l1b.yaml
+++ b/satpy/etc/readers/mersi3_l1b.yaml
@@ -438,7 +438,7 @@ datasets:
         standard_name: counts
   '19':
     name: '19'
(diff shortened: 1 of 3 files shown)


mraspaud (approved):
Thanks a lot for fixing this reader! In principle this looks good to me. However I would like someone who knows more about this format to have a look, maybe @simonrp84 ?

MengXin001:
I read the official document incorrectly, the part of geolocation data was originally correct, so I reverted the corresponding changes. The remaining enhancement to get coefficients is ok

PR #2843, opened 2024-06-28
feat: Enable to read for the Q4 coverage and the IQTI files for the fci l1c data

ClementLaplace:
-  [x] Add support for RSS imagery (Q4 coverage)
-  [x]  test support for RSS imagery (Q4 coverage)  
- [x] Add support for IQTI files 
- [x] Test for IQTI files

PR #2866, opened 2024-08-01
Fix FCI L1c reader for African products

ameraner:
This PR fixes the `fci_l1c_nc` reader for the African products, after https://github.com/pytroll/satpy/pull/2843 introduced/exposed a bug.

https://github.com/pytroll/satpy/pull/2843 changes the filename pattern from `erraneous_count_in_repeat_cycle` to `count_in_repeat_cycle`. With this, this line: https://github.com/pytroll/satpy/blob/bf0d9cd8b2732fff1839c90d8e724c1f46a920ac/satpy/readers/yaml_reader.py#L1192
does not use the `get` default `1` value anymore, but uses the actual value coming from the actual filepattern, which is `0`. This then activates the padding mechanism, which then fails due to `fci_l1c_nc.FCIL1cNCFileHandler.get_segment_position_info` not supporting the African products (as it shouldn't need to). 
The test that was specifically implemented to check that `get_segment_position_info` is not called, and that should have failed in the https://github.com/pytroll/satpy/pull/2843 PR, was faulty due to an extra unnecessary mocking that was invalidating the test.

This PR 
- fixes the root cause of the unwanted `get_segment_position_info` execution by assigning the chunk number `1` to the African products, so that the padding does not get activated (1 chunk is present and one is expected). 
- fixes the test to actually do what it should (the new test would have failed with https://github.com/pytroll/satpy/pull/2843, and it does not now due to the reader fix).

On top
- it fixes the filename pattern for the watervapour channels, that instead of `WV` use `IR` in the filename
- it splits the VIS0.6 channel into two separate filetypes for the 1km and 3km version, so that they can be loaded together and satpy will automatically pick the 1km file if available, as done for the normal files.
- on a side, it cleans up the test and reader file for many new style issues related to formatting, typos, static methods declarations, etc. (even though pre-commit was passing)
- it also updates and clarifies some more docstrings, variable names, etc.

**Sorry for the noise of the reformatting, to make the review easier I have marked with comments the actual relevant code changes.**

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/fci_l1c_nc.py
+++ b/satpy/readers/fci_l1c_nc.py
@@ -229,7 +229,7 @@ class FCIL1cNCFileHandler(NetCDF4FsspecFileHandler):
 
         if self.filename_info["facility_or_tool"] == "IQTI":
             self.is_iqt = True
-        else :
+        else:
             self.is_iqt = False
 
         self._cache = {}
@@ -239,7 +239,7 @@ class FCIL1cNCFileHandler(NetCDF4FsspecFileHandler):
         """Get nominal repeat cycle duration."""
         if "Q4" in self.filename_info["coverage"]:
             return 2.5
-        elif self.filename_info["coverage"] in ["FD","AF"]:
+        elif self.filename_info["coverage"] in ["FD", "AF"]:
             return 10
         else:
             raise NotImplementedError(f"coverage for {self.filename_info['coverage']}"
@@ -250,7 +250,7 @@ class FCIL1cNCFileHandler(NetCDF4FsspecFileHandler):
         """Get nominal start time."""
         rc_date = self.observation_start_time.replace(hour=0, minute=0, second=0, microsecond=0)
         return rc_date + dt.timedelta(
-            minutes=(self.filename_info["repeat_cycle_in_day"]-1)*self.rc_period_min)
+            minutes=(self.filename_info["repeat_cycle_in_day"] - 1) * self.rc_period_min)
 
     @property
     def nominal_end_time(self):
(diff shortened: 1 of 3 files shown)


satpy/readers/fci_l1c_nc.py
```diff
@@ -219,12 +219,18 @@ def __init__(self, filename, filename_info, filetype_info):
...
         if self.filename_info["coverage"] == "Q4":
-            # change number of chunk so that padding gets activated correctly on missing chunks
+            # change the chunk number so that padding gets activated correctly for Q4, which corresponds to the upper
+            # quarter of the disc
             self.filename_info["count_in_repeat_cycle"] += 28
 
+        if self.filename_info["coverage"] == "AF":
+            # change number of chunk from 0 to 1 so that the padding is not activated (chunk 1 is present and only 1
+            # chunk is expected), as the African dissemination products come in one file per full disk.
+            self.filename_info["count_in_repeat_cycle"] = 1
```
ameraner, line 229: this sorts out the padding activation bug for the African products

satpy/tests/reader_tests/test_fci_l1c_nc.py
```diff
@@ -897,24 +921,30 @@ def test_get_segment_position_info(self, reader_configs, fh_param, expected_pos_
             segpos_info = filetype_handler.get_segment_position_info()
             assert segpos_info == expected_pos_info
 
-    @mock.patch("satpy.readers.yaml_reader.GEOVariableSegmentYAMLReader")
```
ameraner, line 900: this mocking was preventing the test to actually test the execution

satpy/tests/reader_tests/test_fci_l1c_nc.py
```diff
@@ -897,24 +921,30 @@ def test_get_segment_position_info(self, reader_configs, fh_param, expected_pos_
...
             reader = _get_reader_with_filehandlers(fh_param["filenames"], reader_configs)
-            reader.load([channel])
+            try:
+                # attempt to load the channel
+                reader.load([channel])
+            except KeyError:
+                # If get_segment_position_info is called, the code will fail with a KeyError because of the mocking.
+                # So we catch the error here for now, but the test will still fail with the following assert_not_called
+                pass
             gspi.assert_not_called()
```
ameraner, line 937: this let the code run accepting the errors due to the mocking, but would still fail when checking the function call

satpy/tests/reader_tests/test_fci_l1c_nc.py
```diff
@@ -998,70 +1027,92 @@ def test_platform_name(self, reader_configs, fh_param):
...
+                                                                                         "2023-10-16 13:00:00",
+                                                                                         "%Y-%m-%d %H:%M:%S"))),
+                              ])
+    def test_count_in_repeat_cycle_rc_period_min(self, reader_configs, fh_param, compare_tuples):
+        """Test the rc_period_min value for each configuration."""
+        self._compare_rc_period_min_count_in_repeat_cycle(fh_param["filetype"], fh_param,
+                                                          reader_configs, compare_tuples)
+
+    @pytest.mark.parametrize(("channel", "resolution", "compare_tuples"),
+                             [("vis_06", "3km", (1, 10,
```
ameraner, line 1080: changing from 0 to 1 here checks that the chunk numbering is modified correctly

satpy/tests/reader_tests/test_fci_l1c_nc.py
```diff
@@ -897,24 +921,30 @@ def test_get_segment_position_info(self, reader_configs, fh_param, expected_pos_
...
             fh_param = FakeFCIFileHandlerAF_fixture
             reader = _get_reader_with_filehandlers(fh_param["filenames"], reader_configs)
-            reader.load([channel])
+            try:
+                # attempt to load the channel
+                reader.load([channel])
+            except KeyError:
+                # If get_segment_position_info is called, the code will fail with a KeyError because of the mocking.
+                # So we catch the error here for now, but the test will still fail with the following assert_not_called
+                pass
```
djhoese, line 936: Is there something we can do besides except->pass? Maybe https://docs.python.org/3/library/contextlib.html#contextlib.suppress? Or what about no `except` at all? I guess maybe I'm confused why a KeyError would get to this level of execution and we shouldn't care about it?

ameraner: I switched to suppress, that's a good idea, thanks! I also tried to clarify a bit more my comment in the code for why the test is setup like this. 

In other words: without the fix of this PR, `get_segment_position_info` would be called (which we don't want), and would then currently fail with `KeyError`. My point is that we want to test that the function is not called at all, which is what the test is doing. If instead we would check for the `KeyError` to happen, someone could technically change `get_segment_position_info` to avoid the `KeyError`and not fail anymore, the test would then pass, but the bug/unexpected behaviour would likely still be there.

mraspaud (approved):
LGTM, thanks for taking the time to clean up the code also. I really like when we get rid of mocking :)

PR #2867, opened 2024-08-02
Fix LI L2 reader for accumulated products from archive

### LINKED ISSUE #2859: LI L2 Accumulated products retrieved from archive ("ARC" 10-min files) have faulty reading
**Describe the bug**
The `li_l2_nc` reader currently produces faulty results when reading LI L2 accumulated products (AF, AFA, AFR) that are retrieved from the archive (containing `ARC` in the filename, with 10 minutes sensing window). 

This is because the reader is unaware that these files contain actually 20 separate time steps of the accumulated products; currently, when used with the default `with_area_definition=True` regridding mode, the reader will just regrid all pixels inside the file to the 2km grid with a "last pixel on top" approach, giving faulty results.

**NOTE: This issue is not impacting the 30-seconds files as disseminated on Eumetcast. The reading of these files is correct.**

**Expected behavior**
When loading the main variable (e.g. `accumulated_flash_area`) without further inputs, the reader should sum up all time steps inside the file. UPDATE: this is fixed by https://github.com/pytroll/satpy/pull/2867

Then, ideally, the reader should know how to split the data inside the files into 20 separate datasets, so that each time step can be accessed singularly. A clean solution for this would likely go over the DataID/DataQuery and/or a `Scene.load` kwarg. This is captured in issue https://github.com/pytroll/satpy/issues/2878:  

**Actual results**
One grid of accumulated products with a faulty and confusing "last pixel on top" mechanism.

ameraner:
This PR solves the bug identified in #2859 .

It implements two parallel solutions:
- [x] the nominal main product (e.g. `accumulated_flash_area`) is cumulated, so that it contains the entire observations for the 10-min period of the ARC products
- [ ] new sub-datasets are added in order to be able to differentiate  between the single 30s repeat cycles inside the 10-min ARC products. The start and end times are included in the dataset names.

The first solution allows the interchangeable usage of normal and ARC files, including the generation of composites. The usage of the new sub-datasets requires more expert handling (e.g. the usage of `scn.available_dataset_names()` to retrieve the list of sub-datasets).

 - [x] Closes #2859 
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/li_l2_nc.py
+++ b/satpy/readers/li_l2_nc.py
@@ -130,21 +130,35 @@ class LIL2NCFileHandler(LINCFileHandler):
         cols = self.get_measured_variable("x")
         attrs = data_array.attrs
 
-        rows, cols = da.compute(rows, cols)
-
         # origin is in the south-west corner, so we flip the rows (applying
         # offset of 1 implicitly)
         # And we manually offset the columns by 1 too:
         rows = (LI_GRID_SHAPE[0] - rows.astype(int))
         cols = cols.astype(int) - 1
 
-        # Create an empyt 1-D array for the results
-        flattened_result = np.nan * da.zeros((LI_GRID_SHAPE[0] * LI_GRID_SHAPE[1]), dtype=data_array.dtype)
-        # Insert the data. Dask doesn't support this for more than one dimension at a time, so ...
-        flattened_result[rows * LI_GRID_SHAPE[0] + cols] = data_array
+        # Create an empty 1-D array for the results
+        li_grid_flat_size = LI_GRID_SHAPE[0] * LI_GRID_SHAPE[1]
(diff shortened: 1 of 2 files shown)


satpy/etc/composites/li.yaml
```diff
@@ -1,5 +1,8 @@
 ---
-sensor_name: visir/li
+# we use li only here, and not visir/li, since the second can cause dependency issues when creating composites
+# combined with imagers in a multi-reader Scene. visir composites do not apply to LI anyway.
+sensor_name: li
```
djhoese, line 4: We could maybe define (I don't know if it errors) a new type of instrument. Like `li/li` for "lightning imager" type and the instrument is called "lightning imager". Or `lightning/li`? Not a huge deal for now but something to think about.

ameraner: True, yes. For now we don't have any, but it would be indeed nice to design some shared composites that work also with GLM and/or other networks.

mraspaud (approved):
LGTM

Issue #2300, opened 2022-11-28
to_image changes image mode


gerritholl:
**Describe the bug**

Calling `to_image` changes a dataset with mode P to one with mode L.

**To Reproduce**

```python
import xarray as xr
from trollimage.xrimage import XRImage
from satpy.writers import to_image
da = xr.DataArray(
        data=[[[1, 2, 3], [4, 5, 6]]],
        dims=("bands", "y", "x"),
        coords={"bands": ["P"]},
        attrs={"mode": "P"})
im = to_image(da)
print(im.mode)
# direct does retain it
im = XRImage(da)
print(im.mode)
```

**Expected behavior**

Expected output:

```
P
P
```

**Actual results**

Actual output:

```
L
P
```

**Environment Info:**
 - OS: openSUSE Leap 13.0
 - Satpy Version: v0.38.0-65-g6dfb9b93
 
**Additional context**

This happens due to the `.squeeze()` call in `satpy.writers.to_image`:

https://github.com/pytroll/satpy/blob/6dfb9b9353fa64e91a1f7553783abd859042da56/satpy/writers/__init__.py#L483

This deletes the `bands` dimension for mode P or L.  Then `XRImage.__init__` adds a `bands` dimension again with mode `L`.

The conversion presents an obstacle for handling P-mode images.

djhoese:
Wow nice catch! This looks like it was added in https://github.com/pytroll/satpy/pull/83 in one of the first PRs adding dask/xarray support to Satpy.

I don't think we need the squeeze. Maybe try removing it and if tests pass then :+1: 

CC @mraspaud who originally added the code.

PR #2870, opened 2024-08-05
Include gradient/axisintercept for mode p

### LINKED ISSUE #2869: ninjogeotiff writer should write gradient for P mode images
## Feature Request

The ninjogeotiff writer omits gradient and axis intercept for images of mode P, RGB, or RGBA, because those tags are not meaningful for categorical or RGB images.  However, for mode P images, NinJo still expects a gradient set to 1.  As it is now, mode P images produced by Satpy cannot be correctly displayed in NinJo, because the absence of a gradient is interpreted as a gradient of 0, leading to all-black images.

For mode P images, NinJo ignores the physical quantity/unit.  Colormaps and legends are defined entirely in terms of pixel values.  Therefore, regardless of the physical quantity/unit, it needs that gradient=1 and axisintercept=0.y**.

The Satpy ninjogeotiff writer should (optionally) include a gradient and axis intercept for mode P images.  The user should have a way to force those values to 1 and 0, overriding other methods that satpy and trollimage use for determining the correct gradient and axis intercept.

If we default to include gradient and axis intercept in the future, any users who rely on their absence will have to change their workflow.  I expect no such users exist.

Overriding the automatically computed values should of course be optional, so this part will not affect any existing user workflow.

**Additional context**

The alternative would be to change NinJo, but changing Satpy is 2–3 orders of magnitude faster.

gerritholl:
NinJo expects palette images to include gradient and axisintercept with values 1 and 0, respectively.

Needs https://github.com/pytroll/satpy/pull/2874 to be merged first.

I needed to include some workarounds for #2300.

 - [x] Closes #2869  
 - [x] Tests added 
 - [x] Fully documented 


mraspaud (approved):
LGTM

PR #2879, opened 2024-08-09
Initial implementation of EUMETSAT IASI-NG reader

roche-emmanuel:
This PR introduces the initial implementation of the IASI-NG reader for Satpy.

The changes provided here are self-contained in three files:
  * `satpy/readers/iasi_ng_l2_nc.py`: The implementation file for the reader.
  * `satpy/etc/readers/iasi_ng_l2_nc.yaml`: The YAML configuration file for the reader.
  * `satpy/tests/reader_tests/test_iasi_ng_l2_nc.py`: The unit tests for the reader.

There are still a few points I need to address (this list may be extended progressively):
  - [x] Check support for products with two-letter names (O3_ and CO_), as I don't think they will pass the current file pattern.
  - [x] Confirm with the EUMETSAT team whether the processing logic needs to be adjusted (awaiting clarification on some points concerning the product file specs).
  - [x] Extend the unit tests to perform more comprehensive checks on different products (currently only testing with TWV).
  - [x] Clean the code and remove debug outputs (e.g., `print(...)` statements).
  - [x] Clarify what needs to be done/added concerning the documentation of this reader.
  - [x] Add my name to the `AUTHORS.md` file.

The idea here is to start this PR early in the development phase to get early feedback and review notes from you, ensuring we are on the right track for eventual PR acceptance.

If you have any feedback on this PR already, please let me know 😊! In the meantime, I will continue working on the points mentioned above and will provide additional details on the changes as they are made on the feature branch.

Thanks for your help 🙏!


roche-emmanuel:
Alright, so I see from the workflows above that there are some failures due to the formatting of the file, I should maybe start with this before moving forward with the remaining code changes 😉.

roche-emmanuel:
Hi everyone, 

We now think that this MR is ready for a review by the Satpy team, so please let me know if you have any feedback on it. Thanks 😉!

roche-emmanuel:
Bump :-)!

Hi @djhoese , @mraspaud ,

I haven't heard back from you concerning this MR. So I was just wondering if there was anything else I could do on my side to help getting this one into the "processing pipeline" ? Please let me know if you have anything to suggest ;-)

Thanks! 🙏

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,418 @@
...
+        # dataset aliases:
+        self.dataset_aliases = self.filetype_info.get("dataset_aliases", {})
+
+        # Transform the aliases on regex patterns:
+        self.dataset_aliases = {
+            re.compile(key): val for key, val in self.dataset_aliases.items()
+        }
+
+        # broadcasts timestamps flag:
+        self.broadcast_timestamps = self.filetype_info.get("broadcast_timestamps", False)
```
mraspaud, line 71: please don’t use comments on every line, good variables names are explanatory enough! and the variable names don’t need to be short…

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,418 @@
...
+        ds_infos.update(desc)
+
+        self.dataset_infos[ds_name] = ds_infos
+
+    def process_dimension(self, key, value):
+        """Process a dimension entry from the file_content."""
+        dim_name = key.split("/")[-1]
+
+        if dim_name in self.dimensions_desc and self.dimensions_desc[dim_name] != value:
+            # This might happen if we have the same dim name from different groups:
```
mraspaud, line 132: if you put this condition in its own function that is called eg `same_dim_names_for_different_groups`, then you don’t need a comment.

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,418 @@
...
+            raise KeyError(f"Detected duplicated dim name: {dim_name}")
+
+        self.dimensions_desc[dim_name] = value
+
+    def process_attribute(self, key, value):
+        """Process a attribute entry from the file_content."""
+        var_path, aname = key.split("/attr/")
+
+        if var_path not in self.variable_desc:
+            # maybe this variable is ignored, or this is a group attr.
```
mraspaud, line 142: same here

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,418 @@
...
+            return
+
+        self.variable_desc[var_path]["attribs"][aname] = value
+
+    def process_variable(self, key):
+        """Process a variable entry from the file_content."""
+        shape = self.file_content[f"{key}/shape"]
+
+        if np.prod(shape) <= 1:
+            # Ignoring scalar variable.
```
mraspaud, line 152: using `np.isscalar` would more explicit, allowing to remove the comment.

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,418 @@
...
+            return
+
+        # Check if this variable should be ignored:
+        if any(p.search(key) is not None for p in self.ignored_patterns):
+            # Ignoring variable on user request.
```
mraspaud, line 157: if you put this condition in its own function that is called eg `variable_should_be_ignored`, then you don’t need comments.

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,418 @@
...
+
+        desc = {
+            "location": key,
+            "prefix": prefix,
+            "var_name": var_name,
+            "shape": shape,
+            "dtype": f"{dtype}",
+            "dims": dims,
+            "attribs": {},
+        }
```
mraspaud, line 173: that comment here surely means this code bit should be in its own function called eg `prepare_description_for_variable`?

mraspaud (changes requested):
ok I started reviewing this, and it looks sensible, but got distracted by all the comments, so I think we need to address this first.
So let’s start by reading this: https://santim0ren0.medium.com/clean-code-comments-f9eac4ada16d

In this PR I can see different types of comments:
- comments on what the next line is: if the name of the variable or function is good enough, they are not needed
- comments that explain what the following lines do: these lines should then be refactored into a function with a clear name
- comments that complement the docstrings: these should be integrated to the said docstrings

I’m happy to answer any questions if you need help with this :)

satpy/readers/iasi_ng_l2_nc.py
roche-emmanuel: Hi @mraspaud ,

Thank you very much for this initial feedback 😊

I've read the article you mentioned, and even if I don't fully agree to this perspective on code comments I think I see your point, and will start working on the requested changes asap. I'll let you when all the fixes are integrated and you can have another look at this PR.

Thanks 🙏!

roche-emmanuel: Hello @mraspaud ,

I have now updated the content of the iasi_ng_l2_nc.py file removing those excessive comments (just keeping a couple of "dev notes" in the end). 

Now, just to be sure before continuing: you would also expect me to do the same for the test_iasi_ng_l2_nc.py file, right :-)?

Also, I'm not sure if you would expect me to "resolve" each of the points reported above myself, or if you prefer to review the changes first and then resolve those points yourself when you juge the result acceptable (please just let me know if you think I should do it myself).

Thanks!

mraspaud: Thanks a lot, looks much better now!
yes it would be great if you could do the same in the test file.
Regarding the resolution of comments, I like it when the original commenter (in this case me) clicks the resolve button.

roche-emmanuel: Okay 👌copy that: so I'm updating the test file now.

roche-emmanuel: Hi @mraspaud ,

I have now also updated the test_iasi_ng_l2_nc.py file as discussed above. So please let me know when you find some time to give this PR another look, if you notice anything else that we should discuss/change. 

Thanks in advance for your time on this 🙏!

roche-emmanuel: Hi @mraspaud ,

I haven't heard back from you concerning this PR since a good time now, So this is just a "ping message" to check if maybe this just got out of your mind but now you could find some time to allocate to it ;-).  Thanks 🙏!

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,406 @@
...
+
+    # Note: patching the collect_groups_info method below to
+    # also collect dimensions in sub groups.
+    def _collect_groups_info(self, base_name, obj):
+        for group_name, group_obj in obj.groups.items():
+            full_group_name = base_name + group_name
+            self.file_content[full_group_name] = group_obj
+            self._collect_attrs(full_group_name, group_obj)
+            self.collect_metadata(full_group_name, group_obj)
+            self.collect_dimensions(full_group_name, group_obj)
```
mraspaud, line 86: Should this be covered by tests?

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,406 @@
...
+        """Check if a given path is a root attribute path."""
+        return var_path.startswith("/attr")
+
+    def is_property_path(self, var_path):
+        """Check if a given path is a sub-property path."""
+        return var_path.endswith(("/dtype", "/shape", "/dimensions"))
+
+    def is_netcdf_group(self, obj):
+        """Check if a given object is a netCDF group."""
+        return isinstance(obj, netCDF4.Group)
```
mraspaud, line 244: Looks like this trying to revert something that the base netcdf filehandler is building. Does is show that maybe the base netcdf filehandler class is not adapted to this case and that it would make more sense to go with eg an `xarray.open_dataset(…)` call for example?

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,406 @@
...
+
+        # Note: converting the time values to ns precision to avoid warnings
+        # from panda+numpy:
+        data_array = xr.DataArray(
+            data=pd.to_datetime(epoch) + (data_array * 1e9).astype("timedelta64[ns]"),
+            dims=data_array.dims,
+            attrs=data_array.attrs,
+        )
+
+        return data_array
```
mraspaud, line 357: Isn't there an xarray utility function to do this? I think it's supposed to support cf time units

mraspaud (commented):
Thanks a lot for the hard work on the reader. I have mostly questions inline, but the one I’m a bit concerned about in the ability to work with actual 3d data…

satpy/readers/iasi_ng_l2_nc.py
mraspaud: yeah sorry, I had a long holiday. Reviewed now.

satpy/readers/iasi_ng_l2_nc.py
roche-emmanuel: Hi @mraspaud ,

Thank you very much for your time reviewing this PR! Let's see what we can do about those additional points...

Hmmm, well... first thing I notice here after collecting all the latest changes from satpy main branch is that now my python environment seems to be broken (on Windows), since I get this error when trying to run the unit tests on this reader:

ImportError while loading conftest 'D:\Projects\satpy\satpy\tests\reader_tests\conftest.py'.
reader_tests\conftest.py:29: in <module>
    from xarray import DataTree
E   ImportError: cannot import name 'DataTree' from 'xarray' (D:\Projects\NervProj\.pyenvs\eum_resp2_t2_satpy_env\lib\site-packages\xarray\__init__.py)

(even after rebuilding this python env from scratches, so I should have the latest version of xarray, etc) 

=> Strange 🤔... (I'm not completely sure about that but I think the "DataTree" module is a pretty new addition in xarray right ?) Anyway, I'm investigating this further now.

roche-emmanuel: (=> OK xarray datatree error clarified: I was still installing another package called "xarray-datatree" in my env requirements, and because of this the xarray module was stuck on an old version (v2024.7.0). If I remove the "xarray-datatree" package then I get a recent version of xarray and no error for the tests execution).

satpy/readers/iasi_ng_l2_nc.py
roche-emmanuel: Hmmmm, well, from my perspective, the implementation we have in this reader rather makes an appropriate usage of the base Netcdf4Handler class: this reader depends heavily on the "self.file_content" member to analyze the available variables/dimensions/attributes, so I think it really makes sense to use the NetCDF4Handler class to access this instead of rebuilding it completely from scratches (if we were to open the file using only xarray directly for instance). 

And at the same time, we need a few small changes/extensions compared to the features provided by the base netcdf class, so, to me it sounds appropriate to also override the base implementation of "_collect_groups_info" for instance, and provide a few additional helper methods (ie. those listed above).

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,406 @@
...
+
+            data_array = (data_array * scale_factor) + add_offset
+
+            for key in ["valid_range", "valid_min", "valid_max"]:
+                if key in attribs:
+                    attribs[key] = attribs[key] * scale_factor + add_offset
+
+            data_array.attrs.update(attribs)
+
+        return data_array
```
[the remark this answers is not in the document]
roche-emmanuel: I have just tried to enable this when creating the reader

`    def __init__(self, filename, filename_info, filetype_info, **kwargs):
        """Initialize object."""
        super().__init__(
            filename, filename_info, filetype_info, auto_maskandscale=True, **kwargs
        )
`

and then disabling the call to the methods listed above, and this seems to work fine in most cases indeed, but then the  following unit test with fail:

    def test_nbr_iterations_dataset(self, twv_scene):
        """Test loading the nbr_iterations dataset."""
        twv_scene.load(["nbr_iterations"])
        dset = twv_scene["nbr_iterations"]

        assert len(dset.dims) == 2
>       assert dset.dtype == np.int32
E       AssertionError: assert dtype('float64') == <class 'numpy.int32'>

=> the original datatype for this array is "int32" in the netcdf file, but there is also a "missing_value" attribute for it, and it seems that in this case the auto_maskandscale flag will automatically convert the array to "float64", but we don't want that. So I'm not quite sure how we could make use of this properly? (=> unless you have a suggestion on this point I think we will need to keep the code above to deal with the int32 arrays as expected)

satpy/readers/iasi_ng_l2_nc.py
roche-emmanuel: I've tried to investigate this a bit further, and there is indeed an `xr.decode_cf(dset)` function, but this only works on a Dataset object, not a DataArray.

I tried a few things but I don't seem to be able to get it to work as desired in this case. I also tried to perform this decoding directly when opening the xarray dataset using this following: 

`
class IASINGL2NCFileHandler(NetCDF4FsspecFileHandler):
    """Reader for IASI-NG L2 products in NetCDF format."""

    def __init__(self, filename, filename_info, filetype_info, **kwargs):
        """Initialize object."""
        super().__init__(
            filename,
            filename_info,
            filetype_info,
            xarray_kwargs={"decode_cf": True},
            **kwargs,
        )
`

But the datetime checking unit tests are still failing with this 😅. Any suggestion on this point from your side ?

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,406 @@
...
+        )
+
+        return data_array
+
+    def apply_broadcast(self, data_array, ds_info):
+        """Apply the broadcast of the data array."""
+        dim_name = ds_info["broadcast_on_dim"]
+        if dim_name not in self.dimensions_desc:
+            raise KeyError(f"Invalid dimension name {dim_name}")
+        rep_count = self.dimensions_desc[dim_name]
```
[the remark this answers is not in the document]
roche-emmanuel: Hi @mraspaud , I'm not quite sure to really understand the reason behind this suggested code change ? Could you please elaborate a bit on why you think that is an improvement over the existing code ?

I mean, from my current perspective, the end effect is exactly the same in both case, but the try/except version would make the intent a "little bit less easy" to understand, also adding a very small [probably just negligeable sure] performance cost with the try/except block and breaking the overall design consistency of the reader implementation (ie. this reader file doesn't have any other try/except block for now: it just deals with all the known failing conditions directly raising exceptions in those cases, and let the caller deal with other (unexpected) exceptions).

Of course, I can make this change if you really think it's a good move, but if you have a minute I would be curious to have more details on your point of view on this topic first.

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,406 @@
...
+
+        data_array = xr.concat([data_array] * rep_count, dim=data_array.dims[-1])
```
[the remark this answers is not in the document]
roche-emmanuel: Well, it's not that replicating the data is strictly "necessary". The rationale is rather as follow:

- The reader should make is easy for end user to access the data provided by the instrument in the most convient/standard/expected format, and it should reduce the need for technical understanding of the file content structure if possible.
- On this instrument, the generated data will be 3D for most datasets since the dimensions used will be (n_lines, n_for, n_fov)
- But the "onboard_utc" variable will only be 2D of shape (n_lines, n_for), because the same timestamp is applicable on all the elements on the n_fov dim for a given pair in  (n_lines, n_for)
- So to make this dataset match the number of elements in the other datasets, we replicate the source elements "n_fov" times, while still keeping the the data array 2D for user convinience (more on this below in fact).

Note: we have in fact a "configuration entry" in the reader yaml file to control if this transformation should happen or not, and as you can see, this configuration entry is set to True, simply because this seems the most appropriate setting considering a typical usage case where we have tested this reader at EUMETSAT. And for now, I don't see any reason myself why it could be beneficial for the end user to disable this timestamp broadcasting.

satpy/readers/iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,406 @@
...
+        if len(data_array.dims) > 2:
+            data_array = data_array.stack(y=(data_array.dims[1:]))
+
+        if data_array.dims[0] != "x":
+            data_array = data_array.rename({data_array.dims[0]: "x"})
+
+        if data_array.dims[1] != "y":
+            data_array = data_array.rename({data_array.dims[1]: "y"})
+
+        return data_array
```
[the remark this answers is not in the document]
roche-emmanuel: As mentioned above, the idea here is to use the reader as an intermediate layer between the end user and the data file to hide some of the "non-critical technical/acquisition details". 

One of this consideration is that the measurement datasets have a shape of (n_lines, n_for, n_fov) because of different timestamps for the acquisition/processing of different chunks of columns, but in the end the resulting dataset is meant to be used as a (n_lines, n_cols) array (in most cases at least, and otherwise, the user could still use the n_for/n_fov values to reshape the arrays if that is ever needed?).

So that's the reason why we transform the data arrays to this shape here, and also rename the dimensions by default. 

=> Do you really think this is overstepping the role the reader should have in "formating" the data ? Considering we are already filtering the data content itself with valid min/max, replacing missing values with NaN or transforming double values into timestamps for instance, I would have considered this reshaping transformation to be at a similar level of "data pre-processing" actually.

satpy/tests/reader_tests/test_iasi_ng_l2_nc.py
```diff
@@ -0,0 +1,640 @@
...
+        from satpy._config import config_search_paths
+
+        self.reader_configs = config_search_paths(
+            os.path.join("readers", self.reader_name + ".yaml")
+        )
+
+    @pytest.fixture(autouse=True, scope="class")
+    def fake_handler(self):
+        """Wrap NetCDF4 FileHandler with our own fake handler."""
+        patch_ctx = mock.patch.object(
```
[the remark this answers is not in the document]
roche-emmanuel: Okay, and finally, concerning this one, I have now replaced the Mock framework with actual writing of temp netcdf files as suggested above, and it still seems to work as expected with this change.

satpy/readers/iasi_ng_l2_nc.py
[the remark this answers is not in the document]
mraspaud: Just some links for now, i can give my full opinion later
https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html
https://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain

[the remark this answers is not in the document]
roche-emmanuel: Hello @mraspaud ,

Thank you very much for those links. So I see here that you are basing this suggestion on the "EAFP" principle.  => So I have now just updated the code accordingly.

satpy/readers/iasi_ng_l2_nc.py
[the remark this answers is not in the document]
mraspaud: I'm wondering if a 2d coordinate could be a good match here?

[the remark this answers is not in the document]
roche-emmanuel: Hi @mraspaud ,

Sorry, but I don't really understand what you mean above with "a 2d coordinate". Could you please clarify if you really think this current implementation should be changed (and how you want it to be in that case), before this pull request could be accepted ? Thanks.

roche-emmanuel:
Hello @mraspaud,

I wanted to follow up on the discussion points raised above. Would you be able to provide feedback on these open items? I'd appreciate any guidance on what really needs to be addressed to move this PR forward.

Thank you for your time 👍!

satpy/readers/iasi_ng_l2_nc.py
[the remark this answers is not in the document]
mraspaud: what we do in readers is provide the data in an xarray.DataArray form that is consumable by the user for whatever work they want to do. The data filtering is to make sure the data is presented in a way that is consistent with the intent of the data provider. For example, for storage purposes, netcdf data is often packed, ie floats are converted to ints, because it takes less space. That's why we need min/max and missing values. I consider this merely as restoring approximately the data to the state it was before writing.
Now, if the data provider has an array in 3d, I am guessing the intent is for it to stay in 3d, otherwise it would have been transformed to 2d to start with, as it would have taken less space on disk for example.
I'm not familiar with the data, but the general philosophy in satpy is to try not to guess what the user wants and provide the data as is after unpacking. If you think this 2d array is useful to the users, I'm all for having it, but then it's a derived product that can be available (on top of the 3d one), but maybe not straight from the reader, but rather as a composite.

satpy/readers/iasi_ng_l2_nc.py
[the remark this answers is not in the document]
mraspaud: yes, having a "missing_value" or "_FillValue" attribute means that that value should be replaced by NaN, according to xarray and CF iirc. I can see that this variable is inherently an integer, but integers don't support NaNs. I would be fine with having this as float personally, but I understand if you want to have it as int.
I wonder what is the best way then. I mean the code here does the job, but I would argue it duplicates the code in xarray. Is there a way instead of just deactivating mask_and_scale just for this variable? or alternatively call xarray's mask and scale function manually for the relevant variables?

satpy/tests/reader_tests/test_iasi_ng_l2_nc.py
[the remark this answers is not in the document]
mraspaud: Looks great, thanks.

mraspaud:
> > Hello @mraspaud,
> > I wanted to follow up on the discussion points raised above. Would you be able to provide feedback on these open items? I'd appreciate any guidance on what really needs to be addressed to move this PR forward.
> > Thank you for your time 👍!
> 
> Hello Martin @mraspaud!
> 
> The new IASI-NG reader will be part of a new version of EUMETSAT's MONALiSA (https://www.eumetsat.int/media/50321) which we are working on in a "new features project". We are in the final phase of the project and according to plan we should be finished next week (first week of March 2025).
> 
> The IASI-NG Satpy reader has now jumped onto the critical path of the project. Currently I check if we are still on track and if we are able to land this aircraft on time 🙂 And if there are any crosswinds during landing 😉
> 
> Do you have an idea when you have time to check the remaining items updated by Emmanuel?
> 
> Kind regards, Rudi Lauster

Hi Rudi,
As I wrote on slack, this PR is not easy to review as it is quite large and introduces a complex kind of data in satpy. I want to make sure this is done correctly so that both your team and the pytroll community get the best out of this. As such, reviewing or answering this takes easily one or two hours of focused time, which is hard to come by for me nowadays, hence the delay, for which I apologise.
I would also like to say that if there is a deadline to respect it would be nice to tell us about it from the start so that we can prioritise this somehow, or maybe involve more of the core developers to the review.
I see no inherent problem with this PR though, just some things that need to integrate better with the usual satpy or xarray workflow, so I’m confident we will get this merged eventually.

RudiLauster:
> > > Hello @mraspaud,
> > > I wanted to follow up on the discussion points raised above. Would you be able to provide feedback on these open items? I'd appreciate any guidance on what really needs to be addressed to move this PR forward.
> > > Thank you for your time 👍!
> > 
> > 
> > Hello Martin @mraspaud!
> > The new IASI-NG reader will be part of a new version of EUMETSAT's MONALiSA (https://www.eumetsat.int/media/50321) which we are working on in a "new features project". We are in the final phase of the project and according to plan we should be finished next week (first week of March 2025).
> > The IASI-NG Satpy reader has now jumped onto the critical path of the project. Currently I check if we are still on track and if we are able to land this aircraft on time 🙂 And if there are any crosswinds during landing 😉
> > Do you have an idea when you have time to check the remaining items updated by Emmanuel?
> > Kind regards, Rudi Lauster
> 
> Hi Rudi, As I wrote on slack, this PR is not easy to review as it is quite large and introduces a complex kind of data in satpy. I want to make sure this is done correctly so that both your team and the pytroll community get the best out of this. As such, reviewing or answering this takes easily one or two hours of focused time, which is hard to come by for me nowadays, hence the delay, for which I apologise. I would also like to say that if there is a deadline to respect it would be nice to tell us about it from the start so that we can prioritise this somehow, or maybe involve more of the core developers to the review. I see no inherent problem with this PR though, just some things that need to integrate better with the usual satpy or xarray workflow, so I’m confident we will get this merged eventually.

Hello Martin @mraspaud,
Thank you very much for your efforts!
Well I didn't want to make "deadline pressure" from the beginning, as August was yet quite far away from the project end. Then I lost track and when Manu commented last week I kind of woke up 😁
Kind regards!

satpy/readers/iasi_ng_l2_nc.py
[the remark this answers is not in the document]
roche-emmanuel: Hello @mraspaud ,

I have now updated this part of the code, completely removing this "apply_broadcasting" operation: so we will now keep the array dimensions as is and perform the transformation we need externally. So this should now match your expectation on this point 😊.

satpy/readers/iasi_ng_l2_nc.py
roche-emmanuel: Hi @mraspaud ,

I've been trying to reconsider this implementation for the "convert_to_datetime" transformation, but unfortunately, so far I could not find anything better than what we currently have above. 

Do you have any idea/suggestion to try yourself ? Otherwise, do you also prefer that we remove this operation from the reader (keeping it as an external transformation on our side), and just provide the raw data from this reader (or could we keep this transformation as is ?)

mraspaud: I thought there was something that would work out of the box, but clearly there isn't. Thanks for investigating.

satpy/readers/iasi_ng_l2_nc.py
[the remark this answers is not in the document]
mraspaud: that sounds great!

mraspaud (approved):
I think this is good for a first implementation, it can be refined when the next versions of test dataset come out, when applications are starting to get developped.
Just one thing: do you have a link to the format documentation that you could paste in the module docstring for example?

roche-emmanuel:
Hi @mraspaud ,

I'm really glad to see that the PR is getting close to acceptance now 😊!

As requested just above, I have now also added a link to the IASI-NG l2 format specification document directly in the IASI-NG reader module docstring. Let me know if you need anything else, and as usual, thanks for your time working on this 🙏!

mraspaud:
Thanks for adding the link! I was about to merge this but saw the error in the automated tests:
```
FAILED satpy/tests/reader_tests/test_iasi_ng_l2_nc.py::TestIASINGL2NCReader::test_onboard_utc_dataset - TypeError: unsupported operand type(s) for +: 'Timestamp' and 'Array'
```
?

mraspaud:
@roche-emmanuel we could merge this if the test error was fixed, and release it on friday...

roche-emmanuel:
Hi @mraspaud ,

Sorry for the delay on this (I was on vacation ;-)). => I'm looking into this error today.

mraspaud:
Thanks for fixing the macos tests! Merging

PR #2913, opened 2024-09-27
Update `check_satpy` to use new `show_version` to display package versions

verduijn:
MR to add `show_versions` to `satpy.utils` to allow for use in `satpy.utils.check_satpy`. 
Use case is to help in figuring out run-time environment factors while debugging issues.

### DIFF
--- a/satpy/tests/test_utils.py
+++ b/satpy/tests/test_utils.py
@@ -278,13 +278,17 @@ class TestCheckSatpy:
         """Test 'check_satpy' with specific features provided."""
         from satpy.utils import check_satpy
         with mock.patch("satpy.utils.print") as print_mock:
-            check_satpy(readers=["viirs_sdr"], extras=("cartopy", "__fake"))
-            checked_fake = False
-            for call in print_mock.mock_calls:
-                if len(call[1]) > 0 and "__fake" in call[1][0]:
-                    assert "ok" not in call[1][1]
-                    checked_fake = True
-            assert checked_fake, "Did not find __fake module mentioned in checks"
+            check_satpy(readers=["viirs_sdr"], packages=("cartopy", "__fake"))
+            checked_fake = any("__fake: not installed" in c[1] for c in print_mock.mock_calls if len(c[1]))
+            assert checked_fake, "Did not find __fake package mentioned in checks"
+
+class TestShowVersions:
+    """Test the 'show_versions' function."""
+
+    def test_basic_show_versions(self):
+        """Test 'check_satpy' basic functionality."""
+        from satpy.utils import show_versions
+        show_versions()
 
 
 def test_debug_on(caplog):
--- a/satpy/tests/test_utils.py
+++ b/satpy/tests/test_utils.py
@@ -291,36 +291,25 @@ class TestShowVersions:
         from satpy.utils import show_versions
         show_versions()
 
-    def test_show_specific_version(self):
+    def test_show_specific_version(self, capsys):
         """Test 'show_version' works with installed package."""
         from satpy.utils import show_versions
-        with mock.patch("satpy.utils.print") as print_mock:
-            show_versions(packages=["pytest"])
+        show_versions(packages=["pytest"])
+        out, _ = capsys.readouterr()
 
-            # no regex or `.__version__` based checks to prevent edge case failures
-            pytest_mentioned = any(
-                "pytest:" in c[1][0] for c in print_mock.mock_calls if len(c[1])
-            )
-            pytest_installed = all(
-                "pytest: not installed" not in c[1][0]
-                for c in print_mock.mock_calls
-                if len(c[1])
-            )
-            check_pytest = pytest_mentioned and pytest_installed
-            assert check_pytest, "pytest with package version not in print output"
+        pytest_mentioned = "pytest:" in out
+        pytest_installed = "pytest: not installed" not in out
+        check_pytest = pytest_mentioned and pytest_installed
+        assert check_pytest, "pytest with package version not in print output"
 
-    def test_show_missing_specific_version(self):
+    def test_show_missing_specific_version(self, capsys):
         """Test 'show_version' works with missing package."""
         from satpy.utils import show_versions
+        show_versions(packages=["__fake"])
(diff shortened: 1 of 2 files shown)


djhoese (commented):
I don't know how I feel about these changes and I haven't felt great about the discussion since it was restarted in the related issue. This may mean that I'm just missing something from the feature request, but overall it feels like a lot of flexibility is being added and things changed for something that I consider rather static and single-use.

1. Why does `show_versions` need to have a `packages` keyword argument? In my opinion (others can feel free to disagree) this should just be a hardcoded list. This function is for us, the maintainers, not for users. Users run it to tell us what we need to know.
2. Why should `check_satpy` call `show_versions`? It feels like they should be separate. There are times when duplicated code is acceptable and if these changes were done to reduce duplication I'm not sure it is worth it.
3. Why did `extras` need to be renamed?

It seems like both `check_satpy` and `show_versions` were implemented/re-implemented to be even more flexible and to be a generic "check this import for me" and I'm just not sure that task is complicated enough that Satpy needs to be providing a generic utility for users to use. What I tried to say above is that I think of these types of functions as tools to assist maintainers in helping users, not as generic tools for the user.

verduijn:
Thanks for the feedback!

I figured that `check_satpy` was in the default issue template, so it could be useful to have the versions information in there. 

From there:  
- I thought i could expose the `show_versions`  functionality  itself without much downside
- `check_satpy` was flexible with `extras` so I just did the same
- since `extras` and `packages` was essentially the same I just renamed it

rounding up I think we should either:
- remove `show_versions` and add it, with hardcoded packages, in `check_satpy`
- Having a hard coded `show_versions` function and reverting the `check_satpy` function if it's not desirable to add the versions in `check_satpy`.

djhoese:
Good point with it being in the issue template. We wouldn't want users to have to run multiple things to show similar information. In that case I agree that the versions after the readers/writers in `check_satpy` does make more sense than the "ok" that is there now. Ok I'm coming around. Thanks for explaining your reasoning.

satpy/utils.py
```diff
@@ -476,6 +478,19 @@ def _check_yaml_configs(configs, key):
...
+
+    Returns:
+        the version number if available else `None`.
+    """
+    try:
+        return importlib.metadata.version(package_name)
+    except importlib.metadata.PackageNotFoundError:
+        return None
 
 def _check_import(module_names):
```
djhoese, line 495: It's hard to tell from github's diff display, but is this function used anymore?

verduijn: good point! I've removed the now unused `_check_imports`.

verduijn:
I guess it's ready for review?

djhoese (approved):
From a basic inspection, these changes seem good to me. I'd like some other reviews though. (@mraspaud @pnuu @sfinkens maybe?)

satpy/tests/test_utils.py
```diff
@@ -278,13 +278,17 @@ def test_specific_check_satpy(self):
...
-            checked_fake = False
-            for call in print_mock.mock_calls:
-                if len(call[1]) > 0 and "__fake" in call[1][0]:
-                    assert "ok" not in call[1][1]
-                    checked_fake = True
-            assert checked_fake, "Did not find __fake module mentioned in checks"
+            check_satpy(readers=["viirs_sdr"], packages=("cartopy", "__fake"))
+            checked_fake = any("__fake: not installed" in c[1] for c in print_mock.mock_calls if len(c[1]))
+            assert checked_fake, "Did not find __fake package mentioned in checks"
+
```
pnuu, line 284: Two empty lines required between classes
```suggestion


```

satpy/utils.py
```diff
@@ -476,30 +478,75 @@ def _check_yaml_configs(configs, key):
...
+
+    Returns:
+        the version number if available else `None`.
+    """
+    try:
+        return importlib.metadata.version(package_name)
+    except importlib.metadata.PackageNotFoundError:
+        return None
 
+def show_versions(packages=None):
```
pnuu, line 495: Two empty lines required between functions.
```suggestion

def show_versions(packages=None):
```

pnuu (changes requested):
Just couple of styling issues that I think linter should have caught.

satpy/tests/test_utils.py
verduijn: Good suggestion, I relied on the pre-commit script and formatting the diff but that doesn't catch these

satpy/tests/test_utils.py
```diff
@@ -278,13 +278,18 @@ def test_specific_check_satpy(self):
...
+
+class TestShowVersions:
+    """Test the 'show_versions' function."""
+
+    def test_basic_show_versions(self):
+        """Test 'check_satpy' basic functionality."""
+        from satpy.utils import show_versions
+        show_versions()
```
sfinkens, line 292: I think it would be nice to add a test with a package that doesn't exist, `show_versions(packages=["foo"])` and check that the output contains "not installed".

sfinkens (commented):
Nice work, thanks! I have one wish for the tests.

satpy/tests/test_utils.py
verduijn: this is somewhat implicitly done in the test for check_satpy that has both a missing and installed package. I think it does make sense to explicitly test it so I've added two tests.

sfinkens: Oh, I missed that. Thanks for adding the extra tests! Just curious: Would it be possible to use the [capsys fixture](https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-capsys) instead of mocking the print method?

verduijn: that is much cleaner, I've changed the test in this MR. perhaps in a different MR i can apply it to other places where print is being patched?

sfinkens: > I've changed the test in this MR

Awesome, thanks!

> perhaps in a different MR i can apply it to other places where print is being patched

Sure, if you have some spare time, go for it :)

sfinkens (approved):

PR #2915, opened 2024-09-30
Improve SEVIRI metadata documentation

sfinkens:
Add an example showing how to access raw metadata.

Also, pin mamba version to latest release < 2 as a temporary workaround until the upstream issue with conda-incubator is fixed.


mraspaud (approved):
LGTM

PR #2923, opened 2024-10-10
Fix data promotion in `generic_image` reader and `satpy.composites.add_bands` 

pnuu:
While figuring out why `geo_color` was `float64`, I found that both `generic_image` reader (used to read background images) and `satpy.composites.add_bands()` were mishandling the data. This fixes these two bugs in one go, as they both are triggered by `BackgroundCompositor` for certain cases.

Requires also https://github.com/pytroll/trollimage/pull/186 for the `geo_color` to be fully fixed, the `invert` enhancement promotes data to `float64` when the inversion parameters are given for each channel.

The `generic_image` reader tests were converted to `pytest`.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -985,6 +985,7 @@ def add_bands(data, bands):
         alpha = new_data[0].copy()
         alpha.data = da.ones((data.sizes["y"],
                               data.sizes["x"]),
+                             dtype=new_data[0].dtype,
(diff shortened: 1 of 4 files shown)


mraspaud (approved):
LGTM, thanks a lot for refactoring the tests!

PR #2929, opened 2024-10-12
Replace patched `print` with capsys fixture

verduijn:
Small change to replace patched print with nicer capsys fixture. Figured I would replace it other tests, turns out there was only one location.

### DIFF
--- a/satpy/tests/test_utils.py
+++ b/satpy/tests/test_utils.py
@@ -274,13 +274,15 @@ class TestCheckSatpy:
         from satpy.utils import check_satpy
         check_satpy()
 
-    def test_specific_check_satpy(self):
+    def test_specific_check_satpy(self, capsys):
         """Test 'check_satpy' with specific features provided."""
         from satpy.utils import check_satpy
-        with mock.patch("satpy.utils.print") as print_mock:
-            check_satpy(readers=["viirs_sdr"], packages=("cartopy", "__fake"))
-            checked_fake = any("__fake: not installed" in c[1] for c in print_mock.mock_calls if len(c[1]))
-            assert checked_fake, "Did not find __fake package mentioned in checks"
+        check_satpy(readers=["viirs_sdr"], packages=("cartopy", "__fake"))
+        out, _ = capsys.readouterr()
+        checked_fake = "__fake: not installed" in out
+        checked_viirs_sdr = "Readers\n=======\nviirs_sdr" in out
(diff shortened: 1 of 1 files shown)


satpy/tests/test_utils.py
```diff
@@ -274,13 +274,15 @@ def test_basic_check_satpy(self):
...
-        with mock.patch("satpy.utils.print") as print_mock:
-            check_satpy(readers=["viirs_sdr"], packages=("cartopy", "__fake"))
-            checked_fake = any("__fake: not installed" in c[1] for c in print_mock.mock_calls if len(c[1]))
-            assert checked_fake, "Did not find __fake package mentioned in checks"
+        check_satpy(readers=["viirs_sdr"], packages=("cartopy", "__fake"))
+        out, _ = capsys.readouterr()
+        checked_fake = "__fake: not installed" in out
+        checked_viirs_sdr = "Readers\n=======\nviirs_sdr" in out
+        assert checked_fake, "Did not find __fake package mentioned in checks"
+        assert checked_viirs_sdr, "Did not find viirs_sdr in readers mentioned in checks"
```
pnuu, line 285: I don't think these do what are meant to do:

```python
>>> foo = "bar"
>>> assert foo, "bar"
>>> assert foo, "foo"
>>> assert foo == "foo"
AssertionError
```

pnuu, line 285:
```suggestion
        assert checked_fake == "Did not find __fake package mentioned in checks"
        assert checked_viirs_sdr == "Did not find viirs_sdr in readers mentioned in checks"
```

pnuu (changes requested):

satpy/tests/test_utils.py
verduijn: I think this is working as intended (though i'm following along what was there before).
```bash
I /tmp> cat dbg.py 
assert True, "you wont see me"
assert False, "this is a helpful message"
I /tmp> python3 dbg.py 
  File "/tmp/dbg.py", line 2, in <module>
    assert False, "this is a helpful message"
           ^^^^^
AssertionError: this is a helpful message
I /tmp> 
```

pnuu: Oh, right. I misunderstood the assertion as I've never seen that before :see_no_evil:

pnuu (approved):
LGTM, thanks and sorry about the noise!

PR #2938, opened 2024-10-17
Update Landsat reader for compatibility with Pyspectral.

simonrp84:
The Landsat reader used different nomenclature for the band names to that used by Pyspectral. This PR updates the reader for compatibility with the pyspectral LUTs.

Note: Only Landsat-8 is supported by pyspectral at present, and the thermal bands are not included. A separate PR for pyspectral will include both thermal and Landsat-9 support.

 - [x] Tests added


mraspaud (approved):
LGTM

Issue #2897, opened 2024-09-09
generic_image reader returns data as float64 for PNG images


strandgren:
**Describe the bug**
The `generic_image` reader returns the image data as `float64` for PNG images. This leads to **corrupt** images with the PNG and JPEG writers when trying to save the image back to a file again, while the geotiff writer falls over completely when trying to save the image back to a TIF file.

**To Reproduce**

```python
from satpy import Scene

infiles = ['input.jpg', 'input.png']
outfiles =  ['output.jpg', 'output.png']
for infile, outfile in zip(infiles, outfiles):
    scn = Scene(filenames=[infile], reader='generic_image')
    scn.load(['image'])
    print(scn['image'].dtype)
    scn.save_dataset('image', outfile, enhance=False, fill_value=0)

```

**Expected behavior**
```
uint8
uint8
```
and one copy each of the input images.

**Actual results**
```
uint8
float64
```
and a corrupt output.png image as shown in the screenshot below.
![image](https://github.com/user-attachments/assets/b2e8d1f2-daa4-4bf3-9100-d64f679950c9)

**Environment Info:**
 - OS: Linux
 - Satpy Version: 0.51.0

**Additional context**
When trying to save `input.png` as `output.tif` the geotiff writer fails with the following traceback:
```
Traceback (most recent call last):
  File "/tcenas/home/strandgren/git/eum/py/mtgi_imagery_stream/te_generic_reader.py", line 9, in <module>
    scn.save_dataset('image', outfile, enhance=False, fill_value=0)
  File "/tcenas/proj/optcalimg/miniconda3/envs/pysat/lib/python3.10/site-packages/satpy/scene.py", line 1234, in save_dataset
    return writer.save_dataset(self[dataset_id],
  File "/tcenas/proj/optcalimg/miniconda3/envs/pysat/lib/python3.10/site-packages/satpy/writers/__init__.py", line 885, in save_dataset
    return self.save_image(img, filename=filename, compute=compute, fill_value=fill_value, **kwargs)
  File "/tcenas/proj/optcalimg/miniconda3/envs/pysat/lib/python3.10/site-packages/satpy/writers/geotiff.py", line 263, in save_image
    raise ValueError("Image must be in 'L' mode for floating "
ValueError: Image must be in 'L' mode for floating point geotiff saving
```

djhoese:
If you run `gdalinfo input.png`, what do you get?

djhoese:
[...]

PR #2940, opened 2024-10-18
Make it possible to set fill value when reading images

### LINKED ISSUE #2928: Allow generic_image reader to set fill value on reading
While looking into https://github.com/pytroll/satpy/pull/2923 I went through the `generic_image` reader code. All the images with `A` channel are converted to floats and `NaN`s are placed to the no data pixels. If there's no `A` but the image has a `nodatavals` metadata item present and `nodata_handling` is set to `"fill_value"` the image is kept as integer data. I'm not sure PNG supports the no data value (or fill value).

Maybe the `generic_image` reader should have a feature to force a `fill_value` that would replace the `A` band whether there is a fill value in the original image or not.

_Originally posted by @pnuu in https://github.com/pytroll/satpy/issues/2897#issuecomment-2406692422_

pnuu:
This PR makes it possible to set fill value when reading for LA and RGBA images with `generic_image` reader. When this is done, the `A` band is dropped. Previously RGBA images were converted to float, which caused problems in https://github.com/pytroll/satpy/issues/2897 .

 - [x] Closes #2928
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/generic_image.py
+++ b/satpy/readers/generic_image.py
@@ -54,11 +54,15 @@ logger = logging.getLogger(__name__)
 class GenericImageFileHandler(BaseFileHandler):
     """Handle reading of generic image files."""
 
-    def __init__(self, filename, filename_info, filetype_info):
+    def __init__(self, filename, filename_info, filetype_info, set_fill_value=None, nodata_handling=None):
         """Initialize filehandler."""
         super(GenericImageFileHandler, self).__init__(
             filename, filename_info, filetype_info)
         self.finfo = filename_info
+        self.set_fill_value = set_fill_value
+        if self.set_fill_value is not None and not isinstance(self.set_fill_value, (list, tuple)):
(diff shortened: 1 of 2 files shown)


pnuu:
This is still a very rough draft, so comments are wellcome.

strandgren:
Thanks for working on this, I can confirm that by setting `reader_kwargs={"set_fill_value": 255}` the issue I originally had in https://github.com/pytroll/satpy/issues/2897 is resolved :)

Would it be possible though to read a PNG image with alpha channel and store it again with `Scene.save_datasets` and retain the alpha channel? So far I have not found a way.

djhoese:
`set_fill_value` -> `fill_value`?

Seems odd to me to have a verb in a variable name. Then again I am probably guilty of doing `use_x_feature=True` or `filter_x=True`. But in this case the action you're doing is the variable you're specifying.

Issue #2932, opened 2024-10-16
Remove mocking in eum_l2_grib reader tests


strandgren:
## Feature Request

Nowadays the preferred approach is to write a synthetic testing file to tmp_path and work with that instead. However the unit tests for the `eum_l2_grib` reader (`test_eum_l2_grib.py`) still uses mocking for creating test files. It shall be assessed if we can write synthetic test files also for the grib format and therefor get rid of the mocking.

PR #2943, opened 2024-10-21
Fix seviri_l2_grib end_time property bug.

### LINKED ISSUE #2942: 0.52.0 breaks `seviri_l2_grib`-reader with 'EUML2GribFileHandler' object has no attribute '_ssp_lon'
**Describe the bug**
`0.52.0` breaks `seviri_l2_grib`-reader

**To Reproduce**

```python
# Your code here
scene = satpy.Scene(reader='seviri_l2_grib', filenames=files)
```

**Expected behavior**
To not crash.

**Actual results**
````
  File "/Users/username/PycharmProjects/data-engineering-exchange/MTG_FCI_L1/open_and_read.py", line 23, in <module>
    scene = satpy.Scene(reader='seviri_l2_grib', filenames=files)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/geo/lib/python3.11/site-packages/satpy/scene.py", line 155, in __init__
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/geo/lib/python3.11/site-packages/satpy/scene.py", line 176, in _create_reader_instances
  File "/opt/homebrew/Caskroom/miniforge/base/envs/geo/lib/python3.11/site-packages/satpy/readers/__init__.py", line 572, in load_readers
    reader_instance.create_storage_items(
  File "/opt/homebrew/Caskroom/miniforge/base/envs/geo/lib/python3.11/site-packages/satpy/readers/yaml_reader.py", line 617, in create_storage_items
    return self.create_filehandlers(files, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/geo/lib/python3.11/site-packages/satpy/readers/yaml_reader.py", line 629, in create_filehandlers
    filehandlers = self._new_filehandlers_for_filetype(filetype_info,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/geo/lib/python3.11/site-packages/satpy/readers/yaml_reader.py", line 612, in _new_filehandlers_for_filetype
    return list(filtered_iter)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/geo/lib/python3.11/site-packages/satpy/readers/yaml_reader.py", line 596, in filter_fh_by_metadata
    filehandler.metadata["end_time"] = filehandler.end_time
                                       ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/geo/lib/python3.11/site-packages/satpy/readers/eum_l2_grib.py", line 81, in end_time
    delta = SEVIRI_REPEAT_CYCLE_DURATION_RSS if self._ssp_lon == 9.5 else SEVIRI_REPEAT_CYCLE_DURATION
                                                ^^^^^^^^^^^^^
AttributeError: 'EUML2GribFileHandler' object has no attribute '_ssp_lon'
```

**Environment Info:**
 - OS: all
 - Satpy Version: 0.52.0
 - PyResample Version: independent

**Additional context**
Add any other context about the problem here.

strandgren:
As described in #2942, the changes in https://github.com/pytroll/satpy/pull/2717 lead to the `end_time` property not working for SEVIRI data, since we now use the sub-satellite point to determine the scanning mode and thus the temporal resolution used to determine the `end_time` from the `start_time` (prior to https://github.com/pytroll/satpy/pull/2717 it was always set to 15 minutes, which is incorrect for RSS data). However, the `self._ssp_lon` attribute is only available after a dataset and its metadata have been loaded and thus an `AttributeError` was raised during `Scene` initialization. This was not covered by the unit test and a human error lead to the bug being unnoticed also when testing with real data.

This PR adapts the `end_time` property to return `None` if no dataset has been loaded, meaning that `_ssp_lon` is not available. As soon as a dataset has been loaded, both `scn.end_time` and `scn[dataset].attrs['end_time']` will have valid and correct `datetime.datetime` objects. The unit test has been adapted to cover this bug.

 - [X] Closes #2942 
 - [X] Tests adapted to capture this bug

### DIFF
--- a/satpy/tests/reader_tests/test_eum_l2_grib.py
+++ b/satpy/tests/reader_tests/test_eum_l2_grib.py
@@ -133,11 +133,15 @@ def test_seviri_data_reading(da_, xr_, setup_reader):
 
             dataset_id = make_dataid(name="dummmy", resolution=3000)
 
+            # Check that end_time is None for SEVIRI before the dataset has been loaded
+            assert reader.end_time is None
+
             common_checks(ec_, reader, mock_file, dataset_id)
 
-            # Check end_time
+            # Check that end_time is now a valid datetime.datetime object after the dataset has been loaded
             assert reader.end_time == datetime.datetime(year=2020, month=10, day=20,
                                                         hour=19, minute=50, second=0)
+
 
             # Checks the correct execution of the _get_global_attributes and _get_metadata_from_msg functions
             attributes = reader._get_attributes()
@@ -234,11 +238,11 @@ def test_fci_data_reading(da_, xr_, setup_reader):
 
             dataset_id = make_dataid(name="dummmy", resolution=2000)
 
-            common_checks(ec_, reader, mock_file, dataset_id)
-
             # Check end_time
             assert reader.end_time == datetime.datetime(year=2020, month=10, day=20,
                                                         hour=19, minute=50, second=0)
+
+            common_checks(ec_, reader, mock_file, dataset_id)
 
             # Checks the correct execution of the _get_global_attributes and _get_metadata_from_msg functions
             attributes = reader._get_attributes()
--- a/satpy/readers/eum_l2_grib.py
+++ b/satpy/readers/eum_l2_grib.py
@@ -78,8 +78,12 @@ class EUML2GribFileHandler(BaseFileHandler):
     def end_time(self):
         """Return the sensing end time."""
         if self.sensor == "seviri":
-            delta = SEVIRI_REPEAT_CYCLE_DURATION_RSS if self._ssp_lon == 9.5 else SEVIRI_REPEAT_CYCLE_DURATION
-            return self.start_time + dt.timedelta(minutes=delta)
+            try:
+                delta = SEVIRI_REPEAT_CYCLE_DURATION_RSS if self._ssp_lon == 9.5 else SEVIRI_REPEAT_CYCLE_DURATION
+                return self.start_time + dt.timedelta(minutes=delta)
+            except AttributeError:
+                # If dataset and metadata (ssp_lon) have not yet been loaded, return None
+                return None
         elif self.sensor == "fci":
             return self.filename_info["end_time"]
 


pnuu:
I have a feeling that either https://github.com/pytroll/trollimage/pull/179 or https://github.com/pytroll/trollimage/pull/181 cause the test failure, I created a new release today with these two included. Although there's no new `conda-forge` package which I would have thought the CI would be using :thinking:

pnuu:
Trollimage seems to be coming from PyPI: https://github.com/pytroll/satpy/blob/main/continuous_integration/environment.yaml#L64

@gerritholl could you perhaps have a look at the failing tests that seem to be handling paletted images and enhancement history?
* satpy/tests/enhancement_tests/test_enhancements.py::TestEnhancementStretch::test_palettize
* satpy/tests/enhancement_tests/test_viirs.py::TestVIIRSEnhancement::test_viirs

pnuu:
I think it should be as simple as updating the expected results.

pnuu:
Looks like the enhancement test function is used in many places other the two failing tests, so needs a bit more involved solution. I'll have a go.

pnuu:
@strandgren the fix to the failing tests has been merged, so please merge `main` to this PR.

strandgren:
@pnuu Thanks, now the tests passed :)

mraspaud (approved):
I think I can live with this solution :)
However, I want to point out that it feels strange to have a property return different results depending on whether the file has already been open or not. Ideally, the call to `end_time` should trigger reading just what is needed in the file to give a sensible result.
But since I don't know the format, I think we can merge this.

strandgren:
> I think I can live with this solution :) However, I want to point out that it feels strange to have a property return different results depending on whether the file has already been open or not. Ideally, the call to `end_time` should trigger reading just what is needed in the file to give a sensible result. But since I don't know the format, I think we can merge this.

Yes I agree, but the metadata are provided for each grib message i.e. dataset meaning that at Scene initialization it would be ambiguous from which message/dataset to read the `ssp_lon` from. Of course it will most likely be the same for all datasets and we could read it from the first message, but I still didn't like that solution, also since it invokes some duplicate reading.

satpy/tests/reader_tests/test_eum_l2_grib.py
```diff
@@ -133,12 +133,16 @@ def test_seviri_data_reading(da_, xr_, setup_reader):
...
+            assert reader.end_time is None
+
             common_checks(ec_, reader, mock_file, dataset_id)
 
-            # Check end_time
+            # Check that end_time is now a valid datetime.datetime object after the dataset has been loaded
             assert reader.end_time == datetime.datetime(year=2020, month=10, day=20,
                                                         hour=19, minute=50, second=0)
 
+
```
pnuu, line 145: Could this test case be split into several specific tests?

strandgren: hm, I've been trying to refactor this into something better, but so far with no satisfying results. I have previously created an issue to work on removing the mocking #2932, do you think we could move this test optimization to the PR that will deal with the mocking (since it will anyway lead to quite some refactoring)?

pnuu: Yeah, sounds good! Feel free to merge.

pnuu (approved):
LGTM

satpy/tests/reader_tests/test_eum_l2_grib.py
strandgren: Thanks! Nah, you do it, you're the reviewer ;)

PR #2944, opened 2024-10-21
Fix tests using `palettize`

pnuu:
After merging https://github.com/pytroll/trollimage/pull/179 and https://github.com/pytroll/trollimage/pull/181 and creating a new Trollimage release, two tests using `palettize` and `water_detection` (also does `palettize`) were failing because of the added `_FillValue` attribute. The fix gets a bit ugly because `colorize` also passes `palettes` but doesn't add the `_FillValue` attribute.

 - [x] Tests added


gerritholl (approved):
Looks good to me.

PR #2946, opened 2024-10-22
Update MODIS L1b reader with additional geoinfo datasets

simonrp84:
The MODIS L1b reader misses a few useful datasets from the MOD03 files. This PR adds the most commonly used of those datasets:
`land_water_mask`, `height`, `range`, `water_present`.

 - [ ] Tests passing

### DIFF
--- a/satpy/readers/hdfeos_base.py
+++ b/satpy/readers/hdfeos_base.py
@@ -333,6 +333,10 @@ class HDFEOSGeoReader(HDFEOSBaseFileReader):
         "satellite_zenith_angle": ("SensorZenith", "Sensor_Zenith"),
         "solar_azimuth_angle": ("SolarAzimuth", "SolarAzimuth"),
         "solar_zenith_angle": ("SolarZenith", "Solar_Zenith"),
+        "water_present": "WaterPresent",
+        "land_sea_mask": "Land/SeaMask",
+        "height": "Height",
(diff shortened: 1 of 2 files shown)


satpy/readers/hdfeos_base.py
```diff
@@ -333,6 +333,10 @@ class HDFEOSGeoReader(HDFEOSBaseFileReader):
         "satellite_zenith_angle": ("SensorZenith", "Sensor_Zenith"),
         "solar_azimuth_angle": ("SolarAzimuth", "SolarAzimuth"),
         "solar_zenith_angle": ("SolarZenith", "Solar_Zenith"),
+        "water_present": "WaterPresent",
+        "land_sea_mask": "Land/SeaMask",
+        "height": "Height",
+        "range": "Range",
```
djhoese, line 339: I'm tempted to say these should be assigned via the `file_key` in the YAML definition, but then again the way you've done this isn't disallowed... :man_shrugging:

djhoese (approved):

satpy/readers/hdfeos_base.py
simonrp84: I'm just copying the format of what was already there, tbh!

djhoese (approved):
Looks simple enough to me. :+1:

PR #2950, opened 2024-10-24
Fix dtype promotion in `SunZenithReduction`

pnuu:
The `SunZenithReduction` modifier used for example in the built-in FCI `true_color` composite has a division with `np.log(2)` that changes the `float32` input data to `float64`. This PR is a simple fix for that.

The upcasting is hidden behind some other part of the code because even without this fix the completed product was `float32`.

 - [x] Tests added 

This small change also reduced the memory usage for my 12-composite set by roughly 800 MB.

With `main`:
Screenshot 2024-10-24 at 10-01-12 Bokeh Plot

With this PR:
Screenshot 2024-10-24 at 10-01-18 Bokeh Plot

### DIFF
--- a/satpy/modifiers/angles.py
+++ b/satpy/modifiers/angles.py
@@ -583,7 +583,7 @@ def _sunzen_reduction_ndarray(data: np.ndarray,
 
     # invert the reduction factor such that minimum reduction is done at `limit` and gradually increases towards max_sza
     with np.errstate(invalid="ignore"):  # we expect space pixels to be invalid
-        reduction_factor = 1. - np.log(reduction_factor + 1) / np.log(2, dtype=data.dtype)
+        reduction_factor = 1. - np.log2(reduction_factor + 1)
 
     # apply non-linearity to the reduction factor for a non-linear reduction of the signal. This can be used for a
     # slower or faster transision to higher/lower fractions at the ndvi extremes. If strength equals 1.0, this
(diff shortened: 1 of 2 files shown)


pnuu:
The other *two* things doing upcasting for True color composite that hid this one were fixed in the merged #2954 and #2956 PRs.

djhoese (approved):
So this fixes some extra memory being used inside a dask task, but doesn't change the final dtype of the output of this "angles" step (which was correct), right?

Side note: Based on this math website, can't this equation be rewritten as log-base2 of `reduction_factor + 1`?

https://www.cuemath.com/algebra/log-rules/

```
change of base rule: loga b = (logc b) / (logc a)
```

pnuu:
With the current  `main` `float32` input data resulted in output of `float64` because of `np.log(2)` promoting `reduction_factor` to `float64` and finally causing the same to `res`. With this fix the output will also be `float32`. `float64` input will still yield `float64` result.

pnuu:
And well spotted, the `log2` is much simpler in every way. Fixed in https://github.com/pytroll/satpy/pull/2950/commits/619304061e195d74ba28c7b498603f19c0fe38c2

pnuu:
> > And well spotted, the log2 is much simpler in every way. Fixed in [6193040](https://github.com/pytroll/satpy/commit/619304061e195d74ba28c7b498603f19c0fe38c2)
> 
> Hopefully that is still mathematically equivalent and I'm not misreading it.

It is, the tests pass (at least locally).

pnuu:
I restarted CI, MacOS tests failed on setup.

PR #2958, opened 2024-10-30
Remove co2_corrected modifier for METimage 3.74 µm

gerritholl:
Remove the co2_corrected modifier for the METimage 3.74 µm channel.  The CO₂ correction was developed specifically for the SEVIRI 3.9 µm channel and should not be applied to channels where the CO₂ impact is different / (mostly) absent.


adybbroe (approved):
Perfect! We definitely should not have the legacy (impirically based SEVIRI biased) CO2 correction here! Thx

adybbroe:
rsr_band_0350_0400

Here we see how SEVIRI is really different from the "rest" concerning this channel

PR #2965, opened 2024-11-04
Fix MODIS readers chunking compatibility with newer dask

djhoese:
Triggered by changes in dask (see https://github.com/dask/dask/issues/11341), but I think the main issue is that I was specifying `previous_chunks` with `-1` which I can understand would be confusing if dask assumes "yeah this is what the chunks are now".

This should fix unstable tests related to MODIS readers.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


mraspaud (approved):
LGTM

PR #2970, opened 2024-11-07
Remove attempt to Rayleigh correct the VIIRS I02 band 

adybbroe:
Remove attempt to Rayleigh correct the VIIRS I02 band for the false-color RGB. The band is beyond 0.8 micron and there is effectively no rayleigh scattering in this band. Pyspectral cannot correct this anyhow.

We get warning messages in our production, like this:
`WARNING  pyspectral.rayleigh] Effective wavelength for band I02 outside nominal 400-800 nm range!`

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/composites/viirs.yaml
+++ b/satpy/etc/composites/viirs.yaml
@@ -232,7 +232,7 @@ composites:
       modifiers: [sunz_corrected, rayleigh_corrected]
     optional_prerequisites:
     - name: I02
-      modifiers: [sunz_corrected_iband, rayleigh_corrected_iband]
+      modifiers: [sunz_corrected_iband]
     standard_name: false_color
     high_resolution_band: green
 


adybbroe:
Here is the spectral response function of I2 for reference:

rsr_band_I2

mraspaud (approved):
LGTM!

djhoese:
I've updated the title to reflect that this is for the false color RGB specifically.

satpy/etc/composites/viirs.yaml
```diff
@@ -232,7 +232,7 @@ composites:
       modifiers: [sunz_corrected, rayleigh_corrected]
     optional_prerequisites:
     - name: I02
-      modifiers: [sunz_corrected_iband, rayleigh_corrected_iband]
```
djhoese, line 235: Should rayleigh be removed from the M05 band too? I suppose that is still in the "correctable" range?

PR #2972, opened 2024-11-07
Remove the application of Rayleigh correction above 0.8 micron on the Geo sats

adybbroe:
This PR removes the attempt to apply Rayleigh correction above 0.8 micron on the Geo sats, where it has no effect anyhow.

It also adds back the rayleigh correction at two places, both for the natual-color RGB on ABI and AHI, where it was commented out - don't know for what reason it was taken away? Maybe that needs clarification?

This PR is similar to #2970 - that did the same for VIIRS

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/composites/ahi.yaml
+++ b/satpy/etc/composites/ahi.yaml
@@ -218,7 +218,7 @@ composites:
     - wavelength: 0.85
       modifiers: [sunz_corrected]
     - wavelength: 0.635
-      modifiers: [sunz_corrected, rayleigh_corrected]
+      modifiers: [sunz_corrected]
     high_resolution_band: blue
     standard_name: natural_color
 
--- a/satpy/etc/composites/ahi.yaml
+++ b/satpy/etc/composites/ahi.yaml
@@ -214,11 +214,11 @@ composites:
     compositor: !!python/name:satpy.composites.SelfSharpenedRGB
     prerequisites:
     - wavelength: 1.63
-      modifiers: [sunz_corrected] #, rayleigh_corrected]
+      modifiers: [sunz_corrected]
     - wavelength: 0.85
-      modifiers: [sunz_corrected] #, rayleigh_corrected]
+      modifiers: [sunz_corrected]
     - wavelength: 0.635
-      modifiers: [sunz_corrected] #, rayleigh_corrected]
+      modifiers: [sunz_corrected, rayleigh_corrected]
     high_resolution_band: blue
     standard_name: natural_color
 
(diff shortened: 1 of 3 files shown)


adybbroe:
Here the spectral responses for the three sats affected, for reference:

rsr_band_0070_0090

adybbroe:
@gerritholl and @pnuu You make imagery from both AHI and ABI right?
Do you use the `nartual-color` RGB? In case you do, is there a reason not to include the Rayleigh correction on the 0.6 micron channel? I guess the effect is really small, but to be consistent with how we do elsewhere I think it should also be corrected for the false-color RGB?

gerritholl:
@adybbroe We process ABI and AHI, but we produce only one RGB (airmass), so have no experience with Rayleigh correction for those sensors.

This would be a good case to test our image comparison tests!  See https://github.com/pytroll/image-comparison-tests-dev/pull/2

pnuu:
> have we ever applied rayleigh correction to natural_colors?

No, I don't think we have.

adybbroe:
> have we ever applied rayleigh correction to natural_colors?

I just saw that it was there but commented out. So, I suspect it was at some point?

adybbroe:
Can it be for computational speed it is avoided (and the fact that it doesn't probably have a big positive/negative effect either or)?

satpy/etc/composites/ahi.yaml
```diff
@@ -214,11 +214,11 @@ composites:
...
     prerequisites:
     - wavelength: 1.63
-      modifiers: [sunz_corrected] #, rayleigh_corrected]
+      modifiers: [sunz_corrected]
     - wavelength: 0.85
-      modifiers: [sunz_corrected] #, rayleigh_corrected]
+      modifiers: [sunz_corrected]
     - wavelength: 0.635
-      modifiers: [sunz_corrected] #, rayleigh_corrected]
+      modifiers: [sunz_corrected, rayleigh_corrected]
```
adybbroe, line 221:
```suggestion
      modifiers: [sunz_corrected]
```

satpy/etc/composites/ami.yaml
```diff
@@ -137,11 +137,11 @@ composites:
...
     prerequisites:
       - name: NR016
-        modifiers: [sunz_corrected] #, rayleigh_corrected]
+        modifiers: [sunz_corrected]
       - name: VI008
-        modifiers: [sunz_corrected] #, rayleigh_corrected]
+        modifiers: [sunz_corrected]
       - name: VI006
-        modifiers: [sunz_corrected] #, rayleigh_corrected]
+        modifiers: [sunz_corrected, rayleigh_corrected]
```
adybbroe, line 144:
```suggestion
        modifiers: [sunz_corrected]
```
I think it is best not to include the correction at this point. But to be considered.

PR #2978, opened 2024-11-13
support bz2 compression in avhrr_l1b_eps

k3a:
EUMETCast distributes global METOP AVHRR data bz2-compressed.
This commit adds decompression support by reusing already-tested `unzip_context` helper function.


mraspaud (changes requested):
Thanks a lot for this new functionality! Can you find a way to add a small unit test for this?

pdebuyl:
Hello @mraspaud , as I was involved in the unzip context I though of giving this PR a boost.

But the tests are not file-based. I propose to convert the test file to not use mocks anymore. But I can't make sense of the test `test_get_full_angles_twice`... 

Also, the failing tests seem to rely on parsing the log of satpy and checking for the appearance of "WARNING" in there. Probably, this PR (here, #2978) can't proceed before the tests are completely updated.

PR #2993, opened 2024-11-21
fix: li missing platform name

BENR0:
This fixes the `DataArrays` returned by the li reader missing the `platform_name` attribute.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


ameraner:
Thanks @BENR0 , I've noticed this as well recently. Could we take the platform info from the filename and use the mapping as done for FCI here? https://github.com/pytroll/satpy/blob/1a7a34201f05b37b301881c200ff656e74b6b9f8/satpy/readers/fci_l1c_nc.py#L205-L209

Although the question of the platform name and the operational renaming to e.g. MET-12 is very valid...

BENR0:
@ameraner Maybe we just wait till it is renamed to fix it? Shouldn't be that long till then I guess? ;-)
And yes I think a mapping would make it more future proof.

ameraner:
Yeah, should only be a couple of weeks or so :) As you prefer, I'm ok also with merging this and update again later. 
Then I would use a mapping/naming similar to what is used for SEVIRI, like `Meteosat-XX` https://github.com/pytroll/satpy/blob/d22d2b6f17c850ce45f399d828435520530b4ff2/satpy/readers/seviri_l1b_native.py#L317

BENR0:
@ameraner Do you know if the name will eventually also be changed on https://space.oscar.wmo.int/satellites/ from MTG-I1 to Meteosat-12?

ameraner:
I assume so, as the MSG satellites are also listed as Meteosat-XX.

BENR0:
@ameraner @mraspaud I changed the platform name for FCI/LI according to the renaming after MTG went operational.

ameraner:
Thanks! the changes look good to me, however it seems that both the FCI and LI tests need to be updated..

ameraner (approved):
LGTM, thanks!

satpy/readers/fci_l1c_nc.py
```diff
@@ -182,6 +182,18 @@ def _get_channel_name_from_dsname(dsname):
...
+# MTI3, and MTI4, but we want to use names such as described in WMO OSCAR
+# MTG-I1, MTG-I2, MTG-I3, and MTG-I4.
+#
+# Not sure how the numbering will be considering MTG-S1 and MTG-S2 will be launched
+# in-between.
+_platform_name_translate = {
+    "MTI1": "Meteosat-12",
+    "MTI2": "MTG-I2",
+    "MTI3": "MTG-I3",
+    "MTI4": "MTG-I4"}
```
gerritholl, line 196: I would move this to `fci_base.py` so it can also be used by `fci_l2_nc.py` (see #3026).  Or even an `mtg_base.py` to be used by both FCI and LI.

satpy/readers/li_base_nc.py
```diff
@@ -191,6 +191,7 @@
 import xarray as xr
 from pyproj import Proj
 
+from satpy.readers.fci_l1c_nc import _platform_name_translate
```
gerritholl, line 194: Since we're now using this across modules, should we remove the `_` prefix?

gerritholl (commented):
Since you're at it anyway, would you mind expanding this to FCI L2 and fix #3026 in the same go?  It fits thematically with this PR.  Or we can fix it in a new PR.

satpy/readers/fci_l1c_nc.py
BENR0: Good idea. Wasn't aware of that otherwise would have done it directly :-).

satpy/readers/li_base_nc.py
BENR0: From my view yes. But will wait for other feedback.

djhoese: Without viewing the rest of this PR, I agree that something imported across modules should not have the `_` prefix.

mraspaud (approved):
LGTM

pnuu (approved):

PR #3002, opened 2024-11-27
ISCCP-NG level 1g reader

ninahakansson:
This is a reader for the ISCCP-NG L1g format https://cimss.ssec.wisc.edu/isccp-ng/. Nominal calibration us used for SEVIRI. It might be good to recalibrate SEVIRI. But I can not find information in the file on the coefficients used. So these would need to be supplied separately.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented


mraspaud (commented):
Looks good I think. One question inline, but also:
Looks like there is no area definition generated by this reader. Would it be possible to have one? in general, areas are easier and faster than lon/lat-based swath definitions…

PR #3008, opened 2024-12-05
Add notes about platform name and sensor to dev guide

BENR0:
- Add guideline for readers dev to add `platform_name` and `sensor` attributes according to https://space.oscar.wmo.int/ to documentation
- fix typo


doc/source/dev_guide/custom_reader.rst
```diff
@@ -485,7 +485,9 @@ needs to implement a few methods:
 
    This method has to return an xarray.DataArray instance if the loading is
    successful, containing the data and :ref:`metadata <dataset_metadata>` of the
-   loaded dataset, or return None if the loading was unsuccessful.
+   loaded dataset, or return None if the loading was unsuccessful. If the reader
+   is reading satellite data the returned xarray.DataArrays should also have the
+   attributes ``platform_name`` and ``sensor`` with names according to https://space.oscar.wmo.int/.
```
djhoese, line 490: I'm not sure this is true. Sure we try to match OSCAR but they've also changed names in their system in the past. Also we typically have sensor being lowercase and it can be a single string or a `set` object. I think we've also had some special cases but don't recall which ones off the top of my head.

pnuu: For platform name I think that is pretty much ok, but for sensors we've used lowercase *and* have also replaced `/` with `-`. As an example `AVHRR/3` is `avhrr-3`.

BENR0: While I personally would like it to be like this, I did this PR to get this discussion started.

@pnuu replacing the `/` with `-` is good which is what space oscar uses. And uppercase/lowercase does not matter as well `-` or `_`. So at least for the URL it does not matter if you have `AVHRR-3`, `avhrr-3`, `AVHRR_3` or `avhrr_3`

pnuu (approved):

PR #3011, opened 2024-12-07
Drop dependency on xarray-datatree

avalentino:
DataTree is now part of the xarray since v2024.10.0. An explicit dependency on the externap package
xarray-datatree is no lionger needed.

 - [ ] Tests added 


djhoese (approved):
Looks good thanks. Am I seeing this correctly on mobile that you removed yourself from the authors list?

djhoese:
Ah I see. Duplicate.

avalentino:
Thanks @djhoese

PR #3017, opened 2024-12-13
Fix SLSTR angle interpolation.

### LINKED ISSUE #3015: extrapolation of satellite relative azimuiths?
**Describe the bug**
The  SLSTR L1B RBT product geometry files (viewing angles) are in tie point grids. When reading from satpy (which should ideally give interpolated values for an asked grid), the values seem to be getting extrapolated (values more than 360 degrees).

```python
scn = Scene(filenames=glob(os.path.join(swath_dir, "*.nc")), reader="slstr_l1b")    
dq = DataQuery(name='satellite_azimuth_angle', stripe='a', view='nadir', resolution=500)
var_ret = load_data(dq, scn)
np.nanmax(var_ret.data)
```

**Expected behavior**
values should always in between 0 to 360 degrees for sun-earth geometry

**Actual results**
out: 428.0610558123558

**Screenshots**
-

**Environment Info:**
 - OS: [ Linux]
 - Satpy Version: Version: 0.53.0
 - PyResample Version: Version: 1.31.0

**Additional context**
- check  _sat_azimuth_tn_ in _geometry_tn.nc_
if interested, file in https://drive.google.com/file/d/1CQFgiUtfOXw6KzcOL3zUnu4pU4mSCXt-/view?usp=sharing

simonrp84:
This PR adds a new method for interpolating angles from the SLSTR tie grid, which should prevent the unrealistically low and/or high values that sometimes showed up before, primarily in granules near the poles.

 - [x] Closes #3015
 - [x] Fully documented


mraspaud (commented):
Interesting method, thanks for implementing it! How about adding a test or two for it?

mraspaud (changes requested):
Could you add a test for this?

simonrp84:
@mraspaud Test added!

PR #3028, opened 2025-01-09
add missing dependency for vii_l1b_reader

BengtRydberg:
This PR adds a missing dependency for the vii_l1b_reader.

I ran into this issue when I was trying to use the reader.


mraspaud (approved):
LGTM

PR #3030, opened 2025-01-13
Fix sdist tarball including unnecessary files

djhoese:
As discussed on slack, I discovered that hatch/hatchling does not use the MANIFEST.in template file for what to include in an sdist. As such, it seems we have been including all files in the git repository when run in CI. Fortunately since this is from a fresh clone this was mostly only including a few extra files plus all the documentation source files, benchmark files, etc. Hatchling *always* includes READMEs and LICENSE files.

In this PR I've removed the unused MANIFEST.in file and make a custom section in pyproject.toml to explicitly say what files to include. Normally this would be `packages = ["satpy"]` but to preserve what the manifest file had I also included authors, changelog, security, and citation. ~The one major difference I've configured here is to *NOT* include the documentation. I can be convinced otherwise, but I just don't think they are necessary.~

The wheel file is unchanged and only includes the python package source.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


mraspaud (approved):
LGTM

PR #3041, opened 2025-01-22
Fix MVIRI tests with unstable dependencies

sfinkens:
Convert scanline timestamps to nanoseconds before averaging them. Looks like a bug in xarray (https://github.com/pydata/xarray/issues/9975), but converting to nanosecond doesn't hurt


mraspaud (approved):
LGTM

PR #3048, opened 2025-02-03
Change sensor attribute to lowercase in FCI L2 NetCDF reader

strandgren:
Following a discussion on slack (https://pytroll.slack.com/archives/CK9TPLCG0/p1737127470400239 and https://pytroll.slack.com/archives/C0LNH7LMB/p1737561433643359), this PR changes the `sensor` name dataset attribute in the FCI L2 NetCDF reader (`fci_l2_nc`) from uppercase to lowercase. This is needed in order to identify instrument specific composites and enhancements. I also modified the tests to mimic the real files where there sensor and platform information are stored in uppercase in the file metadata, such that we can also check that the `sensor` attribute is still returned with lowercase.

As discussed on slack, it could be worthwhile updating the documentation as well. However, I'm a bit confused and therefore not sure what and how to update. In the documentation for adding custom readers (https://satpy.readthedocs.io/en/v0.53.0/dev_guide/custom_reader.html) it says that the sensor(s) listed in the yaml file, "must be all lowercase letters for full support throughout in Satpy". However, this was already the case before the PR:

```
reader:
  name: fci_l2_nc
  short_name: FCI L2 NetCDF4
  long_name: MTG FCI L2 data in netCDF4 format
  description: Reader for EUMETSAT MTG FCI L2 files in NetCDF4 format.
  status: Alpha
  supports_fsspec: false
  sensors: [fci]
  reader: !!python/name:satpy.readers.yaml_reader.GEOFlippableFileYAMLReader
```

Hence, it seems like this is not enough and that the `sensor` attribute of the dataset also has to be in lowercase, which is not mentioned here: https://satpy.readthedocs.io/en/stable/reading.html#dataset-metadata (`sensor` is not even listed as mandatory metadata/attribute). Hence, my question is whether there is any link between the sensor(s) listed in the yaml file and the sensor attribute added as metadata to a dataset? To me it seems like it's not the case, and if so we might want to add a note under the dataset metadata (https://satpy.readthedocs.io/en/stable/reading.html#dataset-metadata), similar to what we have for the reader yaml-file, e.g. something like this:

"`sensor` - name of the sensor. For full support throughout in Satpy this must be all lowercase letters."

 - [X] Tests adapted
 - [ ] Fully documented

### DIFF
--- a/doc/source/reading.rst
+++ b/doc/source/reading.rst
@@ -300,6 +300,9 @@ time etc. The following attributes are standardized across all readers:
   :class:`~pyresample.geometry.SwathDefinition` if data is geolocated. Areas are used for gridded
   projected data and Swaths when data must be described by individual longitude/latitude
   coordinates. See the Coordinates section below.
+* ``sensor``: The name of the sensor that recorded the data. For full support through Satpy this
+  should be all lowercase. If the dataset is the result of observations from multiple sensors a
+  ``set`` object can be used to specify more than one sensor name.
 * ``reader``: The name of the Satpy reader that produced the dataset.
 * ``orbital_parameters``: Dictionary of orbital parameters describing the satellite's position.
   See the :ref:`orbital_parameters` section below for more information.
--- a/satpy/readers/fci_l2_nc.py
+++ b/satpy/readers/fci_l2_nc.py
@@ -48,7 +48,7 @@ class FciL2CommonFunctions(object):
     @property
     def sensor_name(self):
         """Return instrument name."""
-        return self.nc.attrs["data_source"]
+        return self.nc.attrs["data_source"].lower()
 
     @property
     def ssp_lon(self):
--- a/satpy/tests/reader_tests/test_fci_l2_nc.py
+++ b/satpy/tests/reader_tests/test_fci_l2_nc.py
@@ -68,8 +68,8 @@ class TestFciL2NCFileHandler(unittest.TestCase):
             nc.createDimension("maximum_number_of_layers", 2)
 
             # add global attributes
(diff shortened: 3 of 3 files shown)


djhoese:
My guess is that the documentation about the YAML `sensor` was written when it was assumed that file handlers and the base reader was using that YAML sensor. As more readers were developed I think this quickly became not true as most could get the sensor information from the file content.

I think your suggestion of adding the sensor information to the metadata section would be good. I think we should leave the sensor information about the YAML as that is still sometimes used I think with `find_files_and_readers`. Your suggestion was:

> sensor - name of the sensor. For full support throughout in Satpy this must be all lowercase letters.

Small grammar fix: remove the "in" before Satpy.

I wonder if we should add something about needing it to be consistent across readers? Also there is support for multiple sensors as a `set` object. Maybe:

> sensor - name of the sensor that recorded the data. For full support through Satpy this should be all lowercase. If the data is the result of observations from multiple sensors a `set` object can be used to specify more than one sensor name.

strandgren:
> I wonder if we should add something about needing it to be consistent across readers? Also there is support for multiple sensors as a `set` object. Maybe:
> 
> > sensor - name of the sensor that recorded the data. For full support through Satpy this should be all lowercase. If the data is the result of observations from multiple sensors a `set` object can be used to specify more than one sensor name.

I like your suggestion, I've added it to the docs now. The docs build fails for the PR though, however it also failed before I edited the docs file, so I think it's something unrelated?

djhoese (approved):

djhoese:
Thanks! The RTD failure was the one that should be fixed in main so I've merged this.

Issue #3053, opened 2025-02-09
No conversion matrix found for platform GOES-19


ustropics:
This issue is for the new GOES-19 satellite that is now having data processed on AWS bucket (sample here - https://noaa-goes19.s3.amazonaws.com/index.html#ABI-L1b-RadC/2025/039/00/). The available composites function after downloading bands 1-3 states true_color_reproduction is available as a composite. However, I run into the following error:

> [983](file:///C:/Users/ustro/.conda/envs/GOES_main/lib/site-packages/satpy/scene.py:983) # regenerate anything from the wishlist that needs it (combining
>     [984](file:///C:/Users/ustro/.conda/envs/GOES_main/lib/site-packages/satpy/scene.py:984) # multiple resolutions, etc.)
>     [985](file:///C:/Users/ustro/.conda/envs/GOES_main/lib/site-packages/satpy/scene.py:985) if generate:
> --> [986](file:///C:/Users/ustro/.conda/envs/GOES_main/lib/site-packages/satpy/scene.py:986)     new_scn.generate_possible_composites(unload)
>     [988](file:///C:/Users/ustro/.conda/envs/GOES_main/lib/site-packages/satpy/scene.py:988) return new_scn
> ...
> --> [653](file:///C:/Users/ustro/.conda/envs/GOES_main/lib/site-packages/satpy/enhancements/__init__.py:653)     raise KeyError(f"No conversion matrix found for platform {platform}")
>     [655](file:///C:/Users/ustro/.conda/envs/GOES_main/lib/site-packages/satpy/enhancements/__init__.py:655) output = da.dot(img_data.T, ccm.T)
>     [656](file:///C:/Users/ustro/.conda/envs/GOES_main/lib/site-packages/satpy/enhancements/__init__.py:656) return output.T
> 
> KeyError: 'No conversion matrix found for platform GOES-19'

**To Reproduce**

```python
from satpy import Scene
from glob import glob

dat_files = glob('data/*.nc')

scn = Scene(reader='abi_l1b', filenames=dat_files)

recipe = 'true_color_reproduction'

scn.load([recipe])
new_scn = scn.resample(scn.min_area(), resampler='native')
new_scn[recipe]

```
I believe this may be related to GOES-19 current position/not officially operational (it will soon replace GOES-16). Does this impact the matrix lookup needed for corrections (like rayleigh and sunz)?

pnuu:
Hi!

[...]

PR #3054, opened 2025-02-09
Add GOES-19 RGB2XYZ matrix for `true_color_reproduction`

simonrp84:
Adds theRGB-to-XYZ conversion matrix required by the `true_color_reproduction` composite for the GOES-19 spacecraft. This satellite was previously missing. Note: Matrix is not optimal as GOES/ABI misses a green band, and hence here we use a hybrid matrix that *assumes* the value of the green band computes for GOES/ABI.

 - [ ] Closes #xxxx


simonrp84:
NOTE: I haven't had a chance to try this on actual G-19 data, but as the values are close to those for the other GOES satellites I expect it to work fine. I suggest not merging until we get feedback either from the user in #3053 or from someone else who gives this a go with the real data.

gerritholl:
Could you post before/after images?

simonrp84:
I slightly updated the matrix due to an error in my code. New one committed now!

It made very little difference to the images compared to those shown in my (now deleted) previous comment. Top, `true_color`. Bottom, `true_color_reproduction`.
tc

tcr

mraspaud (approved):
Looks good to me!

PR #3057, opened 2025-02-11
Fix LightningTimeCompositor failing when data outside of the time range is passed

ameraner:
This PR fixes the LightningTimeCompositor (implemented in  https://github.com/pytroll/satpy/pull/2895) that was causing failures when the input data contained lightning times outside of the time range.

The compositor was using `drop=True` when filtering lightnings outside the defined time_range. This was causing a mismatch in dimension between the lightning data array and the latitude/longitudes arrays due to the dropped entries, so that resampling operations were failing due to shape/chunks mismatch.

By removing  `drop=True` also the explicit computation of the condition array could be removed.

On a side, this PR also reduces the handling of empty data from an error to a warning, and fixes some language/typos in the docstrings.

 - [x] Tests added

### DIFF
--- a/satpy/composites/lightning.py
+++ b/satpy/composites/lightning.py
@@ -28,11 +28,11 @@ LOG = logging.getLogger(__name__)
 
 
 class LightningTimeCompositor(CompositeBase):
-      """Class used to create the flash_age compositor usefull for lighting event visualisation.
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM

PR #3059, opened 2025-02-12
Add more datasets to IASI L2 reader

pnuu:
This PR adds datasets from `/INFO/` and `/Maps/` groups to the IASI L2 reader.

Also the IASI L2 tests were converted to pytest.

 - [x] Tests added

### DIFF
--- a/satpy/etc/readers/iasi_l2.yaml
+++ b/satpy/etc/readers/iasi_l2.yaml
@@ -156,6 +156,49 @@ datasets:
     resolution: 12000
     coordinates: [longitude, latitude]
 
+  amsu_instrument_flags:
+    name: amsu_instrument_flags
+    file_type: iasi_l2_hdf5
+    units: "1"
+    resolution: 12000
+    coordinates: [longitude, latitude]
+
+  iasi_instrument_flags:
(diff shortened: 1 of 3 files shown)


pnuu:
Seems like `bilinear` resampling doesn't work with the current `("y", "x", "level")` dimension order. If I change this to `("bands", "y", "x")`. The `nearest` works also with the current ordering and naming, so the problem is within the `bilinear` resampler.

djhoese (approved):
My one small request that is not necessary would be rewriting most if not all of the test setup as a pytest fixture. You could have one that just produces the filename (`Path` object?) and is module scoped. Then have one that produces a Scene and one that produces a file handler (replaces `self.reader`) that are test scoped. That way any caching inside the Scene or file handler are not preserved between tests, but the test file is only created once.

mraspaud (approved):
LGTM

pnuu:
Ok, found the `scope="module"` fix from the other PR, a added that in https://github.com/pytroll/satpy/pull/3059/commits/4053955ed2aa7235f4463857468e66d12c777685

mraspaud (approved):
LGTM

PR #3069, opened 2025-02-20
Prevent call to obj.__dict__ when obj is a compound variable.

### LINKED ISSUE #3066: NetCDF4FileHandler faults when using obj.__dict__ called when object is a compound variable.
**Describe the bug**

I have written a Sentinel4/5/5P reader (UVNS.py) for my organisation.  It uses the NetCDF4FileHandler to interrogate the file for variables rather than specify them in the yaml file.  Sentinel 5P Tropomi L1B netcdf files have some compound variables (with compound attributes) - but I believe the error will happen for any netCDF4 file with compound variables.

I am aware that satpy does not really handle compound variables and consensus on compound attributes seems far away, however, if I hack satpy I can load the ordinary Dataset variables and, with a little effort, the compound variables (but not the compound variable attributes).  As satpy stands, though, I cannot even load the Tropomi Dataset variables as it aborts.  This is also important for other members of the community who will want to load these data and have access to reliable readers.

My uvns_reader inherits NetCDF4FileHandler.
On initialisation, without a list of required variables, the NetCDF4FileHandler calls self.collect_metadata("", file_handle), which eventually calls self._get_object_attrs(obj).  Validity of obj is not tested and, in this method, there is a call to obj__dict__  which causes satpy to abort when obj is a compound variable (i.e. it tries to extract a dict from the structure space).

On running, it initially it prints out a warning:
_.../satpy/readers/netcdf_utils.py:126: UserWarning: WARNING: unsupported Compound type, skipping...
  return netCDF4.Dataset(self.filename, "r")
but continues but is followed by some KeyErrors, then it aborts with: 'KeyError: "attribute b'units' has unsupported datatype"' (in this case with Tropomi data).

**To Reproduce**

```python
I use it as:
filename='S5P_OFFL_L1B_RA_BD1_20191114T092139_20191114T110308_10811_01_010000_20191114T124851.nc'
filenames = {'uvns':[filename]}
scn = Scene(filenames=filenames)

but can also be produced by:
from satpy.readers.netcdf_utils import NetCDF4FileHandler
fh = NetCDF4FileHandler(filename, {}, {})
```

**Expected behavior**
I would like _get_object_attrs to do something like:

```python
def _get_obj_attrs(self, obj):
     if obj.dtype.kind != 'V':
            return obj.__dict__
     return {}
```

**Actual results**

```python
fh = NetCDF4FileHandler(filename, {}, {})
/tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:126: UserWarning: WARNING: unsupported Compound type, skipping...
  return netCDF4.Dataset(self.filename, "r")
KeyError                                  Traceback (most recent call last)
File src/netCDF4/_netCDF4.pyx:1586, in netCDF4._netCDF4._get_att()

KeyError: 34

KeyError                                  Traceback (most recent call last)
File src/netCDF4/_netCDF4.pyx:6350, in netCDF4._netCDF4._read_compound()

KeyError: 12

KeyError                                  Traceback (most recent call last)
File src/netCDF4/_netCDF4.pyx:1591, in netCDF4._netCDF4._get_att()

File src/netCDF4/_netCDF4.pyx:6361, in netCDF4._netCDF4._read_compound()

KeyError: 'compound field of an unsupported data type'

RuntimeError                              Traceback (most recent call last)
File src/netCDF4/_netCDF4.pyx:1596, in netCDF4._netCDF4._get_att()

File src/netCDF4/_netCDF4.pyx:6588, in netCDF4._netCDF4._read_enum()

File src/netCDF4/_netCDF4.pyx:2034, in netCDF4._netCDF4._ensure_nc_success()

RuntimeError: NetCDF: Not a valid data type or _FillValue type mismatch

KeyError                                  Traceback (most recent call last)
Cell In[24], line 1
----> 1 fh = NetCDF4FileHandler(filename, {}, {})

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:116, in NetCDF4FileHandler.__init__(self, filename, filename_info, filetype_info, auto_maskandscale, xarray_kwargs, cache_var_size, cache_handle)
    114     self._collect_listed_variables(file_handle, listed_variables)
    115 else:
--> 116     self.collect_metadata("", file_handle)
    117     self.collect_dimensions("", file_handle)
    118 self.collect_cache_vars(cache_var_size)

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:145, in NetCDF4FileHandler.collect_metadata(self, name, obj)
    143 # Look through each subgroup
    144 base_name = name + "/" if name else ""
--> 145 self._collect_groups_info(base_name, obj)
    146 self._collect_variables_info(base_name, obj)
    147 if not name:

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:157, in NetCDF4FileHandler._collect_groups_info(self, base_name, obj)
    155 self.file_content[full_group_name] = group_obj
    156 self._collect_attrs(full_group_name, group_obj)
--> 157 self.collect_metadata(full_group_name, group_obj)

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:145, in NetCDF4FileHandler.collect_metadata(self, name, obj)
    143 # Look through each subgroup
    144 base_name = name + "/" if name else ""
--> 145 self._collect_groups_info(base_name, obj)
    146 self._collect_variables_info(base_name, obj)
    147 if not name:

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:157, in NetCDF4FileHandler._collect_groups_info(self, base_name, obj)
    155 self.file_content[full_group_name] = group_obj
    156 self._collect_attrs(full_group_name, group_obj)
--> 157 self.collect_metadata(full_group_name, group_obj)

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:145, in NetCDF4FileHandler.collect_metadata(self, name, obj)
    143 # Look through each subgroup
    144 base_name = name + "/" if name else ""
--> 145 self._collect_groups_info(base_name, obj)
    146 self._collect_variables_info(base_name, obj)
    147 if not name:

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:157, in NetCDF4FileHandler._collect_groups_info(self, base_name, obj)
    155 self.file_content[full_group_name] = group_obj
    156 self._collect_attrs(full_group_name, group_obj)
--> 157 self.collect_metadata(full_group_name, group_obj)

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:146, in NetCDF4FileHandler.collect_metadata(self, name, obj)
    144 base_name = name + "/" if name else ""
    145 self._collect_groups_info(base_name, obj)
--> 146 self._collect_variables_info(base_name, obj)
    147 if not name:
    148     self._collect_global_attrs(obj)

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:162, in NetCDF4FileHandler._collect_variables_info(self, base_name, obj)
    160 for var_name, var_obj in obj.variables.items():
    161     var_name = base_name + var_name
--> 162     self._collect_variable_info(var_name, var_obj)

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:169, in NetCDF4FileHandler._collect_variable_info(self, var_name, var_obj)
    167 self.file_content[var_name + "/shape"] = var_obj.shape
    168 self.file_content[var_name + "/dimensions"] = var_obj.dimensions
--> 169 self._collect_attrs(var_name, var_obj)

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:235, in NetCDF4FileHandler._collect_attrs(self, name, obj)
    233 def _collect_attrs(self, name, obj):
    234     """Collect all the attributes for the provided file object."""
--> 235     for key in self._get_object_attrs(obj):
    236         fc_key = f"{name}/attr/{key}"
    237         value = self._get_attr_value(obj, key)

File /tcenas/proj/acteam/individual/malcolmt/GIT/satpy-dev/satpy.dev/satpy/readers/netcdf_utils.py:215, in NetCDF4FileHandler._get_object_attrs(cls, obj)
    214 def _get_object_attrs(cls, obj):
--> 215     return obj.__dict__

File src/netCDF4/_netCDF4.pyx:4915, in netCDF4._netCDF4.Variable.__getattr__()

File src/netCDF4/_netCDF4.pyx:1599, in netCDF4._netCDF4._get_att()

KeyError: "attribute b'units' has unsupported datatype"

[DEBUG: 2025-02-19 16:22:41 : asyncio] Using selector: EpollSelector
```

**Screenshots**

**Environment Info:**
 - OS: Linux
 - Satpy Version: 0.1.dev12704+ge338294
 - PyResample Version:

**Additional context**

mataeui:
- [ ] Closes #3066  -->
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -211,8 +211,8 @@ class NetCDF4FileHandler(BaseFileHandler):
             self.file_content[fc_key] = global_attrs[key] = value
         self.file_content["/attrs"] = global_attrs
 
-    @classmethod
-    def _get_object_attrs(cls, obj):
+    @staticmethod
+    def _get_object_attrs(obj):
         """Get object attributes using __dict__ but retrieve recoverable attributes on failure."""
         try:
             return obj.__dict__
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -224,7 +224,6 @@ class NetCDF4FileHandler(BaseFileHandler):
                     atts[attname] = obj.getncattr(attname)
                 except KeyError:
                     LOG.warning(f"Warning: Cannot load object ({obj.name}) attribute ({attname}).")
-                    pass
             return atts
 
     def _collect_attrs(self, name, obj):
--- a/satpy/readers/netcdf_utils.py
+++ b/satpy/readers/netcdf_utils.py
@@ -211,8 +211,21 @@ class NetCDF4FileHandler(BaseFileHandler):
             self.file_content[fc_key] = global_attrs[key] = value
         self.file_content["/attrs"] = global_attrs
 
-    def _get_object_attrs(self, obj):
-        return obj.__dict__
+    @classmethod
+    def _get_object_attrs(cls, obj):
+        """Get object attributes using __dict__ but retrieve recoverable attributes on failure."""
+        try:
+            return obj.__dict__
+        except KeyError:
+            # Maybe unrecognised datatype.
+            atts = {}
+            for attname in obj.ncattrs():
+                try:
+                    atts[attname] = obj.getncattr(attname)
+                except KeyError:
+                    LOG.warning(f"Warning: Cannot load object ({obj.name}) attribute ({attname}).")
+                    pass
+            return atts
 
     def _collect_attrs(self, name, obj):
         """Collect all the attributes for the provided file object."""


satpy/readers/netcdf_utils.py
```diff
@@ -211,9 +211,20 @@ def _collect_global_attrs(self, obj):
...
-    def _get_object_attrs(self, obj):
+    @classmethod
+    def _get_object_attrs(cls, obj):
+        '''
+        Do not try to use __dict__ if obj is a compound variable.
+        Return empty dict if it is, otherwise return the object's attributes in a dict.
+        '''
+        try:
+            if obj.dtype.kind == 'V':
+                print(f'Warning: Cannot load attributes for compound variable ({obj.name})')
```
djhoese, line 222: We would normally log a warning with the `LOG.warning` or using `warnings.warn`. I'm not sure which one is best here.

satpy/readers/netcdf_utils.py
```diff
@@ -211,9 +211,20 @@ def _collect_global_attrs(self, obj):
...
+                return {}
+        except (AttributeError, KeyError):
```
djhoese, line 224: In what cases do we hit these exceptions? When does `obj` not have a `dtype` or `dtype` not have a `kind`...or is `obj.name` the problem?

djhoese: Additionally you could probably replace the `except` and `pass` with a `else: obj.__dict__`. But actually, if the exception isn't possible in the real world then maybe removing the try/except altogether would be good.

djhoese (commented):
Thanks for starting this fix. I have a few comments below.

satpy/readers/netcdf_utils.py
mataeui: When obj is a group there is no dtype == KeyError
When obj is variable, and variable is vlen, then dtype = str == AttributeError string has no attribute kind.

satpy/readers/netcdf_utils.py
mataeui: O.K.  I've added LOG.warning to my code.
Now my problem is that I don't know how to update the code (this pull request) without starting over.

djhoese: Make your commit and `git push`. After this is merged you'll likely want to reset your fork's `main` branch to match `pytroll/satpy`'s `main` branch. Otherwise future pull requests may be hard to get right. After this is merged pytroll's main will have a commit for this PR being merged but your main will not and they will be out of sync.

mataeui:
I assume that:

[docs/readthedocs.org:satpy](https://readthedocs.org/projects/satpy/builds/27260610/) — Read the Docs build failed!

Is nothing to do with me.

djhoese:
Correct. It's a bug/memory leak in the new version of sphinx that I'm working on fixing.

satpy/readers/netcdf_utils.py
djhoese: `.dtype` is a KeyError? Strange, but ok. I was just wondering, would it be more flexible to do the inverse of these conditions and try first to return `obj.__dict__` and if there's an attribute error then warn and return if compound variable otherwise raise the original exception? That way, if it is possible that netcdf4-python adds the ability to read attributes in the future then things just work. Or is that too much wishful thinking on my part?

mataeui: O.K. on looking deeper into the problem with Tropomi, the problem is that the attributes 'units' and 'comment' are compound types of strings whereas '_FillValue' is an appropriate compound type of appropriate numerical value.

I can load the _FillValue attribute (getncattr) and get a tuple consistent with the number of elements in the compound data.  I can load the data (variable[:]).  however, with the 'units' and 'comment' attributes, I get e.g. 'KeyError: "attribute b'comment' has unsupported datatype".  

I have tried to search online (see github.com/Unidata/netcdf4-python/issues/773) and it appears that 'it is not possible to create compound types with variable length arrays' (2018) (numpy issue?).  I cannot find anything more recent.  So, reading of strings in compound types, as such, seems not to be supported in netCDF4.py, however, ncdump (see below) and panoply (panoply without data loading however) seem to manage - don't understand how.

So, for now,
1. i) the simplest would be to replace obj.__dict__ in _get_object_attrs with a call to getncattr from loop over ncattrs with a try/except KeyError to return the loadable attributes and drop the others.
   ii) put a try/except on __dict__ as you wanted, and then implement (i) on failure.
2. otherwise solve the string issue somehow.

This would then parse Tropomi and I would be happy - but it doesn't address any deeper issues that satpy might have with compound data.

From ncdump:
        compound msmt_to_det_row_table_type {
          short det_start_row ;
          short det_end_row ;
        }; // msmt_to_det_row_table_type

        compound msmt_to_det_row_table_type_str {
          string det_start_row ;
          string det_end_row ;
        }; // msmt_to_det_row_table_type_str

        msmt_to_det_row_table_type measurement_to_detector_row_table(time, scanline, ground_pixel) ;
        msmt_to_det_row_table_type_str measurement_to_detector_row_table:units = {"1", "1"} ;
        msmt_to_det_row_table_type measurement_to_detector_row_table:_FillValue = {-32767, -32767} ;
        msmt_to_det_row_table_type_str measurement_to_detector_row_table:comment = {"Detector start row for measurement row", "Detector end row for measurement row"} ;

djhoese: So overall I think I like option 1(ii) best as far as getting most support. **BUT** as you point out, the rest of satpy isn't going to like having multiple values for things it assumes should be scalars (`_FillValue` is typically deleted, `units` must be a string, etc).

Panoply is written in Java, ncdump is written in C, so whatever those versions of the NetCDF library are using to store variable length data must allow for these variable length strings. That makes sense given the lower-level nature of those languages.

I'm not sure I've asked this specifically, but what is/are the compound types in any of the files you've seen (ex. tropomi)? I mean, why did they decide a compound type was needed and could we (satpy's readers) split those variables into individual variables?

Any idea if the netcdf4-python library makes it possible to extract one of the "units" elements at a time? Like `my_nc_file["my_variable"].getncattr("units", index=0)`?

mataeui: Of the current satellite sensors that I have come across, Tropomi L1B data are the first I have seen with compound data types (recently).  In fact compound datatypes are not listed in the cf convention's datatypes and, in fact, files which include these do not pass the cf-checker afaik.  I have come across compound types in point-source station data iirc.

When KNMI decided to do this they 'expected that many software products will be upgraded in time to support the features of the enhanced data model.' (Input-output-data-specification-for-the-TROPOMI-L01b-data-processor.pdf, 2014), basically the compound data in Tropomi is just metadata information on the processing/instrument status.

Splitting into individual variables is the simplest and most obvious way forward without interlacing code to manage compound data into satpy - but the consequence would be a lot of single point variables, possibly with duplicate naming conflicts.

 As far as I can see, getncattr, getattr, __getatrr__, __getattribute>__  are, currently, relatively primitive and don't appear to have options.

djhoese: > but the consequence would be a lot of single point variables, possibly with duplicate naming conflicts.

What do you mean by this? My understanding of what you said, coming from a place of mostly Visible/IR satellite instrument understanding, is that tropomi has a compound variable. One part of each element is the actual measurement, the rest of the information in a single element is mostly metadata that (hopefully) Satpy's basic usage doesn't need. If we can load the data and some/most of the attributes, then could we pull out just one portion of the compound type that has the observation measurement and throw away the rest to produce the single 2D "image" array that Satpy expects?

mataeui: Yes, you are correct, sorry I was being distracted by other thoughts.

mataeui: Can we advance this as a temporary fix (there are people on my side waiting for this fix ...) - or are you wanting the more complete solution.

satpy/readers/netcdf_utils.py
```diff
@@ -211,8 +211,21 @@ def _collect_global_attrs(self, obj):
...
+            return obj.__dict__
+        except KeyError:
+            # Maybe unrecognised datatype.
+            atts = {}
+            for attname in obj.ncattrs():
+                try:
+                    atts[attname] = obj.getncattr(attname)
+                except KeyError:
+                    LOG.warning(f"Warning: Cannot load object ({obj.name}) attribute ({attname}).")
+                    pass
```
djhoese, line 227:
```suggestion
                    LOG.warning(f"Warning: Cannot load object ({obj.name}) attribute ({attname}).")
```

djhoese:
Thanks @mataeui!

mataeui:
Thanks for the updates.  What happens now?

PR #3068, opened 2025-02-19
Update sphinx to use new sphinx.ext.apidoc extension

djhoese:
Updates sphinx to 8.2 which adds a sphinx.ext.apidoc extension and replaces sphinxcontrib.apidoc. I also turned off documenting most test modules which cuts off ~30 seconds (50%) of execution time and ~2GB of memory usage (80%).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 
 - [ ] Add your name to `AUTHORS.md` if not there already

PR #3071, opened 2025-02-21
Fix deprecated codecov action parameter use and limit sphinx in documentation rendering

djhoese:
Replacement for #3068. In that PR I started updating sphinx usage to 8.2.0 and then realized there is a memory leak in 8.2.0 that causes RTD builds to fail. In this PR I keep the CI warning fix, but limit sphinx so 8.2.0 isn't used.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -139,7 +139,7 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           flags: behaviourtests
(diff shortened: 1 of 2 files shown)


djhoese:
I'm going to close this. It shouldn't be needed now that sphinx was patched. #3068 should be the PR to be merged.

PR #3082, opened 2025-03-18
Add workaround so dask arrays are optimized in Delayed writing

djhoese:
This speeds up writers like "simple_image" significantly.

This is the simplest workaround I could think of that works for all cases. This addresses the problem described in my dask discourse discussion post [here](https://dask.discourse.group/t/optimal-graph-optimization-when-mixing-dask-objects/3867).

Essentially we're saying "I know this is a Delayed object, but look at the task graph as a series of Array operations, not basic Delayed tasks". This means more complex graph optimizations are performed by dask rather than only the simple culling style optimization done for Delayed objects.

Note: This does not fix the rare issue caused by `da.store` pre-optimizing for Arrays and satpy users combining multiple writing (`.save_datasets`) calls into a single call using `compute_writer_results`. In some cases like this some Array tasks are re-computed because the pre-optimized tasks were renamed/merged/fused separately and dask thinks they are separate unique tasks. This comes down to https://github.com/dask/dask/issues/8380 and https://github.com/dask/dask/pull/9732 and my discussion post linked above and that they are not resolved upstream (yet :crossed_fingers: ).

As noted on slack, for a simple singe MODIS band to PNG this sped things up from 54s to 39s with only a slight but expected increase in memory as more tasks are computed in parallel and in bigger groups of operations...or at least I think that's why.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/writers/__init__.py
+++ b/satpy/writers/__init__.py
@@ -23,6 +23,7 @@ import os
 import warnings
 from typing import Optional
 
+import dask
 import dask.array as da
 import numpy as np
 import xarray as xr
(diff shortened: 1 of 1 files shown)


simonrp84:
For my own understanding, you say memory use has increased slightly due to more parallel tasks - does that mean that a user could, if needed, reduce memory use again by limiting the number of parallel tasks?

djhoese:
Yes, but that would be done in the traditional way of limiting number oof dask workers or reducing chunk size. This PR is just making it more likely that more mini-tasks (a numpy function call) are happening during one merged task and using more memory. That's my best guess at least.

mraspaud (approved):
LGTM, makes sense

PR #3084, opened 2025-03-18
Updated the grid size to read claas3 data

jequierz:
I have updated the cmsaf_claas2 reader so that it handles now the larger grid of the claas3 (changed from 3636x3636 to 3712x3712). 

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/cmsaf_claas2.py
+++ b/satpy/readers/cmsaf_claas2.py
@@ -29,12 +29,12 @@ FULL_DISK_WITH_OFFSET = _adjust_area_to_match_shifted_data(FULL_DISK)
 class CLAAS2(NetCDF4FileHandler):
     """Handle CMSAF CLAAS-2 files."""
 
+    grid_size = 3636
 
     def __init__(self, *args, **kwargs):
         """Initialise class."""
         super().__init__(*args, **kwargs, cache_handle=False,
                          auto_maskandscale=True)
-        self.grid_size = self.file_content["y/shape"][0]
 
     @property
     def start_time(self):
@@ -104,10 +104,16 @@ class CLAAS2(NetCDF4FileHandler):
         than the full disk (excludes most of the space pixels).
         """
         full_disk = self._get_full_disk()
-        if _is_georef_offset_present(self.start_time.date()):
-            offset = int((full_disk.width - self.grid_size) // 2)
-            return full_disk[offset:-offset, offset:-offset]
-        return full_disk
+        try:
+            print('been there')
+            grid_size =  self.file_content["y/shape"][0]
+        except KeyError:
+            print('done that')
+            grid_size = self.grid_size
+        offset = int((full_disk.width - grid_size) // 2)
+        if offset == 0:
+            return full_disk
+        return full_disk[offset:-offset, offset:-offset]
 
     def _get_full_disk(self):
         if _is_georef_offset_present(self.start_time.date()):
--- a/satpy/readers/cmsaf_claas2.py
+++ b/satpy/readers/cmsaf_claas2.py
@@ -33,7 +33,7 @@ class CLAAS2(NetCDF4FileHandler):
         """Initialise class."""
         super().__init__(*args, **kwargs, cache_handle=False,
                          auto_maskandscale=True)
-        self.grid_size =  self.file_content["/dimension/y"]
+        self.grid_size = self.file_content["/dimension/y"]
 
     @property
     def start_time(self):
--- a/satpy/readers/cmsaf_claas2.py
+++ b/satpy/readers/cmsaf_claas2.py
@@ -29,12 +29,11 @@ FULL_DISK_WITH_OFFSET = _adjust_area_to_match_shifted_data(FULL_DISK)
 class CLAAS2(NetCDF4FileHandler):
     """Handle CMSAF CLAAS-2 files."""
 
-    grid_size = 3636
-
     def __init__(self, *args, **kwargs):
         """Initialise class."""
         super().__init__(*args, **kwargs, cache_handle=False,
                          auto_maskandscale=True)
+        self.grid_size =  self.file_content["/dimension/y"]
 
     @property
     def start_time(self):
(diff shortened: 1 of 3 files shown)


sfinkens:
Tests fail, because the dummy file we use for testing doesn't have coordinates. That's actually a valuable feedback, because we can make the check a little simpler:

```
self.grid_size = self.file_content["/dimension/y"]
```

Then the tests will fail again, because the area definition from the (2x2) test file doesn't match the (3636x3636) expectation any more. Let me know when you are at this point, then we can continue together.

sfinkens:
Here's a proposal for an updated test that runs for two time stamps and two grid sizes

```python
class TestAreaDef:
    @pytest.fixture(params=[3636, 3712])
    def grid_size(self, request):
        return request.param

    @pytest.fixture
    def fake_file(self, fake_dataset, encoding, tmp_path, start_time, grid_size):
        """Write a fake dataset to file.

        Only add y dimension, so that the file is small but we can still check the
        real-life area definition.
        """
        filename = tmp_path / "CPPin20140101001500305SVMSG01MD.nc"
        ds = netCDF4.Dataset(filename, mode="w")
        ds.createDimension("y", grid_size)
        ds.setncattr("time_coverage_start", start_time.strftime("%Y%m%dT%H%MZ"))
        ds.close()
        return filename

    @pytest.fixture
    def file_handler(self, fake_file):
        """Return a CLAAS-2 file handler."""
        from satpy.readers.cmsaf_claas2 import CLAAS2
        return CLAAS2(fake_file, {}, {})

    @pytest.fixture
    def area_extent_exp(self, start_time, grid_size):
        """Get expected area extent."""
        if start_time < datetime.datetime(2017, 12, 6):
            extents = {
                3636: (-5454733.160460291, -5454733.160460292, 5454733.160460292, 5454733.160460291),
                3712: (-5568748.485046371, -5568748.485046371, 5568748.485046371, 5568748.485046371)
            }
        else:
            extents = {
                3636: (-5456233.362099582, -5453232.958821001, 5453232.958821001, 5456233.362099582),
                3712: (-5570248.686685662, -5567248.28340708, 5567248.28340708, 5570248.686685662)
            }
        return extents[grid_size]

    @pytest.fixture
    def area_exp(self, area_extent_exp, grid_size):
        """Get expected area definition."""
        proj_dict = {
            "a": 6378169.0,
            "b": 6356583.8,
            "h": 35785831.0,
            "lon_0": 0.0,
            "proj": "geos",
            "units": "m",
        }
        return AreaDefinition(
            area_id="msg_seviri_fes_3km",
            description="MSG SEVIRI Full Earth Scanning service area definition with 3 km resolution",
            proj_id="geos",
            projection=proj_dict,
            area_extent=area_extent_exp,
            width=grid_size,
            height=grid_size,
        )

    def test_get_area_def(self, file_handler, area_exp):
        """Test area definition."""
        area = file_handler.get_area_def(make_dataid(name="foo"))
        assert area == area_exp

```

satpy/tests/reader_tests/test_cmsaf_claas.py
```diff
@@ -183,15 +184,69 @@ def file_handler(self, fake_file):
...
+
+    def test_start_time(self, file_handler, start_time):
+        """Test start time property."""
+        assert file_handler.start_time == start_time
+
+    def test_end_time(self, file_handler):
+        """Test end time property."""
+        assert file_handler.end_time == datetime.datetime(2085, 8, 13, 13, 15)
+
+class TestAreaDef:
```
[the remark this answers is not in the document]
sfinkens: Ok, more docstring problems reported by ruff. Try installing pre-commit hooks in your working copy

```
pip install ruff pre-commit
pre-commit install
ruff check test_cmsaf_claas.py
```

pnuu:
pre-commit.ci autofix

pnuu (changes requested):
These updates seem to be fine, just needs the one docstring that I added a suggestion to. 

The tests are failing though: https://github.com/pytroll/satpy/actions/runs/15050403317/job/42303560714?pr=3084#step:9:201

We can have a debugging session tomorrow to see what's going on.

satpy/tests/reader_tests/test_cmsaf_claas.py
```diff
@@ -183,15 +184,70 @@ def file_handler(self, fake_file):
...
+    def fake_file(self, tmp_path, start_time, grid_size):
+        """Write a fake dataset to file.
+
+        Only add y dimension, so that the file is small but we can still check the
+        real-life area definition.
+        """
+        filename = tmp_path / "CPPin20140101001500305SVMSG01MD.nc"
+        ds = netCDF4.Dataset(filename, mode="w")
+        ds.createDimension("y", grid_size)
+        ds.setncattr("time_coverage_start", start_time.strftime("%Y%m%dT%H%MZ"))
```
sfinkens, line 224: I used the wrong time format here, this should fix the problem:

```suggestion
        ds.setncattr("time_coverage_start", start_time.strftime(""%Y-%m-%dT%H:%M:%SZ""))
```

(`fromisoformat` doesn't support all iso formats in Python-3.10; and this is the format in the CLAAS files)

lahtinep: Yeah! We found out that independently on our debugging session with @jequierz :grin:

pnuu: It was also weird that the test passed with Python 3.12 locally. I guess the NetCDF4 library has added some datetime handling to newer versions to make it work even with the "wrong" time string.

satpy/readers/cmsaf_claas2.py
```diff
@@ -100,12 +99,14 @@ def get_area_def(self, dataset_id):
...
         """Get subset of the full disk.
 
-        CLAAS products are provided on a grid that is slightly smaller
+        CLAAS2 products are provided on a grid that is slightly smaller
         than the full disk (excludes most of the space pixels).
         """
         full_disk = self._get_full_disk()
         offset = int((full_disk.width - self.grid_size) // 2)
-        return full_disk[offset:-offset, offset:-offset]
+        if offset>0:
```
pnuu, line 107:
```suggestion
        if offset > 0:
```

PR #3085, opened 2025-03-19
Refactor "compute_relative_azimuth" to be more flexible

djhoese:
Similar to #3081, while debugging something unrelated I got annoyed at how many tasks were created for a simple task and decided to wrap the code in a map_blocks call. One complication is that @simonrp84 wrote this function to work for users calling it alone and passing `DataArray` objects, but the only use in Satpy itself is passing dask arrays. I thought about adding a decorator to handle the xarray -> dask conversion and back, but decided the if statements inside the function weren't that bad.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/modifiers/angles.py
+++ b/satpy/modifiers/angles.py
@@ -350,6 +350,24 @@ def compute_relative_azimuth(
     Relative azimuth is 0 when sun and satellite are aligned on one side of a pixel (back scatter).
     Relative azimuth is 180 when sun and satellite are directly opposite each other (forward scatter).
     """
+    xarray_dims = getattr(sat_azi, "dims", None)
+    xarray_coords = getattr(sat_azi, "coords", None)
+    if xarray_dims is not None:
+        sat_azi = sat_azi.data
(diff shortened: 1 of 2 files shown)


djhoese:
Note @simonrp84, if you were using this with straight numpy arrays or DataArrays with numpy (instead of dask) arrays inside, then it might not work anymore. I'm not sure what dask `map_blocks` does when provided numpy arrays.

simonrp84:
Apologies for the basic question, but what does this actually achieve? I mean, I get that it reduces the number of dask tasks but does that have any impact from the user side, like faster processing or reduced memory use?

Anyway, can't test the code right now but the changes look OK to me.

djhoese:
Sorry, yes, that's why I lazily mentioned the MODIS PR. This does almost nothing for performance except give dask less tasks to manage. So yes, faster, but by maybe a second or two at most. This is mostly to ease debugging dask graphs and performance related to dask graphs.

PR #3086, opened 2025-03-21
Add reader for OCI onboard NASA's PACE mission.

simonrp84:
This PR adds a reader for Level-1b data from the OCI sensor aboard NASA's PACE mission.

OCI is a hyperspectral sensor, which presents some problems in terms of managing the datasets. Here I use a dynamic dataset approach to generate the datasets for each wavelength measured by OCI.

Todo: Few fixes with the angles and adding some tests.

 - [x] Tests added 
 - [x] Fully documented


mraspaud (approved):
This looks good to me. Anything you want to change before we merge?

simonrp84:
No! All good from my side.

PR #3087, opened 2025-03-25
Add ABI Blowing Snow RGB

djhoese:
Adds the NASA/CIRA "Blowing Snow" RGB to the ABI composites. I only had one winter-ish dataset on my local computer so here's what the top looks like:

image

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


mraspaud (approved):
LGTM

PR #3088, opened 2025-03-25
Deprecate "cira_day_convection" in favor of "convection"

djhoese:
Added in https://github.com/pytroll/satpy/pull/1251 but was identified as a duplicate from the convection RGB added in #564 but was merged anyway out of impatience it seems.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/composites/abi.yaml
+++ b/satpy/etc/composites/abi.yaml
@@ -353,7 +353,8 @@ composites:
 
   cira_day_convection:
     compositor: !!python/name:satpy.composites.GenericCompositor
-    standard_name: cira_day_convection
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM

djhoese:
@gerritholl I agree it should have "day" in the name, but all the other "convection" recipes don't have it and all the other ABI "official" RGBs that are "Day X RGB" don't have "Day" in the composite name so I went with consistency rather than what makes more sense to me. If EUM and other core dev folks are onboard I can make a PR trying to add "day_" to the composites it makes sense for.

PR #3091, opened 2025-03-28
Fix `RatioCompositor` `standard_name` handling

pnuu:
The `RatioCompositor` was keeping the `standard_name` of the input data. This PR changes it so that it is taken from the compositor attrs if present. This way the `standard_name` given in the composite definition YAML is replacing the `standard_name` of the combined input data.

Also, the enhancement configs for the ESSL Low Level Moisture where this was noticed have now `standard_name`s defined and the extra comments are removed.

 - [x] Tests added

### DIFF
--- a/satpy/composites/__init__.py
+++ b/satpy/composites/__init__.py
@@ -300,6 +300,8 @@ class RatioCompositor(CompositeBase):
         projectables = self.match_data_arrays(projectables)
         info = combine_metadata(*projectables)
         info["name"] = self.attrs["name"]
+        if "standard_name" in self.attrs:
+            info["standard_name"] = self.attrs["standard_name"]
 
         proj = projectables[0] / projectables[1]
         proj.attrs = info
--- a/satpy/etc/enhancements/generic.yaml
+++ b/satpy/etc/enhancements/generic.yaml
@@ -1230,7 +1230,7 @@ enhancements:
           max_stretch: [50, 50, 100]
 
   essl_low_level_moisture:
-    name: essl_low_level_moisture
+    standard_name: essl_low_level_moisture
     operations:
       - name: linear_stretch
         method: !!python/name:satpy.enhancements.stretch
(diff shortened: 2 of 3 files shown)


satpy/composites/__init__.py
```diff
@@ -300,6 +300,8 @@ def __call__(self, projectables, nonprojectables=None, **info):
         projectables = self.match_data_arrays(projectables)
         info = combine_metadata(*projectables)
         info["name"] = self.attrs["name"]
+        if "standard_name" in self.attrs:
+            info["standard_name"] = self.attrs["standard_name"]
```
djhoese, line 304: Should this do `info.update(self.attrs)` instead to take any metadata that may be defined on YAML?

pnuu: Oooh, yeah, true! Nice catch!

pnuu: Updated in https://github.com/pytroll/satpy/pull/3091/commits/3cc1c6f8cb4eb1b8504a035b8fbed7b90f20f824

djhoese:
Restarting failed CI. Seemed like a OS hiccup.

satpy/etc/enhancements/generic.yaml
```diff
@@ -1241,9 +1241,7 @@ enhancements:
     operations: []
 
   essl_colorized_low_level_moisture:
-    # this enhancement is only found if using name but not standard_name
-    # The colormap was developed by the European Severe Storms Laboratory (ESSL).
```
gerritholl, line 1245: Could have kept this line...

djhoese: Oops PR welcome :wink:

pnuu: I'm pretty sure the reason the comment was there was fixed and the enhancement was found, and thus the comment was removed.

pnuu: Actually it was exactly what this PR did.

djhoese: @pnuu he means:

> # The colormap was developed by the European Severe Storms Laboratory (ESSL).

PR #3100, opened 2025-04-04
Add hrv_clouds composite to FCI

guidocioni:
This feature adds the possibility of creating the hrv_clouds composite also with FCI data. 

 - [x] Closes [#2935](https://github.com/pytroll/satpy/issues/2935)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -20,6 +20,7 @@ The following people have made contributions to this project:
 - [Pierre de Buyl (pdebuyl)](https://github.com/pdebuyl)
 - [Eric Bruning (deeplycloudy)](https://github.com/deeplycloudy)
 - [Manuel Carranza (manucarran)](https://github.com/manucarran)
+- [Guido Cioni (guidocioni)](https://github.com/guidocioni)
 - [Lorenzo Clementi (loreclem)](https://github.com/loreclem)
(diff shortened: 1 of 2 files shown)


pnuu:
For SEVIRI the HRV channel covers the wavelength range from 0.5 to 0.9 um while `vis_06` for FCI is only from 0.590 to 0.690 um. There's a lot of the NIR range that is not covered, so I wonder how well the images match in more vegetated areas? I remember testing some variants of `hrv_fog` product with the early data, but I don't remember where I put the images.

guidocioni:
> For SEVIRI the HRV channel covers the wavelength range from 0.5 to 0.9 um while `vis_06` for FCI is only from 0.590 to 0.690 um. There's a lot of the NIR range that is not covered, so I wonder how well the images match in more vegetated areas? I remember testing some variants of `hrv_fog` product with the early data, but I don't remember where I put the images.

I could definitely do some comparison on my side if you can tell me an area or period of time. But anyway we know for sure the hrv_clouds will look different in FCI than SEVIRI, I think the idea is just to have it available.

mraspaud (approved):
LGTM.

guidocioni:
> @guidocioni would you mind adding your name in the AUTHORS.md file? (alphabetical ordering)

Donde

PR #3109, opened 2025-04-15
Add quality filtering to VIIRS EDR CldBaseHght

djhoese:
This adds a `filter_cbh` to a custom file handler for the "CloudBase" VIIRS EDR file. This was a requirement for my Polar2Grid project. I assume the QC is not applied as part of the fill value assignment in the algorithm software (I'm asking Kathy about this).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


djhoese:
Turns out this is likely not needed. The fill value seems to serve the same purpose as checking QC for a value of 0. Closing...

PR #3110, opened 2025-04-16
Improve handling of category products in VIIRS EDR reader

djhoese:
While looking at some VIIRS EDR products like "CloudPhase" I noticed that the DataArray was returned as a 32-bit float. This is because xarray doesn't handle integer types when decoding CF data. It assumes if a variable has a `_FillValue` that those fills should be replaced by NaN and therefore the array should be 32-bit or 64-bit floats.

I don't agree with this and traditionally Satpy doesn't do this. In this PR I fixed it so category arrays stay their integer type and continuous variables lose there `_FillValue` attribute.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented 


pnuu (approved):
LGTM

PR #3113, opened 2025-04-24
Fix angles tests computing xarray DataArrays instead of dask arrays

djhoese:
See https://github.com/dask/dask/issues/11901

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese:
Fixing CodeScene is not an easy fix. I'd prefer to just keep it as is.

mraspaud (approved):
LGTM!

PR #3112, opened 2025-04-22
Fci l1c handle other coverage

ClementLaplace:
This PR is dedicated to handle other coverages regarding the `fci_l1c` reader. 

 - [x] Tests applied for the `fci_l1c` reader

PR #3115, opened 2025-04-24
Fix padding for FCI xx coverage

ameraner:
This PR adds a modification to the `fci_l1c_nc` reader to activate the padding mechanism for special scans like `xx` (implemented in #3112 ).

In order for the padding to work in this special case, a modification/bugfix is also applied to the padding mechanism. In the previous implementation, a bigger missing empty segment was generated by concatenating two segments together. However, in some cases this was not enough (when the desired new empty segment was bigger than two concatenated segments). 
--> To solve this, instead of concatenating, we use the `pad` function to add arbitrarily big segments.

On a side, the documentation is also updated for new links and info.

 - [x] Tests added 
 - [x] Fully documented


mraspaud (approved):
LGTM.
As mentionned on slack, that part of the code in a bit obscure so we should refactor it soon.

PR #3119, opened 2025-05-01
Small Update to MultiScene documentation

joleenf:
This PR just adds a missing import statement in the stack example.


djhoese:
I was going to update the references in this documentation, but I was curious why sphinx wasn't warning about them already. I found sphinx's "nitpicky" mode (`-n` flag to `sphinx-build`) and it showed 336 bad references in the Satpy documentation. I think I'll need to make a separate PR for these fixes. I might rename `_blend_funcs` to `blend_funcs` too and change the documentation in this PR to use the full import. I should probably stop using "alias" imports so much as they are harder to document in sphinx.

Issue #3123, opened 2025-05-09
Refactor modules to avoid unnecessary imports


djhoese:
## Feature Request

**Is your feature request related to a problem? Please describe.**
I have a project that uses Satpy but does not need most of the "usual" functionality that Satpy provides. Specifically I need to read ABI and AHI input data, no resampling, no compositing, and write to a NetCDF file using the "awips_tiled" writer. This project involves creating a docker image so reducing the number of unnecessary dependencies installed can make a huge difference on the size of the docker image.

Additionally, the more imports needed to do basic functionality in Satpy means longer import times overall when the functionality using those imports isn't needed.

**Describe the solution you'd like**
I'd like to refactor a few key modules in Satpy into a series of submodules. As part of that I'd also like to remove "alias" imports as much as possible. I started this with the "satpy.writers" sub-package in #3122, but naming still needs to be finalized. Other modules that should be looked at:

- [ ] `satpy/tests/utils.py` - This one maybe doesn't make sense if we think that public users should not be importing stuff from tests. In my case I wanted to use the `make_dataid` function but the `utils` module was importing a lot of stuff that wasn't necessary.
- [x] `satpy/readers/__init__.py` - Hopefully not as difficult as the others described here since most of the functionality is using builtin python features. https://github.com/pytroll/satpy/pull/3127 and https://github.com/pytroll/satpy/pull/3124
- [x] `satpy/composites/__init__.py` - We have a ton of composite classes here. They should really be divided up. https://github.com/pytroll/satpy/pull/3157
- [x] `satpy/enhancements/__init__.py` - All the generic builtin enhancements are in here and require too many imports (trollimage) that most don't use. https://github.com/pytroll/satpy/pull/3155
- [x] `satpy/multiscene/__init__.py`: These are all alias imports when I refactor `multiscene.py` into separate modules. They should be removed. Additionally `_blend_funcs.py` should be made public. https://github.com/pytroll/satpy/pull/3236
- [x] `satpy/writers/` - https://github.com/pytroll/satpy/pull/3122
- [ ] I'm sure others, but these are the ones I've been bumping into.

After these are refactored the `__init__.py` modules could get a `__getattr__` function that produces a warning when someone tries to access the old import.

**Describe any changes to existing user workflow**
Users will need to change their imports.

djhoese:
[...]

PR #3124, opened 2025-05-12
Refactor resampler module to submodules

pnuu:
This PR refactors the `satpy/resample.py` module to a `satpy.resample` subpackage and splits the contents to several modules.

At least the warnings in the tests should be fixed, but this is the first version to see if there are other things to consider. Also some of the imports could be moved to where they are used. And the docstring should be fixed with the new import paths.

Triggered by https://github.com/pytroll/satpy/issues/3123

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/test_resample.py
+++ b/satpy/tests/test_resample.py
@@ -681,3 +681,28 @@ class TestBucketFraction(unittest.TestCase):
         assert "categories" in res.coords
         assert "categories" in res.dims
         assert np.all(res.coords["categories"] == np.array([0, 1, 2]))
+
+
+@pytest.mark.parametrize("name",
+                         ["KDTreeResampler",
+                          "BilinearResampler",
+                          "NativeResampler",
+                          "BucketResamplerBase",
+                          "BucketAvg",
+                          "BucketSum",
+                          "BucketCount",
+                          "BucketFraction",
+                          "resample",
+                          "prepare_resampler",
+                          "resample_dataset",
+                          "get_area_file",
+                          "get_area_def",
+                          "add_xy_coords",
+                          "add_crs_xy_coords",
+                          ]
+                         )
+def test_moved_import_warns(name):
+    """Test that imports done directly from satpy.resample sub-package issue a warning."""
+    import satpy.resample
+    with pytest.warns(UserWarning):
+        _ = getattr(satpy.resample, name)


satpy/resample/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+            "BucketAvg",
+            "BucketSum",
+            "BucketCount",
+            "BucketFraction"
+    ):
+        from . import bucket
+        new_submod = "bucket"
+        obj = getattr(bucket, name)
+    elif name in (
+            "hash_dict",
```
djhoese, line 188: Is `hash_dict` used anywhere else? Should it be "private-ized"? It would need handling in this if statement, but better to make things private now rather than later.

djhoese:
The other thing about this PR that I'm realizing is that some of the resamplers (ewa maybe? kdtree nearest maybe?) exist in pyresample in a way that should be useable. Or at least that was my intention before I stopped working on pyresample 2.0. Maybe it never got all the way there (as we already know I'm wrong about the caching functionality implemented in pyresample).

satpy/resample/__init__.py
pnuu: Didn't find any calls to it outside the new "base" module, so made it private and removed from the `__getattr__` handling in https://github.com/pytroll/satpy/pull/3124/commits/44ee39b9c122da80218cebfd5ed078ecf4f495e2

pnuu:
I think this is now ready. Unless we come up with a better name for the `base` module.

satpy/resample/__init__.py
```diff
@@ -0,0 +1,209 @@
...
+    elif name in (
+            "get_area_file",
+            "get_area_def",
+            "add_xy_coords",
+            "add_crs_xy_coords",
+            "update_resampled_coords",
+            "resample",
+            "prepare_resampler",
+            "get_fill_value",
+            "resample_dataset"
```
djhoese, line 196: Are all of these used outside of the resample code? I know Scene calls `prepare_resampler` and `resample_dataset`, but are the other functions used? Maybe they could be made private?

djhoese (changes requested):
One more comment for making things private, but also what do you think about moving `get_area_file/def` functions and `add_xy_coords` and `add_crs_xy_coords` to a `utils.py` module? Or even go as far as the `get_area_*` functions to one module outside of the resample subpackage (since they are used by readers and users) like `satpy.area` or `satpy.area_utils` or `satpy.area_config`? And `add_*_coords` to another module inside or outside of resampling? Like `satpy.coords` or `satpy.coords_utils`?

pnuu:
> One more comment for making things private, but also what do you think about moving `get_area_file/def` functions and `add_xy_coords` and `add_crs_xy_coords` to a `utils.py` module? Or even go as far as the `get_area_*` functions to one module outside of the resample subpackage (since they are used by readers and users) like `satpy.area` or `satpy.area_utils` or `satpy.area_config`? And `add_*_coords` to another module inside or outside of resampling? Like `satpy.coords` or `satpy.coords_utils`?

I'll have a look at this today evening after I reach the hotel in Milan, where I need to stay overnight due to a cancelled flight.

satpy/resample/__init__.py
pnuu: I made `update_resampled_coords` and `get_fill_value` private. The former is used only in AWIPS tests outside of `resample` package, the latter only within resamplers. The others have more uses also elsewhere.

satpy/coords_utils.py
```diff
@@ -0,0 +1,137 @@
...
+    return data_arr
+
+
+def _add_crs(area, data_arr):
+    # add CRS object if pyproj 2.0+
+    try:
+        from pyproj import CRS
+    except ImportError:
+        LOG.debug("Could not add 'crs' coordinate with pyproj<2.0")
+        crs = None
```
djhoese, line 126: We now require pyproj>2.2. Do you want to remove this import check and either:

1. Move the import to the top of the module (OK with me as I see pyproj a hard requirement of satpy).
2. Make it an in-function import as it is now.

Either way, the `else:` block can be de-dented.

djhoese (changes requested):
Awesome. I noticed one small thing that could be changed and cleaned up. The only other discussion we should maybe have is how many `satpy/X_utils.py` modules do we have before making things a series of `utils/X.py` modules? Or would it be better to drop the `_utils.py` part of the name and just call them `satpy/area.py` and `satpy/coords.py`?

satpy/area_utils.py
```diff
@@ -0,0 +1,43 @@
...
+def get_area_file():
+    """Find area file(s) to use.
+
+    The files are to be named `areas.yaml` or `areas.def`.
+    """
+    paths = config_search_paths("areas.yaml")
+    if paths:
+        return paths
+    else:
+        return get_config_path("areas.def")
```
djhoese, line 30: In a near-future PR should we drop support for the old areas format?

satpy/resample/base.py
```diff
@@ -0,0 +1,217 @@
...
+#              }
+
+def get_all_resampler_classes():
+    """Get all available resampler classes."""
+    resamplers = {}
+    try:
+        from .native import get_native_resampler_classes
+        resamplers.update(get_native_resampler_classes())
+    except ImportError:
+        pass
```
djhoese, line 96: I don't think it needs to be done in this PR but this seems like another case where it would be easy to have a dictionary or list of strings to import and use `importlib.import_module` and `getattr` to get access to them. You could also use `with suppress(ImportError):` to avoid the "uglier" `except ImportError`.

satpy/resample/ewa.py
```diff
@@ -0,0 +1,11 @@
...
+
+from pyresample.ewa import DaskEWAResampler, LegacyDaskEWAResampler
+
+
+def get_ewa_resampler_classes():
+    """Get bucket resampler classes."""
+    return {
+        "ewa": DaskEWAResampler,
+        "ewa_legacy": LegacyDaskEWAResampler,
+    }
```
djhoese, line 11: Is the idea that this entire module would go away if/when Resamplers can be loaded directly from a pyresample interface? Or should something like this be replace with a dict/list of strings like I mentioned in the `base.py` comment? That is a series of strings to be imported.

satpy/scene.py
```diff
@@ -29,13 +29,14 @@
...
 
+from satpy.area_utils import get_area_def
 from satpy.composites import IncompatibleAreas
 from satpy.composites.config_loader import load_compositor_configs_for_sensors
 from satpy.dataset import DataID, DataQuery, DatasetDict, combine_metadata, dataset_walker, replace_anc
 from satpy.dependency_tree import DependencyTree
 from satpy.node import CompositorNode, MissingDependencies, ReaderNode
 from satpy.readers.core.loading import load_readers
-from satpy.resample import get_area_def, prepare_resampler, resample_dataset
+from satpy.resample.base import prepare_resampler, resample_dataset
```
djhoese, line 39: Since pyresample is a hard requirement this probably doesn't matter, but I'm wondering if this import should be in-function to avoid import Resampler classes. :man_shrugging: Eh that will probably go away naturally when more things are moved to pyresample.

djhoese (approved):
Awesome! Just. Awesome.

I think the most obvious win from this PR is how many things are loading from the area and coords utils instead of the resample module/subpackage.

I made a couple inline comments about long term design decisions. The only other remaining question for me is should `satpy/area_utils.py` and `coords_utils.py` have the `_utils` removed?

satpy/area_utils.py
pnuu: Yeah, I think it's well due time. At the latest in 1.0.

satpy/resample/ewa.py
pnuu: The idea was to just kind of follow the structure of the other resample modules and provide a single function to get a dict of the available resamplers to replace the `RESAMPLERS` dict in `__init__.py` but still have the ability to get all of them.

satpy/resample/base.py
pnuu: Refactored in https://github.com/pytroll/satpy/pull/3124/commits/9f9a3854c8d2c5d92a04d0e597c194f450245ecb . As a side-effect, all the resampler classes could be moved to Pyresample.

pnuu:
> The only other remaining question for me is should `satpy/area_utils.py` and `coords_utils.py` have the `_utils` removed?

These were renamed in https://github.com/pytroll/satpy/pull/3124/commits/5bbd4f124667f2b890a0ab0dcbe42a2daed110eb and https://github.com/pytroll/satpy/pull/3124/commits/0b2be453b412f74db7b4c39ee9ac21372d87ff39

satpy/scene.py
pnuu: ... which doesn't really change anything because the functions a imported anyway and thus import from Pyresample :man_facepalming:

pnuu: Moved the imports in https://github.com/pytroll/satpy/pull/3124/commits/7ff4f6df8d4dcc81db0aefecd5dd57c105712305

PR #3125, opened 2025-05-12
Fix ACSPO reader using binary operations on signed integers

djhoese:
Fix bug in the ACSPO reader where newer versions of xarray raise an exception if you try to do a bitwise operation on a signed integer. Not sure why but the ACSPO "l2p_flags" variable is a signed integer instead of an unsigned integer even though it is a bit mask.

This PR fixes the tests (which failed) to represent the data in real files and then fixes the bug by casting the flags to the unsigned version of their signed type.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/acspo.py
+++ b/satpy/readers/acspo.py
@@ -145,7 +145,8 @@ class ACSPOFileHandler(NetCDF4FileHandler):
 
         if ds_info.get("cloud_clear", False):
             # clear-sky if bit 15-16 are 00
-            clear_sky_mask = (self["l2p_flags"][0] & 0b1100000000000000) != 0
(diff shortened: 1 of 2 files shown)


pnuu (approved):
> l2p_flags is usually signed 16-bit
Usually? Should there be, just in case, a test for uint16 flag values?

Anyways, LGTM, merge when you feel like it.

djhoese:
Good catch @pnuu. I think I phrased that that way because I am not really sure. I only have so much test data and it is all from when I originally wrote the reader. All the data I have and have seen looks like it is signed 16-bit integers for this "l2p_flags" variable so I think I'll ignore the unsigned case for now.

PR #3126, opened 2025-05-13
Replace deprecated utcnow

### LINKED ISSUE #2752: Replacement of deprecated `datetime.datetime.utcnow()`

pjuda:
Removes deprecated `datetime.datetime.utcnow()` and `.utcfromtimestamp()` according to option 3 (start using timezone-aware datetimes and use `dt.datetime.now(dt.datetime.timezone.utc`) as described in #2752 

 - [x] Closes #2752


pnuu:
I had a quick look at the test failure in the evening, but couldn't figure out why the `get_observer_look` now gets a different number of calls. I might explore a bit more later today.

pnuu:
Oooh, the failures were only on Windows. Harder for me to debug. But I started the CI tasks again, just in case...

pnuu (approved):
Looks good to me, thank you!

PR #3135, opened 2025-05-16
Fix more invalid sphinx object references

djhoese:
This is a continuation of #3120. This one is mostly targeting type annotations. A couple things I've done:

1. Switched some docstring types to be type annotations. Sphinx and its extensions were getting confused.
2. Started using a third-party "sphinx_autodoc_typehints" which seems to have better handling of things.
3. Took some liberties with type annotations that I wasn't sure about or were causing errors. The biggest one was the use of `numbers.Number` in satpy.utils as mypy got mad that we were multiplying a `Number` with an `int`. :shrug: 

Things I've noticed:

1. The switch to sphinx_autodoc_typehints got mad:

```
WARNING: error while formatting signature for satpy.readers.pmw_channels_definitions.FrequencyDoubleSideBandBase.__new__: Handler <function process_signature at 0x79f4572afe20> for event 'autodoc-process-signature' threw an exception (exception: 'NoneType' object has no attribute 'FrequencyDoubleSideBandBase') [autodoc]
WARNING: error while formatting signature for satpy.readers.pmw_channels_definitions.FrequencyQuadrupleSideBandBase.__new__: Handler <function process_signature at 0x79f4572afe20> for event 'autodoc-process-signature' threw an exception (exception: 'NoneType' object has no attribute 'FrequencyQuadrupleSideBandBase') [autodoc]
WARNING: error while formatting signature for satpy.readers.pmw_channels_definitions.FrequencyRangeBase.__new__: Handler <function process_signature at 0x79f4572afe20> for event 'autodoc-process-signature' threw an exception (exception: 'NoneType' object has no attribute 'FrequencyRangeBase') [autodoc]
```

NOTE: There are still 63 warnings left according to my local execution.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/doc/rtd_environment.yml
+++ b/doc/rtd_environment.yml
@@ -27,6 +27,7 @@ dependencies:
   - setuptools_scm
   - sphinx>=8.2.0
   - sphinx_rtd_theme
+  - sphinx-autodoc-typehints
   - trollsift
   - xarray
   - zarr
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -534,7 +534,7 @@ enhancement functions in Satpy.
 
 .. warning::
    If you define a composite with no matching enhancement, Satpy will by
-   default apply the :func:`~trollimage.xrimage.XRImage.stretch_linear` enhancement with
+   default apply the :meth:`~trollimage.xrimage.XRImage.stretch_linear` enhancement with
    cutoffs of 0.5% and 99.5%.  If you want no enhancement at all (maybe you
    are enhancing a composite based on :class:`DayNightCompositor` where
    the components have their own enhancements defined), you can use the `image_ready` standard name.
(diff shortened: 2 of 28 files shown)


djhoese:
Just so this is recorded somewhere: some of the remaining warnings look like:

```
<unknown>:1: WARNING: py:class reference target not found: numpy.float32 [ref.class]
```

This comes from global or class attributes like:

```python
some_dtype_alias = np.float32
```

where sphinx sees `np.float32` as a class and tries to link to it in numpy. However, numpy's docs incorrectly list these dtype types as attributes or "data" fields so they can't be found. The workaround would be to do `np.dtype(np.float32)`, but @pnuu and I decided it may be better to just wait. The current cases are in:

1. `fy4_base.py` uses a `npt.NDArray[np.float32]` type.
2. `aapp_l1b.py` uses `CHANNEL_DTYPE = np.float32`.
3. `seviri_l1b_native_hdf.py:GSDTRecords` uses `gp_fac_env = np.uint8` and other similar values.

mraspaud (approved):
Lgtm but haven't checked the rendering

pnuu (approved):

Issue #3106, opened 2025-04-10
Orbital parameters missing in MTG-LI data


jequierz:
The lack of the attribute "orbital_parameters" in the MTG data implies that the function`get_satpos()` raises the following error when using it with MTG-LI data: 

 ```
from satpy.utils import get_satpos

get_satpos(mtg_scene['latitude'])
Traceback (most recent call last):
  File "/home/zjequier/Documents/satpy/satpy/utils.py", line 349, in get_satpos
    lon, lat = _get_sat_lonlat(data_arr, lonlat_prefixes)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zjequier/Documents/satpy/satpy/utils.py", line 387, in _get_sat_lonlat
    orb_params = data_arr.attrs["orbital_parameters"]
                 ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'orbital_parameters'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/zjequier/Documents/satpy/satpy/utils.py", line 357, in get_satpos
    raise KeyError("Unable to determine satellite position. Either the "
KeyError: "Unable to determine satellite position. Either the reader doesn't provide that information or geolocation datasets were not available." 
```
When looking at the attributes of the data, the orbital parameters are indeed not given in the metadata. This is likely that this is not a satpy problem but rather that the LI files themselves do not contain the orbital parameter attributes. However, this would still be meaningful to be able to retrieve the information regarding the satellite location, for instance to apply a parallax correction to the lightning data.

ameraner:
Indeed, that metadata is contained only in L1 data, so it's lost in the disseminated LI L2 files. 
Checking FCI is a good workaround though (as discussed on slack). I would expect that the orbital position movement is slow enough that the 10 minutes data you get by using the FCI `orbital_parameter` is enough for usual use cases. If that's not the case, the FCI files actually contain the subsat lat, lon and altitude at a resolution of 1 second (the FCI satpy reader takes an average of those arrays to populate the `orbital_parameters`.
The alternative would be to try to use pyorbital with the according TLEs I guess.

TAlonglong:
Issue conversation proposed an aproach. Closing this issue.

ameraner:
[...]

PR #3137, opened 2025-05-19
MTG LI L2: Hard coded the satellite position with the nominal vallues 

jequierz:
In the MTG LI L2  reader  `li_l2_nc`, I have hard coded the nominal satellite position as given in the L1 FCI dataset. It allows to use functions such as `get_satpos` from `satpy.utils` that is necessary to apply a parallax correction to the data. 

 - [ ] Workaround for #3106  
 - [ ] Tests added:  no test added


satpy/readers/li_base_nc.py
```diff
@@ -733,6 +733,17 @@ def update_array_attributes(self, data_array, ds_info):
...
 
+    def add_orbital_parameters(self, data_array):
+        """Hard code the nominal position of the satellite in the orbital parameters attribute."""
+        logger.info("The satellite position is the nominal position and thus does not" \
+        "correspond to the actual position of the satellite. The exact satellite position can be found in " \
+        "FCI level 1 data. ")
+        data_array.attrs["orbital_parameters"] = {"satellite_nominal_longitude": 0.0,
+                                                  "satellite_nominal_latitude": 0.0,
+                                                  "satellite_nominal_altitude": 35786400.0
+                                                }
```
gerritholl, line 744: If at least the satellite number is known from the metadata (is it?) this hardcoded position could be sorted by satellite.  MTG-I2 / Meteosat-14 will have a different position.

satpy/readers/li_base_nc.py
```diff
@@ -733,6 +733,17 @@ def update_array_attributes(self, data_array, ds_info):
 
         return data_array
 
+    def add_orbital_parameters(self, data_array):
+        """Hard code the nominal position of the satellite in the orbital parameters attribute."""
+        logger.info("The satellite position is the nominal position and thus does not" \
+        "correspond to the actual position of the satellite. The exact satellite position can be found in " \
```
gerritholl, line 739:
```suggestion
        "correspond to the actual position of the satellite. The exact satellite position can be found in "
```

gerritholl (changes requested):
Still needs a unit test, and I suggested some other changes.

satpy/readers/li_base_nc.py
```diff
@@ -733,6 +733,17 @@ def update_array_attributes(self, data_array, ds_info):
...
+        "FCI level 1 data. ")
```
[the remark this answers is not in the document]
jequierz: Thank you for the suggestion. I will change this :)

satpy/readers/li_base_nc.py
jequierz: Good point, I didn't think about the fact that other satellite's products would require the use of the LI reader... But it is true that in any case, it would be better to have a more general solution because apparently the same problem holds for any L2 product of the MTG. 
However, I am not sure to understand, how does the satellite number can be used to for this application? By writing a dict of satellites containing a dict of nominal position for each satellite?

gerritholl: Yes, I would have a dict of satellites.  Those should rarely change, although if they do, it will cause more of a headache as we would need to cover old and new data.

ameraner: yeah, when we start moving the MTGs between the nominal, rapid scan , IODC position etc, we'll have a new headache. But maybe by then we can include the information in the file format... So for now a simple dict mapping the `platform_name` to the `satellite_nominal_longitude` should be good for a while. Note that the `platform_name` was added only recently in https://github.com/pytroll/satpy/pull/2993

PR #3143, opened 2025-06-03
Add page for specific reader documentation

sfinkens:
Simplify the "Reading" documentation page by moving reader/format specific guidance to a new page "Specific Readers and Formats".

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -68,10 +68,10 @@ Documentation
     composites
     resample
     enhancements
-    readers_formats
     writing
     multiscene
     dev_guide/index
+    readers_formats
 
 .. toctree::
     :maxdepth: 1


doc/source/index.rst
```diff
@@ -68,6 +68,7 @@ Documentation
     composites
     resample
     enhancements
+    readers_formats
```
djhoese: I think the location of the document in the TOC is the only thing that I disagree with in this PR. It feels like this should either go right after "reading" or go after the dev_guide/index in the sense that it is an "extra" document. I think I prefer the latter.

djhoese (commented):
Very nice. This does seem like quite an improvement. Of course it is unfortunate that it (a human-maintained document) is needed, but :man_shrugging:

doc/source/index.rst
sfinkens: :+1: I've placed it after the dev_guide/index

PR #3149, opened 2025-06-09
Fix incorrect wavelength for AHI B04

djhoese:
According to https://www.data.jma.go.jp/mscweb/en/himawari89/space_segment/spsg_ahi.html the AHI B04 central wavelength should be 0.86, not 0.85. This PR fixes this value for AHI HSD and HRIT. I assume that this does not break anything with pyspectral's SRFs (CC @adybbroe @simonrp84).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


mraspaud (approved):
LGTM

PR #3155, opened 2025-06-19
Refactor enhancements

pnuu:
This PR continues with the tasks listed in https://github.com/pytroll/satpy/issues/3123 to split `__init__.py` of subpackages to modules by refactoring enhancements.

 - [x] Tests added 
 - [x] Fully documented


pnuu:
No warnings from the doc build with the `-n` in `Makefile`.

pnuu:
I renamed `stretching.py` to `contrast.py` and `color_mapping.py` to `colormap.py`, so they have more consistent naming with the other modules.

satpy/enhancements/convolution.py
```diff
@@ -0,0 +1,57 @@
...
+# Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# satpy is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
+"""Stretching."""
```
djhoese: Different docstring?

satpy/tests/enhancement_tests/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+# version.
+#
+# satpy is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Helpers for enhancement testing."""
```
djhoese: Codecov is complaining about almost every line in this module not being checked during tests. Are you sure these are being used?

djhoese: Oh is that because "utils" is being used instead? Should this be removed?

satpy/composites/__init__.py
```diff
@@ -581,7 +581,7 @@ class ColormapCompositor(GenericCompositor):
         standard_name: cloud_top_height
         operations:
         - name: palettize
-          method: !!python/name:satpy.enhancements.palettize
+          method: !!python/name:satpy.enhancements.color_mapping.palettize
```
djhoese:
```suggestion
          method: !!python/name:satpy.enhancements.colormap.palettize
```

?

djhoese (changes requested):
Made a couple comments inline. "contrast" is not my favorite name, but it isn't wrong and it is probably more meaningful to others compared to "stretch"/"stretching"/etc. The "wrappers" module seems like the best use case for "utils", but maybe "wrappers" is the better more explicit name of generic "utils". Bottom line I think these names are fine.

Awesome job though. Thanks for all this work.

pnuu:
Thanks for the review, I'll have a closer look tomorrow of the two remaining comments when I'm at the work laptop and a bigger screen.

I went with `contrast` as that's what all the functions in the module adjust. Well, it's debatable if `invert()` does that...

satpy/enhancements/convolution.py
pnuu: Fixed in a303370dabf528f76af2d2af53fc3648fafab4ef

satpy/tests/enhancement_tests/helpers.py
pnuu: Hmm, I wonder why this was duplicated. Removed (contrary to the commit message :man_facepalming:) in https://github.com/pytroll/satpy/pull/3155/commits/7c93d063a8dba01a5cbb83b9658c7f204080581d

satpy/enhancements/convolution.py
pnuu: Fixed in https://github.com/pytroll/satpy/pull/3155/commits/c0765feb3cc9834c62884206cd5900abbfbe5383

pnuu:
I moved `trollimage.colormap.Colormap` imports to the functions using it in `enhancements/colormap.py`. Yeah, I know that it is expected it is imported at the module level seeing the name of the module, but the `lookup()` enhancement doesn't need it so thought it might help in some very edgy use-cases to reduce the footprint.

pnuu:
Fixed the conflicts from https://github.com/pytroll/satpy/commit/562ac1ebd580550ba93c46e6eb438cc9e273632e

djhoese (approved):
I *think* I got all the updated references that I saw in the merge conflicts from `main` but maybe there are more in non-conflicted files?

PR #3157, opened 2025-06-23
Refactor composites to separate modules

pnuu:
This PR continues https://github.com/pytroll/satpy/issues/3123 by refactoring the `composites/__init__.py` file to new modules.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 


pnuu:
I did some cleanup to reduce CodeScene complaints.

Once we've agreed on the split and naming, I'll update the composite definitions, imports and documentation to match the new module structure.

I'm not sure why pre-commit.ci is complaining about the exceptions not derived from `BaseException` as they clearly are. My quick search hinted that the `raise` should instantiate the class, but that doesn't seem to help.

pnuu:
I checked all the compositor modules for eager imports at the module level, and they should be as clean as they reasonably can be. I also checked most if not all of the connected imports.

doc/source/composites.rst
```diff
@@ -12,7 +12,7 @@ the key/values of the DataID used to identify it.
 Built-in Compositors
 ====================
 
-.. py:currentmodule:: satpy.composites
+.. py:currentmodule:: satpy.composites.core
```
djhoese, line 15: I *think* since you specify all API objects with a full path now then this line can be removed.

satpy/composites/core.py
```diff
@@ -0,0 +1,535 @@
...
+
+def check_times(projectables):
+    """Check that *projectables* have compatible times."""
+    times = []
+    for proj in projectables:
+        status = _collect_time_from_proj(times, proj)
+        if not status:
+            break
+    else:
+        return _get_average_time(times)
```
djhoese, line 496: Codecov is complaining about this line and many of the lines below this not being covered. I don't think that's the fault of this PR, but it is concerning how much time-related things are not being tested in the composite tests.

satpy/composites/core.py
```diff
@@ -0,0 +1,535 @@
...
+
+
+def _get_average_time(times):
+    # Is there a more gracious way to handle this ?
+    if np.max(times) - np.min(times) > TIME_COMPATIBILITY_TOLERANCE:
+        raise IncompatibleTimes("Times do not match.")
+    return (np.max(times) - np.min(times)) / 2 + np.min(times)
+
+
+class RGBCompositor(GenericCompositor):
```
djhoese, line 523: Maybe not in this PR, but it is probably time to remove this compositor.

satpy/composites/resolution.py
```diff
@@ -0,0 +1,289 @@
...
+        y_size, x_size = r__.shape
+        r__ = da.reshape(r__, (1, y_size, x_size))
+        g__ = da.reshape(g__, (1, y_size, x_size))
+        b__ = da.reshape(b__, (1, y_size, x_size))
+
+        rgb_img.data = da.vstack((r__, g__, b__))
+        return super(LuminanceSharpeningCompositor, self).__call__(rgb_img, *args, **kwargs)
+
+
+class SandwichCompositor(GenericCompositor):
```
djhoese, line 274: This `SandwichCompositor` seems wrong for `resolution.py`. It doesn't actually do anything to sharpen or change the resolution of the image unless I'm misunderstanding stuff. I could see how the placement of `LuminanceSharpeningCompositor` in this module (which does operate on two different resolutions) makes sense and that another compositor that operates on luminance should live next to it.

satpy/composites/lookup.py
```diff
@@ -0,0 +1,208 @@
...
+        data = projectables[0].astype(int)
+        res = data.data.map_blocks(self._getitem, self.lut, dtype=self.lut.dtype)
+
+        new_attrs = data.attrs.copy()
+        self._update_attrs(new_attrs)
+
+        return xr.DataArray(res, dims=data.dims, attrs=new_attrs, coords=data.coords)
+
+
+class ColormapCompositor(GenericCompositor):
```
djhoese, line 81: Wow this module name makes a lot more sense now that I realize that this compositor and the others below it are all deprecated.

satpy/composites/aux_data.py
```diff
@@ -0,0 +1,164 @@
...
+
+import satpy
+from satpy.aux_download import DataDownloadMixin
+
+from .core import GenericCompositor
+
+LOG = logging.getLogger(__name__)
+
+
+class StaticImageCompositor(GenericCompositor, DataDownloadMixin):
```
djhoese, line 32: I'm not sure how I feel about `aux_data` for this module name for this compositor. Yes it takes outside data ("aux" data), but 99% of the time this is to provide a background or mask. I...ugh I don't know.

djhoese (approved):
Awesome work. I had a couple inline comments. Some of my comments and remaining thoughts are wondering on the module naming which is by far the hardest part of this PR. While reviewing this it seemed like your module naming was based on how the composites do what they do, but not necessarily why they do it or how they're used. For example, you have `cloud_products.py` but then in `mask.py` there are cloud related composites. Another example is `aux_data.py` which is for the static image compositor, but maybe `image.py` would describe what it is used for better but at the same time be misleading because you might expect most of the compositors to be image-related.

My other concern is that you went with the easiest "core.py" module, but there are already other modules that could go in a a `core/` subpackage like `config_loader.py`.

Overall I think the naming is pretty good for how hard it is to split these things up. I don't think anything I could suggest would improve on what you've done here.

doc/source/composites.rst
pnuu: I *thought* I removed it after changing to the full paths, but I guess I just removed the original one after I noticed how many times I'd need to use the `currentmodule` thing. (And there were also some places where there were two classes that were moved to different modules.)

satpy/composites/core.py
pnuu: I had the same thought, but thought again and left it for the 1.0 deprecation cleanup.

satpy/composites/resolution.py
pnuu: Hmm. The normal use I've seen for the compositor is to use higher resolution VIS data as luminance on top of lower resolution IR data. Like SEVIRI 1 km HRV data on top of 3 km IR10.8. So it's kind of pansharpening.

pnuu: The composite is desrcibed in https://eumetrain.org/sites/default/files/2024-03/SandwichProduct.pdf

satpy/composites/lookup.py
pnuu: Yeah, that makes the `lookup` name better than the previous `color_mapping` or `colormap`.

satpy/composites/aux_data.py
pnuu: I'm not entirely happy with the name either, but couldn't figure out a better one. None of the builtin composites use this compositor to provide a mask, it is only used to provide a background. Not to say it could be used that way...

pnuu:
> While reviewing this it seemed like your module naming was based on how the composites do what they do, but not necessarily why they do it or how they're used. For example, you have cloud_products.py but then in mask.py there are cloud related composites.

My reasoning was that the compositors in `cloud_products.py` are handling cloud product data as input specifically, while the cloud related compositors in `mask.py` create new (cloud) masks.

pnuu:
I fixed the merge conflicts caused by https://github.com/pytroll/satpy/commit/eeba7ebcf5a57204fa0e9c42dbc7823c97c35c16 . I hope I got it right.

mraspaud (approved):
LGTM, thanks a lot for the huge effort!

djhoese:
I reviewed you last commits including the `main` merge and it all looks good to me. I know there were some comments about module naming but I don't think either of us came up with anything better. Let's merge this and see how much people hate things...

PR #3166, opened 2025-07-14
Fix for radiance conversion in Sentinel-2 / MSI SAFE reader.

simonrp84:
The reflectance-to-radiance calculation in the MSI SAFE (Sentinel-2) reader was incorrect, it wrongly assumed the Sun-Earth distance rather than a correction factor.
This small PR fixes the calculation and updates the tests.

 - [x] Tests added


djhoese (approved):
LGTM from what I understand. Thanks!

Issue #2620, opened 2023-10-26
Large overhead resampling multiple datasets to multiple areas with FCI


gerritholl:
**Describe the bug**

When saving multiple composites to multiple areas, there is a very large overhead before dask gets to work using all nodes.

**To Reproduce**

The following might seem convoluted, but is a pretty close approximation to what trollflow2 does in practice and therefore to what happens operationally.  @pnuu may correct me if I'm wrong.

```python
import hdf5plugin
from satpy import Scene
from glob import glob
from satpy.utils import debug_on; debug_on()
from satpy.writers import compute_writer_results
from dask.diagnostics import Profiler, ResourceProfiler, visualize
import dask.config
fci_files = glob("/media/x21308/MTG_test_data/2022_05_MTG_Testdata/RC0040/*BODY*.nc")
res500 = ["vis_06"]
res1000 = ["vis_08", "nir_16", "ir_38", "ir_105"]
res2000 = ["wv_63", "wv_73", "ir_87", "ir_97", "ir_123", "ir_133", "airmass",
           "convection", "dust", "ash"]
cd = "/data/gholl/cache/pytroll"
with Profiler() as prof, ResourceProfiler(dt=0.25) as rprof:
    sc = Scene(filenames={"fci_l1c_nc": fci_files})
    sc.load(res500 + res1000 + res2000, generate=False)
    ls1 = sc.resample("nqeuro500m", cache_dir="/data/gholl/cache/pytroll")
    ls2 = sc.resample("nqeuro1km", cache_dir="/data/gholl/cache/pytroll")
    ls3 = sc.resample("nqeuro2km", cache_dir="/data/gholl/cache/pytroll")
    comps = []
    for c in res500:
        comps.append(ls1.save_dataset(c, filename=f"{c:s}.tif", compute=False))
    for c in res1000:
        comps.append(ls2.save_dataset(c, filename=f"{c:s}.tif", compute=False))
    for c in res2000:
        comps.append(ls3.save_dataset(c, filename=f"{c:s}.tif", compute=False))
    compute_writer_results(comps)
visualize([prof, rprof], show=False, save=True, filename="fci-dask-multi-dataset.html")
```

**Expected behavior**

I would expect a small overhead and then many dask workers working along nicely.

**Actual results**

The dask graph shows the clearest illustation of the problem:

![grafik](https://github.com/pytroll/satpy/assets/500246/d0a1b37b-1415-4f87-ace2-806d860fa5c4)

[...]

PR #3168, opened 2025-07-17
Track and write measurement time for area

### LINKED ISSUE #3161: Write "valid time" to filename or headers
## Feature Request

Our users have requested that we write a "valid time" header to our ninjogeotiff and NetCDF files and include this timestamp in the filename.  This "valid time" should be as representative as possible of the measurement time for the image being produced.  We have agreed to this request (at least for GEO products).  

Currently, we use the reference time for satellite imagery production, or the start time when this is not available.  For Meteosat, this time is not optimally representative for the measurement time in Central Europe.  In the past, we have artificially added the repeat cycle interval for MSG ZDS (15 minutes) or MSG RSS (5 minutes) and called it "end of scan" in an in-house addition, but this time is not the best possible estimate for the measurement time either (and our implementation was buggy).

I would like that, when writing geotiff, ninjogeotiff, or NetCDF files, it should be possible to write a "valid time" header that is representative of the measurement time for the image.  This could be the measurement time for the centre or an average measurement time over the entire image (those two are probably close to each other in most cases).

Naturally, this will need to consider resampling.

This would be an additional feature.  Therefore, it would not change existing user workflow.

**Additional context**

An alternative would be an ad-hoc in-house solution where we hardcode a fixed number of minutes to add depending on the production.  However, a solution integrated in satpy would be easier to maintain, and possibly of use to others as well.

gerritholl:
Track measurement times such that a representative measurement time can be estimated and written for an area.

To include valid time in file headers for supported writers, add `reader_kwargs: {"track_time": True}` to `Scene.__init__` and
`resample_coords=True` to `Scene.resample`.

When storing as `ninjogeotiff`, time coordinates will be averaged and stored in the `ninjo_ValidTimeID` tag.

An example is included in the PR under `doc/source/examples/valid_time.rst`.

Those two PRs must be merged first:

* #3178 or #3477 
* #3371 

Next steps, either within this PR or within one or more later PRs:

- [ ] Add other readers
- [x] Add other writers
- [ ] Add a way to include the valid time in the filename (deferred)
- [x] Investigate reduce_data performance penalty (see [this comment](https://github.com/pytroll/satpy/pull/3168#issuecomment-4207260455))

The latter is difficult to do without triggering a computation, so doing this efficiently should probably be done as postprocessing beyond the scope of this PR.

```python
import os
from satpy import Scene
from glob import glob
sc = Scene(filenames={"seviri_l1b_hrit": seviri_files}, reader_kwargs={"track_time": True})
sc.load(["IR_108"])
ls = sc.resample("nqeuro3km", resample_coords=True, resampler="gradient_search")
ls.save_dataset(
        "IR_108",
        "{platform_name}-{sensor}-{name}-{area.area_id}-{start_time:%Y%m%d%H%M}-geotiff.tif",
        writer="ninjogeotiff",
        ChannelID="IR 10.8",
        DataType="GORN",
        PhysicValue="Temperature",
        PhysicUnit="K",
        SatelliteNameID="METEOSAT Europe-Atlantic",
        fill_value=0)
```

 - [x] Closes #3161 
 - [x] Tests added 
 - [x] Fully documented 


gerritholl:
To track "valid time" through resampling for geostationary satellite data, I'm trying to think of a good way to approach this.  My current thinking is that if the user creates the scene with reader_kwargs={"track_time": True}, the file handler adds per-pixel time information to the variable attributes.  By putting it inside ancillary variables, dataset_walker and co ensure it gets sliced and resampled along with the parent dataset, so that at the end, an approximate measurement time can be estimated for the resampled scene.  But normally ancillary variables are actual variables available in the data file, they remain text attributes until after segments have been concatenated, and then the file handler is called again for every segment, to replace the text label by actual data in ancillary variables.  Therefore, I'm not sure if ancillary variables is the right place to put it, or if time should be stored there at all.

gerritholl:
Maybe it fits better in a coordinate variable, but not sure if it's valid to have those be not-unique.

gerritholl:
Coordinates can probably be non-unique, but are currently dropped in resampling.  In case of nearest neighbour resampling, this is due to 

https://github.com/pytroll/pyresample/blob/61628530636b596150f3b0f45d0a9e78a2d80f2c/pyresample/kd_tree.py#L1094-L1095

gerritholl:
Addressing the CodeScene complaint about `Scene._resampled_datasets` is too difficult for me.  I've looked at it for too long and tried different things, but I do not manage to improve this, sorry.

https://github.com/gerritholl/satpy/blob/329dab5489adf9dfa0b170be3cd16acc52f1674d/satpy/scene.py#L864-L907 

I am not able to resolve this as part of this PR.

gerritholl:
I find it hard to make sense of the `CodeScene` complaints.  Yes, I increased `save_image` from 90 to 94 LOC, but 57 lines are documentation and the rest is just 13 logical LOCs.  And suppressing the complaint does not work ("Error occurred while suppressing! Please try again").

satpy/readers/seviri_l1b_hrit.py
```diff
@@ -453,6 +456,19 @@ class HRITMSGFileHandler(HRITFileHandler):
...
+    **Time tracking**
+
+    The reader supports adding per-pixel time estimates as coordinates to
+    loaded variables::
+
+        scene = satpy.Scene(filenames,
+                            reader='seviri_l1b_hrit',
+                            reader_kwargs={'track_time': True})
+
+    Those coordinates are kept in resampling iff passing the argument
```
pnuu:
```suggestion
    Those coordinates are kept in resampling if, and only if, passing the argument
```

pnuu: Not sure if the "and only if" is even necessary, seems redundant to me.

satpy/tests/writer_tests/test_ninjogeotiff.py
```diff
@@ -985,3 +991,30 @@ def test_str_ids(test_image_small_arctic_P):
...
+        compress="lzw",
+        predictor=2,
+        dynamic_fields={"valid_time"},
+        PhysicUnit="N/A",
+        PhysicValue="N/A",
+        SatelliteNameID="trollsat",
+        ChannelID="trollchannel",
+        DataType="GORN",
+        DataSource="dowsing rod")
+    src = rasterio.open(fn.replace("{valid_time:%Y%m%d%H%M%S}",
```
pnuu: Should the file be closed?

satpy/tests/writer_tests/test_utils.py
```diff
@@ -0,0 +1,50 @@
...
+import numpy as np
+import pytest
+import xarray as xr
+
+
+def test_get_valid_time():
+    """Test extracting valid time from a dataset."""
+    from satpy.writers.core.utils import get_valid_time
+
+    da = xr.DataArray(
```
pnuu: `da` is typically used for `import dask.array as da` so this variable name can cause conflicts and/or confusement (like it did for me).

satpy/tests/writer_tests/test_utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+    da = xr.DataArray(np.zeros((3, 3)), dims=("y", "x"), name="Liedik")
+    with pytest.raises(ValueError,
+                       match="Dataset Liedik has no time coordinate."):
+        get_valid_time(da)
+
+    da = xr.DataArray(np.zeros((3, 3)), dims=("y", "x"), name="Liedik", attrs={"name": "Gámasčearru"})
+    with pytest.raises(ValueError,
+                       match="Dataset Gámasčearru has no time coordinate."):
+        get_valid_time(da)
```
pnuu: These should be separate test cases.

satpy/writers/core/utils.py
```diff
@@ -0,0 +1,44 @@
...
+"""Utilities shared between writers."""
+
+from xarray.coding.times import decode_cf_datetime
+
+
+def get_valid_time(dataset):
+    """Get the valid time for a dataset.
+
+    If a dataset has time coordinates, get the arithmetic mean time as a
+    representative time for the dataset.
```
pnuu: I wouldn't name this function as "valid" time but "average" or "mean" (observation) time. Or maybe add a kwarg that controls the behaviour? Others might have different definition of "valid" time. The other options could just be `NotImplementedError` for now.

satpy/writers/ninjogeotiff.py
```diff
@@ -418,12 +430,22 @@ def get_date_id(self):
...
         delta = tm.replace(tzinfo=datetime.timezone.utc) - self._epoch
         return int(delta.total_seconds())
 
+    def get_valid_date_id(self):
+        """Calculate the valid date ID.
+
+        That's seconds since UNIX Epoch for a representative time for the
+        image.
+        """
+        dt = get_valid_time(self.dataset)
```
pnuu: `dt` is used for `import datetime as dt`, so this is very confusing.

pnuu (changes requested):
Some comments and change requests inline.

satpy/readers/seviri_l1b_hrit.py
gerritholl: Fixed in c3626c430.

satpy/tests/writer_tests/test_ninjogeotiff.py
gerritholl: I thought garbage collection should do that, but in the spirit of Explicit Is Better Than Implicit, I have changed this and all other `rasterio.open` instances in context managers.  Fixed in c3626c430.

satpy/tests/writer_tests/test_utils.py
gerritholl: Fixed in c3626c430.

satpy/tests/writer_tests/test_utils.py
gerritholl: Split in two in c3626c430, one for has valid time, one for has no valid time.  Or should it even be split in three?  Should each piece of behaviour being tested come with its own testing function, or is that exaggerated?

satpy/writers/core/utils.py
gerritholl: Renamed `get_mean_time` in c3626c430.

satpy/writers/ninjogeotiff.py
gerritholl: Fixed in c3626c430.

djhoese:
Docs still failing:

```
satpy/readers/seviri_l1b_hrit.py:docstring of satpy.readers.seviri_l1b_hrit.HRITMSGFileHandler:27: WARNING: unknown document: 'valid_time' [ref.doc]
/home/docs/checkouts/readthedocs.org/user_builds/satpy/conda/3168/lib/python3.11/site-packages/satpy/writers/ninjogeotiff.py:docstring of satpy.writers.ninjogeotiff.NinJoGeoTIFFWriter.save_image:74: WARNING: unknown document: 'mean_time' [ref.doc]
/home/docs/checkouts/readthedocs.org/user_builds/satpy/checkouts/3168/doc/source/writing.rst:20: WARNING: unknown document: 'mean_time' [ref.doc]
```

satpy/tests/writer_tests/test_utils.py
gerritholl: Split in three in [d619238](https://github.com/pytroll/satpy/pull/3168/commits/d619238e9758916f233a79b87422f9097d2f8ef2).

gerritholl:
Returned to draft because it fails with `reduce_data=False`.  It fails with the `EffectiveSolarPathlengthCorrector` in combination with `generate=False`, because `match_data_arrays` destroys the time coordinate due to its call to `drop_coordinates`:

https://github.com/pytroll/satpy/blob/58ce5367ea81a8083794b1e53c643b4bb85a5005/satpy/composites/core.py#L205-L226

The question is, why is satpy doing this in the first place…

gerritholl:
We need a smarter strategy for handling time coordinates in composites, in case where we want to keep them.  A compositor with a single band shouldn't need to throw them out.  When multiple bands exist with different time coordinates, in cases where the bands do not need resampling anyway due to differing areas, we would need to offer a different option apart from dropping, like taking the first band or averaging them.  Hopefully the differences should be very small!

gerritholl:
Elsewhere, Satpy expects that time coordinates are of dtype datetime64, for example, in `satpy.composites.core`, originally added in https://github.com/pytroll/satpy/commit/c1b65cf18b072915f53ba58cd16bcfd613072f68:

https://github.com/pytroll/satpy/blob/e078c6ed5a068657f3f0d5cdaeb7bc423651fba3/satpy/composites/core.py#L38 

https://github.com/pytroll/satpy/blob/e078c6ed5a068657f3f0d5cdaeb7bc423651fba3/satpy/composites/core.py#L549-L555

This is a problem, because datetime64 cannot be resampled, so this conflicts with the use case for tracking measurement time.

gerritholl:
Elsewhere, satpy expects numerical values:

https://github.com/pytroll/satpy/blob/e078c6ed5a068657f3f0d5cdaeb7bc423651fba3/satpy/composites/core.py#L535 

Why is it skipping time coordinates where the initial value is 0?

gerritholl:
It also reduces the dimensionality of the time coordinate by taking only the first value — why?

https://github.com/pytroll/satpy/blob/e078c6ed5a068657f3f0d5cdaeb7bc423651fba3/satpy/composites/core.py#L536

Seems to be originally added in https://github.com/pytroll/satpy/commit/d41a213afb2c3fcad24872da1bb57137e0f702a5.

gerritholl:
Writing the valid time in the headers now works lazily.  For the filename, lazy evaluation is too hard and I will probably not implement this.  Users who need this should use a second step renaming the file based on information from the header.

gerritholl:
**EDIT:** This comment is out of date.  There is no longer a performance penalty.

Original comment:

---

This PR is strongly increasing the overhead of `reduce_data` when resampling, even if time tracking is not switched on.

Calling `Scene.resample`:

| branch | reduce_data | time tracking | [caching](https://pyresample.readthedocs.io/en/latest/howtos/configuration.html#cache-geometry-slices) | duration
| ------------- | ------------- | ---- | --- | --- | 
| main | False  | N/A | False | 0.97 
| main  | True | N/A | False | 8 
| main | False | N/A | True | 0.96
| main | True | N/A | True | 0.97
| 3168 | False | False | False | 0.96
| 3168 | False | False | True | 0.96
| 3168 | False | True | False | 1.11
| 3168 | False | True | True | 1.20
| 3168 | True | False | False | 42.7
| 3168 | True | False | True | 0.99
| 3168 | True | True | False | 83.4
| 3168 | True | True | True | 1.2

NB: This is just for calling resample — no computation.  See also https://github.com/pytroll/satpy/issues/2620.

I don't yet understand why the overhead for `resample` with `reduce_data=True` increases from 8 to 43 seconds even if no time tracking is actually used.  Needs to be investigated more.

gerritholl:
For ninjogeotiff, the NinJo team have agreed this should go in a tag called `ninjo_ValidTimeID` (not `ValidDateID`).

(In case a NinJo developer or someone with access to the internal NinJo JIRA reads this: details in NinJo JIRA-38947).

gerritholl:
The worsening performance with `_reduce_data` when not using time tracking is due to https://github.com/pytroll/satpy/pull/3178.

The even-more-worsening performance with time tracking is because reduce data is called twice as often.

gerritholl:
Both problems are solved with https://github.com/pytroll/satpy/pull/3178/commits/f70d80f2c2d5a52a222a45920c36a0e36a7d64d1.  The superfluous calls to `_reduce_data` are not a problem any more (little overhead), because on the second run it finds the source areas in the reductions dictionary.  The overhead merely increases from 8.3 to 8.6 seconds (previously from 8.3 to 83.4 seconds).

doc/source/examples/mean_time.rst
```diff
@@ -0,0 +1,39 @@
...
+   sc = Scene(filenames={"seviri_l1b_hrit": seviri_files}, reader_kwargs={"track_time": True})
+   sc.load(["IR_108"])
+
+The time is stored as a coordinate:
+
+.. code-block:: python
+
+   sc["IR_108"].coords["time"]
+
+To retain it upon resampling, pass ``resample_coords=True`` to :meth:`~satpy.scene.Scene.resample`:
```
mraspaud, line 27: Does that mean that lons and lats will be resampled too if they are in the product?

doc/source/examples/mean_time.rst
```diff
@@ -0,0 +1,39 @@
+Tracking measurement time
+=========================
+
+.. versionadded:: v0.58
+
+For some readers and writers, it is possible to keep track of pixel-level
+measurement times and store the average measurement time in the metadata
+for the resampled image.  This can be stored in the filename or the file
```
mraspaud, line 8: So it doesn't work for line-based acquisition time?

satpy/cf/decoding.py
```diff
@@ -57,3 +59,39 @@ def _str2datetime(string):
...
+
+    Args:
+        xrda_encoded (array-like):
+            Xarray data array with units attribute.  The units attribute should
+            be a string describing the time units using "x since timestamp",
+            UDUNITS-style.
+    """
+    # An early iteration of this function was written with the
+    # assistance of GPT-5.4.
+
```
mraspaud, line 79:
```suggestion
```

satpy/cf/decoding.py
```diff
@@ -57,3 +59,39 @@ def _str2datetime(string):
         return dt.datetime.fromisoformat(string)
     except (TypeError, ValueError):
         return None
+
+
+def lazy_decode_cf_time(xrda_encoded):
+    """Lazily decode CF-encoded time in limited situations.
+
```
mraspaud, line 66: Wondering if that function should really be in satpy, or rather in eg xarray

satpy/composites/arithmetic.py
```diff
@@ -53,7 +55,8 @@ def __call__(self, projectables, nonprojectables=None, **info):
         """Generate the composite."""
         if len(projectables) != 2:
             raise ValueError("Expected 2 datasets, got %d" % (len(projectables),))
-        projectables = self.match_data_arrays(projectables)
+        projectables = self.match_data_arrays(projectables,
+                                              drop_coordinates=False)
```
mraspaud, line 59: Does this change the default behaviour of compositing? In that case what is the effect?

satpy/composites/core.py
```diff
@@ -413,7 +456,7 @@ def _concat_datasets(self, projectables, mode):
             data = xr.concat(projectables, "bands", coords="minimal")
             data["bands"] = list(mode)
         except ValueError as e:
-            LOG.debug("Original exception for incompatible areas: {}".format(str(e)))
+            LOG.exception("Original exception for incompatible areas: {}".format(str(e)))
```
mraspaud, line 459: `log.exception` make this an error in the log, which is used in some places to trigger alarms.
Is this really an error? Isn't that expected in some cases?

satpy/composites/core.py
```diff
@@ -503,39 +543,84 @@ def _get_updated_attrs(self, datasets, attrs, mode):
...
+                # this point, we should make sure NO time coordinates are
+                # assigned to the time coordinate or things will get messed up
+                # later.
+                first = timed_projectables[0].coords["time"]
+                new_coords = {}
+                if "y" in first.coords:
+                    new_coords["y"] = first.coords["y"].copy()
+                if "x" in first.coords:
+                    new_coords["x"] = first.coords["x"].copy()
+                da_new_time = sum([x.coords["time"].data for x in timed_projectables]) / len(timed_projectables)
```
mraspaud, line 572: Isn't that code duplicated from `combine_times` above?

satpy/readers/core/seviri.py
```diff
@@ -441,6 +441,41 @@ def add_scanline_acq_time(dataset, acq_time):
...
+
+    Args:
+        dataset (xarray.DataArray): dataset with acq_time coordinate
+        label (str): Label for time coordinate.  Defaults to "time".  Choosing
+            a different label might impact the ability of writers trying to
+            process time coordinate information.
+    """
+    st_time = dataset.attrs["nominal_start_time"]
+    delta_per_line = (dataset["acq_time"].chunk({"y": dataset.sizes["y"]}) - np.datetime64(st_time)).astype("m8[s]")
+    # convert to float, because datetime64 cannot be resampled and ints cannot
```
mraspaud, line 461: Out of curiosity, what makes datetime64 no resamplable?

satpy/writers/core/utils.py
```diff
@@ -0,0 +1,46 @@
...
+# terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# satpy is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
```
mraspaud, line 15: The licence has changed, no headers needed anymore

mraspaud (commented):
Thanks for working on this, interesting problem. Some comments and questions online.

PR #3178, opened 2025-07-28
Refactor `_resampled_scene()` and `_reduce_data()` methods of the `Scene` class

pnuu:
This PR refactors `Scene._resampled_scene()` that @gerritholl reported to be complicated in https://github.com/pytroll/satpy/pull/3168#issuecomment-3109062005 . As there was a change to `_reduce_data()`, I did some additional refactoring to it too.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


pnuu:
I'll do some more refactoring to `_reduce_data()`.

gerritholl:
Thanks!  I've merged this into #3168 but there is still an issue with missing test coverage.

satpy/scene.py
```diff
@@ -878,34 +876,59 @@ def _resampled_scene(self, new_scn, destination_area, reduce_data=True,
...
+            res = self._resample_dataset(source_area, destination_area, dataset, resamplers, resample_kwargs)
+
             new_datasets[ds_id] = res
             if ds_id in new_scn._datasets:
                 new_scn._datasets[ds_id] = res
             if parent_dataset is not None:
                 replace_anc(res, pres)
 
+    @classmethod
+    def _get_new_datasets_from_parent(self, new_datasets, parent_dataset):
```
djhoese: The `self` should be `cls` since these are classmethods now, but given that they are classmethods or could be staticmethods, how about we move these to outside of the Scene? And could they (or should they) be moved to the `satpy.resample` subpackage?

satpy/scene.py
```diff
@@ -927,32 +950,39 @@ def _prepare_resampler(self, source_area, destination_area, resamplers, resample
...
-                            destination_area)
-                    source_area = source_area[slice_y, slice_x]
-                    reductions[key] = (slice_x, slice_y), source_area
-                dataset = self._slice_data(source_area, (slice_x, slice_y), dataset)
-            else:
-                LOG.debug("Data reduction disabled by the user")
+            slice_x, slice_y = self._get_source_dest_slices(source_area, destination_area, reductions, resample_kwargs)
+            source_area = source_area[slice_y, slice_x]
+            reductions[source_area] = (slice_x, slice_y), source_area
+            dataset = self._slice_data(source_area, (slice_x, slice_y), dataset)
```
djhoese: I'm curious if the `_get_source_dest_slices` operation is the only step that raises NotImplementedError or if `_slice_data` also does it? If the former, then maybe `_slice_data` should be moved outside of the try/except. Thoughts?

satpy/scene.py
```diff
@@ -878,34 +876,59 @@ def _resampled_scene(self, new_scn, destination_area, reduce_data=True,
...
             new_datasets[ds_id] = res
             if ds_id in new_scn._datasets:
                 new_scn._datasets[ds_id] = res
             if parent_dataset is not None:
                 replace_anc(res, pres)
 
+    @classmethod
+    def _get_new_datasets_from_parent(self, new_datasets, parent_dataset):
+        if parent_dataset is not None:
+            return new_datasets[DataID.from_dataarray(parent_dataset)]
```
djhoese: `DataID.from_dataarray` returns a single DataID, right? I think I'd prefer a different name for this method. I think the purpose of this chunk of code is to say "if we've resampled the parent already, use the resampled parent", right? Or rather, if the current dataset has a parent, it should have been resampled already, so we should use the resampled version of the parent. I think?

In addition to renaming, it seems that `parent_dataset` is only used in the later steps to check for `is None`. I'm wondering if we can remove the use of `parent_dataset` in favor of `pres` and "bundle" this methods operation with the `dataset_walker` to be something like:

```python
for ds_id, dataset, resampled_parent in resampled_dataset_walker(datasets, new_datasets):
```

Or something like that.

...and if that is done, then there might be an argument for putting `_replace_anc_for_new_datasets` and `_update_area` into the for loop generator too. This changes the purpose of the for loop to be "what datasets do we need to resample" and then the inside of the for loop logic is just "reduce data", "resample data", "store result".

I'll admit the code was ugly and the logic of `new_scn._datasets` and `new_datasets` really isn't helping that.

djhoese: Sorry for all the comments and no regular review. I just keep brainstorming.

One other idea, what if only `new_datasets` gets modified in the for loop and then assigning to `new_scn._datasets` is left for a second for loop (ex. `for ds_id, new_data_arr in new_datasets.items():`)? Maybe that would clean up the code inside the loop. I feel like a lot of this ugliness is caused by new_scn (or rather the `DatasetDict` inside) copying the DataArray and/or making modifications to it.

satpy/scene.py
pnuu: I'll have a look, thanks.

satpy/scene.py
pnuu: `NotImplementedError` or `KeyError`?

pnuu: Nevermind, the try/except was just outside of this snippet. I'll see what can trigger the error.

pnuu:
I made the Fine Choice(tm) of asking Copilot (Claude Sonnet 4.6) to refactor the method. After couple of iterations we ended up with this version that uses a data class to hold resampling context and a method that is called for each resampled dataset.

djhoese:
Interesting...I've never personally used data classes beyond toy programs and watching programming videos. What it is doing here makes me wonder if this deserves a full blown helper *class* where instead of `context` in these sub-methods it is just `self.`.

djhoese:
And sorry, I don't fully remember all the previous context for this PR and past thoughts I had about it.

pnuu:
Unrelated test failure:

```
ERROR satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py::TestFiduceoMviriFileHandlers - satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py::TestFiduceoMviriFileHandlers::test_init: duplicate parametrization of 'projection_longitude'
```

I can't replicate this locally with package versions I have.

pnuu:
Seems to be somehow related to https://github.com/pytest-dev/pytest/issues/13974 that have several links from the few last days.

pnuu:
Possible fix in https://github.com/pytroll/satpy/pull/3415

gerritholl:
Needs more test coverage?

gerritholl:
This PR increases the overhead/runtime of `_reduce_data` by a factor 5.  Test script:

```
import os
from satpy import Scene
from glob import glob
from satpy.tests.utils import CustomScheduler
from satpy.writers.core.compute import compute_writer_results
import dask.config
import time
reduce = True
with dask.config.set(scheduler=CustomScheduler(max_computes=0)):
    sc = Scene(filenames={"seviri_l1b_hrit": seviri_files})
    nms = {"ash", "natural_color", "dwd_RGB_12_12_9i_N", "convection",
           "dwd_IR_108", "dwd_HRV", "airmass", "dust"}
    sc.load(nms, generate=False)
    tic = time.time()
    ls = sc.resample("nqeuro3km",
                     resampler="gradient_search", reduce_data=reduce)
    print("resample", time.time() - tic, "seconds")
    res_all = []
    for nm in nms:
        res_all.append(ls.save_dataset(
                nm,
                "{platform_name}-{sensor}-{name}-{area.area_id}-{start_time:%Y%m%d%H%M}-geotiff.tif",
                writer="ninjogeotiff",
                ChannelID="IR 10.8",
                DataType="GORN",
                PhysicValue="Temperature",
                PhysicUnit="K",
                SatelliteNameID="METEOSAT Europe-Atlantic",
                fill_value=0,
                #dynamic_fields={"mean_time"},
                compute=False))
```

Satpy main: 8.2 seconds

This PR: 44.0 seconds

gerritholl:
Profiling reveals:

* the time per call for `_reduce_data` increases from 1.02 to 5.97 seconds.
* The number of calls to `spherical.py:492(get_next_intersection)` increases from 302 to 1812.
* The number of calls to `geometry.py:2654(get_area_slices)` (first entry point into pyresample) increases from 2 to 12.

Here both profiles:

[profile-results.zip](https://github.com/user-attachments/files/29285455/profile-results.zip)

satpy/scene.py
```diff
@@ -927,32 +963,40 @@ def _prepare_resampler(self, source_area, destination_area, resamplers, resample
...
+            source_area = source_area[slice_y, slice_x]
+            reductions[source_area] = (slice_x, slice_y), source_area
+            dataset = self._slice_data(source_area, (slice_x, slice_y), dataset)
+
         return dataset, source_area
 
+    @staticmethod
+    def _get_source_dest_slices(source_area, destination_area, reductions, resample_kwargs):
+        try:
+            (slice_x, slice_y), source_area = reductions[source_area]
```
gerritholl, line 987: Something is going wrong with the reductions dictionary.  With satpy main, my test scripts goes twice into the `except KeyError` block, but the remaining 10 times it finds the reductions in the dictionary.  With this PR, it never finds an entry in the reductions dictionary, going 12 times into the except block.

satpy/scene.py
```diff
@@ -66,6 +67,19 @@ class DelayedGeneration(KeyError):
...
+class _ResamplingContext:
+    """Holds configuration and mutable state for a single resampling operation."""
+
+    new_scn: Any
+    destination_area: Any
+    reduce_data: bool
+    resample_kwargs: dict
+    resampled_datasets: dict = dataclasses.field(default_factory=dict)
+    resamplers: dict = dataclasses.field(default_factory=dict)
+    reductions: dict = dataclasses.field(default_factory=dict)
```
gerritholl, line 80: I wonder if moving this into a dataclass breaks something?  In any case, the reductions dictionary has stopped working.  I haven't looked deeper yet.

gerritholl (changes requested):
Serious performance regression in `_reduce_data`.  The `reductions` dictionary appears no longer functional, leading to `_reduce_data` being called for every dataset.  For my test script, that increases the time for calling `resample` from ca. 8 seconds to ca. 40 seconds.  Without computing anything.

satpy/scene.py
```diff
@@ -927,32 +963,40 @@ def _prepare_resampler(self, source_area, destination_area, resamplers, resample
...
-                    reductions[key] = (slice_x, slice_y), source_area
-                dataset = self._slice_data(source_area, (slice_x, slice_y), dataset)
-            else:
-                LOG.debug("Data reduction disabled by the user")
+            slice_x, slice_y = self._get_source_dest_slices(source_area, destination_area, reductions, resample_kwargs)
         except NotImplementedError:
             LOG.info("Not reducing data before resampling.")
+        else:
+            source_area = source_area[slice_y, slice_x]
+            reductions[source_area] = (slice_x, slice_y), source_area
```
gerritholl: This is wrong.  The reductions dictionary should contain `source_area`, not `source_area[slice_y, slice_x]`.  Previously this was retained in `key = source_area` but this retaining got destroyed in the refactoring.

satpy/scene.py
```diff
@@ -927,32 +963,40 @@ def _prepare_resampler(self, source_area, destination_area, resamplers, resample
...
+            dataset = self._slice_data(source_area, (slice_x, slice_y), dataset)
```
gerritholl, line 980: This would seem to fix the performance regression:

```suggestion
        else:
            orig_source_area = source_area
            source_area = source_area[slice_y, slice_x]
            reductions[orig_source_area] = (slice_x, slice_y), source_area
            dataset = self._slice_data(source_area, (slice_x, slice_y), dataset)
```

satpy/scene.py
gerritholl: No, that was not the problem.

satpy/scene.py
gerritholl: Fixed in f70d80f2c

gerritholl (approved):

satpy/scene.py
gerritholl: Fixed in 70d80f

pnuu:
Claude agrees the changes are good, but also found a minor regression in caching where the reduced source area is discarded and rebuilt every time for each dataset. Fix incoming.

pnuu (commented):
The latest cache fix looks good to me.

pnuu:
What Dave and Dave's Claude did in https://github.com/pytroll/satpy/pull/3477 is far cleaner, so I'm closing this.

PR #3179, opened 2025-07-28
Fix some warnings produced from running the mitiff writer

djhoese:
In an effort to reduce the number of warnings produced by running satpy tests, I've refactored or rewritten the mitiff tests. The major changes are:

1. Switch from unittest to pytest.
2. Change most test data to `arange` versus `zeros` so that the dynamic linear stretching actually produces an image with data rather than it getting confused by a flat image (no range).
3. Explicitly convert NaNs to 0 with `numpy.nan_to_num`. (@TAlonglong thoughts?)
4. Refactor tests to move some shared logic to helper functions.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/writers/core/image.py
+++ b/satpy/writers/core/image.py
@@ -47,7 +47,7 @@ class ImageWriter(Writer):
 
                     {platform_name}_{sensor}_{name}_{start_time:%Y%m%d_%H%M%S}.tif
 
-            base_dir (str):
+            base_dir (str | Path):
                 Base destination directories for all created files.
             enhance (bool or Enhancer): Whether to automatically enhance
                 data to be more visually useful and to fit inside the file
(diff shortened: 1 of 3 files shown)


satpy/writers/core/image.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, name=None, filename=None, base_dir=None, enhance=None, **kwar
 
                     {platform_name}_{sensor}_{name}_{start_time:%Y%m%d_%H%M%S}.tif
 
-            base_dir (str):
```
djhoese, line 50: This was an attempt to make my IDE less dumb about creating writers where it kept complaining that `base_dir` should be a `dict` or something...it didn't work, but this change in the type annotation is accurate from what I can tell.

TAlonglong (approved):
No problems with this as I see it.

PR #3180, opened 2025-07-28
Fix all fiduceo warnings during tests

djhoese:
Major changes:

1. Redefined how/where the expected xarray warning about duplicate warnings happens. The global filter previously implemented was not working.
2. Refactored one of the fixtures in the tests to pass a callable version of the file handler so that we didn't have to use an un-registered pytest mark (produces a warning).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/mviri_l1b_fiduceo_nc.py
+++ b/satpy/readers/mviri_l1b_fiduceo_nc.py
@@ -153,6 +153,7 @@ References:
 """
 
 import abc
+import contextlib
 import functools
 import warnings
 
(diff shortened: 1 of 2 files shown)


djhoese:
Ah sorry @mraspaud, github said you edited the files recently but it was probably just from your `ruff` refactoring PR.

djhoese:
I might merge this later today since Stephan just went on vacation and I think Gerrit said he didn't do much with the reader itself so he may not have much of an opinion. If anything is controversial I can update it in a later PR.

PR #3176, opened 2025-07-25
Add new QA band filename support to Landsat reader

simonreise:
In the most of the new Landsat products the qa band have a suffix QA_PIXEL.tif instead of just QA.tif

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 
 - [ ] Add your name to `AUTHORS.md` if not there already

PR #3181, opened 2025-07-30
Added more Landsat readers

simonreise:
# Landsat readers

## Overview

This PR adds readers for Landsat OLI-TIRS L2, ETM+ L1-2, TM L1-2 and MSS products.

All these readers are mostly based on the existing Landsat OLI-TIRS L1 reader.

## Readers

Landsat reader classes are hierarchially organized into a nested structure of classes where

- `BaseLandsatReader` class contains `__init__` and `get_dataset` functions. Their logic is mostly copied from the original reader, with few additions and bug fixes.

- Landsat readers for L1 and L2 classes contain calibration logic, which is the same for every sensor, but differs for L1 and L2 products. L1 logic is copied from the existing reader. 

**Note**: I moved the \* 0.001 scaling logic for `TRAD`, `URAD`, `CDIST` etc bands to calibration. Maybe it belongs to `get_dataset` function, just like `ang` bands \* 0.01 scaling in L1 reader? Or should I move `ang` bands \*0.01 scaling to calibrations also?

- The product readers. They mostly contain attributes like spectral and thermal band list and sensor name. But `MSSCHReader` also contains the `available_datasets` functions that dynamically sets B4 wavelength (see yamls section for more info)

## YAMLs

I created YAML files for every reader.

They are all based on the OLI-TIRS L1 YAML file.

What was updated or needs additional checking:

- Reader namings. Should etm+ reader be named "etm_lx_tif" or "etm_plus_lx_tif"?

- `data_identification_keys` section was added to the head of the file to add custom calibration support to L2 products. Is that necessary? Should custom calibrations be implemented in any other way?

- `{collection_category}` was limited to `{collection_category:2s}` because otherwise L2 product and Landsat-7 (which also contains GM bands) search was faulty. AFAIK collection category can contain only 2 symbols, but if not, it can cause errors.

- QA band namings: see #3176

- calibration `standard_names` and `units` probably should be double-checked

- `wavelength` was added or updated according to official Landsat Data Format Control Books. But probably also should be double-checked.

- Again, custom calibrations at band definitions at `datasets` section. Are they added the right way?

- `MSS` bands. In Landsat 1/2/3 MSS products bands are named B4, B5, B6, B7 and in Landsat 4/5 MSS products the same bands are named B1, B2, B3, B4. Now MSS reader just read the bands and their names as-is. The issue with `B4` band being `green` in Landsat 1/2/3 and `nir` in Landsat 4/5 is solved by changing `wavelength` in `available_datasets` function in the reader. But is it the right way to handle such issues? Maybe there is a better way to solve that problem? Or maybe separate readers for Landsat 1/2/3 and Landsat 4/5 MSS products should be added?

- Also, some docs say that Landsat-3 had a thermal band B8, but it failed just after the launch. I am not sure, if any products with B8 actually exist and if it should be added to band list.

## Tests

I added separate test files for every reader. They are mostly just the adapted versions of the OLI-TIRS L1 test file. The differences are:

- L2 product test files test `TRAD` file instead of `sza` file

- Products other than OLI-TIRS do not have `test_loading_badchan` test, because they do not have products with only spectral or only thermal bands, so they can not face such errors

- MSS test tests both Landsat-1 and Landsat-4 products and tests if B4 wavelength is set correctly

- Other small differences

Maybe any other tests should be added?

### DIFF
--- a/satpy/etc/readers/etm_l1_tif.yaml
+++ b/satpy/etc/readers/etm_l1_tif.yaml
@@ -1,11 +1,11 @@
 reader:
-  name: oli_tirs_l1_tif
-  short_name: OLI/TIRS L1 GeoTIFF
-  long_name: Landsat-8/9 OLI/TIRS L1 data in GeoTIFF format.
-  description: GeoTIFF reader for Landsat-8/9 OLI/TIRS L1 data.
+  name: etm_l1_tif
+  short_name: ETM+ L1 GeoTIFF
+  long_name: Landsat-7 ETM+ L1 data in GeoTIFF format.
+  description: GeoTIFF reader for Landsat-7 ETM+ L1 data.
   status: Beta
   supports_fsspec: true
-  sensors: oli_tirs
+  sensors: etm+
   default_channels: []
   reader: !!python/name:satpy.readers.core.yaml_reader.FileYAMLReader
   data_identification_keys:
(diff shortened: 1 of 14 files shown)


simon-sat:
I don't have time for a full review of this until some time in September, as I'm on annual leave and will then be catching up on work for some time thereafter.

But one quick comment, the file `oli_tirs_l1_tif.py` (as well as the associated tests) should be renamed, it's no longer a reader for OLI/TIRS only. Something like `landsat_base.py` might be better...

mraspaud (commented):
That’s a big PR, but thanks for putting it together!
I haven’t checked all the details, but I was wondering if it was possible somehow to refactor the tests, as it seems there is quite some duplication. Maybe using "pytest.mark.parametrize" could help? But as I said, I haven’t checked the details, so maybe my suggestion doesn’t make sense...

simonreise:
I tried to merge all the tests into a single file `test_landsat.py`. 

I used `pytest.mark.parametrize`, just as you suggested, and it really helped to get rid of duplicate code, but the final test file is very large and complicated, and parametrize blocks before functions are sometimes lareger than the function itself due to large number of parameters. 

So I decided to keep the original test files for now and delete them later, because I am not sure if one single test file or separate test files for each sensor is preferrable in this case.

Let's discuss which test implementation will be more suitable in this case! Also, probably tests could be optimized better (maybe somehow getting rid of similar fixtures), or maybe more tests should be added (e. g. to test bands that appear only in specific products, like GM bands for ETM+)

satpy/readers/landsat_base.py
djhoese, line 1: In a recent large refactor PR for readers we made the decision to put "base" reader modules in `satpy/readers/core/` and remove the `_base` portion of the name. Could you move this to `satpy/readers/core/landsat.py`?

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+
+        # OLI-TIRS sensor data sometimes can contain only OLI or only TIRS data
+        if self.sensor == "OLI_TIRS" and key["name"] in self.spectral_bands and self.chan_selector not in ["O", "C"]:
+            raise ValueError(f"Requested channel {key['name']} is not available in this granule")
+        if self.sensor == "OLI_TIRS" and key["name"] in self.thermal_bands and self.chan_selector not in ["T", "C"]:
+            raise ValueError(f"Requested channel {key['name']} is not available in this granule")
+
+        logger.debug("Reading %s.", key["name"])
+
+        data = xr.open_dataarray(open_file_or_filename(self.filename), engine="rasterio",
```
djhoese: Could you add an import to the top of the module for `rioxarray` to make sure users get an accurate error as early as possible. See for example:

https://github.com/pytroll/satpy/blob/d86b21ac3f4c43148dcc9ccacec82564cc61ca00/satpy/readers/generic_image.py#L34

You could also copy the comment from that module that's above `open_dataset`:

https://github.com/pytroll/satpy/blob/d86b21ac3f4c43148dcc9ccacec82564cc61ca00/satpy/readers/generic_image.py#L80-L82

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+                                 chunks={"band": 1,
+                                         "y": "auto",
+                                         "x": "auto"},
+                                 mask_and_scale=False).squeeze()
+
+        # For calibration simplicity convert the fill value to np.nan
+        bands_9999 = ["TRAD", "DRAD", "URAD", "ATRAN", "EMIS", "EMSD", "CDIST", "qa_st", "qa_atmos_opacity"]
+        if self.collection == "02" and key["name"] in bands_9999:
+            # The fill value for several Landsat L2 bands is '-9999'
+            data.data = xr.where(data.data == -9999, np.float32(np.nan), data.data)
```
djhoese: Should these be `da.where`? You're using `data.data` which is a dask array, right?

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        """Get area definition of the image from the metadata."""
+        return self._mda.build_area_def(dsid["name"])
+
+
+class BaseLandsatL1Reader(BaseLandsatReader):
+    """Basic file handler for Landsat L1 files (tif)."""
+
+    def __init__(self, filename, filename_info, filetype_info, mda, **kwargs):
+        """Initialize the reader."""
+        super().__init__(filename, filename_info, filetype_info, mda, **kwargs)
```
djhoese: This method can be removed, right?

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+            data.data = data.data.astype(np.float32)
+            return data
+
+
+class BaseLandsatL2Reader(BaseLandsatReader):
+    """Basic file handler for Landsat L2 files (tif)."""
+
+    def __init__(self, filename, filename_info, filetype_info, mda, **kwargs):
+        """Initialize the reader."""
+        super().__init__(filename, filename_info, filetype_info, mda, **kwargs)
```
djhoese: Same here.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        return ["B10", "B11"]
+
+    @property
+    def sensor(self):
+        """Sensor name."""
+        return "OLI_TIRS"
+
+    def __init__(self, filename, filename_info, filetype_info, mda, **kwargs):
+        """Initialize the reader."""
+        super().__init__(filename, filename_info, filetype_info, mda, **kwargs)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        return ["B10"]
+
+    @property
+    def sensor(self):
+        """Sensor name."""
+        return "OLI_TIRS"
+
+    def __init__(self, filename, filename_info, filetype_info, mda, **kwargs):
+        """Initialize the reader."""
+        super().__init__(filename, filename_info, filetype_info, mda, **kwargs)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        return ["B6_VCID_1", "B6_VCID_2"]
+
+    @property
+    def sensor(self):
+        """Sensor name."""
+        return "ETM+"
+
+    def __init__(self, filename, filename_info, filetype_info, mda, **kwargs):
+        """Initialize the reader."""
+        super().__init__(filename, filename_info, filetype_info, mda, **kwargs)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        return ["B6"]
+
+    @property
+    def sensor(self):
+        """Sensor name."""
+        return "ETM+"
+
+    def __init__(self, filename, filename_info, filetype_info, mda, **kwargs):
+        """Initialize the reader."""
+        super().__init__(filename, filename_info, filetype_info, mda, **kwargs)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        return ["B6"]
+
+    @property
+    def sensor(self):
+        """Sensor name."""
+        return "TM"
+
+    def __init__(self, filename, filename_info, filetype_info, mda, **kwargs):
+        """Initialize the reader."""
+        super().__init__(filename, filename_info, filetype_info, mda, **kwargs)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        return []
+
+    @property
+    def sensor(self):
+        """Sensor name."""
+        return "MSS"
+
+    def __init__(self, filename, filename_info, filetype_info, mda, **kwargs):
+        """Initialize the reader."""
+        super().__init__(filename, filename_info, filetype_info, mda, **kwargs)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        ext_p3 = float(self.root.find(".//PROJECTION_ATTRIBUTES/CORNER_LR_PROJECTION_X_PRODUCT").text) + pixoff
+        ext_p4 = float(self.root.find(".//PROJECTION_ATTRIBUTES/CORNER_UL_PROJECTION_Y_PRODUCT").text) + pixoff
+
+        # Create area definition
+        pcs_id = f"{datum} / UTM zone {utm_str}"
+        proj4_dict = {"proj": "utm", "zone": utm_zone, "datum": datum, "units": "m", "no_defs": None, "type": "crs"}
+        area_extent = (ext_p1, ext_p2, ext_p3, ext_p4)
+
+        # Return the area extent
+        return AreaDefinition("geotiff_area", pcs_id, pcs_id, proj4_dict, x_size, y_size, area_extent)
```
djhoese: I'm wondering if there a more useful name for the area definition here? Like naming it after the UTM maybe? Like "UTM{utm_zone}"? Or is that unhelpful?

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+
+
+class BaseLandsatMDL1Reader(BaseLandsatMDReader):
+    """Metadata file handler for Landsat L1 files (tif)."""
+
+    def __init__(self, filename, filename_info, filetype_info):
+        """Init the reader."""
+        super().__init__(filename, filename_info, filetype_info)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        bt_k2 = float(self.root.find(f".//LEVEL1_THERMAL_CONSTANTS/K2_CONSTANT_BAND_{band}").text)
+        return rad_gain, rad_add, bt_k1, bt_k2
+
+
+class BaseLandsatMDL2Reader(BaseLandsatMDReader):
+    """Metadata file handler for Landsat L2 files (tif)."""
+
+    def __init__(self, filename, filename_info, filetype_info):
+        """Init the reader."""
+        super().__init__(filename, filename_info, filetype_info)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+        bt_add = float(self.root.find(f".//LEVEL2_SURFACE_TEMPERATURE_PARAMETERS/TEMPERATURE_ADD_BAND_{band}").text)
+        return bt_gain, bt_add
+
+
+class OLITIRSMDReader(BaseLandsatMDL1Reader):
+    """File handler for Landsat OLI-TIRS L1 files (tif)."""
+
+    def __init__(self, filename, filename_info, filetype_info):
+        """Init the reader."""
+        super().__init__(filename, filename_info, filetype_info)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+
+        return bdict
+
+
+class OLITIRSL2MDReader(BaseLandsatMDL2Reader):
+    """File handler for Landsat OLI-TIRS L2 files (tif)."""
+
+    def __init__(self, filename, filename_info, filetype_info):
+        """Init the reader."""
+        super().__init__(filename, filename_info, filetype_info)
```
djhoese: Remove.

satpy/readers/landsat_base.py
```diff
@@ -0,0 +1,815 @@
...
+
+        return bdict
+
+
+class ETMMDReader(BaseLandsatMDL1Reader):
+    """File handler for Landsat ETM+ L1 files (tif)."""
+
+    def __init__(self, filename, filename_info, filetype_info):
+        """Init the reader."""
+        super().__init__(filename, filename_info, filetype_info)
```
djhoese: Remove.

satpy/readers/landsat_base.py
simonreise: I am not sure if this name is visible to user, but giving a more meaningful name definitely would be a good addition. Added it

satpy/readers/landsat_base.py
simonreise: Done

satpy/readers/landsat_base.py
simonreise: Done

satpy/readers/landsat_base.py
simonreise: Renamed and moved the file. Is it in the right place now?

simonreise:
I refactored `MDReaders`. Now `band_calibration` and `band_saturation` automatically read all the available parameters, so I completely got rid of product-level `MDReader` classes, leaving only L1 and L2 classes.

We could probably also keep only one `MDReader` class by letting `band_calibration` use `self.process_level`, but it won't simplify the code, because we will have to keep all the coef keys anyway:

Now:
```py
class LandsatL1MDReader(BaseLandsatMDReader):
    """Metadata file handler for Landsat L1 files (tif)."""

    @property
    def band_calibration(self):
        """Return per-band calibration parameters."""
        radcal = self.get_cal_params("LEVEL1_RADIOMETRIC_RESCALING", "RADIANCE_MULT", "RADIANCE_ADD")
        viscal = self.get_cal_params("LEVEL1_RADIOMETRIC_RESCALING", "REFLECTANCE_MULT", "REFLECTANCE_ADD")
        tircal = self.get_cal_params("LEVEL1_THERMAL_CONSTANTS", "K1_CONSTANT", "K2_CONSTANT")
        topcal = viscal | tircal
        return {key: tuple([*radcal[key], *topcal[key]]) for key in radcal}


class LandsatL2MDReader(BaseLandsatMDReader):
    """Metadata file handler for Landsat L2 files (tif)."""

    @property
    def band_calibration(self):
        """Return per-band calibration parameters."""
        viscal = self.get_cal_params("LEVEL2_SURFACE_REFLECTANCE_PARAMETERS", "REFLECTANCE_MULT", "REFLECTANCE_ADD")
        tircal = self.get_cal_params("LEVEL2_SURFACE_TEMPERATURE_PARAMETERS", "TEMPERATURE_MULT", "TEMPERATURE_ADD")
        return viscal | tircal

```

If merged to one class:
```py
@property
def band_calibration(self):
    """Return per-band calibration parameters."""
    if "1" in self.process_level:
        radcal = self.get_cal_params("LEVEL1_RADIOMETRIC_RESCALING", "RADIANCE_MULT", "RADIANCE_ADD")
        viscal = self.get_cal_params("LEVEL1_RADIOMETRIC_RESCALING", "REFLECTANCE_MULT", "REFLECTANCE_ADD")
        tircal = self.get_cal_params("LEVEL1_THERMAL_CONSTANTS", "K1_CONSTANT", "K2_CONSTANT")
        topcal = viscal | tircal
        return {key: tuple([*radcal[key], *topcal[key]]) for key in radcal}
    elif "2" in self.process_level:
        viscal = self.get_cal_params("LEVEL2_SURFACE_REFLECTANCE_PARAMETERS", "REFLECTANCE_MULT", "REFLECTANCE_ADD")
        tircal = self.get_cal_params("LEVEL2_SURFACE_TEMPERATURE_PARAMETERS", "TEMPERATURE_MULT", "TEMPERATURE_ADD")
        return viscal | tircal
    raise ValueError("blahblahblah")

```

simonreise:
Improved `build_area_def` function: now it creates area definition using EPSG code instead of WKT string and now it can handle Arctic and Antarctic scenes that have specific projections. I also added a test for an Antarctic scene

satpy/readers/landsat_base.py
djhoese: Yep, that looks right. Thanks.

FYI I need to finish up some other stuff, but I hope to re-review this and merge it by the end of my day today.

satpy/readers/landsat_base.py
djhoese: Hm it is visible via `scn[product_name].attrs["area"].area_id`. The "EPSG: XXXX" name may not be super filename friendly. This is a selfish request on my part because I have software that puts the Area ID into the default filename for geotiffs.

satpy/readers/core/landsat.py
```diff
@@ -0,0 +1,581 @@
...
+        ext_p1 = float(self.root.find(".//PROJECTION_ATTRIBUTES/CORNER_UL_PROJECTION_X_PRODUCT").text) - pixoff
+        ext_p2 = float(self.root.find(".//PROJECTION_ATTRIBUTES/CORNER_LR_PROJECTION_Y_PRODUCT").text) - pixoff
+        ext_p3 = float(self.root.find(".//PROJECTION_ATTRIBUTES/CORNER_LR_PROJECTION_X_PRODUCT").text) + pixoff
+        ext_p4 = float(self.root.find(".//PROJECTION_ATTRIBUTES/CORNER_UL_PROJECTION_Y_PRODUCT").text) + pixoff
+
+        # Create area definition
+        area_extent = (ext_p1, ext_p2, ext_p3, ext_p4)
+
+        # Return the area extent
+        return AreaDefinition(f"EPSG: {proj_code}", pcs_id, pcs_id, proj_code, x_size, y_size, area_extent)
```
djhoese: @mraspaud @pnuu and anyone else who might care, what do you think about the name of the area *being* the EPSG code. I asked to make it more descriptive but now using the EPSG code in this way seems wrong. I'm not sure what a better name would be and I don't know enough about landsat to suggest something else.

pnuu: I have all my production areas named like `epsg_3035_1km`, so in principle I'm fine with that.

satpy/tests/reader_tests/test_landsat.py
```diff
@@ -0,0 +1,3784 @@
...
+    filename = antarctic_files_path / "LC08_L1GT_132122_20220301_20220308_02_T2_MTL.xml"
+    with open(filename, "wb") as f:
+        f.write(antarctic_metadata_text)
+    return os.fspath(filename)
+
+
+class TestLandsat:
+    """Test Landsat image readers."""
+
+    @pytest.fixture(autouse=True)
```
djhoese: This decorator shouldn't be needed. A `setup_method` is always called for each test method.

satpy/tests/reader_tests/test_landsat.py
```diff
@@ -0,0 +1,3784 @@
...
+    return os.fspath(filename)
+
+
+@pytest.fixture(scope="session")
+def mss_l1_landsat4_all_files(mss_l1_landsat4_b4_file, mss_l1_landsat4_mda_file):
+    """Return all the files."""
+    return mss_l1_landsat4_b4_file, mss_l1_landsat4_mda_file
+
+
+@pytest.fixture(scope="session")
```
djhoese: Are these fixtures used in the other test modules? The scope for all of these should be "module". If these are used in the other landsat test modules then how about:

1. Create a `satpy/tests/reader_tests/landsat_tests/` directory.
2. Move these fixtures to `landsat_tests/conftest.py` module and set scope to "module".
3. Move all the landsat tests to `landsat_tests/` directory.
4. Use the fixtures as usual. Pytest will automatically load them due to their proximity to `conftest.py`.

And yes, they will be recreated for each test module, but in my opinion that's better than them existing for all tests for the whole session.

satpy/readers/core/landsat.py
simonreise: The problem with the "UTM{utm_zone}" name is that Arctic and Antarctic scenes do not have a UTM zone, so I decided to use EPSG code instead. I used it only because that was the first idea I came with, and if there is a more appropriate area definition name, we probably should use it. 

If f"EPSG: {proj_code}" is bad mostly because it is not filename friendly, I can just replace it to f"epsg_{proj_code}"

If "UTM{utm_zone}" approach is preferrable, I can do something like that:
```py
# Reading utm zone or get specific crs for arctic and antarctic
if self.root.find(".//PROJECTION_ATTRIBUTES/UTM_ZONE") is not None:
    utm_zone = self.root.find(".//PROJECTION_ATTRIBUTES/UTM_ZONE").text
    pcs_id = f"{datum} / UTM zone {utm_zone}N"
    proj_code = f"EPSG:326{utm_zone.zfill(2)}"
    name = f"UTM{utm_zone}"
else:
    lat_ts = self.root.find(".//PROJECTION_ATTRIBUTES/TRUE_SCALE_LAT").text
    if lat_ts == "-71.00000":
        # Antarctic
        proj_code = "EPSG:3031"
    if lat_ts == "71.00000":
        # Arctic
        proj_code = "EPSG:3995"
    pcs_id = f"{datum} / EPSG: {proj_code[5:]}N"
    name = f"EPSG_{proj_code[5:]}"
return AreaDefinition(name, ...)
```

Btw, I adapted the code for the Arctic and Antarctic scenes from [here](https://pystac.readthedocs.io/en/stable/tutorials/creating-a-landsat-stac.html#Add-projection-information)

satpy/tests/reader_tests/test_landsat.py
simonreise: Yes, you are right, they are not used in the other test modules and their scope should be "module".

Also, there are two approaches to implement the same tests in the PR right now.

The first one is one test file per sensor (`test_oli_tirs_l1_tif.py` etc.). They are relatively small, and not so complicated, but they contain lot of duplicate code, which is probably not good.

The second one is one large test file for all the Landsat products `test_landsat.py`. This file is really large, complicated and sometimes hard to understand, but it contains almost no code duplication due to use of `pytest.mark.parametrize`.

The tests in both approaches are the same, so we should decide which approach is more appropriate to use and remove the another one.

What do you think, should we use one test file or several test files? Or should we refactor the tests somehow?

satpy/tests/reader_tests/test_landsat.py
simonreise: I am not so good at using pytest, and if I just remove the decorator, the tests crash with error `ERROR satpy/tests/reader_tests/test_landsat.py::TestLandsat - Failed: In test_basicload: function uses no argument 'date'`. 

Now I pass `date`, `level_correction`, `spacecraft` and `data_type` parameters to `setup_method` (as they differ from test to test!) via setting them in `pytest.mark.parametrize` section of each test, and `setup_method` then handles them, but if the `@pytest.fixture(autouse=True)` decorator is removed, pytest fails to pass these parameters from `parametrize` block to `setup_method`

I am not sure if I do it right, maybe there is a less complicated and more reliable way

djhoese: So `setup_method` is a special method in pytest so that's what I thought you were going for. See:

https://docs.pytest.org/en/4.6.x/xunit_setup.html#method-and-function-level-setup-teardown

If you are using this as method to create test data, then you're right this would need to be a fixture. In that case though you should rename the method.

In general, your use of parameters that magically get passed to an autouse fixture is a little too magical for me and that's a bad sign since I'm one of the more fixture-friendly maintainers of Satpy. Overall I'm not sure I see the benefit of this being a fixture. I don't think it is too much to have a function call at the start of every test to get `filename_info`. And `ftype_info` is so small...oh it's static. You could make it an attribute on the test class:

```python
class MyTestClass:
    ftype_info = {"file_type": "granule_B4"}
```

and then still use `self.ftype_info`.

satpy/tests/reader_tests/test_landsat.py
djhoese: Overall I like combining with parametrize. I'm a little surprised how much fits into a single file with only a few `if` statements for specific checks. Normally I'd say parametrize as much as possible for a single reader but each reader gets its own module. However, this seems like a good exception to that "rule". You could also probably get a similar effect by using base classes, but the parametrize version really isn't too bad in my opinion.

I like the all-in-one file.

@mraspaud any strong opinions now that you've seen the end result?

satpy/readers/core/landsat.py
djhoese: eh let's just leave it the way it is. If it ends up mattering in the future I'll make an issue or a PR.

satpy/tests/reader_tests/test_landsat.py
mraspaud: No strong opinion here, so let’s go for the big file.

satpy/tests/reader_tests/test_landsat.py
simonreise: Ok, I moved setting up `filename_info` into functions that require it. It also allowed to simplify the `parametrize` blocks of the functions that not use it by removing unnecessary parameters

djhoese:
Ok we're getting really close to a merge now. The biggest issue I have now is how much CodeScene (the check on this PR) dislikes this implementation. It makes good points about there being a lot of data declarations and there being a lot of function arguments. On the other hand these are tests and you're parametrizing them so you can kind of expect a lot of arguments.

I thought about refactoring the tests for you but ran out of time today. I had the thought that maybe it would be nice to have a base class (ex. `LandsatTestBase`) that defines these test methods and then the subclasses could have the reader and other reader-specific stuff set in class attributes. However, I'm realizing now that doesn't help you with the parametrization that you'd need to do for each reader. :thinking: 

@mraspaud @pnuu how do you feel about just leaving CodeScene mad and merging this implementation with a single test file?

Other consideration: Move the metadata definitions in the tests to text files in the tests directory that are loaded when they are needed (a fixture if it makes sense).

pnuu:
I'll see if I have some time tomorrow to have a look at the complaints. I think at least the ones in `landsat.py` would be nice to clean before merging. With a quick glance adding few helper methods/functions would solve them quite easily.

pnuu:
Pushed the first round of refactorings, lets see what CodeScene thinks about them.

pnuu:
Ok, `landsat.py` should now be ok. There's not much that we can do to the number of arguments in the `__init__()` at the momenta.

pnuu:
With https://github.com/pytroll/satpy/pull/3181/commits/31a6f2353cd0250177d10d981cbae97023d050cd the metadata are moved from the test module to separate text files.

pnuu:
Complex test conditionals refactored in https://github.com/pytroll/satpy/pull/3181/commits/63bbc0add9f6dd6999de423a90d48b95d0ecea72

pnuu:
@mraspaud asked me to rename the metadata files as `.xml`, done in https://github.com/pytroll/satpy/pull/3181/commits/3ea946c526125fdc7490b8c52a72e222425a4f11

djhoese:
Ok I've reduced the number of arguments by quite a bit, but they are still over the threshold for CodeScene. The main changes are:

1. Call `get_filename_info` in the parametrize arguments (4 arguments -> 1)
2. Use lazy fixtures instead of `request.`. We already depend on this pytest plugin and it means we can get rid of the usage of `request`. Reduces the number of arguments by 1. FYI @simonreise this might be a nice thing to keep in mind in your other projects that may use pytest.

I could combine the filename fixtures with the filename_info since the two are typically together but it wouldn't reduce the number of arguments enough for most of the test methods. The next option would be to do my subclassing idea, but I'm not sure I want to put that much work into this.

simonreise:
I thought about @djhoese 's suggestion about per-reader test classes, and maybe in this case we could get rid from parametrization at all, as in the current implementation every parameter set usually corresponds to a single reader? So we could probably do something like this:
```py
class BaseLandsatTest:
    def _basicload(remote):
        if remote:
            all_files = convert_to_fsfile(self.all_files)
        else:
            all_files = self.all_files

        scn = Scene(reader=self.reader, filenames=all_files)
        if thermal_name is not None:
            scn.load([self.spectral_name, self.thermal_name])
        else:
            scn.load([self.spectral_name])

        self._check_basicload_thermal(scn)
        self._check_basicload_mss_l1_tif(scn)
   
    ...
    
    def _ch_startend(self):
        """Test correct retrieval of start/end times."""
        scn = Scene(reader=self.reader, filenames=[self.spectral_file, self.mda_file])
        bnds = scn.available_dataset_names()
        assert bnds == [self.spectral_name]

        scn.load(["B4"])
        assert scn.start_time == self.date_time
        assert scn.end_time == self.date_time

class TestLandsatOLITIRSL1(BaseLandsatTest):
    reader = "oli_tirs_l1_tif"
    spectral_name = "B4"
    thermal_name = "B11"
    all_files = lf("oli_tirs_l1_all_files")
    ...
    
    @pytest.mark.parametrize("remote", [True, False])
    def test_basicload(self, remote):
        self._basicload(remote)
    
    def test_ch_startend(self):
        self._ch_startend()
```

We could use `self.` everywhere we can, so number of parameters is reduced to minimum. We could also use nested class structure just like in the Landsat class itself, it would help us to get rid of complex if-else blocks.

Should I implement it this way?

djhoese:
@simonreise should we maybe merge this and then do a separate refactor PR? I'm actually not sure why we haven't merged this already.

simonreise:
Shouldn't @ simonrp84 also review this? He said he will review the PR in September after his summer vacation in [one of the first messages in the thread](https://github.com/pytroll/satpy/pull/3181#issuecomment-3144595166).

simonreise:
@simonrp84 [here](https://github.com/pytroll/satpy/pull/3181#issuecomment-3144595166) you mentioned that you might review this PR

> I don't have time for a full review of this until some time in September, as I'm on annual leave and will then be catching up on work for some time thereafter.
> 
> But one quick comment, the file `oli_tirs_l1_tif.py` (as well as the associated tests) should be renamed, it's no longer a reader for OLI/TIRS only. Something like `landsat_base.py` might be better...

Are you going to review it, is it needed? If not, I think it can be merged

mraspaud:
I'm good with merging this as is.

PR #3186, opened 2025-08-01
Optimize AWIPS Tiled writer and remove use of dask delayed

djhoese:
This is a major rewrite of how the AWIPS tiled (`awips_tiled`). Previously I had a lot of hacks to workaround xarray or dask hanging. Recent advise from dask maintainers is to not passing Array objects to Delayed functions so in this PR I removed that in favor of a call to `blockwise`. I also removed (hopefully) various hacks that won't be needed any longer. I also caught that dask was computing things multiple times with various solutions I tried to implement and ended up filing a bug report with dask on that, but for some reason this particular implementation in this PR does not suffer from that.

This PR also ignores some expected warnings from xarray about saving coordinate variables to a NetCDF file without a `_FillValue`. This is CF standard so expected.

I still have some type annotations and other things I'd like to add but this is ready for review.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/writers/awips_tiled.py
+++ b/satpy/writers/awips_tiled.py
@@ -1162,7 +1162,7 @@ def _copy_to_existing(dataset_to_save, output_filename):
     #   only sometimes. Limiting dask to 1 worker seems to fix this.
     #   I (David Hoese) was unable to make a script that reproduces this
     #   without using this writer (makes it difficult to file a bug report).
-    existing_dataset = xr.open_dataset(output_filename)
+    existing_dataset = xr.open_dataset(output_filename, cache=False)
     # the below used to trick xarray into working, but this doesn't work
     # in newer versions. This was a hack in the first place so I'm keeping it
     # here for reference.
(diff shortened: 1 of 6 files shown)


satpy/writers/awips_tiled.py
```diff
@@ -1627,71 +1592,56 @@ def save_datasets(self, datasets, sector_id=None,  # noqa: D417
...
+                all_coords=[data_arr.coords for data_arr in data_arrs],
+                template=template, area_def=area_def,
+                tile_info=tile_info, sector_id=sector_id,
+                creation_time=creation_time,
+                shared_attrs=ds_info,
+                extra_global_attrs=extra_global_attrs,
+                compress=self.compress,
+                check_categories=check_categories,
+                output_filename=output_filename,
+            )
```
djhoese: @mraspaud and other reviewers, this is the primary magic of this PR. I plan to refactor this to a utility function for calling blockwise on DataArrays so this will get cleaned up.

The main difficulty here is that the `.attrs` of the DataArrays might include a `valid_range` with dask arrays in it and dask, in any other calling mode, gets mad at that.

Xarray does have some `map_blocks` utilities but I wasn't able to get them to work.

djhoese:
Ugh code scene counts docstring lines as lines in a function. Stupid.

pnuu (approved):

PR #3187, opened 2025-08-03
Fix AWIPS tiled writer not being computed properly when compute=False

djhoese:
The AWIPS tiled writer, as modified in #3186, returns only a list of dask arrays now. This results from switching from Delayed functions to a series of `blockwise` operations as recommended by dask maintainers. This allows dask to better optimize the dask graph including not re-computing tasks it has already computed.

The problem with these changes is that the `compute_writer_results` function that is used when `scn.save_datasets(..., compute=False)` is used doesn't properly handle source-only results. This PR fixes that.

A possible change to this that I'm considering: return the result of `dask.compute` for the source compute and possibly for the delayed compute. This should theoretically allow writers that write files to return the filename that was created. This is something I've wanted for a long time as it allows Satpy-based scripts to inform later processing steps (think bash pipes) of the files that are created. However with threaded dask schedulers you can't get completed tasks in the order they are completed (I think this is distributed only) so I might just update the AWIPS writer to have an option to print the path as part of it's `blockwise` function.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/writer_tests/test_core/test_compute.py
+++ b/satpy/tests/writer_tests/test_core/test_compute.py
@@ -20,6 +20,7 @@ import datetime as dt
 import os
 import shutil
 
+import numpy as np
 import xarray as xr
 from dask import array as da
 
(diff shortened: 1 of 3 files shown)


pnuu (approved):
LGTM.

I need to think about the future plans, but at the moment don't see anything that would prevent returning filenames.

mraspaud (approved):
Do I understand correctly that "targets" is eg open files? I which case, LGTM

djhoese:
Yeah "targets" are array-like objects in that they have to have a `__setitem__` method that supports slice arguments. Dask is essentially doing `target[chunk_offset:chunk_offset + chunk_size] = source_chunk`.

djhoese:
I'm mentioning this here for future Dave but I also mentioned it on slack:

The `compute_writer_results` and `Scene.save_datasets` could return a list of filenames created. In the future if there is a writer that doesn't write to files (ex. stores data in a database) then something else could be returned or None.

Optionally a user could provide a `completion_callback` that is passed to `map_blocks` so it can do something with the created filename. A simple use case would be printing the filenames as they are finished/completed.

PR #3189, opened 2025-08-04
Add quality flags to 'viirs_l1b' reader

zxdawn:
The quality_flags are missing in the viirs_l1b reader yaml file. This PR aims to add them for both i and m data.

 - [x] Tests added


djhoese:
FYI this may be easier in the python code with `available_datasets`:

https://github.com/pytroll/satpy/blob/7a0dc1ad024dc96f30f533413378e61e04cf6abb/satpy/readers/viirs_l1b.py#L243-L261

See the base file handler class for the longer docstring of this method and how it works. I don't remember if you have experience.

zxdawn:
Em ... Do you mean we can apply dynamic datasets for this?

djhoese:
Yeah. If I'm understanding this PR correctly. For every input dataset configured in the YAML, you could check if there is a `{name}_quality_flags` in the data file and if so add the quality flag product as an option. Saves you from all this copy pasting in the YAML...I guess you have it done already, but yeah.

zxdawn:
Aha, yes. I can change the Python code to make it easier ;)

zxdawn:
I have tried something like this:

```
    def available_datasets(self, configured_datasets=None):
        """Generate dataset info and their availablity.

        See
        :meth:`satpy.readers.core.file_handlers.BaseFileHandler.available_datasets`
        for details.

        """
        for is_avail, ds_info in (configured_datasets or []):
            if is_avail is not None:
                # some other file handler said it has this dataset
                # we don't know any more information than the previous
                # file handler so let's yield early
                yield is_avail, ds_info
                continue
            ft_matches = self.file_type_matches(ds_info["file_type"])
            var_path = self._dataset_name_to_var_path(ds_info["name"], ds_info)
            is_in_file = var_path in self
            yield ft_matches and is_in_file, ds_info

            # Attempt to find _quality_flags variable path
            qf_name = ds_info["name"] + "_quality_flags"
            qf_var_path = self._dataset_name_to_var_path(qf_name, ds_info)

            if qf_var_path in self:
                qf_info = {
                    "name": qf_name,
                    "file_type": ds_info["file_type"],
                    "resolution": ds_info.get("resolution"),
                    "coordinates": ds_info.get("coordinates") or [],
                    "units": "1",
                    "standard_name": "quality_flag",
                }
                yield True, qf_info
```

But it still keeps reading nonexistent quality flags. Do you have any idea how to check the path correctly?

djhoese:
> But it still keeps reading nonexistent quality flags.

Your code makes sense to me. What kind of quality flags is it reporting as existing that don't exist?

djhoese:
Ah you'll have to be careful with how you use that variable method:

https://github.com/pytroll/satpy/blob/ccbfaa7a03510ce43d133549906563f6f286d1ad/satpy/readers/viirs_l1b.py#L274-L276

It gets `file_key` from `ds_info` which in this case will always be the original band's variable path.

zxdawn:
This PR seems to be ready now. Is the isort failure fine?

djhoese:
pre-commit.ci autofix

PR #3195, opened 2025-08-11
Fix various warnings encountered during tests

djhoese:
This PR does a couple things but most importantly (and as discussed on slack with @pnuu) it ignores all Numpy invalid value (NaN) warnings during pytest tests. I also document and recommend that users does this.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/doc/source/quickstart.rst
+++ b/doc/source/quickstart.rst
@@ -275,6 +275,33 @@ To subset multi-resolution data consistently, use the :meth:`~satpy.scene.Scene.
   >>> vis006_llbox_meas = vis006_llbox.values
   >>> vis006_llbox_lon, vis006_llbox_lat = vis006_llbox.attrs['area'].get_lonlats()
 
+Warnings and Errors
+===================
+
+Throughout the calculations that Satpy performs you may see various warnings
+or if you have logging enabled (see :ref:`troubleshooting` below) see warning
+or error log messages. There are some warnings emitted as part of Satpy's
+processing that originate in the libraries that Satpy depends on, but are
+mostly expected due to the way Satpy does it's calculations and the data that
+Satpy is working with. Except for in special cases, Satpy does not generally
+hide or ignore these warnigns from dependencies and it is left to the user
+to control how they'd like to handle them.
+
+Many of the data arrays that Satpy works with use NaN values to indicate
+invalid, masked, or bad quality pixels. Some calculations or libraries
+emit a warning or error when they encounter NaNs. For example, Numpy will
+often emit warnings like::
+
+    RuntimeWarning: invalid value encountered in multiply
+
+For the most part these warnings are expected. For single Satpy-based scripts
+it is recommend to ignore these warnings from numpy globally using
+the :func:`numpy.seterr` or :class:`numpy.errstate` functions:
+
+.. code-block:: python
+
+    import numpy as np
+    np.seterr(invalid="ignore")
 
 .. _troubleshooting:
 
--- a/satpy/modifiers/parallax.py
+++ b/satpy/modifiers/parallax.py
@@ -314,7 +314,7 @@ class ParallaxCorrection:
         """
         logger.debug("Calculating parallax correction using heights from "
                      f"{cth_dataset.attrs.get('name', cth_dataset.name)!s}, "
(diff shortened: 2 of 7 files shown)


pnuu:
Two kinds of Numpy warnings are still emitted:

https://github.com/pytroll/satpy/actions/runs/16892247242/job/47854613440?pr=3195#step:9:130
satpy/tests/writer_tests/test_simple_image.py: 2 warnings
RuntimeWarning: divide by zero encountered in divide

https://github.com/pytroll/satpy/actions/runs/16892247242/job/47854613440?pr=3195#step:9:145
satpy/tests/reader_tests/test_msi_safe.py::TestTileXML::test_angles[L1C-satellite_zenith_angle-angle_tag0-expected0]
  /Users/runner/work/satpy/satpy/satpy/readers/msi_safe.py:406: RuntimeWarning: Mean of empty slice
    angles = np.nanmean(np.dstack(arrays), -1)

https://github.com/pytroll/satpy/actions/runs/16892247242/job/47854613440?pr=3195#step:9:204
satpy/tests/reader_tests/test_fci_l2_nc.py::TestFciL2NCReadingByteData::test_byte_extraction
  /Users/runner/work/satpy/satpy/satpy/readers/fci_l2_nc.py:284: RuntimeWarning: Mean of empty slice.
    scale_factor = (x[1:] - x[0:-1]).values.mean()

djhoese:
In my opinion these should still be emitted, but if tests are producing them then the test should be changed. Either better input data or catch the warning if it's expected.

doc/source/quickstart.rst
```diff
@@ -275,6 +275,33 @@ To subset multi-resolution data consistently, use the :meth:`~satpy.scene.Scene.
...
   >>> vis006_llbox_lon, vis006_llbox_lat = vis006_llbox.attrs['area'].get_lonlats()
 
+Warnings and Errors
+===================
+
+Throughout the calculations that Satpy performs you may see various warnings
+or if you have logging enabled (see :ref:`troubleshooting` below) see warning
+or error log messages. There are some warnings emitted as part of Satpy's
+processing that originate in the libraries that Satpy depends on, but are
+mostly expected due to the way Satpy does it's calculations and the data that
```
ameraner:
```suggestion
mostly expected due to the way Satpy does its calculations and the data that
```

satpy/modifiers/parallax.py
```diff
@@ -314,7 +314,7 @@ def corrected_area(self, cth_dataset,
         """
         logger.debug("Calculating parallax correction using heights from "
                      f"{cth_dataset.attrs.get('name', cth_dataset.name)!s}, "
-                     f"with base area {self.base_area.name!s}.")
+                     f"with base area {self.base_area.description!s}.")
```
ameraner, line 317: I wonder why this was needed? I'm not sure we always have an area description, but we'll always have a name, no?

djhoese: The "name" attribute is deprecated. The "description" is the new attribute. There was a deprecation warning and this fixes it.

See https://github.com/pytroll/pyresample/blob/55c28ab0221a2aa6517322e1baea49a26a0fa6d9/pyresample/geometry.py#L1715-L1719

I think there was name and area_id and it was confusing what was what so `name` was changed to `description`.

doc/source/quickstart.rst
djhoese: Is an English as a second language person (English is not your first language I assume) correcting my english...and correctly? Thanks! :man_facepalming:

satpy/modifiers/parallax.py
ameraner: Right, I was clearly still confused about that :)

djhoese:
I'm also tempted to ignore the below warning in tests:

```
UserWarning: The specified chunks separate the stored chunks along dimension "Along_Track_375m" starting at index 8. This could degrade performance. Instead, consider rechunking after loading.
    self.nc = xr.open_dataset(self.filename,
```

Since usually this is because we have unrealistic array shapes for tests so default chunking isn't going to perform well. However, I could also see this being a sign that a reader or operation (ex. resampling we sometimes see this warning) is doing something sub-optimally.

PR #3200, opened 2025-08-17
Remove unnecessary dask task compute check in test utility scheduler

djhoese:
I had added the functionality in the CustomScheduler used in Satpy's tests to make sure that a task wasn't computed multiple times. I had thought that I tested it and that it worked as expected, but it turns out it only checks the top-level key that is requested. It is also causing issues like in #3199 where older versions of dask are providing keys as lists that are unhashable. Overall, let's just get rid of this.

@avalentino if you don't want to wait for the next Satpy release you could take the patch/diff from this PR and apply it. I do think upgrading *dask* would help, but anyway. The breakage is in the tests only, not the Satpy package itself.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/utils.py
+++ b/satpy/tests/utils.py
@@ -18,7 +18,6 @@
 
 import datetime as dt
 import os
-from collections import defaultdict
 from collections.abc import Mapping
 from contextlib import contextmanager
 from typing import Any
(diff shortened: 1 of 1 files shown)


avalentino (approved):

avalentino:
Thansk @djhoese 
I confirm the the following (15) failures have been fixed in my debian branch after applying this patch:
```
-# FAILED satpy/tests/compositor_tests/test_fill.py::TestDayNightCompositor::test_day_only_*
-# FAILED satpy/tests/compositor_tests/test_fill.py::TestDayNightCompositor::test_daynight_*
-# FAILED satpy/tests/compositor_tests/test_fill.py::TestDayNightCompositor::test_night_only_*
-# FAILED satpy/tests/reader_tests/modis_tests/test_modis_l1b.py::TestModisL1b::test_load_longitude_latitude
```

PR #3211, opened 2025-08-21
add treshhold tuned fire temperature rgb for europe

alexhalbig:
This enhancement is tuned for more contrast with case studies over europe. Surface signal are stongly reduced by using a lower limit at 20 °C instead of 0°C and using 0.2 gamma. Intra-fire distiction is improved by lowering the maximum tresholds for yellow and white signals.

FT_Vergleich
On the right is the standard recipe, on the left the tuned one. In the middle is the tuned masked with L2 FIR fire mask product. I systematically looked at the pixel values for forest fires across Europe in histograms and adjusted the thresholds so that a) earlier saturation can take place in channel 2.2 and 1.6 for better intra-fire differentiation. (Previously, the maximum saturation was never reached in all evaluated case studies). b) I adjusted the threshold of 3.8 to avoid the diurnal temperature variation in the red beam over land.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -34,6 +34,7 @@ The following people have made contributions to this project:
 - [Andrea Grillini (AppLEaDaY)](https://github.com/AppLEaDaY)
 - [Blanka Gvozdikova (gvozdikb)](https://github.com/gvozdikb)
 - [Nina Håkansson (ninahakansson)](https://github.com/ninahakansson)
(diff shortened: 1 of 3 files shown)


mraspaud (approved):
This looks good to me, thanks for providing this!
In the future, we may want to have a naming convention for composites that are improvements over standard ones, or even version numbers, but that is another issue.

Merging.

PR #3213, opened 2025-08-22
Remove sun glint masking in VIIRS EDR NDVI/EVI filtering

djhoese:
New versions of the LSR software (NOAA version 1.3) now report sun glint over land and this ends up masking most of the NDVI and EVI products.

See https://github.com/ssec/polar2grid/issues/768 for details.

Note: I made these edits using github's editor so I have no idea if the tests pass or if there are other formatting issues. Let's check what CI has to say.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/viirs_edr.py
+++ b/satpy/readers/viirs_edr.py
@@ -348,7 +348,7 @@ class VIIRSSurfaceReflectanceWithVIHandler(VIIRSJRRFileHandler):
     def _get_veg_index_good_mask(self) -> da.Array:
         # each mask array should be TRUE when pixels are UNACCEPTABLE
         qf1 = self.nc["QF1 Surface Reflectance"]
-        has_sun_glint = (qf1 & 0b11000000) > 0
+        # has_sun_glint = (qf1 & 0b11000000) > 0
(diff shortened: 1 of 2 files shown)


djhoese:
@kathys, if you have time, would you mind looking at the changes to `viirs_edr.py` and hacking your installation of Polar2Grid? It would be commenting out the same lines in the `viirs_edr.py` file and ensuring that the filtering behaves as you expect. If you don't have time, you can point me to the newer EDR files and I can test it.

kathys:
@djhoese I will give it a try today.

kathys:
@djhoese  I did make this code change to the `libexec/python_runtime/lib/python3.11/site-packages/satpy/tests/reader_tests/test_viirs_edr.py` file, but the result was the same.  There is a big area of missing swath data over land.  I checked the other QF bytes that are used, and I am fairly certain it is only the sun glint bit that is causing this. Is there something else I need to change?

kathys:
@djhoese  sorry, I missed something.  Let me try again.

kathys:
Thumbs up @djhoese.  It works as expected.  There are less retrievals apparent, but I think that is because they tightened up the Quality Flags in general.  Thanks.

Issue #3151, opened 2025-06-16
Earth radius missing in sat_xyz computation in parallax modifier


jequierz:
Hello, 

I am working on a parallax correction of the MTG-LI data. Recently, I have realized that my correction for photon travel time is unrealistic and led to wrong results for my matching algorithm with ground-based data. I noticed that maybe the line 158-160 of the script modifiers>parallax.py may have a mistake. Therefore, I start this discussion in case others may have trouble with this as well.  

The lines I am unsure about in the script [parallax.py](https://github.com/pytroll/satpy/blob/main/satpy/modifiers/parallax.py) are the following: 
```
 158 sat_xyz = np.hstack(lonlat2xyz(sat_lon, sat_lat)) * sat_alt
 159 cth_xyz = np.stack(lonlat2xyz(lon, lat), axis=-1) * EARTH_RADIUS*1e3  # km → m
 160 delta_xyz = cth_xyz - sat_xyz
```

In my understanding, the variable ``delta_xyz`` is a vector coming from the satellite to the parallaxed location of the cloud at the earth surface. But it seems that the origin of the ``sat_xyz`` vector is defined at ground level, since it is specified in the code at line 88 that: "This code expects ``sat_alt`` and ``height`` to be in meter above the Earth's surface." In contrast, the ``cth_xyz`` is defined w.r.t. the center of the earth (since we multiply with the earth's radius). 

Finally, I get more realistic results for the photon travel time since I have changed line 158 to: 

` sat_xyz = np.hstack(lonlat2xyz(sat_lon, sat_lat)) * (sat_alt+EARTH_RADIUS) `

Would you have comments or other opinion regarding this part of the modifier? 

Thank you

ameraner:
Hi @jequierz , after reviewing this I agree with you, the radius should be added to the altitude for the calculation to make sense. However I think the new line should include the km-> m conversion, as it's done in line 159, so:
`sat_xyz = np.hstack(lonlat2xyz(sat_lon, sat_lat)) * (sat_alt+EARTH_RADIUS*1e3)`
since `EARTH_RADIUS` is stored in km. Do you agree? How do your travel time computations look with this?

The difference in terms of imagery is anyway rather minor for geostationary data.

gerritholl:
[...]

PR #3217, opened 2025-08-26
Fix the parallax correction code for the unit and computation of satellite altitude

### LINKED ISSUE #3182: Wrong satellite altitude units in parallax correction computation and `get_satpos` docstring
**Describe the bug**
It seems that there is an inconsistent handling of units, between km and m, for the satellite altitude in the parallax correction code and `satpy.utils.get_satpos` with the `orbital_attributes` coming from reader datasets.

_Orbital Parameters documentation and reader usage_ 
The docs indicate that the satellite altitudes in the datasets `orbital_parameters` attributes should be in m: 
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/doc/source/reading.rst?plain=1#L290-L292
evidence of readers actually implementing this are given below.

_Parallax correction and get_satpos_
**Here is the likely issue**: the parallax correction assumes that `satpy.utils.get_satpos` returns the altitude in km, so it converts it to m in the last line here:
https://github.com/pytroll/satpy/blob/0ef189634030bda38bd069de30b60f8769995050/satpy/modifiers/parallax.py#L544-L553 

So, checking `satpy.utils.get_satpos`: the docstring says indeed that the altitude is in km:
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/utils.py#L348-L349
but then the code fetches the values from the dataset `orbital_parameters` without applying any conversion:
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/utils.py#L381-L385
**so the retrieved altitudes from `satpy.utils.get_satpos` will be actually in m, but will be handled as if they were in km by the parallax correction.**

The  "km" indication in the `get_satpos` docstring was added during the parallax correction implementation, so likely there was a misunderstanding at the time: https://github.com/pytroll/satpy/commit/26564f5fd7b2bf70eec0c39c769b90a1f0ea4a46#diff-0890a10be8f77360bce29f11ea97a1aed60a254d729b78cbf20ed2160297e717R345

**Solutions**
Option 1:(my favourite, likely the least disrupting and most consistent across the code): 
- remove the extra unneeded conversion to m in the parallax `_get_satpos_from_cth`
- fix the docstring of `satpy.utils.get_satpos` to say that altitudes are in m
- fix also the TLE-based retrieval of the altitude for consistency (this is the alternative method that `satpy.utils.get_satpos` uses if no orbital_parameters are available, also added during the parallax implementation): https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/utils.py#L432
--> keeps everything consistently in m, fixes the parallax correction, with little impact anywhere else

Option 2: 
- add a conversion from m to km in `satpy.utils.get_satpos` to get the outputs in km. 
--> Would create issues for users using this util elsewhere, and would be maintaining the m/km inconsistency in the code

Let me know what you think and if you agree with Option 1 solution. 

**Additional context**
Here some evidence that m are used in most/all readers:
ABI L1b: 
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/readers/abi_l1b.py#L85
ABI L2: 
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/readers/abi_l2_nc.py#L61
AMI L1b: 
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/readers/ami_l1b.py#L183
AHI HSD:
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/readers/ahi_hsd.py#L634
SEVIRI:
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/readers/core/seviri.py#L776
GOES HRIT:
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/readers/goes_imager_hrit.py#L56
FCI:
https://github.com/pytroll/satpy/blob/a9c531df055f7fc254b36f31d497b03e345b54c8/satpy/tests/reader_tests/test_fci_l1c_nc.py#L960

ameraner:
This PR fixes two aspects in the parallax correction code
- it fixes and uniformises the usage of meters for the satellite/platform altitude
- add the Earth radius to the satellite altitude to correctly compute the cartesian satellite coordinates 

 - [x] Closes #3182 #3151 
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/doc/source/dev_guide/writing_tests.rst
+++ b/doc/source/dev_guide/writing_tests.rst
@@ -21,3 +21,57 @@ that are used in the module::
     # The following fixtures are not defined in this file, but are used and injected by Pytest:
     # - tmp_path
     # - fixture_defined_in_conftest.py
+
+Avoiding Warnings
+=================
+
+Satpy tries to avoid all warnings being emitted during tests. A warning
+(ex. UserWarning) being emitted during tests suggests there is a new
+or upcoming change that will affect Satpy or its users. At the time of
+writing Satpy does not fail when warnings are encountered but this may
+change in the future.
+
+Warnings encountered during testing should be handled in one of a couple
+different ways.
+
+1. Fix the underlying issue. For example, if a dependency is changing behavior
+   then update Satpy's usage to not produce the warning.
+
+2. Catch the specific warning as part of the test. For example, if a test is
+   expecting to produce the warning should be making sure that it is, do::
+
+    with pytest.warns(UserWarning, match="the warning message"):
+        # code being tested
+
+3. Ignore the error at the test level::
+
+    @pytest.mark.filterwarnings("ignore:the warning message:UserWarning")
+    def test_something():
+        # test code
+
+4. Ignore the warning globally. This is typically reserved for dependency
+   changes that are expected to be removed in a future version. These are
+   configured in the ``pyproject.toml`` in the root of the repository in the
+   ``tool.pytest.ini_options`` section. See existing warning filters there for
+   examples.
+
+**Other warnings tips**
+
+* Create semi-realistic test data and avoid ``da.zeros`` or ``da.ones`` when
+  creating test data. A simple option is to use ``arange``::
+
+    test_data = da.arange(100 * 200).reshape((100, 200)).rechunk(50)
+
+* If using pytest's ``parametrize`` functionality and only some of the
+  parameters should produce a warning, use ``contextlib.nullcontext``::
+
+    exp_warning = pytest.warns(...) if condition else contextlib.nullcontext()
+    with exp_warning:
+        # code being tested
+
+Lastly, note that all Numpy "invalid value" warnings are ignored globally in
+the ``pyproject.toml`` file. So if your test involves NaNs in the data and
+you are expecting to see them or expecting to catch them then you may have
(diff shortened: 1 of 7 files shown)


ameraner:
cc @jequierz

jequierz:
Great work! Thank you Andrea :)

satpy/tests/modifier_tests/test_parallax.py
```diff
@@ -205,7 +205,7 @@ def test_get_surface_parallax_displacement(self):
 
         val = get_surface_parallax_displacement(
                 0, 0, 36_000_000, 0, 10, 10_000)
-        np.testing.assert_allclose(val, 2141.2404451757875)
```
gerritholl, line 208: Nice, this tells us how far we were off :) 👍

gerritholl (approved):
Thanks!

Issue #3227, opened 2025-09-05
Cleanup of enhancements for composites using `DayNightCompositor` is needed


pnuu:
## Feature Request

**Is your feature request related to a problem? Please describe.**
There are a lot of unnecessary enhancements defined for composites using `DayNightCompositor` stretching the data from 0 to 1. The compositor it self ensures the data are already in this range by enhancing the sub-composites.

**Describe the solution you'd like**
Remove all the named "from 0 to 1" enhancements created for composites using `DayNightCompositor` and switch the `standard_name`s with the no-op `image_ready`.

**Describe any changes to existing user workflow**
None, unless user has defined their own composites using the currently existing enhancements that would be removed.

**Additional context**
-

gerritholl:
> **Describe any changes to existing user workflow**
> None, unless user has defined their own composites using the currently existing enhancements that would be removed.

Just this week I changed our internal composites which used `fci_day_night_blend` to use `image_ready` instead.  I think the scenario where someone copied an FCI composite from `satpy/etc` and then adapted it locally is not so rare, so I think this needs a deprecation warning like the refactoring of module imports has.

strandgren:
I agree with @gerritholl . I'm quite sure this is the case for EUMETView for example, where we have a local copy of the recipe, but not all the enhancements. So a deprecation warning would be good.

djhoese:
I've said this before and plan to make a related PR today, but I feel like the default enhancement should detect if it has RGB/A data and if it is already integer type, then do a no-op. That way `standard_name` can still be something useful/relevant to the actual image and could be used in an output format (ex. CF NetCDF file)...granted I don't know of any RGB CF standard names but still.

strandgren:
And on that note, I would again be in favour of not using the `standard_name` to identify the enhancement, but father use something more explicit like `enhancement`, with a potential fallback on `standard_name` if `enhancement`  isn't set. That would also allow us to have more useful `standard_name`

djhoese:
@strandgren So this is the thing where enhancements could be defined in the composite definition, right? 👍

strandgren:
@djhoese yes correct. Now that you mention it, I think the argument against using `enhancement` instead of `standard_name` was that it doesn't work for datasets created dynamically like `toa_brightness_temperature`. But allowing to fall back on `standard_name` if `enhancement` is not set, would overcome that issue I think?

[...]

PR #3228, opened 2025-09-05
Static geo color stretching

pnuu:
This PR makes the GeoColor use static stretching for `geo_color_high_clouds` sub-composite and remove the dynamic stretching of `geo_color_low_clouds` sub-composite that already is in the range `[0, 1]`.

There is now real difference in processing time, but the memory usage is lower.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/enhancements/generic.yaml
+++ b/satpy/etc/enhancements/generic.yaml
@@ -1036,8 +1036,8 @@ enhancements:
         method: !!python/name:satpy.enhancements.contrast.stretch
         kwargs:
           stretch: crude
-          min_stretch: [300, 0]
-          max_stretch: [215, 1]
+          min_stretch: [310, 0]
+          max_stretch: [190, 1]
 
   geo_color_low_clouds:
     standard_name: geo_color_low_clouds


pnuu:
Sorry, the comparison images have to wait until monday, I ran out of time today.

satpy/etc/composites/abi.yaml
```diff
@@ -875,7 +875,7 @@ composites:
           - name: C07
       - name: C13
       - compositor: !!python/name:satpy.composites.aux_data.StaticImageCompositor
-        standard_name: land_water_mask
```
strandgren: The previous `standard_name` for the land-water masks (`land_water_mask`) was not intended for the enhancement, but rather as a `standard_name` of the type of data. I guess that means that no enhancement was defined? Does that then mean, that by default dynamic stretching is applied? Given that the land-water mask is "only" used as input to `geo_color_low_clouds` I didn't think an enhancement was needed.

strandgren:
Thanks for spotting this and also working on a solution. However, something is broken and the high clouds are all gone/black. I also saw it in your example on [slack ](https://pytroll.slack.com/archives/C0LNH7LMB/p1757073024102419?thread_ts=1757055251.449109&cid=C0LNH7LMB), but thought it was just a test for the low cloud compositor.

Left satpy main, right this PR:

image

For the background image I don't see much difference compared to your example on slack, but I guess that's because I use FCI full disk, meaning that the data probably covers the full range, meaning that the type of stretch has little impact on the end result, compared to the VIIRS granule you used which has much less data variability?

I also see the slightly lower contrast of the low clouds as expected, but I don't mind too much. Will look more once the high-cloud issues is fixed.

djhoese:
How does this PR change if I (quickly, next week?) change the default enhancement to not enhance RGB(A)s that are already integers?

I'm not really a fan of `standard_name` evolving into something where it describes the data type rather than what the data represents. @strandgren had suggested something like a new field called "enhancement" which we would add to the enhancement decision tree, but that's probably too far for this PR.

pnuu:
One of the reasons I didn't manage to get examples out is that I got lost in my local configs, so also missed something as @strandgren  pointed out.

The data from `generic_image` reader is `float32`, so they'd still need to be enhanced.

The new `enhancement` mentioned in https://github.com/pytroll/satpy/issues/3227#issuecomment-3259262391 would be great. To me, `standard_name` has been *only* used for linking the composite to the enhancement, so that's what I've done here and suggest in doing in https://github.com/pytroll/satpy/issues/3227

satpy/etc/composites/abi.yaml
pnuu: The default linear stretching is applied to everything that has no associated enhancement.

pnuu:
I converted the PR to draft until I figure out what I missed that removed the high clouds.

satpy/etc/composites/abi.yaml
pnuu: Nevermind, you're right. The `StaticImageCompositorr` is used just to read the data and not really used as in the composite as a channel, so it is not enhanced. Reverting my change of the `standard_name`.

pnuu:
Ok, found it: `geo_color_high_clouds` actually does need to be stretched. I've added a first guess of the static stretch that also handles the `invert` there was previously.

The performance boost I reported earlier was mostly me messing up my local configs that caused the usage of `linear` stretch which is still using `Delayed` in current Trollimage 1.26.0. There is still a rather big difference in memory usage in generating the `geo_color` composite:

Current `main`:
Screenshot 2025-09-08 at 12-56-20 Bokeh Plot

This PR:
Screenshot 2025-09-08 at 12-56-39 Bokeh Plot

pnuu:
Here's how it looks like with my VIIRS case:

Current `main` with dynamic low/high cloud stretching:
20170905_0028_sogs_viirs_EPSG_3035_375m_geo_color_orig_scaled-0


This PR:
20170905_0028_sogs_viirs_EPSG_3035_375m_geo_color_scaled-0

@strandgren could you test again with full disk FCI and give your opinion on the static stretch? And if you wrap you test script with

```python

import datetime as dt
from dask.diagnostics import ResourceProfiler, Profiler, CacheProfiler, visualize

with ResourceProfiler() as rprof, Profiler() as prof:
    tic = time.time()

    # code here

    print("Elapsed time:", time.time() - tic)
visualize([rprof, prof]],
          filename=dt.datetime.now(dt.UTC).strftime('/tmp/profile_%Y%m%d_%H%M%S.html'))
...

could you show how the memory usage/processing time changes?

**Comment by strandgren** [comment]
Hm, we need to refine the stretch limits for the high clouds, because now the very high cold clouds are saturated. I can also have a look and see what a good default value would be. I know I initially implemented this, but can you remind me what the different stretched for the high-cloud compositor refers too ([300, 215] vs [0, 1]). Also, why do we need to stretch this one, but now the low-cloud compositor?

main:
image

This PR:
image


Regarding performance I remember some previous discussions that our file system at EUMETSAT is not stable enough for proper profiling (@ameraner may know more), so not sure how useful the analysis is. Anyway here it is, and indeed I also see generally less memory usage.

Main:
Elapsed time: 240.9058346748352
image

This PR:
Elapsed time: 254.62602472305298
image

**Comment by strandgren** [comment]
Also, in your example on [slack](https://pytroll.slack.com/archives/C0LNH7LMB/p1757066107044999?thread_ts=1757055251.449109&cid=C0LNH7LMB), the color of the background black marble image also changed when removing the dynamic stretching, but now it's no longer the case. Was the initial change a result of your issues with local configs?

**Comment by pnuu** [comment]
> Hm, we need to refine the stretch limits for the high clouds, because now the very high cold clouds are saturated. I can also have a look and see what a good default value would be.

Yeah, I kind of expexted that the initial values would cut some of the data.

> I know I initially implemented this, but can you remind me what the different stretched for the high-cloud compositor refers too ([300, 215] vs [0, 1]). Also, why do we need to stretch this one, but now the low-cloud compositor?

The brightness temperature data are as-is, but the alpha channel has been modified in the compositor to the `[0, 1]` interval here: https://github.com/pytroll/satpy/blob/main/satpy/composites/mask.py#L342-L343

For `LowCloudCompositor` the "data" are a temperature difference. So it actually might not be within the `[0, 1]` range and I need to have another look.

**Comment by pnuu** [comment]
> Also, in your example on [slack](https://pytroll.slack.com/archives/C0LNH7LMB/p1757066107044999?thread_ts=1757055251.449109&cid=C0LNH7LMB), the color of the background black marble image also changed when removing the dynamic stretching, but now it's no longer the case. Was the initial change a result of your issues with local configs?

Yeah, I think it was me messing up with the configs.

**Comment by pnuu** [comment]
> Anyway here it is, and indeed I also see generally less memory usage.

10 GB less is not too bad :sweat_smile:

**Comment by strandgren** [comment]
> > Anyway here it is, and indeed I also see generally less memory usage.
> 
> 10 GB less is not too bad 😅

Ah yes :D I mis-read the label I though it went from roughly 5-4GB, rather than 50-40GB. That's very nice indeed!

**Comment by pnuu** [comment]
For `geo_color_low_clouds` I get this with my test data:

```python
print([np.nanmin(vals[0, :, :]), np.nanmax(vals[0, :, :])])
[np.float32(-58.483643), np.float32(19.932663)]
print([np.nanmin(vals[1, :, :]), np.nanmax(vals[1, :, :])])
[np.float32(0.0), np.float32(1.0)]
```

So there we'd also need to come up with a reasonable default `min_stretch` and `max_stretch` values for the temperature difference data. The alpha band is already in the correct range.

pnuu:
Weirdly the low clouds look pretty much identical with or without the `crude` stretch prior to the colorization.

Oh, that must be because there's only one color applied to the data! So basically only the masking done in the compositor and the resulting transparency matters. It doesn't matter how the sub-composite is stretched, because the data values are colorized with a single color and varying transparency (which is already in the `[0, 1]` range.

*edit*: clarified the wording

strandgren:
> Weirdly the low clouds look pretty much identical with or without the `crude` stretch prior to the colorization.
> 
> Oh, that must be because there's only one color applied to the data! So basically only the masking done in the compositor and the resulting transparency matters. It doesn't matter how the sub-composite is stretched, because the data values are colorized with a single color and varying transparency (which is already in the `[0, 1]` range.
> 
> _edit_: clarified the wording

yeah right, so that's why it worked without stretching. And I agree with you that only the alpha matters, since we're mapping any value to the same color. So what do you think is the best way: Apply no stretching at all for this composite, or have a static stretch of at least the alpha to make sure it's always in the range [0,1] and thus consistent over multiple repeat cycles/granules (assuming there may be cases where the alpha range is not covering the full [0,1] range)?

pnuu:
I think we're fine as it is now. The `CloudCompositor` already makes sure the alpha is in the range `[0, 1]` and colorization handles the "data" channel.

pnuu:
Waaaaait a sec. Looking at the graphs again, it was *this* PR that uses *more* memory than `main` :astonished:

strandgren:
> Waaaaait a sec. Looking at the graphs again, it was _this_ PR that uses _more_ memory than `main` 😲

you mean that in my example this PR has higher peak memory after approx. 200s? Because in both sets of plots I see generally less memory use with this PR, no?

pnuu:
> > Waaaaait a sec. Looking at the graphs again, it was _this_ PR that uses _more_ memory than `main` 😲
> 
> you mean that in my example this PR has higher peak memory after approx. 200s? Because in both sets of plots I see generally less memory use with this PR, no?

In https://github.com/pytroll/satpy/pull/3228#issuecomment-3266642832 the first graph is for `main` and the scale goes to ~60 GB while the lower graph for this PR goes to 70 GB.

mraspaud:
Yes, but it's just because of the last spike. The rest of the time, he PR is lower on memory.

mraspaud:
So where are we on this one?

pnuu:
I think @strandgren wanted to have a look at the high cloud stretch limits.

strandgren:
I have done some tests with different ranges now and I would suggest the range ~~310-210K~~ **310-190K** for the high clouds. Extending the thresholds to e.g. 180-320 gets us closer to the current dynamic stretching in main (looking at a full disk scene with deep convection in the tropics), but the image is unnecessarily dark in my opinion. Below are two sets of images comparing `main` and this PR with static stretch between `[190, 310]` for FCI 2025-09-12 18:00z. 

Increasing the lower end (e.g. 190 -> 200 or 210) may be good for local processing of certain areas, but leads to saturation for high cloud tops in the tropics, hence we should keep the lower end quite low for the default stretch.

Results from all tests/ranges I looked at can (temporarily) be found here: https://sftp.eumetsat.int/public/folder/tuzy66nlcuczxo7n90_l2a/geo_color_high_cloud_stretch_tests

**main - dynamic stretch**
a_FCI_main_geo_color_africa_20250912T180000

e_FCI_main_geo_color_europe_20250912T180000

**This PR with static stretch between 310-190K - my recommendation for default stretch**
a_FCI_pr3228-310-190_geo_color_africa_20250912T180000

e_FCI_pr3228-310-190_geo_color_europe_20250912T180000

pnuu:
The default high cloud stretch have been updated to `310...190` as @strandgren suggested between the example images. The range before the images was different, so I hope I took the correct latest suggestion.

> Yes, but it's just because of the last spike. The rest of the time, he PR is lower on memory.

It's the peak that matters when trying to determine the needed resources. For me the memory usage is lower for the static stretch over the whole processing, so I'm happy anyways :grin:

strandgren:
Thanks, `[310, 190]` is indeed correct. I modified the comment above. So unless there are any concerns about the memory use I think we can merge?

pnuu:
I'm okay to merge, memory usage is better for my use case at least.

strandgren:
@mraspaud what is your opinion on this, shall we merge? In my test only the peak memory was a bit higher, but our linux system is not very stable and optimal for profiling, so I would anyway trust @pnuu's results in https://github.com/pytroll/satpy/pull/3228#issuecomment-3265527221 more. Also there is no valid reason why removing one stretch and changing one from dynamic to static would make performance worse.

mraspaud:
> @mraspaud what is your opinion on this, shall we merge? In my test only the peak memory was a bit higher, but our linux system is not very stable and optimal for profiling, so I would anyway trust @pnuu's results in [#3228 (comment)](https://github.com/pytroll/satpy/pull/3228#issuecomment-3265527221) more. Also there is no valid reason why removing one stretch and changing one from dynamic to static would make performance worse.

I agree that this change should improve performance, because it is making things easier computation-wise. One reason the peak memory could be higher is that the data is ready faster, and thus everything is finishing at the same time.
So feel free to merge.

PR #3237, opened 2025-09-18
Add warning to default enhancement

djhoese:
Reviewers, I don't remember who was interested in this topic so forgive me if I missed anyone or added someone who doesn't care.

This PR changes the behavior of the default enhancement in Satpy. That is, the enhancement used when no other enhancement is configured. This PR adds a warning informing the user that there was no other enhancement configured. I should probably add more info about what the user should do about that.

I am also adding a no-op before falling back to the linear stretch if the data is integer type already as that it likely more useful (I think) than a dynamic stretch for data that is likely already image ready.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/enhancements/contrast.py
+++ b/satpy/enhancements/contrast.py
@@ -42,9 +42,10 @@ def warn_if_float_debug_otherwise(img, msg=""):
 
 def stretch_if_floating(img, **kwargs):
     """Perform a regular linear stretch but warn about no other enhancement."""
-    if np.issubdtype(img.data.dtype, np.floating):
-        return stretch(img, **kwargs)
-    return img
+    # if np.issubdtype(img.data.dtype, np.floating):
+    #     return stretch(img, **kwargs)
+    # return img
+    return stretch(img, **kwargs)
 
 
 def stretch(img, **kwargs):
--- a/satpy/enhancements/contrast.py
+++ b/satpy/enhancements/contrast.py
@@ -23,29 +23,26 @@ if typing.TYPE_CHECKING:
 LOG = logging.getLogger(__name__)
 
 
-def warn_if_float_debug_otherwise(img, msg=""):
-    """Perform no enhancement but emit a warning if data is floating point.
+class _MissingAttrs(dict):
+    """Mapping that produces a placeholder for attributes a ``DataArray`` doesn't have."""
 
-    The warning message is specified by the ``msg`` keyword argument
-    and can be a format string. Keyword arguments for the message
-    formatting are taken from the underlying ``DataArray``s ``.attrs``
-    dictionary.
+    def __missing__(self, key):
+        return "<unknown>"
+
+
+def warn(img: XRImage, msg: str = "") -> XRImage:
+    """Perform no enhancement but emit a warning.
+
+    The warning message is specified by the ``msg`` keyword argument and is
+    formatted with :meth:`str.format_map`. Format fields are taken from the
+    underlying ``DataArray``s ``.attrs`` dictionary. Any field that isn't in
+    ``.attrs`` is rendered as ``<unknown>`` so a missing attribute can never
+    turn a warning into an error.
 
     """
-    formatted_msg = msg.format(**img.data.attrs)
-    if np.issubdtype(img.data.dtype, np.floating):
-        warnings.warn(formatted_msg, UserWarning, stacklevel=2)
-    else:
-        LOG.debug(formatted_msg)
+    formatted_msg = msg.format_map(_MissingAttrs(img.data.attrs))
+    warnings.warn(formatted_msg, UserWarning, stacklevel=2)
     return img
-
-
-def stretch_if_floating(img, **kwargs):
-    """Perform a regular linear stretch but warn about no other enhancement."""
-    # if np.issubdtype(img.data.dtype, np.floating):
-    #     return stretch(img, **kwargs)
-    # return img
-    return stretch(img, **kwargs)
 
 
 def stretch(img, **kwargs):
--- a/satpy/enhancements/contrast.py
+++ b/satpy/enhancements/contrast.py
@@ -42,9 +42,9 @@ def stretch_if_not_integer(img, **kwargs):
     # XXX: Should this be all integers or just uint8?
     # XXX: Should it check for "bands" dimension?
     # XXX: Rename to "stretch_if_necessary"
-    if np.issubdtype(img.data.dtype, np.integer):
-        return img
+if np.issubdtype(img.data.dtype, np.floating):
     return stretch(img, **kwargs)
+return img
 
 
 def stretch(img, **kwargs):
--- a/satpy/enhancements/contrast.py
+++ b/satpy/enhancements/contrast.py
@@ -23,8 +23,8 @@ if typing.TYPE_CHECKING:
 LOG = logging.getLogger(__name__)
 
 
-def noop_with_warning(img, msg=""):
-    """Perform no enhancement but emit a warning.
+def warn_if_float_debug_otherwise(img, msg=""):
+    """Perform no enhancement but emit a warning if data is floating point.
 
     The warning message is specified by the ``msg`` keyword argument
     and can be a format string. Keyword arguments for the message
@@ -33,18 +33,18 @@ def noop_with_warning(img, msg=""):
 
     """
     formatted_msg = msg.format(**img.data.attrs)
-    warnings.warn(formatted_msg, UserWarning, stacklevel=2)
+    if np.issubdtype(img.data.dtype, np.floating):
+        warnings.warn(formatted_msg, UserWarning, stacklevel=2)
+    else:
+        LOG.debug(formatted_msg)
     return img
 
 
-def stretch_if_not_integer(img, **kwargs):
+def stretch_if_floating(img, **kwargs):
     """Perform a regular linear stretch but warn about no other enhancement."""
-    # XXX: Should this be all integers or just uint8?
-    # XXX: Should it check for "bands" dimension?
-    # XXX: Rename to "stretch_if_necessary"
-if np.issubdtype(img.data.dtype, np.floating):
-    return stretch(img, **kwargs)
-return img
+    if np.issubdtype(img.data.dtype, np.floating):
+        return stretch(img, **kwargs)
+    return img
 
 
 def stretch(img, **kwargs):
--- a/satpy/enhancements/contrast.py
+++ b/satpy/enhancements/contrast.py
@@ -27,9 +27,9 @@ def noop_with_warning(img, msg=""):
     """Perform no enhancement but emit a warning.
 
     The warning message is specified by the ``msg`` keyword argument
-    and can be a format string. Keyword arguments will be passed
-    to the formatting of the message string with the underlying
-    ``DataArray``s ``.attrs`` dictionary.
+    and can be a format string. Keyword arguments for the message
+    formatting are taken from the underlying ``DataArray``s ``.attrs``
+    dictionary.
 
     """
     formatted_msg = msg.format(**img.data.attrs)
--- a/satpy/enhancements/contrast.py
+++ b/satpy/enhancements/contrast.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import logging
 import typing
+import warnings
 from collections import namedtuple
 from numbers import Number
 from typing import Optional
@@ -20,6 +21,30 @@ if typing.TYPE_CHECKING:
     from trollimage.xrimage import XRImage
 
 LOG = logging.getLogger(__name__)
+
+
+def noop_with_warning(img, msg=""):
+    """Perform no enhancement but emit a warning.
+
+    The warning message is specified by the ``msg`` keyword argument
+    and can be a format string. Keyword arguments will be passed
+    to the formatting of the message string with the underlying
+    ``DataArray``s ``.attrs`` dictionary.
+
+    """
+    formatted_msg = msg.format(**img.data.attrs)
+    warnings.warn(formatted_msg, UserWarning, stacklevel=2)
+    return img
+
+
+def stretch_if_not_integer(img, **kwargs):
+    """Perform a regular linear stretch but warn about no other enhancement."""
+    # XXX: Should this be all integers or just uint8?
+    # XXX: Should it check for "bands" dimension?
+    # XXX: Rename to "stretch_if_necessary"
+    if np.issubdtype(img.data.dtype, np.integer):
+        return img
+    return stretch(img, **kwargs)
 
 
 def stretch(img, **kwargs):
--- a/satpy/etc/enhancements/generic.yaml
+++ b/satpy/etc/enhancements/generic.yaml
@@ -4,11 +4,11 @@ enhancements:
   default:
     operations:
       - name: warn_no_predefined
-        method: !!python/name:satpy.enhancements.contrast.noop_with_warning
+        method: !!python/name:satpy.enhancements.contrast.warn_if_float_debug_otherwise
         kwargs:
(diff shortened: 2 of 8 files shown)


satpy/enhancements/contrast.py
```diff
@@ -38,6 +39,30 @@
 LOG = logging.getLogger(__name__)
 
 
+def noop_with_warning(img, msg=""):
+    """Perform no enhancement but emit a warning.
+
+    The warning message is specified by the ``msg`` keyword argument
+    and can be a format string. Keyword arguments will be passed
+    to the formatting of the message string with the underlying
+    ``DataArray``s ``.attrs`` dictionary.
```
pnuu: This needs rewording. My first try:
```suggestion
    and can be a format string. Keyword arguments for the message
    formatting are taken from the underlying ``DataArray``s ``.attrs``
    dictionary.
```

satpy/enhancements/contrast.py
```diff
@@ -38,6 +39,30 @@
...
+
+    """
+    formatted_msg = msg.format(**img.data.attrs)
+    warnings.warn(formatted_msg, UserWarning, stacklevel=2)
+    return img
+
+
+def stretch_if_not_integer(img, **kwargs):
+    """Perform a regular linear stretch but warn about no other enhancement."""
+    # XXX: Should this be all integers or just uint8?
```
pnuu: I think it depends both on the `dtype` of the data and the output `dtype`. If the data are `uint16` and the output is `uint8`, I guess a warning and truncation (data might still fit the smaller range) is needed. Maybe also the output format makes a difference?

satpy/enhancements/contrast.py
```diff
@@ -38,6 +39,30 @@
...
+    # XXX: Should it check for "bands" dimension?
```
pnuu: Integer data with
* one channel: could be categorical data or B&W image
  * check for palette?
* 2-4 channels: could be an image

I think both cases (categories, image) could (should?) be handled the same.

Are there other cases?

satpy/enhancements/contrast.py
```diff
@@ -38,6 +39,30 @@
...
+    # XXX: Rename to "stretch_if_necessary"
```
pnuu: Maybe. And have suitably named sub-methods for the different cases.

satpy/etc/enhancements/generic.yaml
```diff
@@ -3,9 +3,13 @@
...
-    - name: stretch
-      method: !!python/name:satpy.enhancements.contrast.stretch
-      kwargs: {stretch: linear}
+      - name: warn_no_predefined
+        method: !!python/name:satpy.enhancements.contrast.noop_with_warning
+        kwargs:
+          msg: "No YAML enhancement configured for {name!r}. Falling back to a dynamic linear stretch."
+      - name: stretch
+        method: !!python/name:satpy.enhancements.contrast.stretch_if_not_integer
+        kwargs: {stretch: linear}
```
pnuu, line 12: Do we want `linear` or `crude` as the default stretch? `linear` has the cutoffs, `crude` stretches between min and max values. Meh, I guess the current `linear` is fine to get rid of the outliers.

satpy/enhancements/contrast.py
```diff
@@ -38,6 +39,30 @@
 LOG = logging.getLogger(__name__)
 
 
+def noop_with_warning(img, msg=""):
+    """Perform no enhancement but emit a warning.
+
+    The warning message is specified by the ``msg`` keyword argument
+    and can be a format string. Keyword arguments will be passed
```
gerritholl: There are [at least seven](https://stackoverflow.com/q/13451989/974555) ways to do string formatting in Python.  The documentation should specify which type is meant.  I think it's the type described as method 2b in the linked SO question.

satpy/enhancements/contrast.py
gerritholl: uint8 would be a greyscale image.  B&W would be type bool, such as a cloud mask, but I think even a cloud mask is usually encoded as uint8.  In numpy, `np.bool_` is *not* treated as a subdtype of `np.integer`, in fact it is not even treated as a number.

satpy/enhancements/contrast.py
```diff
@@ -38,6 +39,30 @@
...
+
+
+def stretch_if_not_integer(img, **kwargs):
+    """Perform a regular linear stretch but warn about no other enhancement."""
+    # XXX: Should this be all integers or just uint8?
+    # XXX: Should it check for "bands" dimension?
+    # XXX: Rename to "stretch_if_necessary"
+    if np.issubdtype(img.data.dtype, np.integer):
+        return img
+    return stretch(img, **kwargs)
```
gerritholl: To defend against unusual data (bool, complex, str), I would formulate this the other way around:

```suggestion
if np.issubdtype(img.data.dtype, np.floating):
    return stretch(img, **kwargs)
return img
```

satpy/etc/enhancements/generic.yaml
gerritholl: Doesn't crude require defining numeric values?  That wouldn't work well here because there is no way to define values that work by default.

satpy/enhancements/contrast.py
gerritholl: We don't know the output dtype at this point, do we?

satpy/etc/enhancements/generic.yaml
```diff
@@ -3,9 +3,13 @@
 enhancements:
   default:
     operations:
-    - name: stretch
-      method: !!python/name:satpy.enhancements.contrast.stretch
-      kwargs: {stretch: linear}
+      - name: warn_no_predefined
+        method: !!python/name:satpy.enhancements.contrast.noop_with_warning
+        kwargs:
+          msg: "No YAML enhancement configured for {name!r}. Falling back to a dynamic linear stretch."
```
gerritholl: Adding a link to documentation, and changing `configured` to `found`.  It is possible that the user has tried to configure an enhancement, but that they did something wrong and the enhancement wasn't found.  Also, the warning message is factually incorrect for non-floating data, because it is only falling back to a dynamic linear stretch for floating data.  For integer data no enhancement is defined.

```suggestion
          msg: "No YAML enhancement found for {name!r}. Falling back to a dynamic linear stretch in case of floating point data, or no enhancement otherwise. See https://satpy.readthedocs.io/en/stable/enhancements.html for documentation on defining an enhancement."
```

satpy/enhancements/contrast.py
```diff
@@ -38,6 +39,30 @@
...
+
+    The warning message is specified by the ``msg`` keyword argument
+    and can be a format string. Keyword arguments will be passed
+    to the formatting of the message string with the underlying
+    ``DataArray``s ``.attrs`` dictionary.
+
+    """
+    formatted_msg = msg.format(**img.data.attrs)
+    warnings.warn(formatted_msg, UserWarning, stacklevel=2)
+    return img
```
gerritholl: I think we might want different warnings depending on the mode of the image data.  If image data is mode L, chances are the user is simply doing `sc.load([10.5]); sc.show([10.5])`, or the equivalent for anything that's not a reflectance.  For a quick look at a cloud mask, this is fine and a warning might needlessly scare users.  But if the user has defined their own RGB and the output looks wrong, the warning might tell the user that they have not defined an enhancement or not correctly.

gerritholl (changes requested):
I like this, and I think it's great for user-defined RGBs, where the default enhancement is usually not what they want or should want.  However, I'm a bit worried about the consequences for users looking at single-channel brightness temperatures or categorical data.  If I understand this correctly, the following code will trigger the warning:

```python
>>> sc.load([10.8])
>>> sc.show(10.8)
```

So I think we should go ahead with this, but:

1. The warning message should probably depend on whether this is single-channel or multi-channel data.
2. We should have default enhancements for brightness temperature and perhaps some other common data.
3. When the user is looking at data that do not need an enhancement, such as categorical data (`sc.show("cloud_mask")`), I don't know if we need a warning at all.

satpy/enhancements/contrast.py
djhoese: No we don't. This is all about scaling/normalizing. The assumption here is that someone with a category product probably doesn't want it scaled, but at the same time if you had a category product that used 0-8 then you wouldn't really see much in a geotiff image without additional stretching.

satpy/enhancements/contrast.py
djhoese: I don't think we currently have any explicit support or use of bool-type arrays.

satpy/enhancements/contrast.py
djhoese: So `stretch_if_necessary` and then have a `stretch_if_not_integer` that uses it? Like it passes the boolean condition for it?

```python
def stretch_if_not_integer(img, **kwargs):
    return stretch_if_necessary(img, not np.issubdtype(img.data.dtype, np.integer), **kwargs)

```

Like that?

satpy/etc/enhancements/generic.yaml
djhoese: Agreed. It'd be nice to use `crude` for performance but we just can't.

satpy/enhancements/contrast.py
djhoese: I think "format string" has a specific meaning and is similar to an f-string. That is, you call `.format` on it.

satpy/enhancements/contrast.py
djhoese: Oh I like this idea. Then maybe I should avoid the "not integer" talk and focus more on saying "if floating point".

satpy/enhancements/contrast.py
djhoese: I'm not sure I understand. Or maybe, I'm not sure how likely some of these scenarios are. A single band DataArray that is non-floating (category product) makes perfect sense and shouldn't be scaled, but should probably still have a warning that the user should configure an enhancement...

Oh or maybe there is a difference between a DEBUG log message and a UserWarning depending on how "reasonable" the case is. I guess the main deciding factor/problem is do we want the default/fallback enhancement to be something people can depend on (in specific cases) or should it always tell users they should define their own?

gerritholl: My view:

If users are defining their own composite, they should always specify an enhancement.  This can be their own or a built-in one from satpy.  Not having an enhancement should be at least a warning, but possibly an error.

If users are loading single-channel datasets directly from a reader into a scene, there should be a reasonable default for `sc.show(x)`, without a warning message.

Although we can not reliably tell the two cases apart (user-defined composites with `SingleBandCompositor` look similar to single-channel data), a substitute that works in many cases is to always warn for floating point RGBs without an enhancement defined, but to define defaults for single band floating point with common standard names.  By far the most common example of single band floating point without an enhancement defined would be brightness temperature data.  Maybe radiance too.

djhoese: Yes, of course, the BTs (and maybe radiances) need some logical well-accepted defaults. If we ignore that those are missing currently, I think we're still mostly saying the same thing. Here's what I think I'll try to implement today if I have time:

1. Warn and stretch if float
2. Debug log message if not float and no-op

The mixing of log and UserWarning here makes me think maybe I should *also* log a warning. I don't know.

satpy/enhancements/contrast.py
djhoese: I'm resolving this conversation as yeah there isn't much we can do when we don't have the output type. The other discussion had specified not checking for integers, but checking for not floats which changes how this is handled.

satpy/enhancements/contrast.py
djhoese: Resolving this thread as I think we all agreed all integer data should probably be handled the same (single band or multi-band) but also that we should check for float not integer.

satpy/enhancements/contrast.py
djhoese: I decided to be a little more explicit for now instead of having multiple "helper" versions of the function. Feel free to disagree and suggest improvements.

djhoese:
This should be merged **after** a default BT enhancement is defined.

pnuu (approved):
LGTM.

I was thinking that the douple check of floating point is a bit excessive, but I can't think of a better way to separate the two new functionalities.

djhoese:
I mean I *could* put them in one function but then I kept thinking someone might find them useful.

pnuu:
Agreed.

CI doesn't seem to be happy about ninjogeotiff tags.

djhoese:
No it's that some test data doesn't have "name" defined and the warning message uses it.

djhoese:
@pnuu @gerritholl What if in Satpy <1.0 I keep the existing warning, but don't stop scaling non-float data. I can also add an additional warning telling people of the change upcoming in Satpy 1.0. That change would be no longer scaling non-float data.

pnuu:
I'm fine with that.

djhoese:
Ok I've started updating this PR again hoping to point @mraspaud at it. My last commit implements what I mentioned in my previous comment. That is, in Satpy <1.0 the behavior is the same but you get a warning about:

1. You should define your own explicit enhancement.
2. Satpy 1.0+ will no longer dynamically scale non-floating data as it is likely unexpected/unintended.

However, now I'm debating whether this change where non-floating point data is not going to be scaled. If we're warning people to make an explicit enhancement, then what's the point of careful/data-type-specific scaling anyway?

djhoese:
Talked about this in the meeting today. We decided to always warn about not using the default enhancement but we'll keep scaling the same as it is now.

djhoese:
~TODO: Make sure warning references Satpy documentation on how to create a custom enhancement.~ Looks like this was already done.

djhoese:
As discussed at the end of the meeting, it would be good if this warning message functionality could be moved outside of the individual enhancement operation functions and into the enhancement handling itself (if possible). This way @mraspaud could use it for his WMO RGB recipe work to deprecate certain enhancements.

djhoese:
So I was going to work on this today and thought I had written a comment about what I needed to do, but my last comment does not include the detail I expected and it doesn't match my memory of where we ended with this. So first:

> Talked about this in the meeting today. We decided to always warn about not using the default enhancement but we'll keep scaling the same as it is now.

I still need to do this. That is, don't worry about dtype as I am currently in this branch (float versus integer data) and just always warn if the default enhancement is used.

Then @mraspaud wanted this, I think, to be configured higher up in the enhancement. So not in an operation but at the overall enhancement level. I remember saying I was going to do this in a meeting and then somewhere (I don't see it in the comments above, probably in slack somewhere) telling Martin that I didn't see the benefit of the complexity of changing the `Enhancer` code to handle this and that I thought it was actually possible to use the warnings-as-an-operation style implementation I kind of have here. So @mraspaud can you re-describe to me how you want to have warnings for enhancements? This was to deprecate enhancements for the old untagged RGB recipes right?

djhoese:
@mraspaud If you could take a look at this as it is now and we discuss the current choices that would be good. Right now a lot of tests are going to fail because they are falling through to the default and getting the enhancement warning and therefore pytest is failing. So my choices around that are:

1. Add a filter to the global pytest list so ignore the warning. This could be a problem where it hides someone testing an enhancement matching and not noticing that it is falling through.
2. Add a filter to all the currently failing tests that are supposed to fail. A lot of annoying extra junk to add to the tests. Not hard with Claude, but ugly.
3. Update the tests to they actually match a meaningful enhancement section. Most of this is just laziness when tests were developed and there's no `standard_name` defined.

mraspaud (approved):
LGTM, nice and simple! Thanks!

mraspaud:
> @mraspaud If you could take a look at this as it is now and we discuss the current choices that would be good. Right now a lot of tests are going to fail because they are falling through to the default and getting the enhancement warning and therefore pytest is failing. So my choices around that are:
> 
>     1. Add a filter to the global pytest list so ignore the warning. This could be a problem where it hides someone testing an enhancement matching and not noticing that it is falling through.
> 
>     2. Add a filter to all the currently failing tests that are supposed to fail. A lot of annoying extra junk to add to the tests. Not hard with Claude, but ugly.
> 
>     3. Update the tests to they actually match a meaningful enhancement section. Most of this is just laziness when tests were developed and there's no `standard_name` defined.

I would vote for 3.

PR #3241, opened 2025-09-25
add support for amsr2 l1r

twak:
https://eolp.jaxa.jp/GCOM-W_AMSR2_L1R_TB.html

Adds a missing downstream layer from Jaxa's AMSR2 satellite

 - [x] Closes  
 - [ ] Tests added - no really new code
 - [ ] Fully documented  - none of the existing amsr stuff is documented...?

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -108,3 +108,4 @@ The following people have made contributions to this project:
 - [Yufei Zhu (yufeizhu600)](https://github.com/yufeizhu600)
 - [Clément (ludwigvonkoopa)](https://github.com/ludwigVonKoopa)
 - [Xuanhan Lai (sgxl)](https://github.com/sgxl)
+- [Tom Kelly (twak)](https://twak.org)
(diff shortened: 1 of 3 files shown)


djhoese:
The referenced URL mentions this data is a resampled version of the L1B data. It sounds like it is resampled to the 89GHz A channel, not a uniform grid. Does that sound right? So you still have a "swath" of pixels, right?

twak:
Yes, it is a swath, very similar to l1b, but with some brightness and alignment corrections.

strandgren:
Sorry for closing, accidentally touched the wrong part of the screen..

twak:
I also have code for another modis l2 product - [mod29](https://nsidc.org/data/mod29/versions/61) - let me know if you want a pr!

djhoese:
Looks like you'll also need to fix the issues that pre-commit is talking about. You could also have it do that locally when you make a commit:

```
pip install pre-commit
pre-commit install
```

then every commit after that will run our pre-commit checks before letting you successfully making a commit.

For the MODIS changes, feel free to make a PR to update the modis_l2 reader with the new file type.

PR #3244, opened 2025-09-26
Remove unused default_channels from readers

gerritholl:
Many readers contained a key 'default_channels' that was not used inside or outside satpy.  Remove this.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/etc/readers/abi_l1b.yaml
+++ b/satpy/etc/readers/abi_l1b.yaml
@@ -15,7 +15,6 @@ reader:
   status: Nominal
   supports_fsspec: true
   sensors: [abi]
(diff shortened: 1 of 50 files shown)


djhoese:
CI is failing because of coveralls. Don't worry about it. I added @ameraner to the reviewers list because the only other outside-of-pytroll application I can think of that might care about this configuration is SIFT, but I also think SIFT is defining its own lists of things like this.

ameraner (approved):
LGTM, no usage in SIFT either :)

djhoese (approved):

PR #3252, opened 2025-10-06
Fix 'awips_tiled' writer tests producing warnings on some hosts

djhoese:
While trying to reduce warnings generated while running tests I noticed that on some CI environments there are sometimes warnings from the AWIPS tiled writer. In the writer code, one of the global attributes produced defaults to the hostname of the current system. On some github runner systems the hostname is long enough to trigger a warning in the writer for this particular attribute being greater than 31 characters. To fix this, this PR forces the override environment variable to a short string in a pytest autouse fixture.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/writer_tests/test_awips_tiled.py
+++ b/satpy/tests/writer_tests/test_awips_tiled.py
@@ -158,6 +158,17 @@ def _get_test_lcc_data(dask_arr, area_def, extra_attrs=None):
     return _update_resampled_coords(ds, ds, area_def)
 
 
+@pytest.fixture(autouse=True)
+def short_prod_location(monkeypatch):
+    """Force the environment variable used for "production_location" to a short string.
+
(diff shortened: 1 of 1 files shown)


mraspaud (approved):
I'm not familiar with this writer, but this seems sound. On the testing side, didn't you want to catch warnings to ensure the warning you get isn't raised anymore? Otherwise, feel free to merge.

djhoese:
I went the "catch warnings" side locally first, but it just got annoying. I have one test where I specifically test this warning. The fact that the default behavior (which is a historical decision) is based on the hostname of the running system is just annoying. I have a test function that does specifically test for the length warning. I guess the one thing I'm not testing for here is that the hostname is properly accessed, but its also not something I'm really worried about.

PR #3255, opened 2025-10-07
Fix NIRReflectance not matching optional inputs against required inputs

djhoese:
@tommyjasmin ran into an issue where he was getting a ValueError from some NIRReflectance modifier usage. This was caused by the `optional_prerequisites` not being included in the call to `self.match_data_arrays`. This method is meant to check resolution and array size and make sure that inputs are compatible with each other and if not raise an `IncompatibleAreas` exception so the `Scene` can warn the user that resampling will be needed to generate the composites they want.

~TODO: Rewrite tests to use pytest instead of unittest and remove most (if not all) mocking.~

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/test_modifiers.py
+++ b/satpy/tests/test_modifiers.py
@@ -254,21 +254,18 @@ class TestNIRReflectance:
                          "area": area,
                          "start_time": self.start_time}
 
-        self.nir_arr = nir_arr = np.array([[273.15, 275.15], [277.15, 279.15]], dtype=np.float32)
+        self.nir_arr = nir_arr = np.array([[283.15, 285.15], [287.15, 289.15]], dtype=np.float32)
         self.nir = xr.DataArray(da.from_array(nir_arr), dims=["y", "x"])
         self.nir.attrs.update(self.metadata)
 
-        ir_arr = np.array([[283.15, 285.15], [287.15, 289.15]], dtype=np.float32)
-        self.ir_ = xr.DataArray(da.from_array(ir_arr), dims=["y", "x"])
-        self.ir_.attrs["area"] = area
+        ir_arr = np.array([[273.15, 275.15], [277.15, 279.15]], dtype=np.float32)
+        self.ir_ = xr.DataArray(da.from_array(ir_arr), dims=["y", "x"], attrs={"area": area})
 
-        self.sunz_arr = np.array([[1.0, 2.0], [3.0, 4.0]], dtype=np.float32)
-        self.da_sunz = da.from_array(self.sunz_arr)
-        self.sunz = xr.DataArray(self.da_sunz, dims=["y", "x"])
-        self.sunz.attrs["standard_name"] = "solar_zenith_angle"
-        self.sunz.attrs["area"] = area
+        self.sunz_arr = np.array([[1.0, 20.0], [40.0, 60.0]], dtype=np.float32)
+        self.sunz = xr.DataArray(da.from_array(self.sunz_arr), dims=["y", "x"],
+                                 attrs={"standard_name": "solar_zenith_angle", "area": area})
 
-        co2_arr = np.array([[300.0, 301.0], [302.0, 303.0]], dtype=np.float32)
+        co2_arr = np.array([[240.0, 241.0], [242.0, 243.0]], dtype=np.float32)
         self.co2 = xr.DataArray(
             da.from_array(co2_arr),
             dims=("y", "x"),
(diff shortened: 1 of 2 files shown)


djhoese:
@simonrp84 Let me know what you think of the test values I use in this PR. It looks like my fake stuff is producing negative reflectances so I must be doing something wrong, right?

@pnuu @mraspaud I could go further and do parametrize and fixtures and all this stuff, but this was already not my bug and were not my tests that I wrote and I think I'm leaving it in a better place (once the test data is reviewed for realism). I don't think I can spend much more time on this.

Note: I removed some of the tests as they were only checking some attributes of the output and that's it. I didn't see them really needing their own test.

djhoese:
Ok I lied. I parametrized the tests.

satpy/tests/test_modifiers.py
```diff
@@ -227,176 +227,135 @@ def test_invalid_max_sza(self, sunz_ds1, sunz_sza):
...
+
+        self.start_time = dt.datetime(2020, 1, 1, 12, 0, 0)
         self.metadata = {"platform_name": "Meteosat-11",
                          "sensor": "seviri",
                          "name": "IR_039",
                          "area": area,
                          "start_time": self.start_time}
 
-        nir_arr = RANDOM_GEN.random((2, 2))
+        self.nir_arr = nir_arr = np.array([[273.15, 275.15], [277.15, 279.15]], dtype=np.float32)
```
simon-sat: NIR data is reflectance, so these values should be in the range 0-100

satpy/tests/test_modifiers.py
```diff
@@ -227,176 +227,135 @@ def test_invalid_max_sza(self, sunz_ds1, sunz_sza):
...
         self.nir.attrs.update(self.metadata)
 
-        ir_arr = 100 * RANDOM_GEN.random((2, 2))
+        ir_arr = np.array([[283.15, 285.15], [287.15, 289.15]], dtype=np.float32)
         self.ir_ = xr.DataArray(da.from_array(ir_arr), dims=["y", "x"])
         self.ir_.attrs["area"] = area
 
-        self.sunz_arr = 100 * RANDOM_GEN.random((2, 2))
-        self.sunz = xr.DataArray(da.from_array(self.sunz_arr), dims=["y", "x"])
+        self.sunz_arr = np.array([[1.0, 2.0], [3.0, 4.0]], dtype=np.float32)
```
simon-sat: For better representivity, a larger range of SZA may be better.

satpy/tests/test_modifiers.py
```diff
@@ -227,176 +227,135 @@ def test_invalid_max_sza(self, sunz_ds1, sunz_sza):
...
-        self.refl_from_tbs = mock.MagicMock()
-        self.refl_from_tbs.side_effect = self.fake_refl_from_tbs
-
-    def fake_refl_from_tbs(self, sun_zenith, da_nir, da_tb11, tb_ir_co2=None):
-        """Fake refl_from_tbs."""
-        del sun_zenith, da_nir, da_tb11
-        if tb_ir_co2 is not None:
-            return self.refl_with_co2
-        return self.refl
+        co2_arr = np.array([[300.0, 301.0], [302.0, 303.0]], dtype=np.float32)
```
simon-sat: Is this for correcting the NIR and TIR data? If so, the CO2 values should be colder than the TIR values (as CO2 absorbs signal).

djhoese: Side note: Git says you wrote at least some of these tests.

There are 2 hard requirements: NIR and IR bands. And 2 optional: CO2 and SZA. What does the "T" in TIR stand for?

satpy/tests/test_modifiers.py
djhoese: I should have mentioned, I was very tired doing these updates.

satpy/tests/test_modifiers.py
pnuu: Thermal.

satpy/tests/test_modifiers.py
djhoese: In the ABI composites the C07 channel has this modifier applied and it is in Kelvin.

djhoese: And IR_039 from SEVIRI which this data is mimicking also has units of Kelvin.

satpy/tests/test_modifiers.py
djhoese: Fixed. Thanks.

djhoese:
> Regarding the SZA array, should we have one value in the array larger than the threshold (84 degrees)?

Probably. Let me try.

satpy/tests/test_modifiers.py
```diff
@@ -227,176 +227,132 @@ def test_invalid_max_sza(self, sunz_ds1, sunz_sza):
...
-        """Test NIR reflectance compositor provided only sunz."""
-        calculator.return_value = mock.MagicMock(
-            reflectance_from_tbs=self.refl_from_tbs)
-        sza.return_value = self.da_sunz
+
+    @pytest.mark.parametrize(
+        ("include_sunz", "include_co2", "exp_res"),
+        [
+            (False, False, np.array([[4.3689156, 4.762686], [5.1886106, 5.6510105]], dtype=np.float32)),
+            (True, False, np.array([[3.9977279, 4.6561675], [384.3708, np.nan]], dtype=np.float32)),
```
djhoese, line 284: @simonrp84 See this expected value here. The last two SZA values are 87 and 89 (after the sza threshold and after the limit). Is this really high reflectance value (384.3708) expected?

satpy/tests/test_modifiers.py
simon-sat: Oh, those are not NIR channels! Those are MIR channels! NIR channels are solar, usually between 1 and 2 micron. MIR channels are thermal, usually between 3-4 micron wavelength.

This whole thing probably needs a find-replace on NIR->MIR

djhoese:
Approval was given on slack. Discussion will continue there on whether the large (384) value is expected with the SZA test values used here.

tommyjasmin:
Yay!  Thanks everyone!

PR #3260, opened 2025-10-13
Remove pyspectral downloading from tests and fail if used

djhoese:
I noticed while working on other PRs that some of the CI tests were failing with HTTP hiccups trying to download pyspectral LUTs. Turns out this has been happening for years and is a major waste of resources even though it was nice to have the full processing being tested, it isn't "right" in my opinion. This PR:

* Mocks and removes the usage of pyspectral so it never downloads anything. Additionally it uses `autospec` so Satpy's usage will always match upstream pyspectral.
* Adds an autouse fixture for *ALL* of satpy to mock pyspectral's usage of the `requests` library so tests fail if they make pyspectral download anything.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/hrpt.py
+++ b/satpy/readers/hrpt.py
@@ -127,7 +127,7 @@ class HRPTFile(BaseFileHandler):
         self.channels = {i: None for i in AVHRR_CHANNEL_NAMES}
         self.units = {i: "counts" for i in AVHRR_CHANNEL_NAMES}
 
-        self.year = filename_info.get("start_time", dt.datetime.now(dt.timezone.utc)).year
+        self.year = filename_info["start_time"].year
 
     @cached_property
     def times(self):
--- a/satpy/tests/conftest.py
+++ b/satpy/tests/conftest.py
@@ -21,7 +21,6 @@ This module is executed automatically by pytest.
 
 """
 import os
-from unittest import mock
 
 import pytest
 
@@ -60,7 +59,7 @@ def include_test_etc():
 
 @pytest.fixture(autouse=True, scope="session")
 def _forbid_pyspectral_downloads():
-    with mock.patch("pyspectral.utils.requests") as mock_requests:
-        mock_requests.get.side_effect = RuntimeError(
-            "Pyspectral is attempting a download during tests. Mock pyspectral as necessary to avoid this.")
+    from pyspectral.testing import forbid_pyspectral_downloads
+
+    with forbid_pyspectral_downloads():
         yield
--- a/satpy/tests/conftest.py
+++ b/satpy/tests/conftest.py
@@ -21,6 +21,7 @@ This module is executed automatically by pytest.
 
 """
 import os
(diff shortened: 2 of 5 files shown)


satpy/utils.py
```diff
@@ -418,7 +418,7 @@ def _get_satpos_from_platform_name(cth_dataset):
 
     Needs pyorbital, skyfield, and astropy to be installed.
     """
-    from pyorbital.orbital import tlefile
```
djhoese: This import was wrong and is no longer available in future versions of pyorbital. I removed it in a PR last week to avoid import-time side effects.

djhoese:
With very little effort of just comparing executing times of the unit test steps to a previous execution: CI on the ubuntu 3.12 false environment ran unit tests 50 seconds faster.

pnuu (approved):
I'm not sure I follow completely how this works, but I certainly approve of forcing the download to be disabled.

djhoese:
For hopefully a little clarification: pytest always loads any `conftest.py` modules in every test directory (include subdirectories) looking for fixtures and other similar things. In Satpy's root test directory I create a fixture that has autouse=True on. That means it is always run for every test. So basically *every* test has pyspectral's requests mocked to error out.

I also defined it as a session scope so it should hopefully only run at the start of testing and not be enabled/disabled/enabled/disabled/enabled and so on for every test. I *think* that should be fine even with our "pytest-xdist" usage where it runs tests in parallel but I should confirm that.

djhoese:
Did a quick check and I *think* this should be fine. I think it is run once per worker, but is actually executed so that's fine.

mraspaud (changes requested):
For reference, as I said on slack, I would like the rayleigh and calculator mocks to be outsourced to pyspectral, eg in a "testing" module. See also: https://github.com/pytroll/pyspectral/issues/262

djhoese:
Ok I've rebased this entire branch and then merged in my branch for the AVHRR HRPT test fixes in #3330. So hopefully if this PR is approved and merged then that PR will be detected as merged as well.

djhoese:
Ok I now also merged #3318 because that was the latest failure I saw.

mraspaud:
Is this failing because of the fake pyspectral data we use now?

djhoese:
Probably. I will review when I have time. Permission to merge if it passes after I review what's going on and I add some tolerances?

djhoese:
Hm two other environments passed for the same OS. I'm going to rerun the failed jobs and see what happens. I did not see this failure before my last commit (I think) so I'm a little confused.

mraspaud:
> Probably. I will review when I have time. Permission to merge if it passes after I review what's going on and I add some tolerances?

Yes for me

djhoese:
@mraspaud it looks like restarting the jobs fixed the failure. If they end up being inconsistent I will work on making them more consistent. Merging...

PR #3273, opened 2025-10-23
Update NWC SAF GEO NetCDF reader to support v2025

### LINKED ISSUE #3165: Possible change in some nwcsaf products in v2025 (MTG enabled) version.
It seems that the some nwcsaf products in v2025 (MTG enabled) version have some format changes, compared to the previous versions. As a result, satpy readers may fail. Could they be updated?

pnuu:
This PR updates the NWC SAF GEO NetCDF reader to support v2025 data format

Changes in the data:

 - [x] Handle time dimension in "normal" datasets
 - [x] Handle new structure in HRW files
 - [x] Add ASII ICE datasets
 - [x] Add ASII GW IR dataset
 - [x] Add Cloud top altitude dataset in hectofeet

The template checklist:

 - [x] Closes #3165 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/reader_tests/test_nwcsaf_hrw_nc.py
+++ b/satpy/tests/reader_tests/test_nwcsaf_hrw_nc.py
@@ -21,7 +21,7 @@ import h5py
 import numpy as np
 import pytest
 
-from satpy.readers.nwcsaf_hrw_nc import PRE_V2025_SEVIRI_WIND_CHANNELS
+from satpy.readers.nwcsaf_hrw_nc import POST_V2025_DATASETS, PRE_V2025_SEVIRI_WIND_CHANNELS
 from satpy.tests.utils import RANDOM_GEN
 
 # This is the actual dtype of the trajectory items. We do not support it, so won't add this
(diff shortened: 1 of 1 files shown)


pnuu:
Looks like the HRW format has changed completely and will require pretty much a completely new file handler code and check for the version to switch automatically between the old and new formats.

~Also in the files I have, the EXIM products are separate and not merged to single files with a time dimension. My colleague is checking whether this is configurable or hasn't actually been changed in v2025 of the GEO software.~

pnuu:
Ok, the EXIM forecast steps are indeed stored in a single file. I got confused as I `grep`'d the `time` dimension from `ncdump` output, but the dimension is actually `fct`. But we also don't have support for reading the EXIM datasets, so I'm skipping the implementation in this PR.

pnuu:
HRW tests have been updated to work with the needed changes in the file handler code. I still need to add tests for both "normal" and HRW files that represent the structure of GEO v2025.

pnuu:
Some tests added for both v2025 filehandlers, "normal" and HRW. Might need some refactoring and renaming, but that'll have to happen next week.

pnuu:
I think this is now ready. I did some refactoring based on what CodeScene pointed to. The naming for the new NWC SAF GEO version case isn't great, but I can't think of anything better either...

The tests failed once for the parallax correction because something going wrong in downloading the Pyspectral data.

pnuu:
@athakarag could you if this works for you?

pnuu:
@mraspaud is this close to what you suggested? In the short time I had left today I couldn't figure out if there's a way to keep `NWCSAFGEOHRWFileHandler` as a class.

pnuu:
I created https://github.com/pytroll/satpy/pull/3289 to fix the windows test failure.

athakarag:
> @athakarag could you if this works for you?

@pnuu I will check it out as soon as I find the time. Thank you.

PR #3274, opened 2025-10-24
add time_parameters attribute containing wind_time to EUMETSAT FCI AMV nc reader

### LINKED ISSUE #3037: Modify time attributes for EUMETSAT AMV products
When loading a FCI AMV NetCDF product witht he `fci_l2_nc` reader the time attributes represent the following:

- `start_time`: start_time of the first image/repeat cycle used for the wind tracking
- `end_time`: start_time of the last image/repeat cycle used for the wind tracking

These are the timestamps stored in the FCI L2 AMV filenames and follows the WMO naming convention. Although it would make sense to have `end_time` representing the **end_time** of the last image used for the wind tracking, it's better to follow the WMO convention and keep it as it is.

However, it would be useful to also make the `wind_time` available. This is a separate dataset available in the AMV product and represented the start_time of the second repeat cycle used to extract the  cloud properties used for cloud detection and height assignment of the winds.

This should ideally be done for all EUMETSAT AMV readers: `fci_l2_nc`, `fci_l2_bufr` and `seviri_l2_bufr`.

Note: revised to be in-line with the latest understanding of the AMV product filename convention.

TristanCantatGaudin:
This PR adds a `time_parameters` attribute to all FCI L2 AMV datasets. This attribute is a dictionary, for the moment only containing the key `wind_time`. In the EUMETSAT AMV products this information is given as a dataset, in number of seconds since 2000-01-01. In this new implementation, `wind_time` is a datetime object. This attribute name and structure are consistent with existing implementation for e.g. [satpy/readers/seviri_l1b_nc.py](https://github.com/pytroll/satpy/blob/main/satpy/readers/seviri_l1b_nc.py).

In the future, `time_parameters` may be expanded to fit other relevant time-related values, since AMV products rely on multiple images and the current convention is not able to encode all information in the file name.

 - [x] Closes #3037 
 - [x] Test updated to cover the extra attribute of AMV products (and handle KeyError if missing `wind_time`)
 - [x] Fully documented (docstring of `satpy.readers.fci_l2_nc.FciL2CommonFunctions._get_global_attributes`)
 - [x] Add your name to AUTHORS.md

### DIFF
--- a/satpy/writers/ninjogeotiff.py
+++ b/satpy/writers/ninjogeotiff.py
@@ -439,14 +439,14 @@ class NinJoTagGenerator:
     def get_min_gray_value(self):
         """Calculate minimum gray value."""
         return self.image._scale_to_dtype(
-            self.dataset.min(),
+            self.dataset.min(keep_attrs=False),
             np.uint8,
             self.fill_value).astype(np.uint8)
 
     def get_max_gray_value(self):
         """Calculate maximum gray value."""
         return self.image._scale_to_dtype(
-            self.dataset.max(),
+            self.dataset.max(keep_attrs=False),
             np.uint8,
             self.fill_value).astype(np.uint8)
 


satpy/readers/fci_l2_nc.py
```diff
@@ -469,6 +472,21 @@ def nc(self):
             }
         )
 
+    @property
+    def wind_time(self):
+        """Return wind time as a datetime object.
+
+        This quantity is already present as a wind_time dataset,
+        in units of Seconds since 2000-01-01 00:00:00.0.
```
strandgren, line 480: Maybe also mention what the time represents, like this:

```
This quantity represents the `start_time` of the second image used for the AMV tracking and is already present as a wind_time dataset, in units of seconds since 2000-01-01 00:00:00.0.
```

TristanCantatGaudin: Done. One docstring was formatted wrong, returning a warning (treated as error) in the ReadTheDocs pipeline. Now all tests pass.

pnuu: So this would be the `nominal_time` of the latter FCI image used in the retrieval, not the "actual" sensing time of the location? I assume the retrieval is done as a continuous field with a optical flow approach or such, so that it would be hard to say what's the exact sensing time for each retrieval. So, if I got this right, should the naming be `nominal_wind_time`?

YouvaEUMex: no not optical flow involved in the AMVs retrieval (unlike 3Dwinds). 
Basically, We use 3 MTG FCI images for the MTG AMVs:  RC-1, RC0 and RC1. 

We track backward from RC0 to RC-1 and invert the direction afterwards. Then,  we track onward from RC0 to RC1. 
This allow to start form the same RC0 feature that we can then track in the intermediate and Final product. 

RC0 being the "center" is also why we used this RC for the the height assignment (using cloud product) and hence why it is defined as `wind_time` .

strandgren:
Looks good to me, thanks for adding this information!

Issue #3202, opened 2025-08-19
Rename VII readers and update VII readers for new test data


djhoese:
It has been pointed out that the VII L1b reader no longer works for the publicly available test data, but also doesn't work for the newest EUMETSAT internal test data. The reader needs updates to work for the future versions of the data files.

Additionally, the "VII" name is discontinued and "METimage" should be used instead. This would mean the new reader names should be "metimage_l1b_nc" and "metimage_l2_nc" and the old names deprecated (see [these dictionaries](https://github.com/pytroll/satpy/blob/0beea130092f3270ae1af8a87644d47971acbbc4/satpy/readers/core/config.py#L33-L36)).

I'm assigning this to Andrea, but understand that other EUMETSAT employees may be doing the work.

Completing this work will also close my related Polar2Grid issue: https://github.com/ssec/polar2grid/issues/767

djhoese:
I also marked this as "pre-launch" even though technically things are launched, but they aren't operational.

ameraner:
Ping @ludwigVonKoopa @cesclc

cesclc:
In satpy/readers/vii_base_nc.py:216
The start/end_time format need to be modified to fit latest L1B global attribute (%Y-%m-%dT%H:%M:%S.%f) as such:

```
@property
def start_time(self):
    """Get observation start time."""
    try:
        start_time = dt.datetime.strptime(self["/attr/sensing_start_time_utc"], "%Y-%m-%dT%H:%M:%S.%f")
    except ValueError:
        try:
            start_time = dt.datetime.strptime(self["/attr/sensing_start_time_utc"], "%Y%m%d%H%M%S.%f")
        except ValueError:
            start_time = dt.datetime.strptime(self["/attr/sensing_start_time_utc"], "%Y-%m-%d %H:%M:%S.%f")
    return start_time

@property
def end_time(self):
    """Get observation end time."""
    try:
        end_time = dt.datetime.strptime(self["/attr/sensing_end_time_utc"], "%Y-%m-%dT%H:%M:%S.%f")
    except ValueError:
        try:
            end_time = dt.datetime.strptime(self["/attr/sensing_end_time_utc"], "%Y%m%d%H%M%S.%f")
        except ValueError:
            end_time = dt.datetime.strptime(self["/attr/sensing_end_time_utc"], "%Y-%m-%d %H:%M:%S.%f")
    return end_time
```

ameraner:
@cesclc PRs are very welcome ;)

PR #3277, opened 2025-10-28
VII Reader Fixes and VII Composites Updates

cesclc:
Updates the VII reader stack so recent changes are fully covered and exercised. The reader now leaves the solar-zenith correction to the compositor (i.e., correction was previously performed twice), tolerates multiple filename layouts for L1B NetCDF products, and parses a wider range of sensing start/end timestamps attributes inside NetCDF. Adds the new VII daytime-only and night-time composites along with their YAML wiring, refreshes the reader unit tests to match the calibration and timestamp updates.

- [x] Contributes to #3202 (not closing it yet due to pending name change)
- [x]  Unit tests for VII readers updated
- [x]  AUTHORS.md updated

### DIFF
--- a/satpy/tests/reader_tests/test_vii_base_nc.py
+++ b/satpy/tests/reader_tests/test_vii_base_nc.py
@@ -216,13 +216,8 @@ class TestViiNCBaseFileHandler(unittest.TestCase):
                         equal = global_attributes[key][inner_key] == expected_global_attributes[key][inner_key]
                     assert equal
 
-    @mock.patch("satpy.readers.core.vii_nc.ViiNCBaseFileHandler._perform_geo_interpolation")
-    def test_start_end_time_additional_formats(self, pgi_):
+    def test_start_end_time_additional_formats(self):
         """Test parsing additional datetime formats."""
-        interp_longitude = xr.DataArray(np.ones((10, 100)))
-        interp_latitude = xr.DataArray(np.ones((10, 100)) * 2.)
-        pgi_.return_value = (interp_longitude, interp_latitude)
-
         time_cases = [
             (
                 "2017-09-20T17:30:40.888000",
@@ -238,11 +233,6 @@ class TestViiNCBaseFileHandler(unittest.TestCase):
             ),
         ]
 
-        filetype_info = {
-            "cached_longitude": "data/measurement_data/longitude",
-            "cached_latitude": "data/measurement_data/latitude",
-        }
-
         for start_str, expected_start, end_str, expected_end in time_cases:
             with Dataset(self.test_file_name, "r+") as nc:
                 nc.sensing_start_time_utc = start_str
@@ -251,11 +241,28 @@ class TestViiNCBaseFileHandler(unittest.TestCase):
             reader = ViiNCBaseFileHandler(
                 filename=self.test_file_name,
                 filename_info=self.filename_info,
-                filetype_info=filetype_info,
+                filetype_info={},
             )
 
             assert reader.start_time == expected_start
             assert reader.end_time == expected_end
+
+    def test_bad_start_end_time(self):
+        """Test parsing a bad datetime format."""
+        with Dataset(self.test_file_name, "r+") as nc:
+            nc.sensing_start_time_utc = "201709201730"
+            nc.sensing_end_time_utc = "201709201740"
+
+        reader = ViiNCBaseFileHandler(
+            filename=self.test_file_name,
+            filename_info=self.filename_info,
+            filetype_info={},
+        )
+
+        with pytest.raises(ValueError, match="Unrecognized datetime format"):
+            reader.start_time
+        with pytest.raises(ValueError, match="Unrecognized datetime format"):
+            reader.end_time
 
     @mock.patch("satpy.readers.core.vii_nc.tie_points_interpolation")
     @mock.patch("satpy.readers.core.vii_nc.tie_points_geo_interpolation")


ameraner:
pre-commit.ci autofix

satpy/readers/core/vii_nc.py
```diff
@@ -213,20 +213,41 @@ def _get_global_attributes(self):
...
+        ]
+
+        for fmt in formats:
+            try:
+                return dt.datetime.strptime(raw, fmt)
+            except ValueError:
+                continue  # try next format
+
+        # if none worked, raise an explicit error
+        raise ValueError(f"Unrecognized datetime format: {raw}")
```
ameraner: As CodeScene is suggesting, please refactor this part to avoid the duplication in the `start_time`

satpy/readers/core/vii_nc.py
```diff
@@ -213,20 +213,41 @@ def _get_global_attributes(self):
...
-        except ValueError:
-            end_time = dt.datetime.strptime(self["/attr/sensing_end_time_utc"], "%Y-%m-%d %H:%M:%S.%f")
-        return end_time
+        raw = self["/attr/sensing_end_time_utc"]
+        formats = [
+            "%Y%m%d%H%M%S.%f",        # e.g. 20250924121530.123456
+            "%Y-%m-%dT%H:%M:%S.%f",   # e.g. 2025-09-24T12:15:30.123456
+            "%Y-%m-%d %H:%M:%S",      # e.g. 2025-09-24 12:15:30
+            "%Y-%m-%d %H:%M:%S.%f",   # e.g. 2025-09-24 12:15:30.123456
+        ]
```
ameraner: can we expect that once real data is out and people stop using test data, this workaround of testing multiple patterns can be removed? If that's the case, I'd add a comment to the code indicating this, so that it can be simplified in the future when not needed anymore

satpy/readers/vii_l1b_nc.py
```diff
@@ -143,5 +143,5 @@ def _calibrate_refl(radiance: np.ndarray, angle_factor: np.ndarray, isi: float)
             array containing the calibrated reflectance values.
 
         """
-        refl_values = (np.pi / isi) * angle_factor * radiance * 100.0
+        refl_values = (np.pi / isi) * radiance * 100.0
```
ameraner, line 141: looks like the `angle_factor` is now unused in the method arguments so it should be removed (and anywhere else if not used anymore, up to the init)

satpy/etc/composites/vii.yaml
```diff
@@ -107,3 +107,40 @@ composites:
...
+        True Color during daytime
+      compositor: !!python/name:satpy.composites.fill.DayNightCompositor
+      standard_name: image_ready
+      lim_low: 78
+      lim_high: 88
+      day_night: day_only
+      prerequisites:
+        - true_color_uncorrected
+  
+  night_10695:
```
ameraner: do you need this one? I think you can use the channel `vii_10690` directly, if you don't need a specific enhancement for it

howff:
When I test this with `VII_L1B_CrossRef_TDP_V2` it complains about missing key `data/calibration_data/band_averaged_solar_irradiance`. I can comment that out as a workaround...

However it always crashes with a Segmentation Fault when I try to write an image. Maybe this isn't a fault of the reader, I don't know, but it doesn't crash with any other data. I've tried these tricks and it still crashes:
```
ulimit -n 4096
export DASK_NUM_WORKERS=1
export OMP_NUM_THREADS=1
```

ameraner:
As discussed, we should also apply the name change from `vii_*` to `metimage_*` (in this PR or another one). We should also update the names here then https://github.com/pytroll/satpy/blob/main/doc/source/examples/vii_l1b_nc.rst.

ameraner:
> When I test this with `VII_L1B_CrossRef_TDP_V2` it complains about missing key `data/calibration_data/band_averaged_solar_irradiance`. I can comment that out as a workaround...
> 
> However it always crashes with a Segmentation Fault when I try to write an image. Maybe this isn't a fault of the reader, I don't know, but it doesn't crash with any other data. I've tried these tricks and it still crashes:
> 
> ```
> ulimit -n 4096
> export DASK_NUM_WORKERS=1
> export OMP_NUM_THREADS=1
> ```

Hi @howff , the KeyError is an incompatibility with the currently public test data. We are now adapting the reader so that it will work with real data.
The NetCDF issue is a tricky one - let's keep that discussion in the other issue that you found already.

satpy/readers/vii_l1b_nc.py
```diff
@@ -55,7 +55,6 @@ def __init__(self, filename, filename_info, filetype_info, **kwargs):
         solar_zenith_angle = self["data/measurement_data/solar_zenith"]
         solar_zenith_angle_on_pixels = self._perform_interpolation(solar_zenith_angle)
         solar_zenith_angle_on_pixels_radians = np.radians(solar_zenith_angle_on_pixels)
```
ameraner:
```suggestion
```
these lines are now unused and should be removed (ruff in pre-commit is also flagging this)

satpy/readers/core/vii_nc.py
```diff
@@ -42,6 +42,14 @@ class ViiNCBaseFileHandler(NetCDF4FileHandler):
 
     """
 
+    # In the futurem, only final format will be defined once operational data is available
```
ameraner:
```suggestion
    # In the future, only final format will be defined once operational data is available
```

ameraner:
Codecov is marking that we're not testing the ValueError for unrecognised timestamps - that should be an easy addition, could you maybe quickly add it?

EDIT: I just commited this change myself

Issue #3235, opened 2025-09-16
Request: Access to the `dataset_info` from `get_area_def` in a reader


simonrp84:
## Feature Request

At present, the `get_area_def` function in a satpy reader accepts the dataset id (`dsid`) as an argument. But, for a reader I'm in the process of creating I also need access to the `dataset_info` variable.
The reason for this is that the latter variable contains an attribute loaded from the reader YAML, which give the internal `path` to a particular variable in the file being read. In this case, the variable itself has projection information as an attribute and is therefore required to generate the area definition correctly.
I can hack this, by hardcoding some parameters, but this is not a long term solution - especially as in the next year or so all the Sentinel data will switch to the format I'm generating a reader for, so we'll need a fairly broad-brush solution rather than a hacky one :)

**Describe the solution you'd like**
`dataset_info` to be an additional argument passed to `get_area_def`

**Describe any changes to existing user workflow**
This should be transparent to the user and should not affect existing readers.

djhoese:
I wonder if we can make a solution for this backwards compatible by using the (admittedly hacky and not my preferred solution) of using the `inspect` library to see if the `get_area_def` method being called expects a 3rd argument (self, data_id, ds_info) and only then provide that `ds_info` dictionary.

PR #3287, opened 2025-11-10
Add a reader for EUMETSAT HSAF SEVIRI H60 and H63

### LINKED ISSUE #3282: Add a reader for EUMETSAT HSAF SEVIRI H60 and H63
## Feature Request: Add support for new H SAF product format

The H SAF product owners recently changed the output format of these products, and EUMETSAT requested support for the new format in EUMETView.

We have already implemented a corresponding Satpy plugin (reader + enhancement) for EUMETView’s Image Generation System.

EUMETSAT now asked us to integrate this reader and enhancement into Satpy.  
We’d appreciate support and guidance from the Pytroll team on how to best upstream this contribution.

albertbrotzer:
The H SAF product owners recently changed the output format of these products, and EUMETSAT requested support for the new format in EUMETView. 

 - [X] Closes #3282 
 - [X] Tests added  - not sure id this is sufficient
 - [X] Fully documented


ameraner (changes requested):
Thanks for this! I would propose to actually have only one yaml file for this reader, and you can have different file types inside the same yaml (see e.g. the other PR here https://github.com/pytroll/satpy/pull/2793/files#diff-4ce2f1c002c4dae06506b9ad57f5a0504f278028002abc6263d2cd90ae50c8c6 ). This way it's more user friendly as there is only one reader for all HSAF products, and it's easier to maintain. 

Also, there are a few things to fix from the pre-commit checks.

satpy/tests/enhancement_tests/test_hsaf.py
```diff
@@ -0,0 +1,61 @@
...
+            np.array([[0.0, 0.1, 0.2], [5.0, 10.0, 20.0]]),
+            dims=("y", "x"),
+            name="rr",
+            attrs={"standard_name": "rain_rate", "units": "mm/h"}
+        )
+
+    # def test_detect_hsaf(self):
+    #     enh = Enhancer()
+    #     enh.add_sensor_enhancements(["hsaf"])
+    #     enh.enhancement_tree.print_tree()
```
ameraner: I assume this can be removed?

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,201 @@
...
+# satpy is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+A reader for H-SAF blended SEVIRI / LEO MW precipitation files in NetCDF
+format. (H60B and H63)
```
ameraner: as reader documentation, it's normally enough to have the docstring here describing what the reader does and can/cannot support, or any other useful information. So feel free to expand a bit here if there is more relevant info or caveats.

ameraner:
Also, would it be possible to add support for the H90 product as well, if it's not a lot of work? That way we could close the other PR https://github.com/pytroll/satpy/pull/2793 for good

albertbrotzer:
Hi @ameraner, there’s now support for H90 products. It uses the same reader and enhancements as H60 and H63. Please let us know if there’s anything else missing. Thanks for the support!

satpy/tests/enhancement_tests/test_hsaf.py
ameraner: Done

satpy/readers/hsaf_nc.py
ameraner: done

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,220 @@
...
+    Meteosat IODC (Indian Ocean) disk.
+  * **H90** – Accumulated Precipitation at ground by blended MW and IR IODC
+    over the Meteosat IODC (Indian Ocean) disk.
+
+Notes:
+    - Externally compressed files with the ``.gz`` suffix are automatically
+      uncompressed to the same directory and deleted on reader close.
+    - The reader relies on area definitions provided in the accompanying YAML
+      configuration file.
+    - Supports ``rain_rate`` and ``q_ind#`` datasets.
```
ameraner:
```suggestion
    - Supports ``rain_rate`` and ``q_ind`` datasets.
```
I guess the hashtag is a typo?

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,220 @@
...
+# satpy is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# satpy.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Reader for H SAF blended precipitation products in NetCDF format.
+
+These products provide instantaneous precipitation rate estimates derived from
```
ameraner:
```suggestion
These products provide precipitation estimates derived from
```

satpy/etc/readers/hsaf_nc.yaml
```diff
@@ -0,0 +1,113 @@
+reader:
+  name: hsaf_nc
+  short_name: Hydrology SAF
+  long_name: "H SAF products in NetCDF4 format"
+  description: |
+    Reader configuration for H SAF H60, H63 and H90 NetCDF files containing
```
ameraner:
```suggestion
    Reader configuration for H SAF H60/H60B, H63 and H90 NetCDF files containing
```
the website lists the product as "H60B" now, so let's mention that for clarity

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,220 @@
...
+a blending of geostationary (GEO) infrared and low-Earth-orbit (LEO) microwave
+observations using the H SAF "Rapid Update" algorithm.
+
+Currently, this reader supports the following products:
+  * **H60B** – Blended GEO/IR and LEO/MW instantaneous precipitation over the
```
ameraner:
```suggestion
  * **H60/H60B** – Blended GEO/IR and LEO/MW instantaneous precipitation over the
```

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,220 @@
...
+        data = self._standarize_dims(data)
+
+        # Handle missing values
+        if "missing_value" in data.attrs:
+            data = data.where(data != data.attrs["missing_value"])
+
+        # Add metadata from YAML
+        for key, value in dataset_info.items():
+            if key == "area":
+                data.attrs[key] = _resolve_area(value)
```
ameraner: I'm not sure if the reader can actually work like this, namely by appending the area as attribute manually in `get_dataset` - if it does, I think it's by chance :D ... 
The way it's done for all GEO readers is by having a method called `get_area_def`, that satpy calls automatically to extract and assign the AreaDefinition; see third bullet point here https://satpy.readthedocs.io/en/stable/dev_guide/custom_reader.html#the-python-file and a simple example here: https://github.com/pytroll/satpy/blob/6cdfea4370ae6fe85de427d97b451e2d4be8ee3c/satpy/readers/li_l2_nc.py#L120-L124 

In your case, it can simply return `satpy.area.get_area_def(self.area_name_from_yaml)`, and you have to store the `self.area_name_from_yaml` already in `get_dataset`, where you have access to `dataset_info` ( https://github.com/pytroll/satpy/issues/3235, would have been nice to have this here)

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,220 @@
...
+
+class HSAFNCFileHandler(BaseFileHandler):
+    """Handle H SAF NetCDF files, with optional .gz external compression.
+
+    This file handler handles H SAF Instantaneous Rain Rate products which contain:
+    - rr: rain rate in mm/h
+    - qind: quality index in percent
+
+    The data is on a geostationary projection grid.
+    This file handler is tested with H SAF h60, h63 & h40 data in NetCDF format.
```
ameraner:
```suggestion
    This file handler is tested with H SAF h60/h60b, h63 & h90 data in NetCDF format.
```

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,220 @@
...
+        resolved_area = area_value
+
+    return resolved_area
+
+
+class HSAFNCFileHandler(BaseFileHandler):
+    """Handle H SAF NetCDF files, with optional .gz external compression.
+
+    This file handler handles H SAF Instantaneous Rain Rate products which contain:
+    - rr: rain rate in mm/h
```
ameraner:
```suggestion
    This file handler handles H SAF Rain products which contain:
    - rr: rain rate in mm/h (instantaneous) or mm (accumulated)
```

satpy/readers/hsaf_nc.py
ameraner: could you also please recap which data is available in the three file types regarding the area? it would probably be a good idea to at least check that the `lon_0` or `ssp` or `sub-satellite_longitude` parameter fits the one we extract from the area name..

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,220 @@
...
+    def combine_info(self, all_infos):
+        """Override super class method to prevent AreaDefinitions being replaced with SwathDefinitions."""
+        # Default combination (times, orbits, etc.)
+        combined_info = super().combine_info(all_infos)
+
+        # If 'area' is already an AreaDefinition, keep it
+        if "area" in all_infos[0] and hasattr(all_infos[0]["area"], "area_id"):
+            combined_info["area"] = all_infos[0]["area"]
+
+        return combined_info
```
ameraner: I think with the correct `get_area_def` implementation as described above, this should not be needed

ameraner (changes requested):
Thank you for the updates! I took a closer looks and gave some more comments. The area definition assignments needs to be revised, otherwise only minor things.

ameraner:
Also, seems that the gzip decompression things are not well coverd by tests, according to codecov...
Note that for some readers we delegate the decompression to the user using FSFile, e.g. see
https://github.com/pytroll/satpy/blob/3e2edf21f5f2e745fcdb2fe29afe3b1eb9f34eb9/satpy/readers/hrit_jma.py#L74-L90

maybe the same could be done here for simplicity?

ameraner:
Actually, @mraspaud pointed me to the `generic_open` context manager, that by using Upath it could actually handle the gz files automagically without extra code, see below (it's already used in a few readers):
https://github.com/pytroll/satpy/blob/6cdfea4370ae6fe85de427d97b451e2d4be8ee3c/satpy/readers/core/utils.py#L346-L350

satpy/readers/hsaf_nc.py
albertbrotzer: We will do this. We did not find an ```ssp``` variable in the data - is it same as `sub_satellite_longitude`? And then: Should we raise an exception if the values do not match?

ameraner: yes, it's the same. Is that always available? What is the difference between the files?

In case it differs, I would propose to raise a warning, and edit the `AreaDefinition` object as e.g. below, with `42` being the longitude extracted from the file. 

I'm proposing this since for example the IODC position has changed in the past, so making sure that at least the subsatellite longitude is correct is a good safeguard, if we don't have enough info to create a full AreaDefinition on the fly every time.

```python
from satpy.area import get_area_def

adef = get_area_def("msg_seviri_fes_3km")

pdict_new = adef.proj_dict.copy()
pdict_new['lon_0'] = 42

adef_new = adef.copy(projection=pdict_new)
```

ameraner: (edited my comment above with a more elegant solution)

albertbrotzer:
@ameraner @mraspaud the PR should be ready for review again. I think we addressed all comments and checks. Thanks again for your exceptional support and feedback!

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,157 @@
...
+    format.
+    """
+
+    def __init__(self, filename, filename_info, filetype_info):
+        """Create a wrapper to opens the nc file and store the nc data."""
+        super().__init__(filename, filename_info, filetype_info)
+        self._area_name = None
+        self._lon_0 = None
+        with generic_open(filename, mode="rb", compression="infer") as fp:
+            self._nc_data = xr.open_dataset(fp, chunks="auto", engine="h5netcdf").compute()
```
ameraner: I believe the `compute()` here is breaking the lazy reading of the data... If you remove it, does the reader still work, and we get dask arrays for the main variables? that should be the case..

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _get_global_attributes(self):
+        """Create a dictionary of global attributes to be added to all datasets."""
+        attributes = {
+            "filename": self.filename,
+            "start_time": self.start_time,
+            "end_time": self.end_time,
```
ameraner: I believe these are not needed, as they're added automatically through the property methods

ameraner (commented):
Thanks for the updates! Looks good - only one point on the use of `compute`, and a minor point on the attributes

satpy/readers/hsaf_nc.py
albertbrotzer: Thanks — you’re absolutely right, and I’m glad you pointed this out.

Removing the compute() lets the reader fail, because the file gets closed immediately after leaving the generic_open context, which breaks lazy Dask access.

I’ve now switched to a pattern that is probably used by other readers as well: keeping the file open for the lifetime of the reader and do the cleanup in the __del__ method ( see commit d2c22ef ) . With this approach, lazy loading works correctly, no eager compute() is needed and cleanup/close is performed by the contextmanager.

satpy/readers/hsaf_nc.py
albertbrotzer: Removed them - see 0c0b2ef

satpy/readers/hsaf_nc.py
ameraner: LGTM, @mraspaud or is there another trick?

ameraner (approved):
LGTM, thanks for taking the initiative to add this reader and all the updates!

ReadTheDocs is currently failing, but only due to a warning regarding not being able to fetch a numpy inventory due to HTTP error, so probably just a glitch and nothing related to the PR.

pnuu:
There are some failing tests. I think the others are glitches, but I think this I fixed in an earlier PR that was merged:
```
FAILED satpy/tests/test_utils.py::TestGetSatPos::test_get_satpos_from_satname - ImportError: cannot import name 'tlefile' from 'pyorbital.orbital' (/home/runner/miniconda3/envs/test-environment/lib/python3.11/site-packages/pyorbital/orbital.py)
```
The same error is in few other tests, but all of them hopefully go away if you merge the current `main` branch to your PR.

pnuu:
Ah, the fix is still unmerged in https://github.com/pytroll/satpy/pull/3294

albertbrotzer:
> LGTM, thanks for taking the initiative to add this reader and all the updates!
> 
> ReadTheDocs is currently failing, but only due to a warning regarding not being able to fetch a numpy inventory due to HTTP error, so probably just a glitch and nothing related to the PR.

Thanks to you all for supporting us in the process. I think we can learn a lot from you guys and we will for sure add more as soon as time allows.

pnuu:
Ok, we merged couple of PRs to `main` and main to your PR branch. Hopefully now all the tests will pass.

satpy/readers/hsaf_nc.py
mraspaud: I think this is fine!

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+    - rr or acc_rr: rain rate in mm/h (instantaneous) or mm (accumulated)
+    - qind: quality index in percent
+
+    The data is on a geostationary projection grid.
+    This file handler is tested with H SAF h60/h60b, h63 & h90 data in NetCDF
+    format.
+    """
+
+    def __init__(self, filename, filename_info, filetype_info):
+        """Opens the nc file and store the nc data."""
```
mraspaud:
```suggestion
        """Open the nc file and store the nc data."""
```

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+        super().__init__(filename, filename_info, filetype_info)
+        self._area_name = None
+        self._lon_0 = None
+        # use generic_open contextmanager to handle compression transparently
```
mraspaud:
```suggestion
```

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+        self._cm = generic_open(filename, mode="rb", compression="infer")
+        # manually enter the context manager
```
mraspaud:
```suggestion
```

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+        fp = self._cm.__enter__()
+        # lazily load nc file through h5netcdf without forcing read
```
mraspaud:
```suggestion
```

pnuu (approved):

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+        self._nc_data = xr.open_dataset(fp, engine="h5netcdf", chunks="auto")
+
+    def __del__(self):
+        """Instruct the context manager to clean up and close the file."""
+        try:
+            if hasattr(self, "_cm") and self._cm:
```
mraspaud:
```suggestion
```

I don't think this is needed: `self._cm` will we be present and have a value as long as the class has been instantiated. If not, this `__del__` function will never be called anyway.

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+                self._cm.__exit__(None, None, None)
+        except (AttributeError, RuntimeError, OSError, ValueError):
+            LOG.warning(f"An error occurred while cleaning up the file {self.filename}")
+
+    def _get_global_attributes(self):
+        """Create a dictionary of global attributes to be added to all datasets."""
+        attributes = {
+            "filename": self.filename,
+            "spacecraft_name": platform_translate.get(self._nc_data.attrs["satellite_identifier"], "N/A"),
+            "platform_name": platform_translate.get(self._nc_data.attrs["satellite_identifier"], "N/A"),
```
mraspaud: is `N/A` an acceptable value for these attributes? Does it ever happen that `satellite_identifier` is not defined?
Also, why do we need both `spacecraft_name` and `platform_name`?

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+    def get_dataset(self, dataset_id, dataset_info):
+        """Get a dataset from the file."""
+        # Get the variable
+        var_name = dataset_info.get("file_key", dataset_id["name"])
+        LOG.debug(f"Getting dataset {var_name} from file {self.filename}")
+        data = self._nc_data[var_name]
+
+        data = self._standarize_dims(data)
+
+        # Handle missing values
```
mraspaud:
```suggestion
```

I think the code is self explanatory.

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+            if key == "area":
+                self._area_name = value
+            elif key not in ("file_key",):
+                data.attrs[key] = value
+
+        data.attrs["resolution"] = self.filetype_info["resolution"]
+        self._area_name = self.filetype_info["area"]
+        self._lon_0 = float(self._nc_data.attrs["sub_satellite_longitude"].rstrip("f"))
+
+        # Add global attributes which are shared across datasets
```
mraspaud:
```suggestion
```

satpy/readers/hsaf_nc.py
```diff
@@ -0,0 +1,167 @@
...
+        data.attrs.update(self._get_global_attributes())
+
+        return data
+
+    @property
+    def end_time(self):
+        """Get end time."""
+        return self.start_time + dt.timedelta(minutes=15)
```
mraspaud: 15: So these products will never contain rss or fci data?

satpy/tests/reader_tests/test_hsaf_nc.py
```diff
@@ -0,0 +1,186 @@
...
+        [
+            (FILE_PARAMS[FILE_TYPE_H60], 1),
+            (FILE_PARAMS[FILE_TYPE_H63], 1),
+            (FILE_PARAMS[FILE_TYPE_H90], 1),
+        ],
+    )
+    def test_reader_creation(self, file_type, expected_loadables):
+        """Test that the reader can create file handlers."""
+        with mock.patch("satpy.readers.hsaf_nc.generic_open") as mock_generic_open, \
+                mock.patch("satpy.readers.hsaf_nc.xr.open_dataset") as mock_open_dataset:
```
mraspaud: We have now as policy to try to create a stub file in `tmp_path` instead of mocking. I think here, it's a simple matter of writing the xarray dataset to disk and use that as input file. With the added benefit that you can (g/b)zip it to test the decompression...

satpy/tests/reader_tests/test_hsaf_nc.py
```diff
@@ -0,0 +1,186 @@
...
+        ],
+    )
+    @pytest.mark.skipif(not os.path.exists(FILE_PARAMS[FILE_TYPE_H60]["real_file"]) or
+                        not os.path.exists(FILE_PARAMS[FILE_TYPE_H63]["real_file"]),
+                                           reason="Real HSAF file not present")
+    def test_real_hsaf_file(self, file_type, loadable_ids):
+        """Test the reader with a real HSAF NetCDF file."""
+        # Select files
+        loadables = file_type["reader"].select_files_from_pathnames([file_type["real_file"]])
+        assert loadables, "No loadables found for the real file"
```
mraspaud: when are these real files available? are they easily downloadable from somewhere so that the user can test this?
Otherwise, I think it's as good to create stub files that mimic these files closely.

satpy/etc/enhancements/hsaf.yaml
```diff
@@ -0,0 +1,39 @@
...
+                2,
+                3,
+                5,
+                7,
+                10,
+                15,
+                20,
+                30,
+                50
+              ]
```
mraspaud, line 38: Is this scale valid for both rain rate and accumulated rain rate?

mraspaud (changes requested):
Thanks a lot for the work on supporting these HSAF products!
I especially appreciate the clarity of the code.

I have some comments, questions and suggestions inline, but I think this is almost ready :)

satpy/etc/enhancements/hsaf.yaml
armelzer: The scale is for rain rate valid & is also used for the images in Eumetview for the rain rate (e.g. Blended SEVIRI / LEO MW precipitation and morphologic information - MSG - 0 degree ). We assumed it is also valid for accumulated rain rate and during our test visualisations it looked also good for us.

Here is an image with test visualisations. (Right is H60B and the left image is H90):

Bildschirmfoto vom 2025-11-21 11-10-35

mraspaud: I'm no expert here, @pnuu @ameraner what do you say?

pnuu: In principle the same colormap can work for both. It depends a lot on the period over which the accumulated product is defined for. Over 24 hours it might need some more values at the top end to not saturate too early.

satpy/readers/hsaf_nc.py
armelzer: As far as I could see, `satellite_identifier` was available in every file I opened with Panoply. So I will change it, if `satellite_identifier` is not available in `platform_translate`, the value of `satellite_identifier` will be used instead. 
Would it be better just to use the `platform_name` instead then?

satpy/readers/hsaf_nc.py
ameraner: the products supported so far are all SEVIRI full-discs - when someone will add support for the FCI counterparts, this will need some refactoring

satpy/etc/enhancements/hsaf.yaml
ameraner: good point - I see that we have the same `standard_name` for both variables in the yaml... @armelzer is the `standard_name` defined in the files and is it the same there? I wonder if we should differentiate and use `instantaneous_rain_rate` and `accumulated_rain_rate` or so for `standard_name`, and then have the enhancement work only for `instantaneous_rain_rate` since that's what the current enhancement is designed for. The H90 comes in hourly and daily accumulations, so I also assume that the daily one will saturate early.

armelzer:
@ameraner @mraspaud The `standard_name` has been changed as suggested for `rr` and `acc_rr`, and I think we should also take into account all other comments/suggestions made so far. We have not made any specific improvements to `acc_rr`, as we are not experts in this field and in [another PR ](https://github.com/pytroll/satpy/pull/2793/files#diff-4ce2f1c002c4dae06506b9ad57f5a0504f278028002abc6263d2cd90ae50c8c6) .In addition, the delta value (in minutes) for the ‘end_time’ is now configurable across file types.

So thanks again for your exceptional support and feedback (also to @pnuu)!

satpy/readers/hsaf_nc.py
mraspaud: `platform_name` is what we usually use

satpy/readers/hsaf_nc.py
mraspaud: ok, how do we keep track of this?

ameraner: looks like Alexandra made it configurable now, so it's solved I'd say

mraspaud (approved):
LGTM

Issue #885, opened 2019-08-23
Standardize platform and sensor names


djhoese:
This is in addition to #749.

There is a bit of inconsistency between satpy and pyspectral about what the name of a platform or sensor should be. We also decided early on in Satpy that composite filenames would be based on the sensor name and lowercased all of them. So this means we have readers like "mersi2_l1b", metadata with sensor "mersi-2" and platform names of "FY-3D", composite configs named "mersi-2.yaml", and pyspectral preferring the WMO OSCAR name of "MERSI-2".

As part of 1.0, we should break backwards compatibility and standardize all of this so things are consistent and follow some official resource for naming.

BENR0:
See also #3008 (and also #2994)

PR #3291, opened 2025-11-11
Add config switch for OSCAR compliant attributes

sfinkens:
Add a config switch for toggling OSCAR compliant attributes (`platform_name`, `sensor`) and adapt ABI readers to see what happens.

```python
with satpy.config.set(oscar_compliant_attributes=True):
    scene = Scene(...)
```

The ABI modifications should probably be reverted and applied together with all readers once we've agreed on the way forward.

### Deprecation Strategy

Changing platform/instrument names has a high potential of breaking users' workflows because they are often used in filenames. So we decided to add a config switch for toggling OSCAR compliant attributes. With satpy-1.0 this would default to True, but we keep it for a couple of releases so that users can get the legacy attributes back if they need more time to adapt.

### Implementation

- Change all file handlers to provide OSCAR compliant dataset attributes. In `FileYAMLReader` revert them back to legacy attributes if desired.
- Update platform/sensor name in yaml definitions (readers, composites, enhancements), but keep the filenames as they are. This requires two additions
    1. Translation from OSCAR sensor name to Satpy filename. 
    2. A new `dependencies` attribute in the composites definition. So far dependencies were specified using forward slashes `sensor_name: dep1/dep2/sensor`. But since OSCAR sensor names can contain slashes (`AVHRR/1`) this doesn't work anymore.
- Mark changes with `8< v1.0` sciccors so that they can be removed using a script.

Notes:
- As far as I can see, composites get their sensor attribute from dataset attributes and not from yaml definitions. So the config switch would also work for composites.

### Upstream Changes

Satpy passes dataset attributes to pyspectral which internally works with OSCAR names, but expects lowercase names in the interface. So that needs to be updated/deprecated as well. Ideally Satpy would also set the pyspectral config switch. _For the moment I just lowered the sensor name before passing it to pyspectral._

### Failed Approaches

- Use OSCAR attributes only in dataset attributes and keep yaml files as they are. Problem: `scene.sensor_names` would contain both `ABI` and `abi`.
- Adding `sensor` and `platform_name` properties to the file handlers. Problem: A file may contain data from multiple sensors and sometimes the sensor is only known after loading the dataset. 

### Class Diagram

```mermaid
---
config:
  theme: 'default'
---
classDiagram
    class Scene {
        +sensor_names [ABI]
        +available_composite_names()
    }
    
    class Reader {
        +sensor_names [ABI]
    }
    
    class ReaderYAML["readers/abi_l1b.yaml"] {
        sensors: [ABI]
    }
    
    class DataArray {
        +sensor ABI
        +platform_name GOES-18
    }
    
    class CompositesYAML["composites/abi.yaml"] {
        sensor_name: ABI
        dependencies: [visir] 🆕
    }
    
    class LoadComp["config_loader.py"] {
        +load_compositor_configs_for_sensor()
        +sensor_to_filename() ABI -> abi.yaml 🆕
    }
    
    Scene--> Reader : Collects sensor names from
    Scene --> DataArray : Collects sensor names from
    Scene --> LoadComp : Passes sensor names to
    Reader --> ReaderYAML : Reads sensors from
    LoadComp --> CompositesYAML : Reads composites & dependencies from

```

 - [X] Prepares for #885  
 - [ ] Wait for upstream changes in pyspectral
 - [ ] Remove `sensor.lower()` hack
 - [ ] Remove ABI modifications
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,10 +18,10 @@ jobs:
       fail-fast: true
       matrix:
         os: ["windows-latest", "ubuntu-latest", "macos-latest"]
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.11", "3.12", "3.13"]
         experimental: [false]
         include:
-          - python-version: "3.12"
+          - python-version: "3.13"
             os: "ubuntu-latest"
             experimental: true
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "universal_pathlib",
 ]
 readme = "README.rst"
-requires-python = ">=3.10"
+requires-python = ">=3.11"
 license = { text = "GPLv3" }
 classifiers = [
     "Development Status :: 5 - Production/Stable",
--- a/satpy/_config.py
+++ b/satpy/_config.py
@@ -54,6 +54,7 @@ _CONFIG_DEFAULTS = {
     "readers": {
         "clip_negative_radiances": False,
     },
+    "oscar_compliant_attributes": False,
 }
 
 # Satpy main configuration object
--- a/satpy/_config.py
+++ b/satpy/_config.py
@@ -54,7 +54,9 @@ _CONFIG_DEFAULTS = {
     "readers": {
         "clip_negative_radiances": False,
     },
+    # 8< v1.0
     "oscar_compliant_attributes": False,
+    # >8 v1.0
 }
 
 # Satpy main configuration object
--- a/satpy/composites/config_loader.py
+++ b/satpy/composites/config_loader.py
@@ -192,7 +192,7 @@ def _load_config(composite_configs):
     sensor_modifiers = {}
 
     dep_id_keys = None
-    sensor_deps = sensor_name.split("/")[:-1]
+    sensor_deps = conf.get("dependencies", sensor_name.split("/")[:-1])
     if sensor_deps:
         # get dependent
         for sensor_dep in sensor_deps:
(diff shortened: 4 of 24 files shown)


djhoese:
Is failing CI expected at this point?

sfinkens:
> Is failing CI expected at this point?

No, locally these tests pass. I tried merging main but that didn't help.

pnuu:
Seemed like Pyspectral LUT download had failed, again. I restarted the builds.

mraspaud:
Sorry it took so long for me to have a proper look at this.
I think it looks really nice, and I would vote for going forward with this.

Just one small thing: Would it be possible to have a small private function (eg `_convert_to_pyspectral_name`) that does the conversion to pyspectral names? It would just do `.lower()` for now, but when pyspectral is updated it will make it easier to find the place where to implement the transition.

sfinkens:
> Would it be possible to have a small private function (eg _convert_to_pyspectral_name)

Yes, good idea!

sfinkens:
I've added the pyspectral conversion helper, turned the config switch off for now and removed the ABI modifications. In the next PR I would then update all the readers.

djhoese (commented):
Tests are failing...I assume (hope) that's due to your changes.

sfinkens:
> Tests are failing...

Yes, some readers provide sensor as set type. I've taken this into account for now. Since the sets actually contain only one sensor, it might make sense to update those readers v1.0 to return a string as well.

djhoese:
It is perfectly valid though for readers to return a set for sensors. Obviously a set with a single item is unnecessary, but it should still be supported in my opinion. Some readers provide data for multiple sensors so the set support is needed anyway.

sfinkens:
Hmm, but what if the dataset has been loaded? Can this be multi-sensor, too? I'm just referring to the dataset attribute `sensor` here. And if it is multi-sensor, how can this information be passed to pyspectral?

sfinkens:
Summary from slack discussion: It is possible for higher levels of processing where data from multiple sensors is used to generate a product. Those won't be passed to pyspectral, though. 

Although the set/str mix is not perfectly consistent, storing all sensor attributes as a set is probably confusing new users/contributors and also breaks even more workflows than changing to WMO names.

sfinkens:
Superseded by https://github.com/pytroll/satpy/pull/3390

PR #3294, opened 2025-11-12
Require Numpy 2.0 or later

### LINKED ISSUE #3268: Drop numpy <2.0 support
https://scientific-python.org/specs/spec-0000/

This doesn't change much specific to Satpy which is pure python but would make dependency package support easier/simpler.

pnuu:
This PR drops Numpy 1.x support from Satpy.

There ~are~ were two functions in `satpy.tests.utils` that ~do~ did checks for Skyfield and h5py and their Numpy 2 support which  ~might need checking~ were removed. The removal of Skyfield also revealed a bug in `satpy.utils._get_satpos_from_platform_name` (incorrect import path).

 - [x] Closes #3268

### DIFF
--- a/satpy/readers/fci_l1c_nc.py
+++ b/satpy/readers/fci_l1c_nc.py
@@ -333,11 +333,11 @@ class FCIL1cNCFileHandler(NetCDF4FsspecFileHandler):
         """Load a dataset."""
         logger.debug("Reading {} from {}".format(key["name"], self.filename))
         if "pixel_quality" in key["name"]:
-            return self._get_dataset_quality(key["name"])
+            return self._get_dataset_quality(key, info=info)
         elif "index_map" in key["name"]:
-            return self._get_dataset_index_map(key["name"])
+            return self._get_dataset_index_map(key, info=info)
         elif _get_aux_data_name_from_dsname(key["name"]) is not None:
-            return self._get_dataset_aux_data(key["name"])
+            return self._get_dataset_aux_data(key, info=info)
         elif any(lb in key["name"] for lb in {"vis_", "ir_", "nir_", "wv_"}):
             return self._get_dataset_measurand(key, info=info)
         else:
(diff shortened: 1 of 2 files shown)


pnuu:
Looks like removing the `xfail_skyfield_unstable_numpy2` decorator reveals some incorrect imports and unnecessary TLE download. I'll see if I find a simple fix for those to be included in this PR.

satpy/_compat.py
```diff
@@ -19,9 +19,4 @@
 
 from functools import cache, cached_property  # noqa
 
-try:
-    from numpy.typing import ArrayLike, DTypeLike  # noqa
-except ImportError:
-    # numpy <1.20
-    from numpy import dtype as DTypeLike  # noqa
-    from numpy import ndarray as ArrayLike  # noqa
+from numpy.typing import ArrayLike, DTypeLike  # noqa
```
djhoese: We could/should probably find all imports of these and replace them with the `numpy.typing` import rather than this `_compat` module.

satpy/utils.py
```diff
@@ -418,7 +418,7 @@ def _get_satpos_from_platform_name(cth_dataset):
 
     Needs pyorbital, skyfield, and astropy to be installed.
     """
-    from pyorbital.orbital import tlefile
+    from pyorbital import tlefile
```
djhoese, line 420: Oh this happened from one of my pyorbital PRs (too lazy to find it now). I fixed it in one of my other Satpy PRs that I thought would be merged quickly...and then wasn't. Sorry about that.

Anyway, in `pyorbital/orbital.py` it was importing `tlefile` and instead of importing the actual/original `pyorbital/tlefile.py` this Satpy code was importing the imported instance of the module. I removed that global import of `tlefile` in `orbital.py` until it is actually needed.

satpy/_compat.py
pnuu: https://github.com/pytroll/satpy/pull/3294/commits/0e1ac868c7e1e2d5c27fb3f87dbf7cd72b8765c5 should replace all the uses with direct Numpy imports

satpy/utils.py
pnuu: Ok, definitely a good change. I guess I saw the failures on my local test because I had Pyorbital `main` installed locally, and in the other PRs this emerged just now as I made a Pyorbital release yesterday with your changes in it.

djhoese:
Should this maybe have the backwards incompatible and enhancement labels added to it? I don't think "dependencies" alone will make it show up in the changelog.

PR #3299, opened 2025-11-14
Change area names to new naming scheme and remove some areas.

BENR0:
This PR changes area names for almost all areas to the naming scheme proposed in #2310 and removes areas listed by @mraspaud in https://docs.google.com/document/d/1LOGtyu8qjHiBcEhDToIgUPjUUlJalpGNgo-hpmBsb_k/edit?tab=t.0#heading=h.5adsx7apljox

Further possible areas which could be deprecated/replaced by other areas (please discuss):

- met07globe
- met09globe
- msg_resample_area

  - [x] Closes
 	 -  #2310
 	 -  #1207
 	 -  #1248
 	 -  #819
 	 -  #1940  

 - [ ] Tests added 
 - [ ] Fully documented 

Notes for the "far" ahead future removal of old area names from areas.yaml and the introduced decorator:
- In the areas.yaml file the newly introduced key `old_name` can easily be removed for example with sed because it is only one line to be deleted in the corresponding areas
- once the above is done the decorator and the corresponding decorator function can be removed from area.py


satpy/area.py
```diff
@@ -29,7 +33,37 @@ def get_area_file():
...
+    def area_with_new_name(area_name):
+        try:
+            return fun(area_name)
+        except AreaNotFound:
+            area_file_name = get_area_file()
+            areas_dict = _read_yaml_area_file_content(area_file_name)
+            area_dict = None
+            for new_area_name, v in areas_dict.items():
+
+                if "old_name" in v.keys() and area_name in v["old_name"]:# == area_name:
```
gerritholl, line 50: Comment should be removed.

satpy/area.py
```diff
@@ -29,7 +33,37 @@ def get_area_file():
...
+                    area_dict = areas_dict.get(new_area_name)
+                    del area_dict["old_name"]
+                    break
+
+            if area_dict is not None:
+                warnings.simplefilter("always", DeprecationWarning)
+                warnings.warn(f"This area name is being deprecated in Satpy v1.0. Pleas use the new area name:"
+                              f" {new_area_name}", DeprecationWarning, stacklevel=2)
+                warnings.simplefilter("default", DeprecationWarning)
```
gerritholl, line 59: This should use the [`warnings.catch_warnings`](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) context manager, so any previously existing user settings are not lost.

satpy/area.py
```diff
@@ -29,7 +33,37 @@ def get_area_file():
...
+                return _create_area_def_from_dict(new_area_name, area_dict)
+            else:
+                raise AreaNotFound('Area "{0}" not found in file "{1}"'.format(area_name, area_file_name))
```
gerritholl, line 62: This would be easier to read with an f-strring.

satpy/area.py
```diff
@@ -29,7 +33,37 @@ def get_area_file():
...
+
+    return area_with_new_name
 
+@deprecate_area_name
```
gerritholl, line 66: I find it confusing to define a decorator that is used exactly once and unlikely to be reused elsewhere.  Wouldn't it be clearer to simply integrate this into `get_area_def`, then refactoring that function if it becomes too large?

satpy/area.py
```diff
@@ -29,7 +33,37 @@ def get_area_file():
...
+            for new_area_name, v in areas_dict.items():
+
+                if "old_name" in v.keys() and area_name in v["old_name"]:# == area_name:
+                    area_dict = areas_dict.get(new_area_name)
+                    del area_dict["old_name"]
+                    break
+
+            if area_dict is not None:
+                warnings.simplefilter("always", DeprecationWarning)
+                warnings.warn(f"This area name is being deprecated in Satpy v1.0. Pleas use the new area name:"
```
gerritholl, line 57:
```suggestion
                warnings.warn(f"This area name is being deprecated in Satpy v1.0. Please use the new area name:"
```

satpy/etc/areas.yaml
```diff
@@ -1301,7 +836,8 @@ eurol:
     lower_left_xy: [-3780000.0, -7644000.0]
     upper_right_xy: [3900000.0, -1500000.0]
 
-eurol1:
+eurol1_stere_0.8km:
+  old_name: eurol1
   description: Euro 3.0km area - Europe
```
gerritholl, line 841: Name says 0.8 km, description says 3 km.  Which one is it?

satpy/etc/areas.yaml
```diff
@@ -1765,10 +1014,9 @@ euroasia_10km:
     lower_left_xy: [-3000000.0, -4999000.0]
     upper_right_xy: [9999000.0, 8000000.0]
 
-euroasia_asia:
-  description:
-    Euroasia - optimised for Asia -
-    Global 1km USGS Landuse database
+euroasia_asia_laea_1km:
```
gerritholl, line 1017: Having Asia in the name twice is confusing.

satpy/etc/areas.yaml
```diff
@@ -1782,8 +1030,9 @@ euroasia_asia:
     lower_left_xy: [-8000000.0, -5499000.0]
     upper_right_xy: [4999000.0, 6500000.0]
 
-euroasia_asia_10km:
-  description: Euroasia - optimised for Asia - Global 1km USGS Landuse database
+euroasia_asia_laea_10km:
```
gerritholl, line 1033: Having Asia in the name twice is confusing.

satpy/etc/areas.yaml
```diff
@@ -1917,108 +1174,9 @@ northamerica_10km:
...
-    width: 5600
-  area_extent:
-    lower_left_xy: [-475617.0, 5324430.0]
-    upper_right_xy: [924383.0, 6724430.0]
-
-moll:
-  description: moll
+global_moll_10km:
+  old_name: moll
+  description: Globe in Mollweide projection
```
gerritholl, line 1179: Maybe "world" is a better term here than "globe"?  To me a globe is a spheroid, (almost) unprojected in lat/lon coordinates.

satpy/etc/areas.yaml
```diff
@@ -2031,8 +1189,9 @@ moll:
     lower_left_xy: [-18040095.696147293, -9020047.848073646]
     upper_right_xy: [18040095.696147293, 9020047.848073646]
 
-robinson:
-  description: robinson
+greenland_robin_1km:
+  old_name: robinson
+  description: Greenland in Robinson projection
```
gerritholl, line 1194: I would say that if the Robinson projection contained only Greenland and not the whole world, this counts as a bug...

satpy/etc/areas.yaml
```diff
@@ -2141,7 +1255,8 @@ msg_resample_area:
     upper_right_xy: [5567248.074173444, 5570248.477339261]
 
 # Global lat / lon gridded areas
-EPSG_4326_36000x18000:
+EPSG_4326_0.01deg:
+  old_name: EPSG_4326_36000x18000
```
gerritholl, line 1259: Then this too should have `world` in the name?

gerritholl (changes requested):
Looks good, some small corrections to be implemented.  The use of a decorator function is a little odd to me.

I wonder if we can do something about deleted areas.

PR #3300, opened 2025-11-14
Landsat MSS reader `_get_matched_dataset_info` bug fix

simonreise:
Function `_get_matched_dataset_info` should return only dataset info, not a tuple of (True, info)

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/core/landsat.py
+++ b/satpy/readers/core/landsat.py
@@ -454,7 +454,7 @@ class MSSCHReader(BaseLandsatL1Reader):
     def _get_matched_dataset_info(self, ds_info):
         if ds_info.get("name") == "B4":
             return self._get_modified_wavelength_info(ds_info)
-        return True, ds_info
+        return ds_info
 
     def _get_modified_wavelength_info(self, ds_info):
         # Modify the dataset's wavelength dynamically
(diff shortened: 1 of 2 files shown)


djhoese:
Thanks! ...but how has this ever worked?

simonreise:
Tests were not failing, because we test only B4 band, and for B4 it worked correctly, hahah

djhoese:
1. If you have it, could you paste the error message you were getting before you fixed this?
2. Would it be possible to add a test for this?

simonreise:
I already do not have an exact error message, but it crased on this line from `satpy/readers/core/yaml_reader.py`
```py
coordinates = ds_info.get("coordinates")
```
with error like `tuple does not have a "get" method`

We can add additional band tests, like testing also B5 for Landsat 1 and B3 for Landsat 4

simonreise:
I added testing "B5" band for Landsat 1 and "B3" for Landsat 4, by adding the corresponding fixtures for these bands and trying to open them in `test_basicload`. But now probably `thermal_name` arg should be renamed to something more general, as previously it was used only to open thermal bands, and now it is also used to open additional bands for MSS products

mraspaud (approved):
LGTM

PR #3301, opened 2025-11-16
Add OSCAT-3 support to scatsat1b reader

### LINKED ISSUE #2628: Adapt scatsat1b reader to oscat-3
## Feature Request

There is a new satellite OceanSat-3 https://space.oscar.wmo.int/satellites/view/oceansat_3_eos_06 with oscat-3 https://space.oscar.wmo.int/instruments/view/oscat_3 instrument. This is a technically improved version of the OSCAT instrument flown onboard Oceansat-2 and ScatSat-1.

I want to adapt this reader to also read this new data.

No testdata yet, but filename is given like this:
`E06SCTL2B2023306_04934_04935_SN_25km_2023-306T11-47-49_v1.0.2.h5.bz2`

Please assign this to me for the PWC this December

TAlonglong:
- [X] Closes #2628  
 - [X] Tests added 
 - [ ] Fully documented 


TAlonglong:
An additional consideration is: this reader can now also read data from other scatterometers, here from the satellite EOS-06 https://space.oscar.wmo.int/satellites/view/oceansat_3_eos_06

So maybe it should be renamed; ie from scatsat1 to scatsat.

If we agree to do this renaming, how should that be done. SHould we retire the scatsat1 reader and add a scatsat reader? Or just do the renmae. Both will be breaking changes I guess.

djhoese (commented):
I love that just switching the base class to the HDF5 helper makes this work.

My biggest concern is the `resolution` being moved from YAML to the `get_dataset` method. If you do `scn.available_dataset_ids()` my guess is the list of datasets won't include the resolution. If that is true, then I think you'd need to define a `available_datasets` method on the file handler and look at the `BaseFileHandler.available_datasets` docstring for help on how to implement it. In that method you would add resolution to the dataset definitions based on the `cell_spacing`.

As for renaming and support for other satellites, I think that could be done. Would the available datasets be exactly the same?

TAlonglong:
pre-commit.ci autofix

TAlonglong:
Not sure what happened with the tests now.

The old data is the same format, but 25 and 50km. I added this to reflect that.

PR #3302, opened 2025-11-17
Fix links to EUMETSAT documents

sfinkens:
EUMETSAT has moved documents, update links accordingly.


sfinkens:
@ameraner If you have time, can you please check that these are the official links?

ameraner (approved):
LGTM, thank you for looking into this! The links look like the correct ones to me, it seems that there is only one version of each document on the User Portal anyway.

djhoese (approved):
If these links are still valid since this is a couple months old then LGTM!

djhoese:
I just checked and all links work. Thanks @sfinkens! Squashing to avoid an extra merge commit in main's history.

PR #3306, opened 2025-11-21
Add `fsspec` support to `msi_safe` reader

simonreise:
The changes added in this PR are nearly identical to #3173

Zip archives are still the default distribution format for Sentinel-2 MSI_SAFE imagery, however, now SatPy `msi_safe` reader does not support it.

`fsspec` can be used to read data from archives, or, e.g. remote data sources like AWS or Planetary Computer.

The msi_safe reader file reading functionality is nearly identical to the one of Landsat readers, so I tried to implement fsspec support the same way.

However, I am stuck on adding tests, because in the current testing workflow no files are being used. For `XML` files the `StringIO` and `BytesIO` objects are being used (_however, we can write the content of these objects to files in the fsspec filesystem to test if it works_). For `jp2` files no reading is performed at all, instead of actual reading the function is just being mocked by `mock.path`, which returns the pre-defined fake data instead of loading something from file (_so, we either have to just leave it as-is or add actual loading functionality to the tests_).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Fully documented 

### DIFF
--- a/satpy/etc/readers/msi_safe.yaml
+++ b/satpy/etc/readers/msi_safe.yaml
@@ -4,7 +4,7 @@ reader:
   long_name: Sentinel-2 A and B MSI L1C data in SAFE format
   description: SAFE Reader for MSI L1C data (Sentinel-2)
   status: Nominal
-  supports_fsspec: false
+  supports_fsspec: true
   sensors: [sen2_msi]
   reader: !!python/name:satpy.readers.core.yaml_reader.FileYAMLReader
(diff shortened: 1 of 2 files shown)


simon-sat:
FYI, the Sentinel-2 file format will be changing next year to Zarr - so if this PR is problematic for you then I'd recommend considering how much time you're willing to spend on something that'll be obselete in 6 months :)

PR #3307, opened 2025-11-26
Fix VIIRS ERFDNB composite not properly rechunking data before map_blocks call

djhoese:
Test data was not sufficiently chunked to reveal an issue in the chunking before the map_blocks call. This used to be a Delayed function call but was switched recently to map_blocks and at the time I didn't realize that `min_val` and `max_val` were arrays. I assumed they were scalars. The tests were using a single chunk for the test data which allowed things to "just work" until @kathys tried using the compositor with real data.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/composites/viirs.py
+++ b/satpy/composites/viirs.py
@@ -230,8 +230,7 @@ class ERFDNB(CompositeBase):
                                                 False)
         super(ERFDNB, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


djhoese:
I need this and the fix is pretty simple. If someone reviews this after the fact and needs changes I'll implement them quickly.

PR #3317, opened 2025-12-16
Fix VII (METimage) reader angles names

ameraner:
This PR fixes the names of satellite and solar angles in the yaml file of the VII/METimage reader, so that they adhere to the convention and are correctly used in predefined modifiers, such as the `sunz_corrected` defined in `generic.yaml`. As discussed in https://pytroll.slack.com/archives/C0LNH7LMB/p1765897278746739 . 

Before this PR, the sunz_correction was not using the in-file angles as they could not be found due to the different name, causing issues when correcting Scenes with granules that were far apart in time; see the bright overcorrection over Europe, solved by this PR:  
before  
image
after  
image

### DIFF
--- a/satpy/etc/composites/vii.yaml
+++ b/satpy/etc/composites/vii.yaml
@@ -5,22 +5,9 @@ modifiers:
     prerequisites:
       - name: 'vii_10690'
     optional_prerequisites:
-      - solar_zenith
+      - solar_zenith_angle
       - name: 'vii_13345'
     sunz_threshold: 85.0
-
-  rayleigh_corrected:
-    modifier: !!python/name:satpy.modifiers.PSPRayleighReflectance
-    atmosphere: us-standard
-    aerosol_type: rayleigh_only
-    prerequisites:
-      - wavelength: 0.67
-        modifiers: [ sunz_corrected ]
-    optional_prerequisites:
-      - observation_azimuth
-      - observation_zenith
-      - solar_azimuth
-      - solar_zenith
 
 composites:
   true_color_uncorrected:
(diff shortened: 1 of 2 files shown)


pnuu (approved):
LGTM

Just wondering whether "observation" and "satellite" zenith angles are the same thing? It could be argued that observation zenith angle is the angle in which the satellite makes the measurement, while satellite zenith angle is the angle the satellite is seen from the observed location.

ameraner:
Yeah, I think "sensor", "viewing", "observation" and "satellite" _zenith are used interchangeably in the community - even in satpy each one of these is used. I went for "satellite" since that's what the default modifiers use, e.g. https://github.com/pytroll/satpy/blob/579e797091af6bfaa30b0923d4735fe4e402b502/satpy/etc/composites/visir.yaml#L52-L63

I checked and at the center of the swath the value is 0 (not 90), so that's as expected I'd say.

ameraner:
To confirm even more that the file variable indeed contains the commonly used viewing/satellite angle (the angle under which the satellite is seen from the ground): the METimage scanner goes up to +- 54.21°, while the "observation angle" data in the file goes up to e.g. 66.8° in my test granule. So I think we're good :)

pnuu:
Thanks for confirming!

CI build for Windows is failing on unrelated `E   rasterio._err.CPLE_AppDefinedError: Cannot open TIFF image` error. I'll try restarting the tests, even if this isn't the typical Zenodo or Conda hickup.

djhoese (approved):
I force pushed a rebase of this PR. All looks good to me. I'll merge it once CI passes.

PR #3318, opened 2025-12-18
Fix test to not save two geotiffs with the same name

djhoese:
Windows started failing in CI because this test was trying to save 2 DataArrays to the same output filename. That is not a use case we want to support, but I think was done accidentally as the test class evolved. This PR updates the test to be explicit about what dataset is being saved and Windows CI passes now.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/writer_tests/test_core/test_base.py
+++ b/satpy/tests/writer_tests/test_core/test_base.py
@@ -68,7 +68,7 @@ class TestBaseWriter:
 
     def test_save_dataset_static_filename(self):
         """Test saving a dataset with a static filename specified."""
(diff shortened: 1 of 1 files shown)


pnuu (approved):

pnuu:
To be more specific, we don't want to support saving two datasets to one *image* file. Saving to NetCDF4 is a valid usecase.

pnuu:
Zenodo is again acting up, I restarted the tests.

pnuu:
Another restart, same reason.

PR #3319, opened 2025-12-23
Refactor Landsat tests

simonreise:
Currently Landsat tests work, but tests code is extremely complicated and almost unreadable. So, it is hard to add new tests

I tried to fully refactor Landsat tests to switch to class-based approach

Most of the functionality is now moved to base Landsat tests class `BaseLandsatTest`, it includes creating file fixtures and most of the tests.

Also, there are classes that include level-specific and sensor-specific tests and fixtures. These classes not only add the specific functionality, but also override the methods that exist in base class, but are not needed for this specific product (e.g. MSS test class override `thermal_band` fixture that is not needed here, but needed in all the other classes)

The actual test classes are the classes that inherit from base class and specific classes. They include mostly setting the data variables that are unique for each test

This class-based approach will not only make tests more readable, but also will make it much easier to add new tests

This PR also fixes the error mentioned in #3300 and adds a specific test for MSS sensors that test wavelengths of a B4 band and one other band


satpy/tests/reader_tests/test_landsat.py
```diff
@@ -155,1052 +81,778 @@ def create_tif_file(data, name, area, filename, date):
...
+            {"standard_name": "test_data", "units": "test_units"},
+        )
 
-@pytest.fixture(scope="module")
-def tm_l1_sza_file(tm_l1_files_path, sza_rad_data, tm_l1_area):
-    """Create the file for the Landsat TM L1 sza."""
-    filename = tm_l1_files_path / "LT04_L1TP_143021_19890818_20200916_02_T1_SZA.TIF"
-    create_tif_file(sza_rad_data, "sza", tm_l1_area, filename, tm_l1_date)
-    return os.fspath(filename)
+    def test_loading_badfil(self, spectral_file, mda_file):
```
mraspaud:
```suggestion
    def test_loading_badfile(self, spectral_file, mda_file):
```

mraspaud (approved):
LGTM in general. I wouldn't have though of organising the tests like this, but if it works...

I think there is a type (comment inline)

PR #3327, opened 2026-01-20
Add cloud_type_chmi

nedelceo:
- [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

Hi, my colleague Martin Setvák recently modified enhancements of **FCI Cloud type** RGB so it is not oversaturated and it better detects cirrus clouds.  @ameraner  suggested to create PR for it. This PR is adding cloud_type_chmi. 

However I am not sure how to properly call the new product, so I renamed original CIMSS to CHMI. The bigger issue might be that in `composites.yml` there are now two same composites that differ only in `name` and `standard_name`. But in my opinion this is the only way of keeping both products.

**Martin's description:**
The difference is namely in the range for NIR1.3 band - we use broader range not to over-saturate the brightest parts of clouds, to maintain their structure, morphology, and also partial transparency (to keep the lower clouds underneath somewhat better visible). To compensate for the brightness loss resulting from the broader range, we use stronger gamma. I'm aware that our product settings results also in somewhat increased noise in darkest parts, but I prefer this negative impact, rather than loosing some of the thinner-most cirrus clouds. 

**Image examples:**
 CIMSS Cloud type
CloudType_20251130_0800_CIMSS
CHMI Cloud type
CloudType_20251130_0800_CHMI


gerritholl:
Thanks, this looks good!

About the name: according to https://github.com/pytroll/satpy/issues/3269, the existing Cloud Type RGB should anyway be renamed `day_cloud_type`.  So I would name your RGB `day_cloud_type_chmi`.

Have you tried the Li and Shibata correction as an alternative to sunz_corrected + sunz_reduced?  Li and Shibata is also a WMO recommendation from the RGB workshop.  It is available in satpy under `modifiers.geometry.EffectiveSolarPathlengthCorrector`.

nedelceo:
Thanks Gerrit for your suggestions to follow WMO guidliness. I have changed the names, and replaced sun corrections as you suggested.  Based on my very simple test the change improved the plotting of the transition zone.

ameraner:
Indeed, it looks very nice! Much less flashy saturation all over. Are these settings good for the full-disk or rather tuned for mid latitudes?
cc @strandgren

nedelceo:
Martin tuned it based on Europe area. However, I would expect much less (if any), sensitivity to tropospheric height compared to the IR channels. And it seem so be working properly for the full-disk. Example here (on CHMI cloud): https://ecloud.chmi.cz/s/4RSNoDigpz7a6DJ

strandgren:
Looks really good indeed, thanks for sharing the recipe. Although I like the intense colours of the CIMSS version, this recipe definitely seems to be more useful showing more cloud structures🙂  Thanks for changing the solar correction modifiers as well, with that we clearly get more data at the terminator, without any obvious artefacts.

nedelceo:
Based on discussions on 3T group I am adding  two more adding examples: 
**1) Example suggested by @gerritholl :**
original cloud_type
mtg-cloud_type-CE_1160x800-202411280840_-_
day_cloud_type_chmi
mtg-day_cloud_type_chmi-CE_1160x800-202411280840_-_



**2: Example of supercooled doplets:** 

New version here has quite a lot of red noise due to ESL ( ? in combiation with old version of IDPF ? ).

original cloud type:
mtg-cloud_type-CE_1160x800-202501140930_-_rayl
day_cloud_type_chmi:
mtg-day_cloud_type_chmi-CE_1160x800-202501140930_-_rayl

nedelceo:
I am also not sure how to handle the tick boxes above. I think this PR does not solve any issue, needs test or need to be documented.

djhoese:
Yep, I agree. Your tick boxes are fine as-is.

djhoese:
That said, you may need to rebase or merge with `main` to get the latest fixes for CI to start passing or get closer to passing.

mraspaud (approved):
Looks great, thanks for submitting this!

mraspaud:
pre-commit.ci autofix

Issue #3043, opened 2025-01-23
RuntimeError: NetCDF: HDF error OR Segmentation fault


MachineMoose:
**Describe the bug**
Hi,

I just got started with satpy and want to try it out with the FCI Level 1c Normal Resolution Image Data - MTG - 0 degree EUMETSAT dataset. 

I found an example in the documentation for EUMETSAT MTG (Third generation) data, the fci_l1c_nc reader. However, when I try to use the boilerplate code in the exampleI get two error messages, they alternate randomly from run to run.

Error messages: 
- RuntimeError: NetCDF: HDF error
- Segmentation fault

The program runs as expected up until 
vis_04_values = scn['vis_04'].values

**To Reproduce**

```python
import hdf5plugin

from satpy.scene import Scene
from satpy import find_files_and_readers
from satpy.utils import debug_on

path_to_data = '/Users/mjodas/Downloads/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--x-x---x_C_EUMT_20250120120254_IDPFI_OPE_20250120120007_20250120120924_N__O_0073_0000'

# find files and assign the FCI reader
files = find_files_and_readers(base_dir=path_to_data, reader='fci_l1c_nc')

# create an FCI scene from the selected files
scn = Scene(filenames=files)

# print available dataset names for this scene (e.g. 'vis_04', 'vis_05','ir_38',...)
print(scn.available_dataset_names())

# print available composite names for this scene (e.g. 'natural_color', 'airmass', 'convection',...)
print(scn.available_composite_names())

# load the datasets/composites of interest
scn.load(['natural_color','vis_04'], upper_right_corner='NE')
# note: the data inside the FCI files is stored upside down. The upper_right_corner='NE' argument
# flips it automatically in upright position.

# you can access the values of a dataset as a Numpy array with
vis_04_values = scn['vis_04'].values

# resample the scene to a specified area (e.g. "eurol1" for Europe in 1km resolution)
scn_resampled = scn.resample("eurol", resampler='nearest', radius_of_influence=5000)

# save the resampled dataset/composite to disk
scn_resampled.save_dataset("natural_color", filename='./fci_natural_color_resampled.png')

```

**Expected behavior**
I expect a successful  run of the script producing a png with natural colors europe. 

**Actual results**

```
Traceback (most recent call last):
[...]

PR #3331, opened 2026-02-03
Address h5netcdf scalar issue

mraspaud:
This PR address caching scalar values in the netcdf util when using h5netcdf >= 1.7.0.
The solution (catching a ValueError on top of the IndexError) was already available, but in a duplicated part of the code that was not used for the caching part.

While setting up the test case corresponding to this issue and adding more h5netcdf engine tests (through simply parametrising existing tests to also use that engine), it became clear that netcdf4 and h5netcdf cannot be imported at the same time in some environments, hence the changes in the netcdf file creation fixture, the removal of netcdf4 as top level import, and the usage of pytest-forked in ci. For information, I am running on my RHEL9 laptop inside a uv environment with python 3.14.

 - [x] Tests added

### DIFF
--- a/satpy/_config.py
+++ b/satpy/_config.py
@@ -107,12 +107,12 @@ if _ancpath is not None and _data_dir is None:
 config = Config("satpy", defaults=[_CONFIG_DEFAULTS], paths=_CONFIG_PATHS)
 
 
-def get_config_path_safe():
+def get_config_path_safe() -> list[str]:
     """Get 'config_path' and check for proper 'list' type."""
     config_path = config.get("config_path")
     if not isinstance(config_path, list):
-        raise ValueError("Satpy config option 'config_path' must be a "
-                         "list, not '{}'".format(type(config_path)))
+        raise ValueError("Satpy config option 'config_path' must be a " +
+                         f"list, not '{type(config_path)}'.")
     return config_path
 
 
@@ -155,7 +155,7 @@ def _entry_point_module(entry_point):
         return entry_point.value.split(":")[0].strip()
 
 
-def config_search_paths(filename, search_dirs=None, **kwargs):
+def config_search_paths(filename: str, search_dirs: list[str]|None = None, **kwargs) -> list[str]:
     """Get series of configuration base paths where Satpy configs are located."""
     if search_dirs is None:
         search_dirs = get_config_path_safe()[::-1]
--- a/satpy/_config.py
+++ b/satpy/_config.py
@@ -155,7 +155,7 @@ def _entry_point_module(entry_point):
         return entry_point.value.split(":")[0].strip()
 
 
-def config_search_paths(filename: str, search_dirs: list[str]|None = None, **kwargs) -> list[str]:
+def config_search_paths(filename: str, search_dirs: list[str] | None = None, **kwargs) -> list[str]:
     """Get series of configuration base paths where Satpy configs are located."""
     if search_dirs is None:
         search_dirs = get_config_path_safe()[::-1]
--- a/satpy/readers/core/loading.py
+++ b/satpy/readers/core/loading.py
@@ -63,7 +63,7 @@ def load_readers(filenames=None, reader=None, reader_kwargs=None):
         loadables = reader_instance.select_files_from_pathnames(readers_files)
         if loadables:
             # WARN: This is very confusing, but it seems to work: The reader_kwargs, when passed without a reader key,
-            # are pud into a dictionary where keys are individual letters of the reader name, and the value is the
+            # are put into a dictionary where keys are individual letters of the reader name, and the value is the
             # reader kwargs; however, notice the `reader[idx]` part in the call here, as idx is 0 and thus the first
             # letter of the reader is used to index `reader_kwargs_without_filter`, thus retrieving the correct
             # reader_kwargs.
--- a/satpy/readers/core/loading.py
+++ b/satpy/readers/core/loading.py
@@ -62,6 +62,11 @@ def load_readers(filenames=None, reader=None, reader_kwargs=None):
 
         loadables = reader_instance.select_files_from_pathnames(readers_files)
         if loadables:
+            # WARN: This is very confusing, but it seems to work: The reader_kwargs, when passed without a reader key,
+            # are pud into a dictionary where keys are individual letters of the reader name, and the value is the
+            # reader kwargs; however, notice the `reader[idx]` part in the call here, as idx is 0 and thus the first
+            # letter of the reader is used to index `reader_kwargs_without_filter`, thus retrieving the correct
+            # reader_kwargs.
             reader_instance.create_storage_items(
                     loadables,
                     fh_kwargs=reader_kwargs_without_filter[None if reader is None else reader[idx]])
--- a/satpy/readers/smos_l2_wind.py
+++ b/satpy/readers/smos_l2_wind.py
@@ -26,6 +26,7 @@ SMOS_WIND_DS_PDD_20191107_signed.pdf
 
 import datetime as dt
 import logging
+from functools import cached_property
 
 import numpy as np
 from pyresample.geometry import AreaDefinition
@@ -102,16 +103,16 @@ class SMOSL2WINDFileHandler(NetCDF4FileHandler):
         except KeyError:
             return data
 
-    def _adjust_lon_coord(self, data):
+    def _normalize_lon_coord(self, data):
         """Adjust lon coordinate to -180 .. 180 ( not 0 .. 360)."""
-        data = data.assign_coords(lon=(((data.lon + 180) % 360) - 180))
-        return data.where(data < 180., data - 360.)
+        if "lon" in data.dims:
+            return data.assign_coords(lon=self.normalized_lon)
+        return data
 
     def _rename_coords(self, data):
-        """Rename coords."""
+        """Rename coordinates."""
         rename_dict = {}
         if "lon" in data.dims:
-            data = self._adjust_lon_coord(data)
             rename_dict["lon"] = "x"
         if "lat" in data.dims:
             rename_dict["lat"] = "y"
@@ -133,33 +134,54 @@ class SMOSL2WINDFileHandler(NetCDF4FileHandler):
             data = data.roll(lon=720, roll_coords=True)
         return data
 
-    def get_dataset(self, ds_id, ds_info):
+    def get_dataset(self, dataset_id, ds_info):
         """Get dataset."""
-        data = self[ds_id["name"]]
+        if  dataset_id["name"] == "lon":
+            lon = self.centered_lon
+            return lon.rename(dict(lon="x"))
+        elif dataset_id["name"] == "lat":
+            lat = self.trimmed_lat
+            return lat.rename(dict(lat="y"))
+
+        data = self[dataset_id["name"]]
         data.attrs = self.get_metadata(data, ds_info)
         data = self._remove_time_coordinate(data)
+        data = self._normalize_lon_coord(data)
         data = self._roll_dataset_lon_coord(data)
         data = self._rename_coords(data)
         data = self._mask_dataset(data)
         if len(data.dims) >= 2 and all([dim in data.dims for dim in ["x", "y"]]):
             # Remove the first and last row as these values extends beyond +-90 latitude
-            # if the dataset contains the y dimmension.
+            # if the dataset contains the y dimension.
             # As this is data over open sea these has no values.
             data = data.where((data.y > -90.0) & (data.y < 90.0), drop=True)
-        elif len(data.dims) == 1 and "y" in data.dims:
-            if ds_id["name"] == "lat":
-                data = data.where((data > -90.0) & (data < 90.0), drop=True)
-            else:
-                # Is there such a case?
-                data = data.where((data.y > 0) & (data.y < len(data.y) - 1), drop=True)
         return data
+
+
+    @cached_property
+    def normalized_lon(self):
+        """Normalize longitudes.
+
+        That means constrain them to the [-180, 180[ interval.
+        """
+        lon = (((self["lon"]+ 180) % 360) - 180)
+        return lon.assign_coords(lon=lon)
+
+    @cached_property
+    def centered_lon(self):
+        """Center the longitudes around 0° east."""
+        return self.normalized_lon.roll(lon=720, roll_coords=True)
+
+    @cached_property
+    def trimmed_lat(self):
+        """Trim the latitudes to remove the problematic -90 and 90° north."""
+        lat = self["lat"]
+        return lat.where((lat > -90.0) & (lat < 90.0), drop=True)
 
     def _create_area_extent(self, width, height):
         """Create area extent."""
         # Creating a meshgrid, not needed actually, but makes it easy to find extremes
-        _lon = self._adjust_lon_coord(self["lon"])
-        _lon = self._roll_dataset_lon_coord(_lon)
-        latlon = np.meshgrid(_lon, self["lat"][1:self["lat/shape"][0] - 1])
+        latlon = np.meshgrid(self.centered_lon.data, self.trimmed_lat.data)
         lower_left_x = latlon[0][height - 1][0] - 0.125
         lower_left_y = latlon[1][height - 1][0] + 0.125
         upper_right_y = latlon[1][1][width - 1] - 0.125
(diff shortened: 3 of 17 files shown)


satpy/readers/core/loading.py
```diff
@@ -62,6 +62,11 @@ def load_readers(filenames=None, reader=None, reader_kwargs=None):
 
         loadables = reader_instance.select_files_from_pathnames(readers_files)
         if loadables:
+            # WARN: This is very confusing, but it seems to work: The reader_kwargs, when passed without a reader key,
+            # are pud into a dictionary where keys are individual letters of the reader name, and the value is the
```
pnuu:
```suggestion
            # are put into a dictionary where keys are individual letters of the reader name, and the value is the
```

satpy/tests/reader_tests/test_netcdf_utils.py
```diff
@@ -68,22 +69,30 @@ def get_test_content(self, filename, filename_info, filetype_info):
 @pytest.fixture
 def netcdf_file(tmp_path):
     """Create a test NetCDF4 file."""
-    from netCDF4 import Dataset
     filename = tmp_path / "test.nc"
-    with Dataset(filename, "w") as nc:
+    from multiprocessing import Process
+    p = Process(target=_create_netcdf_file, args=(filename, ))
+    p.start()
+    p.join()
```
pnuu: Why this complication, does it help us in some way?

pnuu (approved):
Just one question and one typo fix inline. In general I'm fine with this.

satpy/tests/reader_tests/test_netcdf_utils.py
mraspaud: yes, see the text I wrote in the PR note. Basically, it’s because in some environments (like mine), writing with netcdf4 and reading with h5netdf from the same process crashes, making all the h5netcdf tests fail in my case.

pnuu: Ah, ok. I didn't see the connection between the comment and the fixture.

satpy/readers/core/loading.py
```diff
@@ -62,6 +62,11 @@ def load_readers(filenames=None, reader=None, reader_kwargs=None):
 
         loadables = reader_instance.select_files_from_pathnames(readers_files)
         if loadables:
+            # WARN: This is very confusing, but it seems to work: The reader_kwargs, when passed without a reader key,
+            # are put into a dictionary where keys are individual letters of the reader name, and the value is the
+            # reader kwargs; however, notice the `reader[idx]` part in the call here, as idx is 0 and thus the first
+            # letter of the reader is used to index `reader_kwargs_without_filter`, thus retrieving the correct
+            # reader_kwargs.
```
djhoese, line 69: So basically this logic is incorrect and works by luck/chance and needs to be fixed?

mraspaud: I'm not sure if it was the intended behaviour or not... It's working in the end, so maybe it is? But yeah, it would definitely benefit from a dedicated test and then a refactor...

djhoese: So what did you do to run into the issue? You called it from the Scene or noticed it within a test you already had?

djhoese:
> it became clear that netcdf4 and h5netcdf cannot be imported at the same time in some environments

Is this intended to not work? Is this a bug in h5netcdf then? Is it something that can or should be fixed upstream? The forking with the pytest flag I can kind of agree with...maybe, but the process creation inside the test file creation is too much for me when all of this comes down to an HDF5 issue from what we can tell.

djhoese:
Oh and if you have error messages could you please put then in the PR description so in the future this PR is easily discoverable.

djhoese:
I'm guessing you're running into the same issues users describe here when installing packages from PyPI:

https://github.com/pytroll/satpy/issues/3043

If you're able to reproduce this easily that'd be great for making progress on that issue.

mraspaud:
> Oh and if you have error messages could you please put then in the PR description so in the future this PR is easily discoverable.

Good point! Here is the error I get from pytest repeatedly in the h5netcdf cases (and I tried creating coordinate variables, doesn't help, the only thing that helps is when netCDF4 is not loaded).

TL;DR: `RuntimeError: Unspecified error in H5DSis_scale (return value <0)`


```
_______________________________________________________________________________________________________ test_get_data_as_xarray_h5netcdf _______________________________________________________________________________________________________

tmp_path = PosixPath('/tmp/pytest-of-a001673/pytest-3/test_get_data_as_xarray_h5netc0')

    def test_get_data_as_xarray_h5netcdf(tmp_path):
        """Test getting xr.DataArray from h5netcdf variable."""
        import numpy as np
    
        from satpy.readers.core.netcdf import get_data_as_xarray
    
        data = np.array([1, 2, 3])
        fname = tmp_path / "test.nc"
>       fid = _write_test_h5netcdf(fname, data)

data       = array([1, 2, 3])
fname      = PosixPath('/tmp/pytest-of-a001673/pytest-3/test_get_data_as_xarray_h5netc0/test.nc')
get_data_as_xarray = <function get_data_as_xarray at 0x7fc017c50930>
np         = <module 'numpy' from '/home/a001673/usr/src/satpy/h5netcdf-cache-fix/.venv/lib/python3.14/site-packages/numpy/__init__.py'>
tmp_path   = PosixPath('/tmp/pytest-of-a001673/pytest-3/test_get_data_as_xarray_h5netc0')

satpy/tests/reader_tests/test_netcdf_utils.py:389: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
satpy/tests/reader_tests/test_netcdf_utils.py:401: in _write_test_h5netcdf
    fid.dimensions = {"y": data.size}
    ^^^^^^^^^^^^^^
        data       = array([1, 2, 3])
        fid        = <h5netcdf.File 'test.nc' (mode r+, backend h5py)>
Dimensions:
Groups:
Variables:
Attributes:
        fname      = PosixPath('/tmp/pytest-of-a001673/pytest-3/test_get_data_as_xarray_h5netc0/test.nc')
        h5netcdf   = <module 'h5netcdf' from '/home/a001673/usr/src/satpy/h5netcdf-cache-fix/.venv/lib/python3.14/site-packages/h5netcdf/__init__.py'>
.venv/lib/python3.14/site-packages/h5netcdf/core.py:1257: in dimensions
    self._dimensions.update(value)
        self       = <h5netcdf.File 'test.nc' (mode r+, backend h5py)>
Dimensions:
Groups:
Variables:
Attributes:
        value      = {'y': 3}
<frozen _collections_abc>:971: in update
    ???
        key        = 'y'
        kwds       = {}
        other      = {'y': 3}
        self       = <h5netcdf.Dimensions: >
.venv/lib/python3.14/site-packages/h5netcdf/dimensions.py:43: in __setitem__
    self._objects[name] = Dimension(self._group, name, size, create_h5ds=True)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        name       = 'y'
        self       = <h5netcdf.Dimensions: >
        size       = 3
.venv/lib/python3.14/site-packages/h5netcdf/dimensions.py:108: in __init__
    self._create_scale()
        create_h5ds = True
        name       = 'y'
        parent     = <h5netcdf.File 'test.nc' (mode r+, backend h5py)>
Dimensions:
Groups:
Variables:
Attributes:
        phony      = False
        self       = <h5netcdf.Dimension 'y': size 3>
        size       = 3
.venv/lib/python3.14/site-packages/h5netcdf/dimensions.py:228: in _create_scale
    if not self._root._h5py.h5ds.is_scale(self._h5ds.id):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
        dimid      = 0
        dimlen     = b'         3'
        kwargs     = {'track_order': True}
        scale_name = b'This is a netCDF dimension but not a netCDF variable.         3'
        self       = <h5netcdf.Dimension 'y': size 3>
        size       = 3
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
h5py/h5ds.pyx:34: in h5py.h5ds.is_scale
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   RuntimeError: Unspecified error in H5DSis_scale (return value <0)


h5py/defs.pyx:4326: RuntimeError
```

satpy/readers/smos_l2_wind.py
```diff
@@ -147,7 +147,11 @@ def get_dataset(self, ds_id, ds_info):
             # As this is data over open sea these has no values.
             data = data.where((data.y > -90.0) & (data.y < 90.0), drop=True)
         elif len(data.dims) == 1 and "y" in data.dims:
-            data = data.where((data.y > 0) & (data.y < len(data.y) - 1), drop=True)
+            if ds_id["name"] == "lat":
+                data = data.where((data > -90.0) & (data < 90.0), drop=True)
+            else:
+                # Is there such a case?
+                data = data.where((data.y > 0) & (data.y < len(data.y) - 1), drop=True)
```
mraspaud: @TAlonglong That's the part I'm unsure about...

mraspaud:
@joleenf @djhoese I did a refactor of the mimic and geocat tests, if you could have a look that would be great!

satpy/tests/reader_tests/test_geocat.py
```diff
@@ -37,127 +35,130 @@
...
-
-        attrs = ("_FillValue", "flag_meanings", "flag_values", "units")
-        convert_file_content_to_data_array(
-            file_content, attrs=attrs,
-            dims=("z", "lines", "elements"))
-        return file_content
-
-
-class TestGEOCATReader(unittest.TestCase):
+@pytest.fixture(scope="session")
```
djhoese: If these aren't used outside of this module then the scope should be module, not session.

mraspaud: Thanks, great point! Fixed

satpy/tests/reader_tests/test_netcdf_utils.py
```diff
@@ -65,25 +66,25 @@ def get_test_content(self, filename, filename_info, filetype_info):
         raise NotImplementedError("Fake File Handler subclass must implement 'get_test_content'")
 
 
-@pytest.fixture
-def netcdf_file(tmp_path):
+@pytest.fixture(scope="session")
```
djhoese: module scope?

satpy/tests/reader_tests/test_netcdf_utils.py
```diff
@@ -65,25 +66,25 @@ def get_test_content(self, filename, filename_info, filetype_info):
...
+def netcdf_file(tmp_path_factory):
     """Create a test NetCDF4 file."""
     from netCDF4 import Dataset
-    filename = tmp_path / "test.nc"
-    with Dataset(filename, "w") as nc:
+    filename = tmp_path_factory.mktemp("data") / "test.nc"
+    with closing(Dataset(filename, "w")) as nc:
```
djhoese: Is `closing` needed? Isn't it done automatically by `Dataset`?

satpy/tests/reader_tests/test_netcdf_utils.py
```diff
@@ -65,25 +66,25 @@ def get_test_content(self, filename, filename_info, filetype_info):
...
         # Create Group
         g1 = nc.createGroup("test_group")
-
         # Add datasets
         ds1_f = g1.createVariable("ds1_f", np.float32,
                                   dimensions=("rows", "cols"))
         ds1_f[:] = np.arange(10. * 100).reshape((10, 100))
         ds1_i = g1.createVariable("ds1_i", np.int32,
                                   dimensions=("rows", "cols"))
+        ds1_f.set_auto_scale(True)
```
djhoese: This line uses `ds1_f` but it is below `ds1_i` (`ds1_f` is used above that). What is the intended outcome? And doesn't this only matter if `scale_factor` and `add_offset` are defined? I don't see them for any variables here.

satpy/tests/reader_tests/test_netcdf_utils.py
mraspaud: I looked at the documentation, and there no mention at all of the context manager behaviour, hence the change.
But I have now looked at the code, and you're right, it is done automatically:
https://github.com/Unidata/netcdf4-python/blob/master/src/netCDF4/_netCDF4.pyx#L2547-L2551

Fixing

satpy/readers/core/loading.py
mraspaud: no, I was debugging why my "engine" keyword arg was not being taken into account. I was another issue, but I spent a while here trying to understand what was happening…

satpy/tests/reader_tests/test_netcdf_utils.py
mraspaud: fixed in 99ff436

satpy/tests/reader_tests/test_netcdf_utils.py
mraspaud: fixed in 99ff436

satpy/tests/reader_tests/test_netcdf_utils.py
mraspaud: Well, this is an oversight on my part, I should have put them in the correct order. The set_auto_scale I added because it seemed to be needed to make the existing tests pass ~~after converting to a file-based fixture~~ (edit: it already was a file-based fixture).

mraspaud:
For reference, I could confirm yesterday that the errors I was getting happened when running the tests in a pypi-based environment, and disappeared when I switched to a conda-based environment.
I'm suspecting the netcdf4 and h5netcdf/h5py each provide their own compiled version of the hdf5 library in the Pypi wheels, leading to binary incompatibilities.

mraspaud:
@djhoese any more comments here?

satpy/readers/core/loading.py
djhoese: Can you file an issue for this so eventually someone works on it?

satpy/readers/smos_l2_wind.py
djhoese, line 1: I'm unfamiliar with this reader, but I'm having trouble tracking down all the changes. The logic doesn't seem to be equivalent. Is that intended? Especially the 1D "y" variable handling, the trimmed lat operations, and the lon/lat `get_dataset` handling at the top of the method.

satpy/_config.py
```diff
@@ -155,7 +155,7 @@ def _entry_point_module(entry_point):
         return entry_point.value.split(":")[0].strip()
 
 
-def config_search_paths(filename, search_dirs=None, **kwargs):
+def config_search_paths(filename: str, search_dirs: list[str]|None = None, **kwargs) -> list[str]:
```
djhoese:
```suggestion
def config_search_paths(filename: str, search_dirs: list[str] | None = None, **kwargs) -> list[str]:
```

djhoese (commented):
I made a couple comments individually but after those are answered this looks good. It seems much more reasonable where things are imported and how things are checked. Thanks for doing this and all the cleanup.

satpy/readers/smos_l2_wind.py
mraspaud: Yes it is different, sorry for not mentioning it earlier. Turns out some of the tested logic was wrong but worked with the mock. I got hold on some actual files, so I was able to check and fix the behaviours we need for this reader. In the end, the changes are for the better afaics.

PR #3332, opened 2026-02-05
Add QC filtering to the WSPD variable for AMSR2 L2 GAASP

djhoese:
While testing things for Polar2Grid @kathys noticed that the AMSR2 GAASP algorithm's "OCEAN" files don't seem to have as much quality filtering for the "WSPD" variable like many of the other variables do. This PR adds an optional flag (off by default) to use "WSPD_QC" to convert bad quality pixels to NaN. From what we can tell the QC variables in the GAASP output are inconsistent so a generic solution didn't seem feasible.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Fully documented 

### DIFF
--- a/satpy/readers/amsr2_l2_gaasp.py
+++ b/satpy/readers/amsr2_l2_gaasp.py
@@ -149,7 +149,8 @@ class GAASPFileHandler(BaseFileHandler):
         add_offset = attrs.pop("add_offset", 0.)
         scaling_needed = not (scale_factor == 1 and add_offset == 0)
         if scaling_needed:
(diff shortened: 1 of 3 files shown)


pnuu (commented):
Example how the filtering is turned on would be nice.

djhoese:
🤦 of course. I'll add that when I'm back at my computer.

djhoese:
Updating from the conversation on slack: Martin approved of the PR but wanted real files to be used in the tests. This has been implemented and some things refactored. Merging now...

Issue #3239, opened 2025-09-18
Load and write generic image changes the looks


gerritholl:
**Describe the bug**

When I load an image with the `generic_image` reader and immediately write it out again, the image changes.  This happens because the `generic_image` reader returns float data, and then a default enhancement gets produced.

**To Reproduce**

```python
from satpy import Scene
from glob import glob
seviri_files = glob("/media/nas/x21308/scratch/SEVIRI/202103300900/H-000*")
sc = Scene(filenames={"seviri_l1b_hrit": seviri_files})
sc.load(["natural_color"])
sc.save_datasets(writer="geotiff", filename="/tmp/natural-color-1.tif")

sc2 = Scene(filenames={"generic_image": ["/tmp/natural-color-1.tif"]})
sc2.load(["image"])
sc2.save_datasets(writer="geotiff", filename="/tmp/natural-color-2.tif")
```

**Expected behavior**

I expect both images to look the same.

**Actual results**

There is a difference, because the second image is getting re-enhanced:

<details>

```
[DEBUG: 2025-09-18 11:22:19 : satpy.readers.core.yaml_reader] Reading ('/home/gholl/checkouts/satpy/satpy/etc/readers/seviri_l1b_hrit.yaml',)
[...]

PR #3334, opened 2026-02-09
Cleanup up composite utility 'add_bands' function

### LINKED ISSUE #3333: Implicit Eager Evaluation in `add_bands` causing MemoryError
## Describe the bug
I discovered a performance bottleneck in the `add_bands` function within `satpy/composites/core.py`. The function calls `.compute()` on the `bands` DataArray. This triggers an immediate evaluation of the entire dask graph and loads the data into memory. In satellite data processing, where arrays are often massive, this causes a `MemoryError` and breaks the lazy evaluation paradigm that Satpy is built upon.

The computation is unnecessary because only the coordinate names (the band identifiers) are needed for the logic, not the actual pixel values.

## To Reproduce
To be sure, I did a little experiment at colab to prove the idea, you can try it yourself

    import xarray as xr
    import dask.array as da
    import numpy as np
    import psutil
    import os
    from satpy.composites.core import add_bands

    def get_ram_usage():
        """Returns the current RAM usage in MB."""
        process = psutil.Process(os.getpid())
        return process.memory_info().rss / (1024 ** 2)

    # I am simulating a large-scale satellite imagery scenario.
    # Using a 30,000 x 30,000 grid to test how 'add_bands' handles lazy data.
    print("Initializing large-scale dask arrays (Lazy Loading)...")

    shape = (30000, 30000)
    chunks = (5000, 5000)

    data_arr = xr.DataArray(
        da.random.random(shape, chunks=chunks),
        dims=('y', 'x'),
        attrs={'mode': 'L'}
    ).expand_dims(bands=['L'])

    bands_target = xr.DataArray(
        da.random.random(shape, chunks=chunks),
        dims=('y', 'x')
    ).expand_dims(bands=['R', 'G', 'B'])

    print(f"RAM usage before calling add_bands: {get_ram_usage():.2f} MB")

    try:
        print("Executing add_bands...")
        result = add_bands(data_arr, bands_target)
        print(f"RAM usage after execution: {get_ram_usage():.2f} MB")
    except MemoryError:
        print("!!! Session Crashed: MemoryError due to implicit .compute() !!!")

## Expected behavior
The `add_bands` function should remain lazy. It should only inspect the coordinates of the `bands` object to determine the mode and append the necessary channels. RAM usage should remain stable, and the returned object should still be backed by a Dask array.

## Actual results
The system attempts to compute the entire array, causing a massive RAM spike. In my Google Colab environment, this leads to an immediate session crash as it exceeds the available RAM.

## Environment Info
- OS: Google Colab (Linux)
- Satpy Version: Latest
- PyResample Version: Latest

## Additional context
The fix seems straightforward. Instead of calling `bands = bands.compute()` at line 222, the code should only access the coordinate values, for example `bands_names = bands.bands.values` or `bands.coords['bands']`. This would allow helper functions like `_check_mode_p` to perform their checks without loading the data into memory.
### Proposed Fix

I am happy to submit a **Pull Request (PR)** with this fix if you would like me to proceed. Please let me know if this aligns with the project's contribution guidelines!

## Screenshots
this is before running
Image
just a second before crash
Image

djhoese:
See the conversation in #3333 (CC @SalimMessaad1). The summary is that the `add_bands` utility function in `satpy.composites.core` is a little confusing to follow. Mostly, it isn't clear due to the lack of documentation what the second argument should be. If we look at the actual uses of it in Satpy the second argument should always be `some_data_arr.coords["bands"]`. However, there was a `bands = bands.compute()` call to work around a dask bug so it kind of seemed like it should be a regular image data `DataArray` that is passed. But I don't think a "bands" coordinate has any reason to ever be a dask array (versus a numpy array) so I'm not sure why this was ever needed. Additionally, all the tests (either added by @mraspaud or @pnuu, or maybe just refactored to the current location :man_shrugging: ) always pass a `DataArray` where the bands information is in the `.data` part, not just the `.coords["bands"]`. This is never the way this function is used in Satpy.

So this PR:

* Adds docstring information
* Adds type annotations
* Allows "bands" (the second argument) to be a Sequenced (ex. list) or a DataArray and describes in the docstring that the DataArray should come from `.coords["bands"]` of an image array.
* Remove the dask workaround of `bands = bands.compute()` as it is from a 2020 bug and never should have been needed(?)
* Tests have been parametrized
* Tests now properly use the `.coords["bands"]` for the bands argument
* Tests enforce no dask computation
* Fix other various mypy issues (mostly satpy.composites.viirs)

Remaining question for (@mraspaud @gerritholl @pnuu):

The tests use 32-bit float input and make sure it remains 32-bit float data. I think this is reasonable as the actual use cases in Satpy are all with pre-enhanced data I think. So everything is normalized to 0-1 floating point numbers except for maybe some edge cases with integers. In the cases where an Alpha band is added it is set to all 1s. Do we want to support the case of integers where alpha is set to max-int for that integer type? Or maybe we just fail for now and handle/update the code when someone has a legitimate use case for that?

 - [x] Closes #3333 
 - [x] Tests added 
 - [x] Fully documented 

### DIFF
--- a/satpy/composites/core.py
+++ b/satpy/composites/core.py
@@ -350,7 +350,9 @@ def _check_alpha_band(data: xr.DataArray, bands: Sequence) -> xr.DataArray:
         new_data = [data.sel(bands=band) for band in data["bands"].data]
         # Create alpha band based on a copy of the first "real" band
         alpha = new_data[0].copy()
-        # TODO: This will be all 1s as floats, but this is not accurate for integer data. Right?
+        # WARN: This will be all 1s as floats, but this is not accurate for
+        # integer data. Integer data needing an alpha channel is not currently
+        # supported.
         alpha.data = da.ones(
             (data.sizes["y"], data.sizes["x"]),
             dtype=new_data[0].dtype,
--- a/satpy/composites/core.py
+++ b/satpy/composites/core.py
@@ -280,25 +280,54 @@ def _apply_palette_to_image(img):
     return img
 
 
-def add_bands(data, bands):
-    """Add bands so that they match *bands*."""
+def add_bands(data: xr.DataArray, bands: Sequence | xr.DataArray) -> xr.DataArray:
+    """Add image bands so that input data is compatible with the provided ``bands``.
+
+    This function will add additional copies of input data arrays to allow the
+    number of elements in the "bands" dimension to be more compatible with
+    the desired number of bands from the ``bands`` argument. This does not
+    always mean producing the exact number of bands. For example, if the
+    input data is an "L" image with an alpha channel (``["L", "A"]``) and the
+    desired bands are ``["R", "G", "B"]`` then the result will be
+    ``["R", "G", "B", "A"]`` where the R, G, and B data are copies of the original
+    L band data and the A channel is copied from the input.
+
+    If an alpha channel needs to be added to match the desired set of bands then
+    a fully opaque alpha channel is created.
+
+
+    Args:
+        data: Input xarray DataArray with a "bands" dimension and coordinate.
+            The DataArray should also have a "mode" attribute in ``.attrs``
+            with the image mode of the data. So if the "bands" coordinate
+            variable is ``["R", "G", "B"]`` then ``.attrs["mode"]`` will
+            be ``"RGB"``.
+        bands: The bands that the result should be compatible with. If taken
+            from another data DataArray via ``.coords["bands"]`` then this
+            will be a ``DataArray``. Otherwise an sequence supporting
+            ``__contains__`` should be provided. Elements of this argument
+            should be individual string characters, one for each band.
+
+    """
+    desired_bands = bands.values.tolist() if isinstance(bands, xr.DataArray) else bands
     # Add R, G and B bands, remove L band
-    bands = bands.compute()
-    data = _check_mode_p(data, bands)
-    data = _check_mode_l(data, bands)
+    data = _check_mode_p(data, desired_bands)
+    data = _check_mode_l(data, desired_bands)
     # Add alpha band
-    data = _check_alpha_band(data, bands)
+    data = _check_alpha_band(data, desired_bands)
     return data
 
 
-def _check_mode_p(data, bands):
-    if "P" in data["bands"].data or "P" in bands.data:
+def _check_mode_p(data: xr.DataArray, bands: Sequence) -> xr.DataArray:
+    if "P" in data["bands"].data or "P" in bands:
         raise NotImplementedError("Cannot mix datasets of mode P with other datasets at the moment.")
     return data
 
 
-def _check_mode_l(data, bands):
-    if "L" in data["bands"].data and "R" in bands.data:
+def _check_mode_l(data: xr.DataArray, bands: Sequence) -> xr.DataArray:
+    new_data: tuple[xr.DataArray, ...]
+
+    if "L" in data["bands"].data and "R" in bands:
         lum = data.sel(bands="L")
         # Keep 'A' if it was present
         if "A" in data["bands"]:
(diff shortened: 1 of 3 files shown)


SalimMessaad1:
This is great! ,I'm glad to see the dask computation being enforced as lazy and the old workaround removed. Thank you for taking the time to refactor this and improve the documentation. It makes the function much safer for large-scale data processing.
How can I add my name to AUTHORS.md? Should I make a small PR for that or can you add me?

satpy/composites/core.py
```diff
@@ -316,21 +345,23 @@ def _check_mode_l(data, bands):
...
+def _check_alpha_band(data: xr.DataArray, bands: Sequence) -> xr.DataArray:
+    if "A" not in data["bands"].data and "A" in bands:
         new_data = [data.sel(bands=band) for band in data["bands"].data]
         # Create alpha band based on a copy of the first "real" band
         alpha = new_data[0].copy()
-        alpha.data = da.ones((data.sizes["y"],
-                              data.sizes["x"]),
-                             dtype=new_data[0].dtype,
-                             chunks=new_data[0].chunks)
+        # TODO: This will be all 1s as floats, but this is not accurate for integer data. Right?
```
gerritholl: I've never heard of an integer alpha layer.  That would just be a binary mask.  Is an integer alpha layer ever a real-world use case?

gerritholl (approved):
Looks good.  I wasn't actively aware of this function.  The old docstring was certainly not very helpful.

> Do we want to support the case of integers where alpha is set to max-int for that integer type?

As commented in-line, I cannot think of a use case with an integer alpha.  We can take that if there is a need.

satpy/composites/core.py
pnuu: On-disk geotiffs at least have integer values in the alpha layer. The "binary mask" is also the default we have in Satpy for missing data. I guess we can have cases where also integer (categorical) data also need the alpha layer :thinking:

gerritholl: But we define alpha as 0–1, don't we?  It only gets stretched to 0–256 on writing (in case of most image file formats, at least).  Categorical data usually has a category defined as "missing", not an extra layer.  I don't think I've ever tried reading an RGBA GeoTIFF with satpy, but in general reading GeoTIFF has limitations (for example, see https://github.com/pytroll/satpy/issues/3239).  The `generic_image` reader returns float data, so it won't result in an integer alpha layer at least as currently written.

djhoese: I think there are cases where an integer image could make itself to the compositor but it is very unlikely. So do I handle that case just because maybe it could happen or error because it is unlikely/unexpected?

pnuu (approved):
LGTM, thanks!

mraspaud:
> Do we want to support the case of integers where alpha is set to max-int for that integer type?
No need to take care of this now.

satpy/composites/core.py
mraspaud:
```suggestion
        # WARN: This will be all 1s as floats, but this is not accurate for integer data. Integers are not supported for the alpha channel.
```

satpy/composites/viirs.py
```diff
@@ -89,26 +89,26 @@ def __call__(self, datasets, **info):
         output_dataset.attrs = info
         return output_dataset
 
-    def _run_dnb_normalization(self, dnb_data: np.ndarray, sza_data: np.ndarray) -> np.ndarray:
+    def _run_dnb_normalization(self, dnb_data_np: np.ndarray, sza_data_np: np.ndarray) -> np.ndarray:
```
mraspaud: Do we need the "_np" suffix when we have type annotations? (they are added to the rendered documentation, right?)

mraspaud (approved):
LGTM, thanks for coming back to this after the issue was raised. The documentation changes alone are worth it!

Just a couple of inline comments, but nothing critical

satpy/composites/viirs.py
djhoese: This was me being lazy. Mypy got mad that I used the same variable name for different types of values (numpy arrays, DataArrays, etc) so the least amount of changes was to add the `_np` suffix. This is a "private" method so I didn't think much of it, but I can change these back to normal names and rename the internal variables.

satpy/composites/core.py
djhoese: I went with a slightly different wording, but same idea. Thanks. Completed.

satpy/composites/viirs.py
djhoese: Swapped. Done.

PR #3336, opened 2026-02-12
Add support for some MWS composite image generation

adybbroe:
Add support for some MWS composite image generation (based on AWS) and fix file pattern

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


gerritholl:
Do you have some example images on what those look like?  Many people will have never seen MW composites before.

PR #3339, opened 2026-02-17
Make sure FCI L2 reader returns cloud type and other categorical data as int

### LINKED ISSUE #3338: FCI L2 reader returns categorical data as float
**Describe the bug**

Reading categorical data from FCI L2, such as cloud type, results in a float32.

This should be int8 or uint8.

**To Reproduce**

```python
# Your code here
from satpy import Scene
sc = Scene(
        filenames=["/media/nas/x23352/MTG/FCI/L2-cases/202602121200/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-2-CT--FD------NC4E_C_EUMT_20260212121354_L2PF_OPE_20260212120000_20260212121000_N__C_0073_0000.nc"],
        reader="fci_l2_nc")
sc.load(["cloud_type"])
print(sc["cloud_type"].dtype)
```

**Expected behavior**

I expect an integer dtype.

**Actual results**

```
[DEBUG: 2026-02-13 10:31:47 : satpy.readers.core.yaml_reader] Reading ('/home/gholl/checkouts/satpy/satpy/etc/readers/fci_l2_nc.yaml',)
[DEBUG: 2026-02-13 10:31:47 : satpy.readers.core.yaml_reader] Assigning to fci_l2_nc: ['/media/nas/x23352/MTG/FCI/L2-cases/202602121200/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-2-CT--FD------NC4E_C_EUMT_20260212121354_L2PF_OPE_20260212120000_20260212121000_N__C_0073_0000.nc']
[DEBUG: 2026-02-13 10:31:47 : rasterio.session] Could not import boto3, continuing with reduced functionality.
/home/gholl/checkouts/satpy/satpy/readers/fci_l2_nc.py:185: UserWarning: The specified chunks separate the stored chunks along dimension "number_of_rows" starting at index 4096. This could degrade performance. Instead, consider rechunking after loading.
  self.nc = xr.open_dataset(
/home/gholl/checkouts/satpy/satpy/readers/fci_l2_nc.py:185: UserWarning: The specified chunks separate the stored chunks along dimension "number_of_columns" starting at index 4096. This could degrade performance. Instead, consider rechunking after loading.
  self.nc = xr.open_dataset(
[DEBUG: 2026-02-13 10:31:47 : satpy.composites.config_loader] Looking for composites config file fci.yaml
[DEBUG: 2026-02-13 10:31:47 : pyorbital.tlefile] Path to the Pyorbital configuration (where e.g. platforms.txt is found): /home/gholl/miniforge3/envs/py313/lib/python3.13/site-packages/pyorbital/etc
[DEBUG: 2026-02-13 10:31:47 : satpy.composites.config_loader] Looking for composites config file visir.yaml
[DEBUG: 2026-02-13 10:31:47 : satpy.readers.fci_l2_nc] Reading in file to get dataset with key cloud_type.
[DEBUG: 2026-02-13 10:31:47 : satpy.readers.core.yaml_reader] Requested orientation for Dataset cloud_type is 'native' (default). No flipping is applied.
float32
```

**Environment Info:**
 - OS: openSUSE Leap 15.6
 - Satpy Version: v0.59.0-117-g5e5e33cbf

**Additional context**

When writing the output as mode L geotiff, trollimage will scale the pixel values.  This results in pixel values that are wrong.  This wouldn't happen if the data were int.

gerritholl:
Make sure that the FCI L2 reader returns categorical data, such as cloud type, cloud phase, cloud mask, pixel quality etc. as an exact dtype, de facto an integer dtype (in the source data those are enum).

This change may break backwards compatibility for users expecting a cloud mask or a cloud type as a float.

 - [x] Closes #3338
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/fci_l2_nc.py
+++ b/satpy/readers/fci_l2_nc.py
@@ -233,7 +233,7 @@ class FciL2NCFileHandler(FciL2CommonFunctions, BaseFileHandler):
         if dataset_info["file_type"] == "nc_fci_test_clm":
             variable = self._decode_clm_test_data(variable, dataset_info)
 
-        if "fill_value" in dataset_info:
+        if "fill_value" in dataset_info and np.issubdtype(variable.dtype, np.inexact):
             variable = self._mask_data(variable, dataset_info["fill_value"])
 
         variable = self._set_attributes(variable, dataset_info)
--- a/satpy/tests/reader_tests/test_fci_l2_nc.py
+++ b/satpy/tests/reader_tests/test_fci_l2_nc.py
@@ -234,7 +234,6 @@ class TestFciL2NCFileHandler(unittest.TestCase):
         assert attributes["flag_values"] == [0,1]
         assert "flag_meanings" in attributes
         assert attributes["flag_meanings"] == ["False","True"]
-        assert dataset.dtype == np.uint8
 
     def test_enum_with_fill_value_remains_int(self):
         """Test that enum with a fill value (such as cloud type) remains uint8."""
@@ -244,7 +243,7 @@ class TestFciL2NCFileHandler(unittest.TestCase):
                                        "file_type": "test_file_type",
                                        "import_enum_information": True,
                                        "fill_value": -127})
-        assert dataset.dtype == np.int8
+        assert np.issubdtype(dataset.dtype, np.integer)
 
     def test_units_from_file(self):
         """Test units extraction from NetCDF file."""


strandgren:
With the proposed fix you effectively disable the fill_value functionality for the integer data. In that case we should probably remove all `fill_value=-127` entries in the `fci_l2_nc.yaml` file as well in order to be consistent and avoid confusion. That means that it would also be removed from the xarray.DataArray attribute, but since it's not really correct I don't see that as a problem?

Another solution would be to treat float and integer data independently in `_mask_data` (https://github.com/gerritholl/satpy/blob/bug-fci-cloud-type/satpy/readers/fci_l2_nc.py#L157). If done properly we could then set the requested fill values to NaN for float data and to the corresponding fill value for different integer data (-127, 255, -32767, 65535, etc). I have read somewhere about determining the relevant integer fill-value for a given data type in satpy, but I can't recall where. On the other hand I don't see the added value of such a solution, since I don't think we currently want to convert any integer values to a given (different) fill-value (which we also don't do now).

Hence, I'd vote for the first alternative (which you have started), but also cleaning up the corresponding `fill_value=-127` entries in the yaml-file.

Issue #3259, opened 2025-10-12
Relicense to Apache Version 2


djhoese:
## Overview

The maintainers of this project and some leading members of the Pytroll
organization would like to re-license the satpy Python package.
We would like to first ask for opinions and permission from existing
contributors of the project before making this change official.
The software source code would change from a GPL Version 3.0 or Later
license to an Apache Version 2 license. A simple non-exhaustive
summary of these licenses and their differences can be found
at the following URLs along with the full text of the license:

https://www.tldrlegal.com/license/gnu-general-public-license-v3-gpl-3

https://www.tldrlegal.com/license/apache-license-2-0-apache-2-0

## Purpose

By moving away from the current GPLv3 license to a more permissive
license we hope to encourage more collaboration. We also hope this
opens the doors for more funding opportunities as some grants have
explicitly not funded non-permissive/copyleft licenses like GPLv3.

This re-licensing does not change any understanding or assignment
of copyright for this project. Copyright notices will remain with
the collective "satpy developers" identifier.

## Why not license X?

The Apache Version 2 license was chosen because it is a permissive
and open license like the MIT or BSDv3 licenses. It has the additional
requirement that any modified files must "carry prominent notices" stating
that changes were made to the original work. Lastly, Apache Version 2
is a very popular license even if other licenses may
provide similar permissions and restrictions.

## What do we want from you (an existing contributor)?

Please comment on this GitHub issue or contact
a project maintainer (see contributor list below) with your approval
(ex. "I approve") or disapproval (ex. "I do not approve") or
other feedback about this planned re-licensing. We can contact
contributors by email if necessary or requested.

## Contributor and Approval List

- [x] @djhoese (maintainer)
- [x] @adybbroe (maintainer)
- [x] @ameraner (maintainer)
- [x] @gerritholl (maintainer)
- [x] @mraspaud (maintainer)
- [x] @pnuu (maintainer)
- [x] @sfinkens (maintainer)
- [x] @strandgren (maintainer)
- [ ] @AppLEaDaY
- [x] @ArrayBolt3
- [x] @BENR0
- [x] @BengtRydberg
- [x] @ClementLaplace
- [x] @ColinDuff
- [x] @Graenni
- [x] @Isotr0py
- [x] @katherinekolman
- [x] @KatrinPoepp
- [ ] @Kexin828
- [x] @LocutusOfBorg
- [x] @LTMeyer
- [x] @Nalin7parihar
- [x] @StefanSnippetCoder
[...]

PR #3342, opened 2026-02-28
Make Scene bilinear reduction halo-aware and single-stage

Zaczero:
- [x] Tests added 

### DIFF
--- a/satpy/scene.py
+++ b/satpy/scene.py
@@ -40,6 +40,8 @@ from satpy.utils import convert_remote_files_to_fsspec, get_storage_options_from
 
 LOG = logging.getLogger(__name__)
 
+BILINEAR_REDUCTION_HALO = 1
+
 
 def _get_area_resolution(area):
     """Attempt to retrieve resolution from AreaDefinition."""
@@ -58,6 +60,13 @@ def _aggregate_data_array(data_array, func, **coarsen_kwargs):
     else:
         out = getattr(res, func)()
     return out
+
+
+def _expand_slice(slc: slice, size: int, padding: int):
+    """Expand a slice by a symmetric padding within array bounds."""
+    start = 0 if slc.start is None else max(0, slc.start - padding)
+    stop = size if slc.stop is None else min(size, slc.stop + padding)
+    return slice(start, stop, slc.step)
 
 
 class DelayedGeneration(KeyError):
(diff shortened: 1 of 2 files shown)


Zaczero:
I bundled 3 low impact commits in this PR. The reduce_data is a bugfix, and limit_output/include_scanline_acq_time are optimization related tunables. I hope that's okay!

djhoese:
FYI @Zaczero Satpy is in the process of relicensing from a GPLv3 license to an Apachev2 license. Since your recent PRs are your first contribution I assume you're OK with this? We've asked for approval from past contributors in this issue #3259.

Zaczero:
djhoese ye no problem

pnuu:
I'm thinking this was an intentional omission by me to be sure to get data also for the border areas just outside of the target area. Without those there will be missing values. Might be worth to try few different input and output area combinations.

I haven't used the resampler in years since I only use contiguous data (or VIIRS) and gradient search (and EWA) is so much more efficient.

Zaczero:
@pnuu this clarified a lot. I didn't notice the duplicated reduction logic on scene/resampler stages.

So by default it was doing the worst of both worlds: the scene crop would trim out the 1px halo and perform double processing. The correct approach when working with the bilinear resampler was to turn off outer reduce_data.

I changed the approach now, and when scene detects a bilinear resampler, it disables its internal reduction and accounts for 1px padding.

## 1. Old default

- Timing:
    - Sample 1: 1.514s
    - Sample 2: 1.537s
- Accuracy vs full no-reduction baseline:
    - bt108: lost valid pixels 135 / 135
    - bt120: lost valid pixels 135 / 135

## 2. Old reduce_data=False

- Timing:
    - Sample 1: 3.070s
    - Sample 2: 3.058s
- Accuracy vs full no-reduction baseline:
    - bt108: lost valid pixels 0 / 0
    - bt120: lost valid pixels 0 / 0

## 3. New default

- Timing:
    - Sample 1: 1.532s
    - Sample 2: 1.449s
- Accuracy vs full no-reduction baseline:
    - bt108: lost valid pixels 0 / 0
    - bt120: lost valid pixels 0 / 0

## 4. New reduce_data=False

- Timing:
    - Sample 1: 3.116s
    - Sample 2: 3.143s
- Accuracy vs old reduce_data=False:
    - bt108: lost valid pixels 0 / 0
    - bt120: lost valid pixels 0 / 0

PR #3343, opened 2026-02-28
Add optional include_scanline_acq_time flag to SEVIRI native reader

Zaczero:
- True: mean 0.1455s
      - False: mean 0.0306s
      - Delta: -0.1149s, -78.95% mean (n=20 each)

When disabled, reduces scene.load(BT_QUERIES) runtime by 80%.

 - [x] Tests added 
 - https://github.com/pytroll/satpy/pull/3342 Add your name to `AUTHORS.md` if not there already

### DIFF
--- a/satpy/readers/core/seviri.py
+++ b/satpy/readers/core/seviri.py
@@ -134,6 +134,15 @@ The SEVIRI L1.5 readers provide the following metadata:
       scn['IR_108']['y'] = mi
       scn['IR_108'].sel(time=np.datetime64('2019-03-01T12:06:13.052000000'))
 
+  Computing this coordinate can add a performance penalty for some readers.
+  For ``seviri_l1b_native`` it can be disabled with:
+
+  .. code-block:: python
+
+      scene = satpy.Scene(filenames,
+                          reader='seviri_l1b_native',
+                          reader_kwargs={'include_scanline_acq_time': False})
+
 * HRIT and Native readers can add raw metadata from the file header, such
   as calibration coefficients, to dataset attributes. Use the reader keyword
   argument ``include_raw_metadata``. Here's an example for extracting
--- a/satpy/readers/seviri_l1b_native.py
+++ b/satpy/readers/seviri_l1b_native.py
@@ -173,6 +173,13 @@ class NativeMSGFileHandler(BaseFileHandler):
 
     See :mod:`satpy.readers.core.seviri`.
 
+    **Scanline acquisition time**
+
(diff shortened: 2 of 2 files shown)


sfinkens:
Thanks for your contribution! I can confirm the performance penalty.

But I'm curious: What is your use case where this matters? Compared to the time it takes to load the image data (~5s for a single channel on my machine) the extra 0.1 seconds seems small to me.

Zaczero:
I'm processing large amounts of historical data for research. With all the PRs I submitted and all the wins here and there, I achieved 2x less CPU usage. It will be open source in a few weeks so there you will see the exact use case.

sfinkens:
Thanks for sharing! I just realized that I made a timing mistake. It actually takes 0.5s to load the image data, so the extra 0.1s are indeed relevant, especially for large amounts of historical data.

I'm wondering if the timestamp computation can be improved. I quickly changed  [`core/seviri.py:get_cds_time`](https://github.com/pytroll/satpy/blob/f5bd750f41dc694c105d7e14b6859708273a243a/satpy/readers/core/seviri.py#L422) to using plain integer arrays, but that didn't help much. Maybe someone else can spot an obvious performance blocker?

Otherwise I would be ok with adding this new argument. In that case I think it would be nice to add a little bit of documentation in two places

- Class docstring of [`NativeMSGFileHandler`](https://github.com/pytroll/satpy/blob/f5bd750f41dc694c105d7e14b6859708273a243a/satpy/readers/seviri_l1b_native.py#L154C7-L154C27)
- Then there's a section about `acq_time` in the [`core/seviri.py`](https://github.com/pytroll/satpy/blob/f5bd750f41dc694c105d7e14b6859708273a243a/satpy/readers/core/seviri.py#L126) module. I think it would be nice to mention the performance penalty there and that you can turn it off for the Native reader.

sfinkens:
For the record my timing approach

```python
import timeit

filename = "~/MSG4-SEVI-MSG15-0100-NA-20190409121243.927000000Z-20190409121300-1451742.nat"
setup = f"""from satpy import Scene
scene = Scene(['{filename}'], reader="seviri_l1b_native")
"""
expressions = [
    "scene.load(['IR_108'])",
    "scene.load(['IR_108']); scene['IR_108'].compute()"
]
for expression in expressions:
    t = timeit.Timer(expression, setup=setup)
    res = t.repeat(repeat=25, number=1)
    print(min(res))
```

Results
```
                 w/ acq_time  w/o acq_time  Reduction [%]
          load         0.055         0.021             38
load + compute         0.21          0.18              10
```

PR #3313, opened 2025-12-08
Add new composites from the wmo document

mraspaud:
This PR implements the recommendations from the WMO RGB document

 - [ ] Closes #3269 
 - [ ] Tests added 
 - [ ] Fully documented 
 - [ ] Add your name to `AUTHORS.md` if not there already

Todo list:
- [ ] ABI
  - [x] airmass
  - [ ] 24h_microphysics_cloud
  - [ ] 24h_microphysics_dust
  - [ ] 24h_microphysics_cvd_dust
  - [x] 24h_microphysics_ash
  - [ ] night_microphysics
  - [ ] night_microphysics_tropical
  - [ ] day_cloud_phase
  - [ ] day_microphysics
  - [ ] day_cloud_type
  - [ ] day_cloud_type_distinction
  - [ ] day_severe_storms
  - [ ] day_severe_storms_tropical
  - [ ] overshooting_tops
  - [ ] simple_water_vapour
  - [ ] volcanic_emissions
  - [ ] true_colour
  - [ ] fire_temperature
  - [ ] day_fire
  - [ ] day_fire_reflectance
  - [ ] day_land_cloud
  - [ ] day_hrv_clouds
  - [ ] hrv_fog
  - [ ] day_snow_fog
- [ ] AGRI
  - [ ] ~airmass~
  - [ ] 24h_microphysics_cloud
  - [ ] 24h_microphysics_dust
  - [ ] 24h_microphysics_cvd_dust
  - [ ] 24h_microphysics_ash
  - [ ] night_microphysics
  - [ ] night_microphysics_tropical
  - [ ] day_cloud_phase
  - [ ] day_microphysics
  - [ ] day_cloud_type
  - [ ] day_cloud_type_distinction
  - [ ] day_severe_storms
  - [ ] day_severe_storms_tropical
  - [ ] overshooting_tops
  - [ ] simple_water_vapour
  - [ ] volcanic_emissions
  - [ ] true_colour
  - [ ] fire_temperature
  - [ ] day_fire
  - [ ] day_fire_reflectance
  - [ ] day_land_cloud
  - [ ] day_hrv_clouds
  - [ ] hrv_fog
  - [ ] day_snow_fog
- [ ] AHI
  - [x] airmass
  - [ ] 24h_microphysics_cloud
  - [ ] 24h_microphysics_dust
  - [ ] 24h_microphysics_cvd_dust
  - [x] 24h_microphysics_ash
  - [ ] night_microphysics
  - [ ] night_microphysics_tropical
[...]

Issue #3230, opened 2025-09-08
Add enhacement key for defining composite enhancement


strandgren:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Dataset and composite enhancements can be defined/linked to using a set of identifiers:
```
1. name
2. reader
3. platform_name
4. sensor
5. standard_name
6. units
```
In most cases the `standard_name` is used as identifier, especially for pre-configured composite recipes in e.g. `./etc/composites/visir.yaml`. To me this has three drawbacks:

- The link between composite recipe and and enhancement is not very obvious/transparent for new users. I struggled myself in the beginning.
- Using attributes like `name`/`standard_name` may lead to names that are not necessarily useful/optimal for downstream tools using these attributes to e.g. label data/imagery`. For example `standard_name: geo_color_day_night_blend` for `geo_color`. 
- Some enhancements may be duplicated since the `standard_name` of an existing enhancement may not necessarily match the desired `standard_name` of a new composite being added.

**Describe the solution you'd like**
I would like to add a new enhancement identifier called `enhancement` (or similar) and put it at the top of the decision tree. The idea would be that in composite recipes we can use the `name` and `standard_name` attributes purely to attach useful information and then add a new `enhancement` key that tells satpy (and other users) which enhancement to use.

Following the current logic, if the `enhancement` key is not defined, e.g. for a dataset being created dynamically (like `toa_bidirectional_reflectance`) we'd fall back on the `name`/`standard_name` etc. further down the decision tree, thus preserving backward compatibility.

One thing that I don't know how to implement, is how to link the dataset/composite `enhancement` key to a given enhancement defined in a yaml-file. Because we don't wont to add the `enhancement` key to the enhancements itself, since that would require all datasets/composites that should use that enhancement to have the `enhancement` defined in the attributes, which is something that we probably don't want to do, especially for datasets which are created dynamically and get the attributes from the reading part. Perhaps one could link the `enhancement` key to the `name` or `standard_name` of the enhancement, where the latter is already defined for most enhancements making it a small and transparent change.

In that case, one could also skip adding the `enhancement` key to existing composites, where the `standard_name` is already good enough to describe the enhancement (e.g. dust, airmass, etc), although I would still be in favour of adding it for clarity.

Hence, in case of `geo_color` the composite and enhancement could look like this:
```
# Composite
  geo_color:
    compositor: !!python/name:satpy.composites.fill.DayNightCompositor
    lim_low: 78
    lim_high: 88
[...]

PR #3347, opened 2026-03-03
Introduce composite tags

mraspaud:
This PR introduce a system of tagging for composites, allowing switching preferred variants at load time or by configuration.

For example, having the following config:
```yaml
  true_color_crefl:
    tags: ["crefl"]
    compositor: !!python/name:satpy.composites.resolution.SelfSharpenedRGB
    prerequisites: ...
    standard_name: true_color

  true_color_raw:
    tags: ["raw"]
    compositor: !!python/name:satpy.composites.resolution.SelfSharpenedRGB
    prerequisites: ...
    standard_name: true_color

  true_color:
    compositor: !!python/name:satpy.composites.resolution.SelfSharpenedRGB
    prerequisites: ...
    standard_name: true_color

  true_color_nocorr:
    tags: ["nocorr"]
    compositor: !!python/name:satpy.composites.resolution.SelfSharpenedRGB
    prerequisites: ...
    standard_name: true_color
```

We can then load `scn.load(["true_color:crefl"])` directly to load "true_color_crefl". Alternatively, we can set the config with `SATPY_PREFERRED_COMPOSITE_TAGS=crefl` so that `scn.load(["true_color"])` loads "true_color_crefl" by default.

This is needed in order implement WMO RGB recommendations in a backwards- and forwards-compatible manner.

Also, possibility to warn for composites at load time has been added.

Disclaimer: AI was used in the design and implementation of this PR.

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -495,6 +495,164 @@ image) for both of the static images::
           min_stretch: [0, 0, 0]
           max_stretch: [255, 255, 255]
 
+.. _composite_variants:
+
+Composite variants
+------------------
+
+.. versionadded:: 0.60
+
+Satpy supports defining multiple *variants* of a composite (e.g., one that
+applies WMO-recommended recipe and one that does not).  This feature is controlled by optional
+fields in the composite YAML configuration.
+
+Tagging composite variants
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A composite can carry a list of **tags** that describe which processing variant
+it represents.  Tags are plain strings (e.g., ``wmo``, ``crefl``,
+``nocorr``) and have no special meaning to Satpy beyond being matched during
+compositor lookup.
+
+.. code-block:: yaml
+
+    sensor_name: visir
+
+    composites:
+      true_color_wmo:
+        compositor: !!python/name:satpy.composites.SomeCompositor
+        prerequisites:
+          - name: red
+            modifiers: [rayleigh_corrected_wmo]
+          - name: green
+          - name: blue
+        standard_name: true_color
+        tags: [wmo]
+
+      true_color_crefl:
+        compositor: !!python/name:satpy.composites.SomeCompositor
+        prerequisites:
+          - name: red
+            modifiers: [rayleigh_corrected_crefl]
+          - name: green
+          - name: blue
+        standard_name: true_color
+        tags: [crefl]
+
+      true_color:                          # default, no tags
+        compositor: !!python/name:satpy.composites.SomeCompositor
+        prerequisites:
+          - name: red
+          - name: green
+          - name: blue
+        standard_name: true_color
+
+The ``standard_name`` field acts as the *base name* shared by all variants.
+Tag-based resolution searches for a compositor whose ``standard_name`` matches
+and whose ``tags`` list contains the requested tag.
+
+Loading a tagged variant explicitly
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Append ``:<tag>`` to the composite name when calling :meth:`~satpy.scene.Scene.load`:
+
+.. code-block:: python
+
+    scene.load(["true_color:wmo"])   # loads true_color_wmo
+
+This syntax is interpreted as: "find a compositor with
+``standard_name='true_color'`` that has ``'wmo'`` in its ``tags``".  It never
+performs a plain string match, so ``true_color:wmo`` will not accidentally
+resolve to a compositor named ``true_color`` or ``true_color_wmo``.
+
+Multiple tags can be combined with additional colons.  All listed tags must be
+present on the compositor (AND semantics):
+
+.. code-block:: python
+
+    scene.load(["true_color:wmo:pyspectral"])  # compositor must carry wmo AND pyspectral
+    scene.load(["true_color:wmo"])             # also matches a compositor with tags [wmo, pyspectral]
+
+A single-tag request is a subset match, so requesting ``"true_color:wmo"``
+will find a compositor tagged ``["wmo", "pyspectral"]`` just as readily as one
+tagged ``["wmo"]`` alone.
+
+Session-wide tag preferences
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can configure an **ordered** list of preferred tags so that loading an
+unqualified name like ``"true_color"`` automatically selects a tagged variant
+when one is available:
+
+.. code-block:: python
+
+    import satpy
+    with satpy.config.set(preferred_composite_tags=["crefl", "wmo"]):
+        scene.load(["true_color"])   # picks true_color_crefl (crefl listed first)
+
+Resolution order for ``preferred_composite_tags``:
+
+1. Try each tag in the list, in order, looking for a compositor with matching
+   ``standard_name`` and that tag.
+2. If no tagged variant is found, fall back to the normal name-based lookup.
+
+An explicit ``name:tag`` in the load call always overrides the session-wide
+preference for that specific dataset.
+
+The setting can also be provided as an environment variable (comma-separated):
+
+.. code-block:: bash
+
+    export SATPY_PREFERRED_COMPOSITE_TAGS=crefl,wmo
+
+Or as a YAML configuration key:
+
+.. code-block:: yaml
+
+    preferred_composite_tags:
+      - crefl
+      - wmo
+
+Enhancements for tagged variants
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+No extra configuration is needed on the enhancement side.  The enhancement
+decision tree already matches by composite ``name`` (the YAML key) *before*
+it falls back to ``standard_name``.  A composite named ``true_color_wmo`` will
+therefore automatically pick up an enhancement entry keyed ``true_color_wmo``
+if one exists, and fall back to the ``standard_name: true_color`` entry
+otherwise.
+
+Deprecating a composite
+------------------------
+
+To emit a warning when a composite is used, add a ``warnings`` mapping to the
+composite YAML entry.  Each key is a Python warning category name and the
+value is the warning message:
+
+.. code-block:: yaml
+
+    composites:
+      old_true_color:
+        compositor: !!python/name:satpy.composites.SomeCompositor
+        prerequisites:
+          - name: red
+          - name: green
+          - name: blue
+        standard_name: true_color
+        warnings:
+          DeprecationWarning: "old_true_color is deprecated, use true_color_wmo instead."
+
+The warning is emitted only when the compositor is actually *loaded* (i.e.
+when :meth:`~satpy.scene.Scene.load` is called with the deprecated name), not
+during composite discovery calls such as
+:meth:`~satpy.scene.Scene.available_dataset_names`.
+
+Supported warning categories are any that exist in the Python ``builtins``
+module (e.g., ``DeprecationWarning``, ``FutureWarning``,
+``PendingDeprecationWarning``, ``UserWarning``).  If a category name is not
+recognised, ``UserWarning`` is used as a fallback.
+
 .. _enhancing-the-images:
 
 Enhancing the images
--- a/doc/source/composites.rst
+++ b/doc/source/composites.rst
@@ -558,12 +558,39 @@ Append ``:<tag>`` to the composite name when calling :meth:`~satpy.scene.Scene.l
 
 .. code-block:: python
 
-    scene.load(["true_color:wmo"])   # loads true_color_wmo
+    scene.load(["true_color:wmo"])
+
+    # The dataset is stored and accessible using the same tag syntax:
+    data = scene["true_color:wmo"]
 
 This syntax is interpreted as: "find a compositor with
 ``standard_name='true_color'`` that has ``'wmo'`` in its ``tags``".  It never
 performs a plain string match, so ``true_color:wmo`` will not accidentally
 resolve to a compositor named ``true_color`` or ``true_color_wmo``.
+
+Accessing tag-loaded datasets
+""""""""""""""""""""""""""""""
+
+The dataset is stored in the Scene under the *requested* name (including the
+tag suffix) rather than the compositor's internal YAML key.  Use the same
+tag syntax to retrieve it:
+
+.. code-block:: python
+
+    scene.load(["true_color:wmo"])
+    data = scene["true_color:wmo"]
+
+The compositor's YAML key name (e.g. ``"true_color_wmo"``) is preserved in
+``data.attrs["_satpy_compositor_name"]`` and is used automatically for
+enhancement lookups (see below).
+
+This also means that :attr:`~satpy.scene.Scene.missing_datasets` is empty
+after a successful load, and that writing the scene with any writer will use
+``"true_color:wmo"`` as the dataset name.
+
+The same rule applies when a tagged variant is selected automatically via
+``preferred_composite_tags``: the dataset is stored under the plain requested
+name (e.g. ``"true_color"``), not under the compositor's YAML key name.
 
 Multiple tags can be combined with additional colons.  All listed tags must be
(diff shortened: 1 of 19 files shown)


doc/source/composites.rst
```diff
@@ -495,6 +495,164 @@ image) for both of the static images::
...
+
+    import satpy
+    with satpy.config.set(preferred_composite_tags=["crefl", "wmo"]):
+        scene.load(["true_color"])   # picks true_color_crefl (crefl listed first)
+
+Resolution order for ``preferred_composite_tags``:
+
+1. Try each tag in the list, in order, looking for a compositor with matching
+   ``standard_name`` and that tag.
+2. If no tagged variant is found, fall back to the normal name-based lookup.
```
pnuu, line 632: Which variant is used if the tag is not defined in the load and the global env variable is not set and all/some of the following composites are defined?
* `foo_wmo`
* `foo_crefl`
* `foo`

Each of them have the same `standard_name: foo` defined, no `name` is set. Only the first line in the YAML definition is different and for the first two there are tags set.

So what would I get with `scn.load(["foo"])` call in the following cases:
1. all variants are defined (I'd expect `foo` variant without any tags)
2. only the two tagged versions are defined

gerritholl:
This looks great!

This PR adds a special meaning to "standard_name".  Does this break backward compatibility?  If I understand correctly, loading composites will no longer match by the key in the YAML file, but by the contents of the standard name?

We (at least in DWD) currently (ab)use "standard_name" to get a match between a composite and its enhancement.  For example, we have many composites with standard name `image_ready`.  I've always felt a bit uneasy by using the standard name attribute for making this match, and this PR might bite with our current use.

As an example of what we currently use:

```
  day_cloud_type:
    compositor: !!python/name:satpy.composites.fill.DayNightCompositor
    day_night: day_only
    prerequisites:
    - name: cloud_type
    standard_name: image_ready
```

Is this still loadable as `sc.load(["day_cloud_type"])`?  Or would this be matched only by `sc.load(["image_ready"])`?

doc/source/composites.rst
```diff
@@ -495,6 +495,164 @@ image) for both of the static images::
...
+An explicit ``name:tag`` in the load call always overrides the session-wide
+preference for that specific dataset.
+
+The setting can also be provided as an environment variable (comma-separated):
+
+.. code-block:: bash
+
+    export SATPY_PREFERRED_COMPOSITE_TAGS=crefl,wmo
+
+Or as a YAML configuration key:
```
pnuu, line 651: In which config file should this be defined?

pnuu (changes requested):
Some comments inline. I guess the variant loading without defining the variant in any way will need a test where different combinations are defined as options.

pnuu:
Good point Gerrit. I think that is actually the only use-case for `standard_name` for me, linking the composite and enhancement together. I think this is also related to my variant selection comment.

mraspaud:
> This looks great!

Thanks :)

> This PR adds a special meaning to "standard_name". Does this break backward compatibility? If I understand correctly, loading composites will no longer match by the key in the YAML file, but by the contents of the standard name?

It is backwards compatible. Names of compositors that do not contain ":" are processed as usual.
The use of the standard name here is because it was the closest thing to a common name for different variants we had, so I used that instead of introducing an new item in the composite config.

> We (at least in DWD) currently (ab)use "standard_name" to get a match between a composite and its enhancement. For example, we have many composites with standard name `image_ready`. I've always felt a bit uneasy by using the standard name attribute for making this match, and this PR might bite with our current use.
> 
> As an example of what we currently use:
> 
> ```
>   day_cloud_type:
>     compositor: !!python/name:satpy.composites.fill.DayNightCompositor
>     day_night: day_only
>     prerequisites:
>     - name: cloud_type
>     standard_name: image_ready
> ```
> 
> Is this still loadable as `sc.load(["day_cloud_type"])`? Or would this be matched only by `sc.load(["image_ready"])`?

We use the same thing at SMHI, and the current behaviour should be unaffected. I agree it's not really clean to use standard_name for this purpose, would be nice if we had a good solution for that.

doc/source/composites.rst
mraspaud: If tags are not provided, neither at load time or in the configuration, the current behaviour is preserved, ie we will use the name (not standard name) to choose the composite to load.

1. you get the composite name "foo"
2. crash if you don't provide a tag (at load time or as config).

doc/source/composites.rst
mraspaud: at the top of this file, you have your answer :)

**YAML Configuration**

YAML files that include these parameters can be in any of the following
locations:

1. ``<python environment prefix>/etc/satpy/satpy.yaml``
2. ``<user_config_dir>/satpy.yaml`` (see below)
3. ``~/.satpy/satpy.yaml``
4. ``<SATPY_CONFIG_PATH>/satpy.yaml`` (see :ref:`config_path_setting` below)

mraspaud: sorry, it was actually in config.rst

pnuu: Oh, I didn't even know we had a `satpy.yaml` config file :sweat_smile:

doc/source/composites.rst
pnuu: So something like this for the first case? I'm not sure how this works, but hopefully it's close enough:

```python
        pytest.param(
            {"comp1": None, "comp1_wmo": ["wmo"]},
            None,
            "comp1",
            "comp1",
            id="use_plain_version_when_no_tag",
        ),
```

For case 2. the failure might need a completely separate test?

djhoese:
I think I've said this before, but I was never a fan of the `"image_ready"` standard name since, as we're discovering here, it has a more generic but also functional meaning the way it is used. There is already the other issue (PR?) about having a specific enhancement defined for a composite inside the composite YAML and that may be the long term way of doing this explicit composite definition. I think I had hoped that a more generic "if it's an RGB/A with no other enhancement defined, default to no-op" enhancement.

The other thing composite tags is bringing up which @gerritholl had mentioned in the meeting yesterday is what does a `DataID/DataQuery` represent and how do tags differ? For the most part DataID defines what the product **is** and tags are defining a "preference" I guess. The main exception is "modifiers" where these lines are blurred.

We may need to define a "family" for composites or "template" or something like that for tags instead of reusing "standard_name" for this. At least that's my gut feeling right now.

mraspaud:
I agree that the `image_ready` enhancement handling can be improved, however I don't think it is within the scope of this PR.

Regarding the similarity with the data id/data query: Indeed, this is quite similar, however, IMO this PR presents a much more user-friendly interface to a problem that many users with custom configs face: they define a variant for a composite and have a simple way to access it without overwriting the default composite. I am aware that I short circuited the data id entirely here, because I wanted to keep the tags separate from it.

I'm open to feedback of course, if you thing this is the wrong decision.

The last commit added the access to composites in the scene using tags btw, and I'd love to get a review on that.

doc/source/composites.rst
djhoese: The satpy.yaml isn't actually a config file in the "component config" (at least I don't think). It is the YAML that is loaded by donfig as the global Satpy configuration values (not composites, enhancements, etc).

djhoese (commented):
I'm sorry but I don't really understand. At a high level I've never really liked that we depend on a `.attrs["_satpy_id"]` magic attribute, but DataIDs are central to what Satpy does so it seemed like an important exception to allow this "magic". This new "_satpy_compositor_name" seems to be doing a lot for essentially an alias. If I'm not mistaken the two options were:

1. Make the Scene and `Scene.__getitem__` calls understand tag references and store composites with their YAML name.
2. Store things as their requested/tagged name and update enhancements and composites to understand them.

So I think this PR is option number 2, but does it actually handle composites that reference a composite? For example, if a composite depends on `true_color_wmo` but that composite is now sitting in the Scene as `true_color:wmo`, does that work?

I think I'd still prefer a solution where "standard_name" wasn't what tagged names are based on, but rather some other new key. At least I think. Overall I'm just extremely scared of the dependency tree and enhancement configuration stuff needing to handle this as those things were already hard to understand and hold in my brain, not to mention a user's brain.

mraspaud:
> I'm sorry but I don't really understand. At a high level I've never really liked that we depend on a `.attrs["_satpy_id"]` magic attribute, but DataIDs are central to what Satpy does so it seemed like an important exception to allow this "magic". This new "_satpy_compositor_name" seems to be doing a lot for essentially an alias. 

That's correct, it is an alias to be able to go back to the yaml name, so that we can keep the tagged version superficial and be able to go back to the yaml name for the dependency tree for example.
An alternative would be to build a look-up table where we have all the variations for all composites listed with their tagged equivalent. I thought "_satpy_compositor_name" would be more contained.
Another alternative would we to gather all `_satpy_…` attrs into a eg `_satpy_internals` attribute.

> If I'm not mistaken the two options were:
> 
> 1. Make the Scene and `Scene.__getitem__` calls understand tag references and store composites with their YAML name.
> 2. Store things as their requested/tagged name and update enhancements and composites to understand them.
> 
> So I think this PR is option number 2, but does it actually handle composites that reference a composite? For example, if a composite depends on `true_color_wmo` but that composite is now sitting in the Scene as `true_color:wmo`, does that work?

Yes, this is option 2, but it's really just a name translation, so I would expect that use-case to work. But I'll get back to you on this.

> I think I'd still prefer a solution where "standard_name" wasn't what tagged names are based on, but rather some other new key. At least I think. Overall I'm just extremely scared of the dependency tree and enhancement configuration stuff needing to handle this as those things were already hard to understand and hold in my brain, not to mention a user's brain.

Well, I see standard_name as the composite category so to say, so for me it makes sense to use standard_name as a base for the tagging (and the enhancements). What would you call the alternative attribute?

doc/source/composites.rst
```diff
@@ -495,6 +495,205 @@ image) for both of the static images::
           min_stretch: [0, 0, 0]
           max_stretch: [255, 255, 255]
 
+.. _composite_variants:
+
+Composite variants
+------------------
+
+.. versionadded:: 0.60
```
djhoese: TODO: Update to 0.61 across PR...if not 0.62 if we feel we need a 0.61 that doesn't include this.

doc/source/composites.rst
```diff
@@ -495,6 +495,205 @@ image) for both of the static images::
...
+``data.attrs["_satpy_compositor_name"]`` and is used automatically for
+enhancement lookups (see below).
+
+This also means that :attr:`~satpy.scene.Scene.missing_datasets` is empty
+after a successful load, and that writing the scene with any writer will use
+``"true_color:wmo"`` as the dataset name.
+
+The same rule applies when a tagged variant is selected automatically via
+``preferred_composite_tags``: the dataset is stored under the plain requested
+name (e.g. ``"true_color"``), not under the compositor's YAML key name.
```
djhoese: I could be overthinking this and I've only been looking at the documentation before diving into the code but: There seems to be a two separate pieces here that are maybe used for the same thing. There is the composite name (the name via the YAML section) and then the standard_name + tag. This chunk of documentation here says that the dataset is stored "under the plain requested name", but how does that work?

If we have a composite with name "my_tc" and standard_name "true_color" and zero or more tags. If I do `scn.load(["true_color"])` then is that treated as a "no tag"/generic composite request and "my_tc" could be found? Or is that considered a search for the specific composite named "true_color" that may or may not exist? 

Another question, you say that "true_color:wmo" would be used during writing, so a geotiff with file pattern `"{name}.tif"` would be `true_color:wmo.tif`?

gerritholl: I think it would be like for datasets/channels.  A user-configurable priority (with some satpy-proposed default) would determine which composite is loaded or returned if there are multiple.  Although I don't think this is currently user-configurable for channels/datasets, we might want to make it so in the future.

strandgren:
> > I think I'd still prefer a solution where "standard_name" wasn't what tagged names are based on, but rather some other new key. At least I think. Overall I'm just extremely scared of the dependency tree and enhancement configuration stuff needing to handle this as those things were already hard to understand and hold in my brain, not to mention a user's brain.
> 
> Well, I see standard_name as the composite category so to say, so for me it makes sense to use standard_name as a base for the tagging (and the enhancements). What would you call the alternative attribute?

I would actually agree that the `standard_name` would be a good name for grouping different variants of the same RGB for the tagging, but I also agree that using `standard_name` for both this and the enhancement is not optimal. As mentioned in earlier comments I created an issue about this a while ago: https://github.com/pytroll/satpy/issues/3230. But for now that's just an idea and I wouldn't know how to properly implement it in practice - I also agree that it's probably outside the scope of the PR.

djhoese:
Using `standard_name` for this is the 3rd (?) use of standard_name. We have CF standard name which has meaning to the user and is written to NetCDF files but is also not necessarily/strictly CF compliant. We also have the enhancement stuff. We're definitely abusing a single value to mean multiple things, especially when that thing has meaning outside of pytroll/satpy.

doc/source/composites.rst
djhoese: @gerritholl sure, priorities are one thing, but I'm talking about the re-use of the string passed to `Scene.load`. Right now in a basic request you can have `scn.load([str or Number])` where a number means the wavelength and a search is done based on that. The string means the "name" in the `DataID`. Depending on how tags are implemented, with my `my_tc` example I would expect to need to do `scn.load(["my_tc"])` to load it by name or `scn.load(["true_color:wmo"])` to load it by tag (assuming it has a wmo tag). But if it is the most "generic" true color recipe then doing `scn.load(["true_color"])` seems confusing and error prone. That means that the search first by name if no `:` or by standard_name+tag if `:` present, then fallback to standard_name + no tag.

mraspaud: ok, the doc was a bit outdated. I fixed it, and it should answer your questions.

djhoese: Whoa! Ok with those latest changes this is a totally different beast. Before it was a composite having a human-friendly-ish name and a "family" for its recipe (the "standard_name"). Now those two things are the same. I have to think about that.

mraspaud:
Since the comments on this PR, there quite a few updates, which I hope addressed the concerns:

- `standard_name` is no longer used for tagging, so no the overloading. Variants share a name and differ by the new tag DataID key.
- The _satpy_compositor_name magic attribute is gone. The tagged DataID is the identificator.
- tag behaves like any other DataID key, with transitive: false (a tag does not propagate to dependencies) and match_empty: false (a tagged variant never matches an untagged query, so existing configs behave exactly as before).
- Composite definitions use a single `tag:`. The preference list is an ordered list, e.g. preferred_composite_tags: [chmi, wmo], resolved per composite with fallback to the untagged variant. Explicit scn.load(["true_color:wmo"]) is the strict form: it either loads exactly that variant or raises an error. The error now lists the tagged variants that do exist.
- Enhancements can match on tag, so the right enhancement follows the variant actually chosen, including when the preference falls back.
- available_composite_names() and all_composite_names() list one loadable spelling per composite: the plain name when an untagged variant exists, name:tag for tagged-only composites. Everything they return can be passed straight to load(), and load-everything scripts keep getting one product per composite.

@djhoese, to your open questions:
- Composites referencing tagged composites work: a prerequisite can be comp1:wmo (pinned by test_load_composite_with_tagged_prerequisite). Plain-name prerequisites also resolve through preferred_composite_tags, so a site preference reaches nested composites.
- Writer filenames: the default {name} pattern is deliberately unchanged. Switching the preferred variant keeps downstream filenames stable. `tag` is present in the attrs, so {name}_{tag} patterns are available when saving variants side by side (documented, including the caveat for untagged products).

@pnuu, your resolution matrix is now tested: plain wins when both variants exist; loading a tagged-only composite by plain name raises (with the hint above); preference order and untagged fallback are covered in test_dependency_tree.py.

Docs updated throughout (composites.rst, config.rst), including the strict-vs-lenient distinction and the config reference, which still described an older iteration.

The motivating use case is #3313: define the WMO RGB recipes as tag: wmo variants, so sites can switch between e.g. CHMI's variants and the WMO standard at any time via preferred_composite_tags, per composite, without renaming anything downstream.

PR #3355, opened 2026-03-23
Update Sen2/MSI L2A to work properly with composites

simon-sat:
The Sentinel-2 / MSI reader for L2A files does not support any composites as it uses a non-existent reader name.

This PR adds some L2 composites. It uses a separate YAML to the L1 composites, as the L2 composites do not need to have the atmospheric correction (Rayleigh + path length) applied.

### DIFF
--- a/satpy/etc/composites/sen2_msi_l2.yaml
+++ b/satpy/etc/composites/sen2_msi_l2.yaml
@@ -1,0 +1,144 @@
+sensor_name: visir/sen2_msi
+
+composites:
+  true_color:
+    compositor: !!python/name:satpy.composites.core.GenericCompositor
+    prerequisites:
+    - name: 'B04'
+    - name: 'B03'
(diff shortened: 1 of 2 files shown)


djhoese (commented):
What did you mean when you said "as it uses a non-existent reader name"? The reader name shouldn't matter for finding composites, but the sensor name does. The changes here don't really match what we usually do in Satpy.

1. The sensor name at the top of the composite config should match the sensor named used in the composite filename and the sensor the reader returns.
2. Typically to handle some products needing or not needing rayleigh and sunz correction we include the "modifiers" as part of the reader. See the VIIRS SDR reader as an example that includes sunz.
3. The satellite included in the sensor name is not crazy if the sensor is actually different enough than the sensor on other satellites, but I don't think that's the case here. The "l2" being included has never been done before as far as I know.

PR #3359, opened 2026-03-24
Fix `h5netcdf` backend variable caching

pnuu:
This PR fixes the variable caching for `h5netcdf` backend. The difference in the variable handling was missed in the refactoring done in https://github.com/pytroll/satpy/pull/3256

 - [x] Closes #3358  
 - [ ] Tests added


djhoese (approved):
LGTM if it works well with @mraspaud's #3331

PR #3360, opened 2026-03-27
Fix misspelling

andrey-khropov:
subj


djhoese (approved):
Thanks! Interestingly this word was also mistyped in another spot that was caught by @mraspaud in his #3331 PR. Turns out we have trouble typing dimension.

mraspaud:
Thanks for your contribution! We are in the process of changing the license of satpy to Apache v2.0, so the license of your contribution will not be GPLv3 in the long run. Moreover, don't forget to add your name to the authors list.

Issue #3003, opened 2024-11-29
DayNightCompositor with `day_only` does not mask out night part if no alpha layer is present, defeating its purpose


gerritholl:
**Describe the bug**

When using the `DayNightCompositor` with the parameter `day_night: day_only` and `include_alpha: False`, the night part is not set to fill value and the resulting geotiff (mode L or mode RGB) is the same as without using the `DayNightCompositor` at all.

**To Reproduce**

```python
import tempfile
import pathlib
import os
import rasterio

config = """sensor/name: visir

composites:
   day_vis_06:
    compositor: !!python/name:satpy.composites.DayNightCompositor
    prerequisites:
      - vis_06
    day_night: day_only
    include_alpha: False
    standard_name: image_ready
"""
with tempfile.TemporaryDirectory() as td:
    p = pathlib.Path(td)
    fn = p / "composites" / "fci.yaml"
    fn.parent.mkdir(exist_ok=True, parents=True)
    with fn.open(mode="wt", encoding="ascii") as fp:
        fp.write(config)
    os.environ["SATPY_CONFIG_PATH"] = td

    import hdf5plugin
    from satpy import Scene
    from glob import glob
    from satpy.utils import debug_on; debug_on()

    fci_files = glob(f"/media/nas/x23352/MTG/FCI/L1c-cases/20241025-nighttime-convection-mediterranean/09/*FDHSI*C_0055_*.nc")
    sc = Scene(filenames={"fci_l1c_nc": fci_files})
    sc.load(["day_vis_06"])
    ls = sc.resample("eurol")
    out = f"{td:s} / out.tif"
    ls.save_dataset("day_vis_06",
                    out,
                    write="geotiff",
                    fill_value=0)

    src = rasterio.open(out)
    arr = src.read(1)
    print(arr[0:5, 0:5])
```

**Expected behavior**

I expect that those pixels, which are night-time, are set to zero, in accordance to the fill value, so that supporting image viewers such as QGIS will display those pixels as transparent, like they do with space pixels.

**Actual results**

```
[[1 1 1 1 1]
 [1 1 1 1 1]
 [1 1 1 1 1]
 [1 1 1 1 1]
 [1 1 1 1 1]]
```

**Environment Info:**
 - OS: openSUSE Leap 15.3
 - Satpy Version: satpy main (v0.53.0-35-g096d5d9cc)
 
**Additional context**

[...]

PR #3368, opened 2026-04-02
Fix NaN handling in NDVIHybridGreen calculation

djhoese:
While working on generating FCI true colors @kathys noticed that the night portion of the image was always transparent instead of black like other instruments' true colors. I tracked it down to the NDVI Hybrid Green correction. The "problem" comes down to NDVI having `/ 0` in some cases which in numpy results in NaNs. This results in the correction/contribution fraction being NaN and the resulting green pixel value being NaN. In this PR I do an additional "clip" on the NDVI to convert NaNs to the minimum allowed NDVI as configured by the user. Given that the current result is different than what the regular hybrid compositor produces (ex. AHI's current default, black night pixels) and how simple the fix is I think this is the right way to go.

I still need to add a test, but wanted to get people's opinion.

### Old

true_color_20260329_063000 old

### New

true_color_20260329_063000 new

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/composites/spectral.py
+++ b/satpy/composites/spectral.py
@@ -166,7 +166,10 @@ class NDVIHybridGreen(SpectralBlender):
         ndvi = (projectables[2] - projectables[1]) / (projectables[2] + projectables[1])
 
         ndvi = ndvi.clip(self.ndvi_min, self.ndvi_max)
-        ndvi.data = np.nan_to_num(ndvi.data)
+        # dask nan_to_num does not accept kwargs (see https://github.com/dask/dask/issues/12350)
+        # Second argument maps to `copy` kwarg, third maps to `nan` kwarg.
+        # Copy should remain `True` as dask operations require copies to be made
+        ndvi.data = np.nan_to_num(ndvi.data, True, self.ndvi_min)
 
         # Introduce non-linearity to ndvi for non-linear scaling to NIR blend fraction
         if self.strength != 1.0:  # self._apply_strength() has no effect if strength = 1.0 -> no non-linear behaviour
--- a/satpy/composites/spectral.py
+++ b/satpy/composites/spectral.py
@@ -17,6 +17,8 @@
 
 import logging
 
+import numpy as np
+
 from satpy.composites.core import GenericCompositor
 from satpy.dataset import combine_metadata
 
@@ -110,7 +112,7 @@ class NDVIHybridGreen(SpectralBlender):
     """Construct a NDVI-weighted hybrid green channel.
 
     This green band correction follows the same approach as the HybridGreen compositor, but with a dynamic blend
-    factor `f` that depends on the pixel-level Normalized Differece Vegetation Index (NDVI). The higher the NDVI, the
+    factor `f` that depends on the pixel-level Normalized Difference Vegetation Index (NDVI). The higher the NDVI, the
     smaller the contribution from the nir channel will be, following a liner (default) or non-linear relationship
     between the two ranges `[ndvi_min, ndvi_max]` and `limits`.
 
@@ -163,7 +165,7 @@ class NDVIHybridGreen(SpectralBlender):
 
         ndvi = (projectables[2] - projectables[1]) / (projectables[2] + projectables[1])
 
-        ndvi = ndvi.clip(self.ndvi_min, self.ndvi_max)
+        ndvi = np.nan_to_num(ndvi.clip(self.ndvi_min, self.ndvi_max), nan=self.ndvi_min)
 
         # Introduce non-linearity to ndvi for non-linear scaling to NIR blend fraction
         if self.strength != 1.0:  # self._apply_strength() has no effect if strength = 1.0 -> no non-linear behaviour
@@ -181,7 +183,7 @@ class NDVIHybridGreen(SpectralBlender):
         """Introduce non-linearity by applying strength factor.
 
         The method introduces non-linearity to the ndvi for a non-linear scaling from ndvi to blend fraction in
-        `_compute_blend_fraction`. This can be used for a slower or faster transision to higher/lower fractions
+        `_compute_blend_fraction`. This can be used for a slower or faster transition to higher/lower fractions
         at the ndvi extremes. If strength equals 1.0, this operation has no effect on the ndvi.
         """
         ndvi = ndvi ** self.strength / (ndvi ** self.strength + (1 - ndvi) ** self.strength)
--- a/satpy/composites/spectral.py
+++ b/satpy/composites/spectral.py
@@ -165,7 +165,8 @@ class NDVIHybridGreen(SpectralBlender):
 
         ndvi = (projectables[2] - projectables[1]) / (projectables[2] + projectables[1])
 
-        ndvi = np.nan_to_num(ndvi.clip(self.ndvi_min, self.ndvi_max), nan=self.ndvi_min)
+        ndvi = ndvi.clip(self.ndvi_min, self.ndvi_max)
+        ndvi.data = np.nan_to_num(ndvi.data)
 
         # Introduce non-linearity to ndvi for non-linear scaling to NIR blend fraction
         if self.strength != 1.0:  # self._apply_strength() has no effect if strength = 1.0 -> no non-linear behaviour
--- a/satpy/tests/compositor_tests/test_spectral.py
+++ b/satpy/tests/compositor_tests/test_spectral.py
@@ -76,16 +76,17 @@ class TestNdviHybridGreenCompositor:
 
     def setup_method(self):
         """Initialize channels."""
-        coord_val = [1.0, 2.0]
+        y_coord_val = [1.0, 2.0, 3.0]
+        x_coord_val = [1.0, 2.0]
         self.c01 = xr.DataArray(
-            da.from_array(np.array([[0.25, 0.30], [0.20, 0.30]], dtype=np.float32), chunks=25),
-            dims=("y", "x"), coords=[coord_val, coord_val], attrs={"name": "C02"})
+            da.from_array(np.array([[0.25, 0.30], [0.20, 0.30], [0.0, 0.3]], dtype=np.float32), chunks=25),
+            dims=("y", "x"), coords=[y_coord_val, x_coord_val], attrs={"name": "C02"})
         self.c02 = xr.DataArray(
-            da.from_array(np.array([[0.25, 0.30], [0.25, 0.35]], dtype=np.float32), chunks=25),
-            dims=("y", "x"), coords=[coord_val, coord_val], attrs={"name": "C03"})
+            da.from_array(np.array([[0.25, 0.30], [0.25, 0.35], [0.0, 0.0]], dtype=np.float32), chunks=25),
+            dims=("y", "x"), coords=[y_coord_val, x_coord_val], attrs={"name": "C03"})
         self.c03 = xr.DataArray(
-            da.from_array(np.array([[0.35, 0.35], [0.28, 0.65]], dtype=np.float32), chunks=25),
-            dims=("y", "x"), coords=[coord_val, coord_val], attrs={"name": "C04"})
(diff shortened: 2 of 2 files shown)


pnuu (commented):
Apart from the excessive Dask computation, I'm fine with the change. As you said in Slack, there's the `DayNightCompositor` to make the night-side transparent.

gerritholl:
The same happens in `day_microphysics`.

djhoese:
> The same happens in day_microphysics.

I assume that's due to the `nir_reflectance` on the green channel?

pnuu:
Yep, the calculation of reflected portion generally breaks down before the SZA reaches 90 degrees, and those areas are masked out. The cut-off isn't actually even strictly by any SZA limit, also the signal level matters and the edge is often very ragged.

satpy/composites/spectral.py
strandgren, line 64: Even if the problem is in `NDVIHybridGreen` I wonder if it makes more sense to do `np.nan_to_num` here instead? For two reasons:

1. To me it's more intuitive to say that when applying a fraction to the channel data it cannot be NaN and will be converted to 0.0 if not. Doing it already at the NDVI stage is a bit far away from the final output and it may not be clear why it's done. It's also not easy to understand what the correct logic is if the user defines their own `ndvi_min` which may be larger than 0.0 (default value for `np.nan_to_num`).
2. Would make the compositor more future proof for similar developments also inheriting from `SpectralBlender`

satpy/composites/spectral.py
```diff
@@ -164,6 +166,7 @@ def __call__(self, projectables, optional_datasets=None, **attrs):
         ndvi = (projectables[2] - projectables[1]) / (projectables[2] + projectables[1])
 
         ndvi = ndvi.clip(self.ndvi_min, self.ndvi_max)
+        ndvi.data = np.nan_to_num(ndvi.data)
```
strandgren: You write in the description that you "do an additional 'clip' on the NDVI to convert NaNs to the minimum allowed NDVI as configured by the user", but for that you'd need to do `ndvi.data = np.nan_to_num(ndvi.data, nan=self.ndvi_min)`, right?

That being said I don't think that's a wanted behavior since we'd then get something neither black nor transparent I think

strandgren (commented):
Thanks for fixing, I think this change in behavior makes sense! See two comments in-line

satpy/composites/spectral.py
strandgren: I now realize that this may be more computationally expensive since we need to check more than one dataset, so maybe that would be a strong argument against this idea

djhoese: I thought about this as an option or the algernative of doing the NaN check on the `fractions` input. However, it is "valid" for some of the values to be NaN (space pixels, invalid data, etc). By doing it on the NDVI product I'm saying that I don't want the NDVI result to introduce *new* NaN values. There is an equivalent argument in my mind to handle this in the NDVIHybridGreen when NDVI is converted to a fraction but it seemed easier to do it the way I implemented it.

satpy/composites/spectral.py
djhoese: Ah yes, I should re-add that. I got some error from dask or numpy about passing `nan=self.ndvi_min` as a keyword argument so I removed it to get things working then forgot to put it back. For what I was going for I think it makes sense to have it be `ndvi_min` as we're saying "we want the smallest valid/legal value of NDVI to move on in this processing". If we think about it purely as correcting for the denominator of the NDVI being 0 then maybe we do want it to be `ndim_max` and not `ndvi_min` as `X / almost-zero` would be a large number.

djhoese: Thinking about this more, another way to put this is that we're controlling the *fraction* by controlling the NDVI. It isn't that we'd get something that isn't black and isn't transparent, we'd get something that is a blend of the two bands still and the common cases of NaNs in NDVI are NaNs in the data or a divide by 0 where both NDVI inputs are 0 at a pixel. So the end result will typically be 0 (black).

gerritholl:
Our operational true color imagery is black at night.

[MTG_EUROPA_ZENTRAL_true_color_nqcxeur750m_202604060440-geotiff.tif](https://github.com/user-attachments/files/26541149/MTG_EUROPA_ZENTRAL_true_color_nqcxeur750m_202604060440-geotiff.tif)

```
import hdf5plugin
from satpy import Scene
from glob import glob
fci_files = glob("/media/nas/x23352/MTG/FCI/L1c-cases/202309_10-cwg/10/01/08/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-*FDHSI*-FD--CHK-BODY---NC4E_C_EUMT_20231001*_IDPFI_VAL_20231001*_20231001*_N__C_0054_*.nc")
sc.load(["true_color"])
sc.save_datasets(fill_value=0)
```

In fact, my attempts to set the night part transparent using the `DayNightCompositor` have failed, see https://github.com/pytroll/satpy/issues/3003

gerritholl:
> > The same happens in day_microphysics.
> 
> I assume that's due to the `nir_reflectance` on the green channel?

Yes.  In fact, as a workaround for #3003, I've actually experimented with a `DayNightCompositor` setting the **night** part to `day_microphysics`, which works, but it's ugly and relying on that is no better than relying on [spacebar heating](https://xkcd.com/1172/)…  I'm planning to work on a solution for #3003.

djhoese:
@gerritholl what is your composite and enhancement configuration for your operational true color?

pnuu:
> Our operational true color imagery is black at night.

That's what I'd expect with
```
sc.save_datasets(fill_value=0)
```
Right?

djhoese:
Oh yes, I missed that in @gerritholl's code. That is expected as all the NaNs are converted to 0s (the fill value you specified). This PR and this problem is specifically about wanting the space pixels to be NaN (invalid) and the night pixels to be 0 or some other minimal value or more accurately I want them to be whatever the "real" true color would see (a real red channel, a real green channel, a real blue channel)...so black.

gerritholl:
> > Our operational true color imagery is black at night.
> 
> That's what I'd expect with
> 
> ```
> sc.save_datasets(fill_value=0)
> ```
> 
> Right?

No, that's not what I would expect.  Our intention is for the night pixels to become transparent, but satpy "helpfully" sets the night pixels to 1 so that they turn out black, defeating the purpose of them being 0 before (either due to the lack of light, or due to the DayNightCompositor setting them to 0).  This is a long-standing bug in our operational production that is on my list of things to fix in 2026.

1. User sets pixels to 0.
2. User passes `fill_value=0` to `save_datasets`, so that those 0 pixels become transparent.

Step 2 fails.  But that problem is covered by #3003.

gerritholl:
> @gerritholl what is your composite and enhancement configuration for your operational true color?

We use the built-in satpy composite and enhancement `true_color`.

pnuu:
I've always considered `fill_value=X` to mean "set all missing values to X". Missing data are (in most cases at least) set to `NaN` by the file handler / reader. So setting them to zero would give black. There might be differences how the writers handle the kwarg and whether they add the metadata to the final file or not, and in the end it's up to the image display software the show the fill values as transparency, like they would be with the alpha band.

djhoese:
@gerritholl These are two or more separate issues and you have #3003 for the specific DayNightCompositor issue of trying to get the night pixels to be a certain value. For the `save_datasets` with `fill_value=0`, at least for the geotiff and simple_image writer, the NaNs will always be made 0 and all of the valid data (including 0) will always be 1-255 for a uint8 geotiff. The entire valid data space is scaled to 1-255. There is no way to specify `fill_value=0` to the geotiff writer and have NaNs appear in the final out. That is/was the goal of the `fill_value` kwarg.

I mentioned this in the meeting yesterday, but I think this PR is correct and we should generally have psuedo-bands try to use a reasonable value where pixels aren't strictly invalid (0-ish in this case). Theoretically the DayNightCompositor can be used to convert night pixels to NaN but #3003 needs to be figured out. Alternatively there is no way for me to get black night pixels and NaN space pixels with the current way this NDVIHybridGreen compositor works or at least I think it is harder.

The question for this PR still remains from the previous comment thread of what should the "fraction" be based on for night pixels where NDVI is invalid? NDVI min or NDVI max or handle the NaNs in a different portion of the compositor?

satpy/composites/spectral.py
strandgren: I see, I think that makes sense then and that we can keep it like this :)

satpy/composites/spectral.py
strandgren: yes, that's true so we're probably talking about something that doesn't affect the final output. That being said, setting `nan=self.ndvi_min` would correspond to theoretical logic of the current behaviour I think 😄

djhoese: I can do that. Want thought experiment though, let's say the two NDVI are near zero but one negative and one positive. So p2 is 0.01 and p1 is -0.009. I know reflectances should be negative but they happen in the data. That would make `ndvi = (0.019 / 0.001)` which gives us `19.0` on a product that should be 0 to 1. If we flip these values then we get `ndvi = (-0.019 / 0.001)` I think so -19.0. So either ndvi min or max would make sense with this made up example.

I see your documentation says:

```
    In this example, pixels with NDVI=0.0 will be a weighted average with 15% contribution from the
    near-infrared vis_08 channel and the remaining 85% from the native green vis_05 channel, whereas
    pixels with NDVI=1.0 will be a weighted average with 5% contribution from the near-infrared
    vis_08 channel and the remaining 95% from the native green vis_05 channel. For other values of
    NDVI a linear interpolation between these values will be performed.
```

Is the 85%/15% case (NDVI min) the desired outcome or is the 95%/5% (NDVI max) what we want?

strandgren: I don't get what you mean. Are p1 and p2 the reflectance data used to compute ndvi, or something else? I don't see the connection between `So p2 is 0.01 and p1 is -0.009` and `ndvi = (0.019 / 0.001)`

> Is the 85%/15% case (NDVI min) the desired outcome or is the 95%/5% (NDVI max) what we want?

We want to map the lowest valid NDVI value to 85%/15% (i.e. `limits[0]`) and the highest valid NDVI to 95%/5% (`limits[1]`). So if the user says that the smallest NDVI should be 0.1 (`ndvi_min=0.1`), an NDVI of 0.1 should be mapped to 85%/15% (`limits[0]`) and an NDVI of 1.0 to 95%/5% (`limits[1]`) as before. Hence, I think it makes sense to set NaN values to `ndvi_min`. Whatever the final implementation is, I'd still be happy if you can test and verify that the night-side is still black🙂

djhoese: Yes, sorry. `p2` is `projectables[2]` and `p1` is `projectables[1]` from the compositor code. So plugging in the numbers I gave into the NDVI formula of `(p2 - p1) / (p2 + p1)` would give us the values I showed if I did my math correctly. Regardless...I'll set it to `ndvi_min` and make an image.

strandgren: Thanks for the clarification. The theoretical range of NDVI is [-1, 1], but indeed, if the reflectance values (p1 and p2) are very small we can get significantly higher/lower values. Below is an example including the distribution of the NDVI values (log-scale):

image

image

For the sake of the green band correction I decided to limit the NDVI to [0,1] here and it's mainly clear-sky water which has NDVI < 0.0. Furthermore, we now clip the data to the NDVI limits, so with the default configuration there will never be an NDVI outside the range [0,1]. If any user decides to modify this to allow NDVI values up to e.g. +-19 they will have to deal with the consequences😄

djhoese:
With the latest commit using `ndvi_min` as the replacement for NaNs the image looks like this with default Satpy enhancements:

image

djhoese:
And for @kathys here is with the P2G enhancement:

image

which is just a slightly brighter version and shows more of the cloud tops I think.

kathys:
@djhoese  Looks good to me.

strandgren (approved):
Thanks for the images, looks good to me!

djhoese:
Thanks. Any other comments before merging?

PR #3369, opened 2026-04-06
Add snow age RGB recipe for VII/METimage

adybbroe:
Add snow age RGB recipe for VII/METimage and use Oscar naming for channel specs in the reader yaml file

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

The METimage sensor does have compatible bands to VIIRS concerning the otherwise VIIRS unique `snow_age` RGB composite. Below an overview of the relevant spectral responses.

**VIIRS only:**
rsr_band_0060_0240_viirs_only

**Both METimage and VIIRS:**
rsr_band_0060_0240_viirs_metimage

### And here results of two nearly coincident scenes from both sensors over Scandinavia.

**First METimage around 08:40 UTC February 13, 2026:**

**NB! DISCLAIMER - METimage data not shown here are early commissioning data**

**And here VIIRS from around 08:49 from Suomi-NPP the same day:**

Suomi-NPP_20260213_0849_euron1_snow_age_small

### DIFF
--- a/satpy/etc/composites/metimage.yaml
+++ b/satpy/etc/composites/metimage.yaml
@@ -64,6 +64,21 @@ composites:
         modifiers: [ nir_reflectance ]
     standard_name: snow
 
+  snow_age:
+    compositor: !!python/name:satpy.composites.viirs.SnowAge
+    prerequisites:
+    - name: vii_865
+      modifiers: [sunz_corrected]
+    - name: vii_1240
+      modifiers: [sunz_corrected]
+    - name: vii_1375
+      modifiers: [sunz_corrected]
+    - name: vii_1630
+      modifiers: [sunz_corrected]
+    - name: vii_2250
+      modifiers: [sunz_corrected]
+    standard_name: snow_age
+
   convection:
     compositor: !!python/name:satpy.composites.core.GenericCompositor
     prerequisites:
--- a/satpy/etc/composites/vii.yaml
+++ b/satpy/etc/composites/vii.yaml
@@ -64,6 +64,21 @@ composites:
         modifiers: [ nir_reflectance ]
     standard_name: snow
 
+  snow_age:
+    compositor: !!python/name:satpy.composites.viirs.SnowAge
+    prerequisites:
+    - name: vii_865
+      modifiers: [sunz_corrected]
+    - name: vii_1240
+      modifiers: [sunz_corrected]
+    - name: vii_1375
+      modifiers: [sunz_corrected]
+    - name: vii_1630
+      modifiers: [sunz_corrected]
+    - name: vii_2250
+      modifiers: [sunz_corrected]
+    standard_name: snow_age
+
   convection:
     compositor: !!python/name:satpy.composites.core.GenericCompositor
     prerequisites:
--- a/satpy/etc/readers/metimage_l1b_nc.yaml
+++ b/satpy/etc/readers/metimage_l1b_nc.yaml
@@ -48,7 +48,7 @@ datasets:
 # --- Measurement data ---
   vii_4:
     name: vii_443
-    vii_number: 4
+    vii_channel_id: 4
     file_type: nc_vii_l1b_rad
     file_key: data/measurement_data/vii_443
     coordinates: [lat_pixels, lon_pixels]
@@ -60,10 +60,11 @@ datasets:
         standard_name: toa_outgoing_radiance_per_unit_wavelength
     chan_solar_index: 0
     wavelength: [0.428, 0.443, 0.458]
+    resolution: 500
 
   vii_8:
     name: vii_555
-    vii_number: 8
+    vii_channel_id: 8
     file_type: nc_vii_l1b_rad
     file_key: data/measurement_data/vii_555
     coordinates: [lat_pixels, lon_pixels]
@@ -75,10 +76,11 @@ datasets:
         standard_name: toa_outgoing_radiance_per_unit_wavelength
     chan_solar_index: 1
     wavelength: [0.545, 0.555, 0.565]
+    resolution: 500
 
   vii_12:
     name: vii_668
-    vii_number: 12
+    vii_channel_id: 12
     file_type: nc_vii_l1b_rad
     file_key: data/measurement_data/vii_668
     coordinates: [lat_pixels, lon_pixels]
@@ -90,10 +92,11 @@ datasets:
         standard_name: toa_outgoing_radiance_per_unit_wavelength
     chan_solar_index: 2
     wavelength: [0.658, 0.668, 0.678]
+    resolution: 500
 
   vii_15:
     name: vii_752
-    vii_number: 15
+    vii_channel_id: 15
     file_type: nc_vii_l1b_rad
     file_key: data/measurement_data/vii_752
     coordinates: [lat_pixels, lon_pixels]
@@ -105,10 +108,11 @@ datasets:
         standard_name: toa_outgoing_radiance_per_unit_wavelength
     chan_solar_index: 3
     wavelength: [0.7465, 0.7515, 0.7565]
+    resolution: 500
 
   vii_16:
     name: vii_763
-    vii_number: 16
+    vii_channel_id: 16
     file_type: nc_vii_l1b_rad
     file_key: data/measurement_data/vii_763
     coordinates: [lat_pixels, lon_pixels]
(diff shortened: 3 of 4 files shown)


ameraner:
The new composite looks good to me. 

However what is the advantage of using the numbered channel names in the yaml dataset keys, instead of the wavelength names as in the `name`? I see that the important bit is anyway the `name`, since it's what is used in composites and also in `scn.available_dataset_names()`. Maybe we could add the numbered channel name as an extra key in the dataset attrs and keep the key and the name consistent (using the wavelength)? Although they're used in OSCAR, I find the numbered names not very user friendly, and some numbers in between are even missing...

adybbroe:
> The new composite looks good to me.
> 
> However what is the advantage of using the numbered channel names in the yaml dataset keys, instead of the wavelength names as in the `name`? I see that the important bit is anyway the `name`, since it's what is used in composites and also in `scn.available_dataset_names()`. Maybe we could add the numbered channel name as an extra key in the dataset attrs and keep the key and the name consistent (using the wavelength)? Although they're used in OSCAR, I find the numbered names not very user friendly, and some numbers in between are even missing...

@ameraner  Thanks! The official channel naming according to Oscar is the original VII naming, 1 to 40. My understanding is that when the mission was defined (with the SAG?) it was 40 specific spectral bands. The result, as in in METimage is a subset, due to many things, but mainly budget (overruns) I guess? When looking at Oscar I found it helpful to have this mapping from VII name/number to the Satpy name. Anyhow this section name can be anything in the yaml file, the important bit is the `name`. Unless I am wrong? The `vii_4` below could be `jondoe-1` if we wanted right?:
```
# --- Measurement data ---
  vii_4:
    name: vii_443
    file_type: nc_vii_l1b_rad
    file_key: data/measurement_data/vii_443
    ...
```

I thought I was just adding (useful) info. As it was the section naming was just identical to the `name`s.

To the suggestion "Maybe we could add the numbered channel name as an extra key in the dataset attrs": Yes, maybe. But I am less a Satpy specialist to see the advantage and how to do that. Should if then be like:
```
  vii_4:
    name: vii_443
    vii_number: 4
    file_type: nc_vii_l1b_rad
    file_key: data/measurement_data/vii_443
    coordinates: [lat_pixels, lon_pixels]
    calibration:
      reflectance:
        units: "%"
      radiance:
    chan_solar_index: 0
    wavelength: [0.428, 0.443, 0.458]
```
?

ameraner:
The advantage of adding it to the keys is that it would show up as an extra attribute in the dataset attrs (if they are attached during the reading, I assume so). But I'm ok with the current solution too. You just need to update your branch since the filenames have changed on main in the meantime.

adybbroe:
> The advantage of adding it to the keys is that it would show up as an extra attribute in the dataset attrs (if they are attached during the reading, I assume so). But I'm ok with the current solution too. You just need to update your branch since the filenames have changed on main in the meantime.

Thanks. Not sure `vii_number` is an intuitive attribute name though:
```
scn['vii_13345'].attrs
{'end_time': datetime.datetime(2026, 6, 16, 10, 59, 58, 750000), 'filename_start_time': datetime.datetime(2026, 6, 16, 10, 51, 49), 'valid_max': np.uint16(1021), 'file_key': 'data/measurement_data/vii_13345', 'valid_min': np.uint16(0), 'modifiers': (), 'coordinates': ('lat_pixels', 'lon_pixels'), 'chan_thermal_index': 8, 'units': 'K', 'vii_number': 40, 'wavelength': WavelengthRange(min=13.19, central=13.345, max=13.5, unit='µm'), 'calibration': 'brightness_temperature', 'sensor': 'metimage', 'name': 'vii_13345', 'filename_end_time': datetime.datetime(2026, 6, 16, 10, 59, 58), 'spacecraft_name': 'Metop-SG-A1', 'platform_name': 'Metop-SG-A1', 'rows_per_scan': 24, 'start_time': datetime.datetime(2026, 6, 16, 10, 51, 49, 701000), 'file_type': 'nc_vii_l1b_rad', 'ssp_lon': None, 'long_name': '13345 nm spectral radiances', 'standard_name': 'toa_brightness_temperature', 'reader': 'metimage_l1b_nc', 'area': Shape: (6768, 3144)
Lons: <xarray.DataArray 'longitude' (y: 6768, x: 3144)> Size: 170MB
```

ameraner:
Maybe `vii_channel_id_number`? or `vii_channel_id`?

djhoese:
I'm not sure I follow this whole conversation but I'm not a huge fan of the `vii_number` dataset name that gets used as the `DataID(name=X)`. Is that's kind of what's being discussed? Why does `vii_` have to be in the name? With MODIS we had just "<channel number>". With SEVIRI and the like you VIS and IR prefixes which would be useful here too.

adybbroe:
So, @ameraner and @djhoese , I cleaned up the tangled commit history for this PR. It should now be up to date with the current main and only contain the few relevant commits for this PR (currently 3). And to try reply to your input and concerns (David):

1. It was my own "fault" to fiddle with and change the "channel section titles" in the yaml file. The reason was twofold (as far as I remember now) that I don't like the fact that the section titles are identical to the `name`. That may confuse a user on what is actually important. The *important* part is what is in the `name`. Further using the index/id number from Oscar instead gives (perhaps) useful information. Now, thanks to what Andrea suggested this information is also now in the attributes. So, with that new attribute `vii_channel_id` we could have section names like `channel-1`, `channel-2`, or even something else... If I am not totally wrong the section titles do not matter, it could be anything really. I tested: `scn.load([composite_name, "vii_13345", "vii_34"], resolution=500)`, and that gives an error: `KeyError: "Unknown datasets: {DataQuery(name='vii_34', resolution=500)}"`. Whereas, this works fine for instance: `scn.load([composite_name, "vii_13345", 7.3], resolution=500)`
2. I now also added the missing `resolution` since that was missing, and for consistency with other readers it should be there, right? Now that I have been anyhow fiddling with the reader yaml file...

So, please, give your opinion, two questions: 

1. Should I revert back to the old "section title" names (that were identical to the `name` values) for the sake of consistency with many other readers (perhaps all?)?
2. And should we skip the `vii_channel_id` attribute or keep as suggested?

ameraner:
thanks Adam! My vote would go to revert to the old dataset keys, that match the names, for simplicity and consistency. And I would keep the new vii_channel_id key, I'm sure it can be useful if people want to match it with Oscar or other documentation!

adybbroe:
@djhoese Any objections to the suggestion of Andrea? If not I will do as he suggested and merge :-)

djhoese:
I'm fine with his suggestions. I don't see a channel 21 by the way, is that expected? Otherwise, I think my more controversial opinion that the `vii_` should be removed is the only real opinion I have about this.

adybbroe:
@djhoese Any objections to the suggestion of Andrea? If not I will do as he suggested and merge :-)


> I'm fine with his suggestions. I don't see a channel 21 by the way, is that expected? Otherwise, I think my more controversial opinion that the `vii_` should be removed is the only real opinion I have about this.

Great thanks. I have now gone back to the original channel dataset keys. No, there is no VII-21. It was scrapped in the process. Originally the VII mission was defined to be a 40 channel IR/VIS mission. But "only" 20 spectral channels survived. I agree with you (provided I understand you right) that the whole VII thing is actually rather uninteresting and irrelevant now that METimage is flying. We should really only talk about the 20 channels that are there! So, I will support any future effort to clean the reader for any reference to vii!!! Let's take that in another issue/pr though.

djhoese:
I mean sure the channel numbers are a thing, but I don't understand why when I request a channel with `Scene.load` I need to say `vii_13345`. The `vii_` is redundant I guess? I know I'm reading metimage data which used to be called VII. So not only is the `vii` wrong (it should be `metimage_`), but it is repeated for all of the channels. Why not `scn.load(["1"])` for the channel number or `scn.load(["IR_13345"])` or something?

strandgren:
I think the main reason is to simply use the dataset names as they are in the NetCDF file. That way we simply pass on the information as stored in the product.

That being said, I fully agree that the channel names for the metimage reader are ugly and would also be happy to have them changed, even if it means that the satpy dataset names become different to those in the NetCDF file.

djhoese:
:facepalm: I didn't realize the *variable names* in the file also had `vii_` :disappointed: Yeah...ok...leave them as they are.

PR #3371, opened 2026-04-08
Add a unit test for the CO2 corrector

gerritholl:
Add a unit test for `satpy.modifiers.atmosphere.CO2Corrector`.

Also move other atmospheric modifier unit tests to their own test module.

 - [x] Tests added


gerritholl:
CodeScene is complaining about code that I moved but did not otherwise touch.  An attempt to suppress the finding still fails with "Error occurred while suppressing! Please try again."  I don't know if the suppression has ever worked (for me it never has).

djhoese:
Maybe I'm misunderstanding, but we don't want to suppress what CodeScene is finding since it is an actual problem. We just won't let it stop this PR from being merged as it isn't a new issue you are adding.

gerritholl:
> Maybe I'm misunderstanding, but we don't want to suppress what CodeScene is finding since it is an actual problem. We just won't let it stop this PR from being merged as it isn't a new issue you are adding.

Ok :)  My reasoning was that if there are any red checkmarks on a PR that human reviewers might not look at it.  If you're saying it won't stop this PR then that works or me.

djhoese (approved):

djhoese (approved):

djhoese:
I'm going to rebase this first just to make sure everything is still good since it's been a while.

pnuu (approved):
LGTM

PR #3377, opened 2026-04-13
Fix lookup table enhancement removing transparency (NaNs)

Grukank:
In colormap.py in Enhancements, preserves the NaN values lost during lookup and replaces them again where they were lost. For SMHI only relevant for snow_age.

 - [x] Tests added/updated 

### DIFF
--- a/satpy/enhancements/colormap.py
+++ b/satpy/enhancements/colormap.py
@@ -34,24 +34,23 @@ LOG = logging.getLogger(__name__)
 def lookup(img, **kwargs):
     """Assign values to channels based on a table."""
     luts = np.array(kwargs["luts"], dtype=np.float32) / 255.0
-    # Preserve NaNs
-    nans = np.isfinite(img.data)
-    _lookup_table(img.data, luts=luts)
-    # Replace lost NaNs
-    img.data = img.data.where(nans, np.nan)
-    return
+    return _lookup_table(img.data, luts=luts)
 
 
 @exclude_alpha
 @on_separate_bands
 @using_map_blocks
 def _lookup_table(band_data, luts=None, index=-1):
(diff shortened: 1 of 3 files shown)


satpy/enhancements/colormap.py
djhoese, line 53: Your changes are similar to the lines here that were added a while ago. This was done in #3315 by @avalentino to avoid warnings on NaN data and an error on one particular architecture. Your new solution should be fine to handle this as well if re-implemented here. My main question is, is this something we want the user to be able to control? Mainly, should the user be able to have NaN map to a value in the LUT? Technically that's what the `nan_to_num` is allowing but the proposed solution is ensuring that NaNs always remain NaNs.

Regardless, the proposed solution seems like something that could be handled inside this `_lookup_table` function and would therefore have better performance as it happens in parallel per-chunk of the dask array. With your code moved into this function or rather re-implemented with pure-numpy functions/masking/indexing then this `nan_to_num` line could probably be removed.

Grukank: I agree that it would be better to have it in ```_lookup_table``` instead, so I'll move it/re-implement it there. 

Regarding your question about the user mapping NaN to a (custom?) value in the LUT, after speaking about it with @mraspaud, it could be useful but I'm not so sure if it's needed, but in any case it's a bit outside of the scope of this pull request. It would also be a bit more complicated to implement and would change the functionality a bit too much (atleast for right here, right now).

djhoese: I've rewritten this comment at least 3 times now because I kept coming up with a different way of implementing this, but I think I've decided none of those ways were better than what you have here. So 👍

djhoese (approved):

PR #3378, opened 2026-04-22
Rewrite OMPS EDR reader (omps_edr) to support official NOAA NetCDF4 format

djhoese:
The old OMPS EDR reader was written for an old NASA HDF5-based L1b format that as far as I can tell no longer exists. I wrote that original reader at a NASA conference many years ago as a proof of concept on how easily readers could be added to Satpy. I've now gotten a request to be able to read the official NOAA OMPS EDR product so I've rewritten the reader to support these files and removed support for all of the old files.

CC @kathys

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented 


pnuu (approved):
LGTM.

Are the old HDF5 files completely extinct and there's no need to support them?

djhoese:
That was my understanding that they don't exist anymore, but I'll do some searching. I also noticed https://github.com/pytroll/satpy/pull/1126 by @TAlonglong so there's a chance those files still exist.

djhoese:
Some of the files are from NASA DRL which is in "archived" mode so those files are probably harder to find.

djhoese:
I can find similar files, but not the exact files on NASA EarthData, but searching by filename convention also seems difficult.

pnuu:
If someone requests the old version, we can always add the old version back, maybe with a "legacy" name. If it seems unlikely that these files pop up, better to merge so the new files are readable.

djhoese:
I think the difference is the classic NOAA SDR/EDR versus NASA L1b/L2. This reader was misnamed as an "EDR" but these files are actually NASA L2 products. One option would be to rename the old reader to "omps_l2" and keep this new version as "omps_edr" and try to document the backwards incompatibility.

mraspaud (approved):
LGTM, one small thing only:
1. A link to the official documentation for the format is missing.

satpy/etc/readers/omps_edr.yaml
```diff
@@ -1,246 +1,17 @@
...
   sensors: [omps]
 
 file_types:
-  # HDF5 files from NASA GES DISC HTTP
-  # https://search.earthdata.nasa.gov/search/granules
-  # https://snpp-omps.gesdisc.eosdis.nasa.gov/data//SNPP_OMPS_Level2/OMPS_NPP_NMSO2_L2.2/
-  # https://disc.sci.gsfc.nasa.gov/
-  omps_tc_so2_edr_ges_disc:
+  omps_total_ozone:
     file_reader: !!python/name:satpy.readers.omps_edr.EDRFileHandler
```
djhoese, line 14:
```suggestion
    # Product description: https://www.star.nesdis.noaa.gov/jpss/ozone.php
    file_reader: !!python/name:satpy.readers.omps_edr.EDRFileHandler
```

PR #3379, opened 2026-04-27
Update some VIIRS composite definitions to match WMO recommendations

djhoese:
This PR:

* Adds `day_cloud_type_distinction`
* Deprecates `cloud_phase_distinction` in favor of `day_cloud_type_distinction` variables
* Adds `day_cloud_type`
* Replaces `night_microphysics` recipe with `night_fog` recipe

I think this matches WMO suggestions from the RGB Workshop. @adybbroe I'd like your approval on the night microphysics change.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/composites/viirs.yaml
+++ b/satpy/etc/composites/viirs.yaml
@@ -394,9 +394,15 @@ composites:
   night_microphysics:
     compositor: !!python/name:satpy.composites.core.GenericCompositor
     prerequisites:
-    - hncc_dnb
-    - M12
-    - M15
+      - compositor: !!python/name:satpy.composites.arithmetic.DifferenceCompositor
+        prerequisites:
+          - M16
+          - M15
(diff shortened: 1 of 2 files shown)


pnuu (approved):
LGTM.

The only minor thing is the support only for `I` channel versions. For example the compact VIIRSdata we get from EUMETCast has only `M` channels (and partial DNB) available. I don't think we have a good way to switch between resolutions with dedicated modifiers, so the option is to keep these as-is or add "low res" versions.

djhoese:
I think I'd always like the highest (best) resolution version of the composite as the "default" and the lowres version as "_lowres" or something similar. I'm not sure what needs to be changed band-wise to make the low res versions. Like is M05 (0.67) really "equivalent" to I01 (0.64)?

mraspaud (approved):
LGTM

PR #3380, opened 2026-04-27
Cleanup session-scoped pytest fixtures to be more narrowly scoped

djhoese:
While working on another PR I noticed that there were some pytest fixtures that were session-scoped that were only used in their defining module or in the sub-package where they were defined. This PR re-scopes these fixtures to be more accurate for how they are used.

The end result should be no really difference as most of these are files on disk that pytest keeps those around between testing sessions anyway.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


pnuu (approved):
LGTM

mraspaud (approved):
Lgtm, thanks for taking the time!

PR #3382, opened 2026-04-30
Next try with WMO sensor names

sfinkens:
Here's a new try, where the `sensor` attribute holds a set of WMO names. The
attribute itself is not modified. I added normalization and serialization methods
to make use of this attribute in filenames.

We can't use it like this, because changing the type of the `sensor` attribute
would breaks users' workflows. The plan is to add a new `instruments`
attribute that can be used instead. But for now I just used `sensor`
to keep the number of changes small and see if that's the direction we
want to go.

As I was going through the code base I found it easier working with the set
most of the time. For YAML filenames (which only contain a single sensor)
I added a function to normalize the sensor name. Only in the `Writer`
class I needed a serialization function for multiple sensors.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/composites/aux_data.py
+++ b/satpy/composites/aux_data.py
@@ -156,7 +156,7 @@ class StaticImageCompositor(GenericCompositor, DataDownloadMixin):
             if self.area is None:
                 raise AttributeError("Area definition needs to be configured")
             img.attrs["area"] = self.area
-        img.attrs["sensor"] = None
+        img.attrs["sensor"] = set()
         img.attrs["mode"] = "".join(img.bands.data)
         img.attrs.pop("modifiers", None)
         img.attrs.pop("calibration", None)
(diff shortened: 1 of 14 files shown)


sfinkens:
Grrr, I amended to an older commit, will fix this tomorrow, gotta go now

djhoese:
You might need to remake this PR. It has the NDVI Hybrid green change that I made and my commit message, but includes your other changes. Very confusing. Not sure how this happened.

Otherwise, I think the biggest/easiest change I'd like on this is to rename all the helper functions and variables to be "instrument" or "inst" (for the shorter variable names). That's what I took from the slack conversations is that we should do instruments everywhere and get away from sensor.

sfinkens:
Fixed the amended commit.

> rename all the helper functions and variables to be "instrument"

Agreed! I didn't do that yet to make it easier to review. That's the next step once we're sure that this is the direction we want to go

sfinkens:
Superseded by https://github.com/pytroll/satpy/pull/3390

PR #3389, opened 2026-05-05
Cleanup warnings in documentation building

### LINKED ISSUE #3387: Geotiff Writer FAQ link points to xarray
**Describe the bug**
It seems that the issue solved once here https://github.com/pytroll/satpy/pull/1559 for the FAQ link here https://github.com/pytroll/satpy/blob/af9ced56295519d41c1986e9b96713991d69ca50/satpy/writers/geotiff.py#L73 is back.

**To Reproduce**
click on FAQ link here: https://satpy.readthedocs.io/en/stable/api/satpy.writers.geotiff.html#satpy.writers.geotiff.GeoTIFFWriter

**Expected behavior**
Point to Satpy's FAQ

**Actual results**
Points to Xarray's FAQ

**Screenshots**

Image

djhoese:
This might close #3387. Mostly I'm fixing other warnings I encountered. Generating the docs on Python 3.13 locally and I didn't run into the FAQ reference issue so maybe it just works better in newer python? :man_shrugging: Let's see how the RTD CI job builds and renders.

Other fixes:

* Sphinx really doesn't like the `Union` and `Path` types. From what I can tell this is a bug in the python documentation itself based on the issues I found.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Setup Conda Environment
-        uses: conda-incubator/setup-miniconda@v4
+        uses: conda-incubator/setup-miniconda@v3
         with:
           miniforge-version: latest
           python-version: ${{ matrix.python-version }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,7 +59,7 @@ jobs:
         id: cache
 
       - name: Update environment
-        run: mamba env update -n test-environment -f continuous_integration/environment.yaml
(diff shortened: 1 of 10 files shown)


djhoese:
@ameraner it seems good here. I'm not sure what happened in `main` to break it. Maybe a bug in a version of sphinx? Our biggest issue now is that I'm building with Python 3.13 which is not the oldest version of Python that Satpy supports so I guess technically something could succeed in the docs building that wouldn't in Python 3.12 (or 3.11 if we still support that).

pnuu (approved):
LGTM.

PR #3390, opened 2026-05-06
Switch to WMO Instruments

sfinkens:
## The Plan

This is the first of a series of PRs for switching to WMO instrument names (`AVHRR/1`) in user facing places, namely dataset attributes and `scene.sensor_names`. The internal representation (`avhrr-1`) remains the same.

Furthermore, we agreed to deprecate `sensor: str|set[str]` in favor of `instruments: set[str]`. For backwards compatibility, a config switch will be added to get the legacy `sensor` attribute back.

1. (This PR) Change Satpy internals to use sets for instruments and prepare for the transition to `instruments`.
2. Deprecate `sensor` in favor of `instruments`
  2.1 Enhancements: https://github.com/sfinkens/satpy/pull/2
  2.2 File Handlers: https://github.com/sfinkens/satpy/pull/3
3. Rename enhancement/composite YAML files so that they match the internal representation of the WMO sensor name.
  3.1 Rename files: https://github.com/sfinkens/satpy/pull/5
  3.2 Rename `sensor_name` property to `instruments`: https://github.com/sfinkens/satpy/pull/6
4. Rename `sensor` to `instrument` in all method names and docstrings.

I hope this way the changes are less painful to review. PRs 2-4 are relative to this PR, which will be merged into main eventually.

## Changes in this PR

- Add `satpy._instruments` module containing
  - A `StrEnum` with all WMO instrument names. Members behave like strings, even in sets
     ```python
     len({"ABI", OSCAR.ABI}) == 1
     ```
     Internally the enum values are used directly. Dataset attributes are still converted to plain strings because that's easier for users (no custom type, simpler string repr).
  - Helper methods for
    - Getting instruments from dataset attributes. Both `sensor` and `instruments` are supported, with a deprecation warning for `sensor`.
    - Setting instrument attributes
    - Converting instrument names WMO <-> internal
- For backwards compatibility, add `sensor` attribute in `scene.__getitem__` if it doesn't exist. To be removed in v1.0.

### Breaking Changes

None so far. For v1.0 we would have to remove compatibility methods and deprecation warnings wrapped in `8< v1.0`. Then Satpy wouldn't support the `sensor` attribute anymore.

### Notes

It would be nice to have a custom Attributes class with getters and setters for the instrument, but xarray converts it back to plain dicts internally.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


sfinkens:
Would it make sense to move those instrument helper methods to a dedicated module `_instrumenty.py`?

satpy/writers/mitiff.py
```diff
@@ -53,12 +54,9 @@ def _adjust_kwargs(dataset, kwargs):
...
     if "sensor" not in kwargs:
-        kwargs["sensor"] = dataset.attrs["sensor"]
-    # Sensor attrs could be set. MITIFFs needing to handle sensor can only have one sensor
-    # Assume the first value of set as the sensor.
-    if isinstance(kwargs["sensor"], set):
-        LOG.warning("Sensor is set, will use the first value: %s", kwargs["sensor"])
-        kwargs["sensor"] = (list(kwargs["sensor"]))[0]
+        #  MITIFFs needing to handle sensor can only have one sensor
+        # Assume the first value of set as the sensor.
+        kwargs["sensor"] = get_one_instrument_from_attrs(dataset.attrs)
```
sfinkens: I think ``sensor`` here is a keyword argument and not a dataset attribute, so I didn't change it yet

djhoese:
I added the "backwards-incompatibility" label on this PR. Let me know if that isn't accurate. You should rebase this PR on `main` which now includes some fixes for CI.

satpy/composites/aux_data.py
```diff
@@ -156,7 +157,7 @@ def __call__(self, *args, **kwargs):
             if self.area is None:
                 raise AttributeError("Area definition needs to be configured")
             img.attrs["area"] = self.area
-        img.attrs["sensor"] = None
+        set_instruments_attr(img.attrs, set())
```
djhoese: Are we sure this "sensor is None" case isn't needed by something in the `Scene`? Or does it seem to be handled automatically be dealing with sets (ex. iterating over nothing)?

satpy/composites/glm.py
```diff
@@ -97,7 +97,7 @@ def _update_attrs(self, new_data, background_layer, highlight_layer):
         new_data.attrs["units"] = 1
         new_sensors = self._get_sensors((highlight_layer, background_layer))
         new_data.attrs.update({
-            "sensor": new_sensors,
+            "instruments": new_sensors,
```
djhoese: Doesn't this need to use `set_instruments_attr`?

satpy/enhancements/enhancer.py
```diff
@@ -34,16 +40,17 @@ class EnhancementDecisionTree(DecisionTree):
 
     def __init__(self, *decision_dicts, **kwargs):
         """Init the decision tree."""
+        instr_key = get_instruments_key()
         match_keys = kwargs.pop("match_keys",
                                 ("name",
                                  "reader",
                                  "platform_name",
-                                 "sensor",
+                                 instr_key,
```
djhoese: Oh...this is difficult. I wonder if we should have a deprecation cycle or this? Or accept both and display a warning if "sensor" is detected/used?

satpy/writers/mitiff.py
djhoese: Good call.

satpy/utils.py
```diff
@@ -940,3 +942,59 @@ def flatten_dict(d, parent_key="", sep="_"):
...
+    if not instruments:
+        raise KeyError("No 'instruments' dataset attribute")
+    if len(instruments) > 1:
+        logger.warning(f"More than one instrument in dataset attributes, will use the first value: {instruments}")
+    return list(instruments)[0]
+
+
+def get_pyspectral_instrument_name(instrument: str) -> str:
+    """Get instrument name expected by pyspectral."""
+    return normalize_instrument_name(instrument)
```
djhoese: Why is this its own function? Are you concerned that there may be custom conversions needed for pyspectral in the future?

djhoese (commented):
Looks pretty good. I had a couple questions inline. I think my biggest long term fear after seeing these changes is handling the "sensor" in the enhancement decision tree.

satpy/composites/aux_data.py
sfinkens: Yes. In `scene.py` None type sensors were skipped implicitly

```python
for data_arr in self.values():
    if "sensor" not in data_arr.attrs:
        continue
    if isinstance(data_arr.attrs["sensor"], str):
        sensor_names.add(data_arr.attrs["sensor"])
    elif isinstance(data_arr.attrs["sensor"], set):
        sensor_names.update(data_arr.attrs["sensor"])
```
([source](https://github.com/pytroll/satpy/blob/4f1d7236b1de7b9c30a5711f4248b1cfaf87e2fe/satpy/scene.py#L202)).

Now empty sets are handled automatically

```python
for data_arr in self.values():
    sensor_names.update(get_instruments_from_attrs(data_arr.attrs))
```
 ([source](https://github.com/sfinkens/satpy/blob/32ee183d63e51f3f543994ad3e8fd2e962b9d8fd/satpy/scene.py#L204)). 

Apart from that, there was only one explicit None type check in [composites/fill.py](https://github.com/pytroll/satpy/blob/4f1d7236b1de7b9c30a5711f4248b1cfaf87e2fe/satpy/composites/fill.py#L409) that I changed [here](https://github.com/sfinkens/satpy/blob/32ee183d63e51f3f543994ad3e8fd2e962b9d8fd/satpy/composites/fill.py#L410).

satpy/composites/glm.py
sfinkens: Nice catch, thanks!

satpy/utils.py
sfinkens: I thought pyspectral will also require WMO names at some point. In that case we only need to change this method.

satpy/enhancements/enhancer.py
sfinkens: Ah, this is where the `sensor` key from enhancement yamls is actually used? And apparently users can provide their own enhancements, so I think accepting both + warning is a good idea 👍

sfinkens: That makes me think if we should do that everywhere. For example when checking attributes

```python
def get_instruments_from_attrs(attrs):
    # simplified a bit
    legacy = attrs.get("sensor", {})
    if legacy:
        warnings.warn(...)
    return attrs.get("instruments", legacy)
```

and in the [yaml reader](https://github.com/pytroll/satpy/blob/4f1d7236b1de7b9c30a5711f4248b1cfaf87e2fe/satpy/readers/core/yaml_reader.py#L151).

djhoese: Oh maybe. Like if the user is creating their own `DataArray` (either "by hand" or in custom compositor code)? That makes sense.

And yes, this list is what is considered "identifying" information for an enhancement section in an enhancement YAML. It is used, in order, to build up a decision tree to choose which enhancement to use for a particular `DataArray`.

satpy/utils.py
djhoese: Sounds good.

satpy/composites/aux_data.py
djhoese: Sounds good.

satpy/enhancements/enhancer.py
sfinkens: I've reverted this change in the enhancement decision tree, because I think it fits better into PR2 (together with the yaml reader update). But I've updated `get_instruments_from_attrs` as described above.

sfinkens:
I also moved the instrument helpers to their own module.

mraspaud (approved):
Wow, that's a big one! But very clean, I like it!

My only concern is instument vs sensor, not in the naming, but rather that it could be confusing, but I understand that it makes things more distinct. So nothing to change there.

sfinkens:
@martin You mean that there's both `sensor` and `instrument` in the code base? Eventually I would rename all `sensor` occurences to `instrument`.

Ok so I'm going to merge PRs https://github.com/sfinkens/satpy/pull/2 and https://github.com/sfinkens/satpy/pull/3 into this one and continue with renaming enhancement/composite files so that they match the WMO name.

djhoese:
How will you handle backwards compatibility with old filenames? Or search for both?

sfinkens:
My idea was to create a link from the old filename to the new filename

djhoese:
But what will users do before they have time to transition?

djhoese:
Transition their custom YAMLs I mean

sfinkens:
> Transition their custom YAMLs I mean

Hmm, good point. I didn't think of that. Then I'll have so search for both the old and the new name.

PR #3392, opened 2026-05-18
Add new filename pattern for MSG CRM product

strandgren:
This tiny PR simply adds another filename pattern (extension) for the MSG SEVIRI Clear-sky Reflectance Map (CRM) product in order to support files downloaded from the EUMETSAT archive, which use `.nat` as filename extension.

Example filename: `MSG3-SEVI-MSGCRMN-0100-0100-20260517120000.000000000Z-20260517121352-4891512.nat`


pnuu (approved):
LGTM

PR #3393, opened 2026-05-19
WMO Instruments Part 2.2: File Handlers

sfinkens:
Update all file handlers to provide WMO instrument names. To be merged into https://github.com/pytroll/satpy/pull/3390. This is the biggest one.

Summary:

- Add Enum with all WMO instrument names (`satpy._instruments.OSCAR`).
- Replace "sensor" attribute with WMO "instruments" in all file handlers.
- Deprecate "sensors" attribute in ``readers/*.yaml`` in favour of "instruments".
  - YAML reader: If "sensors" is found in the YAML, convert to "instruments"
    and print deprecation warning.
  - ``readers/*.yaml``: Renamed "sensors" to "instruments" in **some** files (case 2 below),
    saved the rest for another PR.
- Add more explicit instrument sets in some places
  - Ocean Color CCI: ``{"SeaWiFS", "MERIS", "MODIS", "VIIRS"}`` instead of "merged". Source: https://docs.pml.space/share/s/2RVhiuK2SWyhbSthqDDoxg
  - geocat: "IMAGER (GOES 12-15)" instead of "goes_imager". The source attribute is just "goes", so there's no way
    to distinguish between 8-11 and 12-15.

Why not change **all** reader YAMLs? Because this would add too much complexity to this PR. Instrument names in ``readers/*.yaml`` serves two purposes, again.

  1. Identify applicable composites/enhancements.
  2. Add them to dataset attributes in the file handler (only some actually do this)
 
 Translating them to WMO names is not always straightforward, so I only changed to WMO names in case 2 and saved
 case 1 for a future PR. In particular:
  - Specifying dependencies via ``instrument_name: visir/<sensor>`` doesn't work anymore,
    because WMO names might contain slashes and parentheses ("AVHRR/2"). Needs a new "dependencies" attribute.
  - Some enhancement/composite YAMLs are not compatible with the normalization scheme and need to be renamed (e.g. `mersi-1.yaml` -> `mersi1.yaml`)
  - In some cases there are multiple WMO names for basically the same instrument ("MWR (Sterna)", "MWR (AWS)" etc),
    but they all use the same composites/enhancements. So instead of copying the composites/enhancements we
    decided to stick to just "MWR". But that's not a valid WMO name.
  - Sometimes the instruments attribute is not even an instrument (e.g. ``scatterometer``, ``hsaf``). The sole purpose
    of that attribute is to link composites/enhancements.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


sfinkens:
Sorry, wrong target branch, correct one: https://github.com/sfinkens/satpy/pull/3

PR #3395, opened 2026-05-19
Change reader and component names from "VII" to "METimage" and fix minor issues

### LINKED ISSUE #3202: Rename VII readers and update VII readers for new test data

ameraner:
This PR 
- changes the name of the METimage instrument reader from `vii_l1b_nc` and `vii_l2_nc` to `metimage_l1b_nc` and `metimage_l2_nc`. The old reader names are put in deprecation (they still work but emit a FutureWarning). The python files and class names are adapted accordingly.
- The platform name is translated to the OSCAR names (Metop-SG-AX)
- With these fixes, the pyspectral RSR work correctly (from https://github.com/pytroll/pyspectral/pull/278 )

On top, two minor fixes are also implemented:
- the longitude arrays are wrapped to the -180, 180 range to avoid failures in resampling corner cases
- the yaml calibration configuration of one channel was faulty

... and the `rows_per_scan` attribute is added to datasets to support EWA resampling.

 - [x] Closes #3202 
 - [x] Tests added 
 - [x] Fully documented 

FYI @cesclc @TAlonglong @ludwigVonKoopa


djhoese:
@ameraner @mraspaud where are we with merging this pull request? Does it need to wait for something? Are there decisions that need to be made?

ameraner:
Hi Dave, from my side it's good to go/ready for review

satpy/readers/core/metimage_nc.py
```diff
@@ -121,6 +125,12 @@ def get_dataset(self, dataset_id, dataset_info):
         variable = self._standardize_dims(variable)
         return variable
 
+    @staticmethod
+    def wrap_longitude(longitude_array):
+        """Wrap longitude between -180 and 180 degrees."""
+        longitude_array = ((longitude_array + 180) % 360) - 180
```
djhoese, line 131: This doesn't have to be worked on in this PR, but I think in some work I did a long time ago on another reader or maybe in pyresample I found that it was actually faster and fewer dask tasks overall to do a `where` operation on the condition of `> 180` where `lon - 360` is what is used if True. ...or whatever the correct math is.

djhoese (approved):
I haven't reviewed every little detail but this looks pretty good to me. It'd be great for one more review @mraspaud @pnuu @gerritholl or someone else.

PR #3402, opened 2026-05-21
Fix cleanup in test_eps_l1b

sfinkens:
I noticed that over time my `/tmp` directory is filling up with stub files from `test_eps_l1b`. The reason was one missing `tearDown` method, but I took the opportunity to migrate that test to pytest and use a class scoped temp dir factory.


djhoese (approved):
Thanks!

PR #3407, opened 2026-05-29
Fix datatype of default time encoding

ninahakansson:
Fix datatype of default time encoding. It silences warning of this type:

```
python3.12/site-packages/satpy/writers/cf_writer.py:309: UserWarning: Times can't be serialized faithfully to int64 with requested units 'days since 1981-03-30T04:23:58.206000'. Resolution of 'milliseconds' needed. Serializing times to floating point instead. Set encoding['dtype'] to integer dtype to serialize to int64. Set encoding['dtype'] to floating point dtype to silence this warning.
    res = ds.to_netcdf(filename,
```

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


mraspaud (approved):
LGTM

PR #3413, opened 2026-06-04
Fix 'omps_edr' not handling an empty list of filters

djhoese:
I'm trying to use the filtering option in the OMPS EDR reader from the command line, but it gets annoying and really difficult that the reader doesn't handle an empty list. This PR fixes that. This also fixes a complaint from mypy on something in aux_download.py.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 


djhoese:
These changes are pretty simple and I need them to continue my work so I'm going to merge this. I can fix anything that needs fixing tomorrow.

PR #3416, opened 2026-06-18
Simplify MVIRI FIDUCEO tests

sfinkens:
This should make the tests compatible with pytest-9.1.0.

### DIFF
--- a/satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
+++ b/satpy/tests/reader_tests/test_mviri_l1b_fiduceo_nc.py
@@ -339,10 +339,10 @@ def fixture_fake_dataset(time_fake_dataset):
     return ds
 
 
-@pytest.fixture(name="projection_longitude", params=["57.0"])
-def fixture_projection_longitude(request):
+@pytest.fixture(name="projection_longitude")
+def fixture_projection_longitude():
     """Get projection longitude as string."""
-    return request.param
+    return "57.0"
 
 
 @pytest.fixture(name="fake_file")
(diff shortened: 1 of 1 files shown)


djhoese:
Is there documentation from pytest showing the old way you were using these fixtures? I can't seem to find anything that doesn't use `pytest_generate_tests`.

djhoese (approved):
I'm OK with this being merged, but I'd still like to learn what changed in pytest to make this not work or how this type of implementation is supposed to work.

djhoese:
This is apparently supposed to be fixed in 9.1.1 and hoped to be released this weekend:

https://github.com/pytest-dev/pytest/issues/14591

sfinkens:
> Is there documentation from pytest showing the old way you were using these fixtures? 

Did you mean this? https://docs.pytest.org/en/stable/example/parametrize.html#indirect-parametrization

Reading this... I find indirect parametrization more elegant than overriding the fixture in a class. Especially now that the original fixture is not parametrized anymore. What do you think?

```python
@pytest.fixture
def projection_longitude():
    return "57.0"
    
@pytest.mark.parametrize("projection_longitude", ["57.0", "5700"], indirect=True)
def test_something(file_handler, projection_longitude):
    ...
```

djhoese:
The newest pytest is out and tests seem to be passing without this PR. If you think your changes are better then feel free to merge. If you liked it the old way then I don't have a strong opinion and this could be closed. I think in general the indirect parameterization is hard to follow. Many people already don't like the magic of fixtures and doing things indirectly is even more magical.

sfinkens:
Ok, then I'm merging it as is

PR #3418, opened 2026-06-24
Update SLSTR L1b reader to support updated processing baseline

### LINKED ISSUE #3417: SLSTR L1B reader returns wrong radiance/reflectance values for products with new baseline collection 005
@simon-sat 

**Describe the bug**
The current SLSTR L1B reader (slstr_l1b) applies automatically channel adjustment factors to compensate well known calibration bias in SLSTR L1B products. Recently, the SLSTR L1B products are delivered under a new baseline collection 005, which includes a correction of this bias. It means that using the current satpy reader with this new baseline is actually doing the correction twice.

- Adjustment factors in current satpy reader: https://github.com/pytroll/satpy/blob/main/satpy/readers/slstr_l1b.py#L46
- Description of the baseline collection 005: https://user.eumetsat.int/news-events/news/new-processing-baseline-for-sentinel-3-slstr-l1

**To Reproduce**
Read any recent SLSTR L1B product (with baseline collection 005). The channels that are the most impacted are the ones with the highest adjustments factors (S5 and S6)

**Expected behavior**
The reader detects the baseline collection, and if higher or equal to 005, the internal calibration adjustment is disabled.

simon-sat:
As discussed in #3417, the new SLSTR processing baseline applies the radiance adjustment factors by default rather than expecting the user to do so themselves. Therefore, data processed with satpy is "double-corrected", as satpy automatically applies these coefficients also.

This PR updates the SLSTR reader to intelligently apply the adjustment factors when an old processing baseline is detected, but not to apply with the latest baseline.

 - [ ] Closes #3417 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/readers/slstr_l1b.py
+++ b/satpy/readers/slstr_l1b.py
@@ -198,11 +198,8 @@ class NCSLSTR1B(BaseFileHandler):
                 self.view != key["view"].name):
             return
         logger.debug("Reading %s.", key["name"])
-        if key["calibration"] == "brightness_temperature":
-            variable = self.nc["{}_BT_{}{}".format(self.channel, self.stripe, self.view[0])]
-        else:
-chan_type = "BT" if key["calibration"] == "brightness_temperature" else "radiance"
-variable = self.nc[f"{self.channel}_{chan_type}_{self.stripe}{self.view[0]}"]
+        chan_type = "BT" if key["calibration"] == "brightness_temperature" else "radiance"
+        variable = self.nc[f"{self.channel}_{chan_type}_{self.stripe}{self.view[0]}"]
         # Processing baseline version 005 and above already include the radiance adjustment factor
         # Therefore, unless user supplies their own, do not apply here.
         if self.baseline < 5 or self.usercalib is not None:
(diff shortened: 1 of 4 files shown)


simon-sat:
This PR should fix the issue raised in #3417. Codescene is moaning about stuff but I think it was also unhappy with the original code.
Unfortunately I will not have time to address issues in the old code, so if it's essential to correct what codescene complains about then I'd value if someone else could take the lead - I won't have time now until late-July and I'd say this is an urgent fix.

satpy/readers/slstr_l1b.py
```diff
@@ -201,7 +202,12 @@ def get_dataset(self, key, info):
             variable = self.nc["{}_BT_{}{}".format(self.channel, self.stripe, self.view[0])]
         else:
             variable = self.nc["{}_radiance_{}{}".format(self.channel, self.stripe, self.view[0])]
```
djhoese: Small cleanup. Let's see is codescene approves.

```suggestion
chan_type = "BT" if key["calibration"] == "brightness_temperature" else "radiance"
variable = self.nc[f"{self.channel}_{chan_type}_{self.stripe}{self.view[0]}"]
```

djhoese:
Ok I made codescene happy and tests pass. I think the logic is equivalent. @simon-sat let me know if you see something but I'll merge this given that it fixes someone's actual issue.

PR #3426, opened 2026-07-11
Add pyresample, trollimage, trollsift, pykdtree to show_versions/chec…

### LINKED ISSUE #3116: Add additional dependencies to `show_versions` and `check_satpy`
@lobsiger pointed out on the mailing list that the output of `check_satpy` (and therefore `show_versions`) don't include obvious dependencies like other pytroll packages.

https://github.com/pytroll/satpy/blob/1b226ab5b56589cff303abc6024dc49232ac7aac/satpy/utils.py#L507-L522

I'd like the following to be added:

1. pyresample
2. trollimage
3. trollsift
4. pykdtree

Can anyone think of any other packages that would be good to include? Pycoast maybe? Pydecorate? Python-geotiepoints (which I think is imported as "geotiepoints")?

Yuxiang-Ren-HUB:
Adds `pyresample`, `trollimage`, `trollsift`, and `pykdtree` to the package
list checked by `show_versions()` (and therefore `check_satpy()`, which
calls it internally).

- [x] Closes #3116
- [ ] Tests added
- [ ] Fully documented

### DIFF
--- a/satpy/utils.py
+++ b/satpy/utils.py
@@ -526,6 +526,10 @@ def show_versions(packages=None):
             "pyhdf",
             "h5netcdf",
             "fsspec",
+            "pyresample",
+            "trollimage",
+            "trollsift",
+            "pykdtree",
         )
         if packages is None
         else packages
--- a/satpy/utils.py
+++ b/satpy/utils.py
@@ -530,6 +530,9 @@ def show_versions(packages=None):
             "trollimage",
             "trollsift",
             "pykdtree",
+            "pycoast",
+            "pydecorate",
+            "python-geotiepoints",
         )
         if packages is None
         else packages


djhoese (changes requested):
I...I...why weren't these included before? I'm so confused. Thanks! As I mentioned in the related issue, what do you think about adding `pycoast` and `pydecorate`? I think `python-geotiepoints` (package name, imported as "geotiepoints") would be good too.

djhoese (approved):
Thanks!

Yuxiang-Ren-HUB:
Thanks for reviewing and merging! Glad the version diagnostics addition was useful — happy to help again if there's anything else in this area.

PR #3428, opened 2026-07-16
Preserve FCI time units from source files

### LINKED ISSUE #3349: FCI L1C reader overwrites units attribute on time metadata, making the data hard to interpret
**Describe the bug**

When loading time data with the FCI L1C NC reader, such as `vis_06_time`, the time attribute says `s`.  In the original NetCDF files, it says `seconds since 2000-01-01 00:00:00.0`.  Without looking at the original file, the user cannot interpret the float data in `vis_06_time`.

**To Reproduce**

```python
import netCDF4
import hdf5plugin
from satpy import Scene
sc = Scene(filenames={"fci_l1c_nc": ["/media/nas/x23352/MTG/FCI/L1c-cases/202601051630/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY--DIS-NC4E_C_EUMT_20260105164035_IDPFI_OPE_20260105163819_20260105163855_N_JLS_O_0100_0036.nc"]})
sc.load(["vis_06_time"])
print(sc["vis_06_time"].attrs["units"])

nc = netCDF4.Dataset("/media/nas/x23352/MTG/FCI/L1c-cases/202601051630/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY--DIS-NC4E_C_EUMT_20260105164035_IDPFI_OPE_20260105163819_20260105163855_N_JLS_O_0100_0036.nc")
print(nc["time"].units)
```

**Expected behavior**

I expect the units are the same:

```
seconds since 2000-01-01 00:00:00.0
seconds since 2000-01-01 00:00:00.0
```

**Actual results**

Satpy turns it into `s`:

```
s
seconds since 2000-01-01 00:00:00.0
```

**Environment Info:**

 - Satpy Version: v0.60.0
 
**Additional context**

I guess this happens because `s` is explicitly set as a unit in the YAML config:

https://github.com/pytroll/satpy/blob/072b27051f9ceddeaac0159ec053be382b514a4e/satpy/etc/readers/fci_l1c_nc.yaml#L797

I have not tested what happens if this field is removed or changed.

sapunyangkut:
## Summary

- remove static `s` overrides from FCI time-dataset configuration
- preserve complete CF reference-time units supplied by the source NetCDF file
- cover every generated FCI time dataset while confirming non-time units remain unchanged

- [x] Closes #3349
- [x] Tests added
- [x] Author entry is already included in open Draft #3427; no duplicate `AUTHORS.md` change is made here

## Validation

- preserved task-owned evidence for this exact upstream base and patch: focused regression 1 passed; full FCI reader module 259 passed; adjacent attribute-cleanup coverage 2 passed; ruff and `git diff --check` passed
- current worktree: `git diff --check`, Python byte-compilation of the changed test, and non-constructing YAML syntax parsing passed

The current checkout is not installed as a package, so a fresh pytest collection stops at the repository's generated `satpy.version` import; `ruff` is also unavailable in this environment. No dependency was installed.

## AI assistance disclosure

This patch, tests, commit message, and pull request text were developed and checked with OpenAI Codex assistance. No independent human self-review is claimed; maintainer review is requested before merge.

### DIFF
--- a/satpy/tests/reader_tests/test_fci_l1c_nc.py
+++ b/satpy/tests/reader_tests/test_fci_l1c_nc.py
@@ -832,6 +832,28 @@ class ModuleTestFCIL1cNcReader:
 class TestFCIL1cNCReader(ModuleTestFCIL1cNcReader):
     """Test FCI L1c NetCDF reader with nominal data."""
 
+    def test_time_units_are_preserved_from_file(self, reader_configs, FakeFCIFileHandlerFDHSI_fixture):
+        """Test that full CF time units from the source file are preserved."""
+        fh_param = FakeFCIFileHandlerFDHSI_fixture
+        reader = _get_reader_with_filehandlers(fh_param["filenames"], reader_configs)
+        file_handler = reader.file_handlers[fh_param["filetype"]][0]
+        expected_units = "seconds since 2000-01-01 00:00:00.0"
+        source_time = xr.DataArray([0.0], attrs={"units": expected_units})
+
+        time_datasets = [(data_id, info) for data_id, info in reader.all_ids.items()
+                         if data_id["name"].endswith("_time")]
+        assert len(time_datasets) == 36
+        assert {data_id["name"] for data_id, _ in time_datasets} == {
+            f"{channel}_time" for channel in LIST_TOTAL_CHANNEL}
+        for data_id, info in time_datasets:
+            loaded_time = source_time.copy()
+            loaded_time.attrs = file_handler._set_and_cleanup_attributes(
+                loaded_time.attrs, info=info)
+            assert loaded_time.attrs["units"] == expected_units
+
+        reflectance_id = reader.get_dataset_key("vis_06", calibration="reflectance", resolution=1000)
+        assert reader.all_ids[reflectance_id]["units"] == "%"
+
     @pytest.mark.parametrize("filenames", [TEST_FILENAMES[filename] for filename in TEST_FILENAMES.keys()])
     def test_file_pattern(self, reader_configs, filenames):
         """Test file pattern matching."""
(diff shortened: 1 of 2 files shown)


ameraner:
Hi @sapunyangkut , thanks for this! Is there a reason for the PR to be still in draft? It looks quite complete to me. We usually mark the PR as Ready for Review when the work is done and the PR is potentially ready to be merged.

satpy/tests/reader_tests/test_fci_l1c_nc.py
```diff
@@ -832,6 +832,28 @@ def _compare_rc_period_min_count_in_repeat_cycle(filetype, fh_param,
...
+        time_datasets = [(data_id, info) for data_id, info in reader.all_ids.items()
+                         if data_id["name"].endswith("_time")]
+        assert len(time_datasets) == 36
+        assert {data_id["name"] for data_id, _ in time_datasets} == {
+            f"{channel}_time" for channel in LIST_TOTAL_CHANNEL}
+        for data_id, info in time_datasets:
+            loaded_time = source_time.copy()
+            loaded_time.attrs = file_handler._set_and_cleanup_attributes(
+                loaded_time.attrs, info=info)
+            assert loaded_time.attrs["units"] == expected_units
```
djhoese, line 852: Is it possible to call the higher-level file handler functions to get these datasets out and compare the `.attrs["units"]` and the actual values? This way if the code is every refactored this test would still pass (not calling a private method)?

Additionally, is the reflectance units check expected below? Or copy/paste from before?

PR #3433, opened 2026-07-31
Add bz2 support to the metimage_l1b_nc reader

### LINKED ISSUE #3432: Support files from Metimage global data dumps
## Feature Request

Metimage global data dump granules (EO:EUM:DAT:0464:COM) are compressed with bz2, and have the filepattern

`W_XX-EUMETSAT-Darmstadt,SAT,SGA1-VII-1B-RAD_C_EUMT_20260727105633_G_O_20260727100659_20260727100800_C_N_T__.nc.bz2`

Those compressed files are not currently supported by the Metimage reader.

Decompressing the file (and removing the `.bz2` suffix) works to make the file readable.

Add support for bz2-compressed files, so metimage global data dump files can be read directly without decompressing them first.

No backward compatibility issues.

**Additional context**

Several other readers can already read compressed data directly, such as SEVIRI L1B HRIT, AHI HRIT, AHI HSD, and the GRIB reader.  Therefore, this should be quite easy to implement.

FabMin94:
Adds bz2-compressed (.nc.bz2) file support to the VII/metimage readers, so compressed granules from the EUMETSAT data hub can be read without manual decompression first.

`ViiNCBaseFileHandler.__init__` now decompresses via the existing `unzip_file()` helper (satpy.readers.core.utils) before opening the file, same approach used by `ahi_l1b_gridded_bin.py`. This lives in the shared base class, so it also applies to `ViiL2NCFileHandler` — L2 wasn't specifically tested here, but the change is in a code path both share identically.

Also updates the `vii_l1b_nc.yaml` file pattern to match `.nc.bz2` filenames.

Trade-off: decompression happens eagerly per-granule (~113MB temp disk per
file); see #3432 for more detail on why this was chosen over a streaming
approach.

 - [x] Closes #3432  
 - [x] Tests added 
 - [x] Fully documented -- no user-facing docs change. 


gerritholl:
Thanks for the PR!  It looks like the tests are failing on Windows ☹.  Unfortunately, I know very little about Windows coding myself.  Maybe it would work with a pytest fixture for temporary files?

satpy/readers/core/vii_nc.py
```diff
@@ -121,6 +126,15 @@ def get_dataset(self, dataset_id, dataset_info):
         variable = self._standardize_dims(variable)
         return variable
 
+    def __del__(self):
+        """Remove the decompressed temp file, if one was created."""
+        try:
+            if getattr(self, "_unzipped", None):
+                os.remove(self._unzipped)
+        except (AttributeError, OSError):
+            logger.warning(f"An error occurred while cleaning up the decompressed file {self.filename}")
```
djhoese: I could be wrong, but I think it is very difficult for logger's to accomplish any logging during the `__del__` step if Python itself is shutting down. Maybe it is fine to leave this here, but I could see it being an issue in the future.

djhoese (commented):
Unfortunately I think @gerritholl is right and it may be easiest to use pytest fixtures and use `tmp_path` to manage the creation and deletion of these files. Extra unfortunately is that this test class is a unittest TestCase so it can't use pytest fixtures directly. It would have to be converted. That would movstly involve changing the setUp and tearDown into fixtures so they can be used by the new fixture(s) needed for your tests.

FabMin94:
Thanks for the pointer — makes sense. I converted the bz2-related tests (test_bz2_reading, test_corrupt_bz2_raises) to plain pytest functions with tmp_path fixtures, so there's no more manual filename/cleanup handling for those. Left the rest of the file (setUp/tearDown, existing TestCase methods) as-is for now — wasn't sure if a full conversion was expected here or as a separate change, happy to do it if you'd like.

Also took your point on the `__del__` logging — removed the logger.warning() call, it just swallows the cleanup failure silently now. Added a small test to check `__del__` doesn't raise even when removing the temp file fails.

On the Windows failures:
- test_bz2_reading: `__del__` was removing the decompressed temp file before the base class's file handle was released — fine on Linux/macOS, not on Windows. Reordered so the handle closes first.
- test_corrupt_bz2_raises: turned out unrelated to my changes. On Windows this raises WinError 32 ("file in use by another
  process") directly from unzip_file() in satpy/readers/core/utils.py, before any VII code runs — confirmed via the log line
  pointing at the same temp file right before the error. Looks like unzip_file() opens/reopens the same temp path in a way
  that's fine on POSIX but not Windows. I've loosened my test to check exception type only (the message is locale- and
  platform-dependent), but this seems like a separate, pre-existing bug worth its own issue — happy to file one if useful.

satpy/tests/reader_tests/test_vii_base_nc.py
```diff
@@ -451,3 +452,136 @@ def test_dataset(self, po_, pi_, pc_):
...
+    pgi_.return_value = (interp_longitude, interp_latitude)
+
+    filetype_info = {
+        "cached_longitude": "data/measurement_data/longitude",
+        "cached_latitude": "data/measurement_data/latitude",
+    }
+    reader = ViiNCBaseFileHandler(str(vii_base_nc_file_bz2), vii_filename_info, filetype_info)
+
+    with mock.patch("os.remove", side_effect=OSError("mock disk error")):
+        reader.__del__()  # must not raise
```
djhoese: I'm tempted to say this test doesn't really need to exist, but maybe other @pytroll/satpy-core can give their opinion?

satpy/tests/reader_tests/test_vii_base_nc.py
```diff
@@ -451,3 +452,136 @@ def test_dataset(self, po_, pi_, pc_):
...
+        var = g2.createVariable("duration_of_data_present", np.double, dimensions=())
+        var[:] = 2.0
+        var = g2.createVariable("duration_of_data_missing", np.double, dimensions=())
+        var[:] = 3.0
+        var = g2.createVariable("duration_of_data_degraded", np.double, dimensions=())
+        var[:] = 4.0
+        var = g2.createVariable("gap_start_time_utc", np.double, dimensions=("gap_items",))
+        var[:] = [5.0, 6.0]
+        var = g2.createVariable("gap_end_time_utc", np.double, dimensions=("gap_items",))
+        var[:] = [7.0, 8.0]
```
djhoese, line 543: I assume this is a duplicate of what the setup does in the unittest test class? Could this logic be moved to a helper function and called from within this fixture *and* be used by the setup method of the unittest class?

djhoese (changes requested):

ameraner:
Hi @FabMin94 , thank you for your work on this. Just note that we have renamed the reader and some class names therein, so you might need to deal with some conflicts now, sorry for that.. See https://github.com/pytroll/satpy/pull/3395

FabMin94:
Thank you all for the input.
- I resolved the conflicts from #3395 by rebasing my commits on top of it.
- In test_metimage_base_nc.py I factored the file creation into a _create_metimage_base_nc_file() helper,
  used by both setUp and the metimage_base_nc_file fixture.

satpy/readers/core/metimage_nc.py
```diff
@@ -208,6 +213,15 @@ def get_dataset(self, dataset_id, dataset_info):
         variable = self._standardize_dims(variable)
         return variable
 
+    def __del__(self):
+        """Remove the decompressed temp file, if one was created."""
+        super().__del__()   # release the netCDF/h5netcdf handle first, so Windows can drop its lock
+        try:
+            if getattr(self, "_unzipped", None):
```
djhoese: Doesn't this object *always* have a `self._unzipped` attribute or does that disappear in some cases when python is cleaning things up?

djhoese: Oh and I think you could use:

https://docs.python.org/3/library/contextlib.html#contextlib.suppress

satpy/tests/reader_tests/test_metimage_base_nc.py
```diff
@@ -441,3 +377,174 @@ def test_dataset(self, po_, pi_, pc_):
...
+        """Test that __del__ never raises, even if removing the decompressed temp file fails."""
+        pgi_.return_value = _interpolated_lonlat()
+
+        filetype_info = {
+            "cached_longitude": "data/measurement_data/longitude",
+            "cached_latitude": "data/measurement_data/latitude",
+        }
+        reader = METimageNCBaseFileHandler(str(metimage_base_nc_file_bz2), metimage_filename_info, filetype_info)
+
+        with mock.patch("os.remove", side_effect=OSError("mock disk error")):
```
djhoese: Is mocking really necessary here? Couldn't you just delete the file beforehand and then call `__del__`? That should create a `FileNotFoundError` which is a subclass of `OSError`.

satpy/tests/reader_tests/test_metimage_base_nc.py
```diff
@@ -441,3 +377,174 @@ def test_dataset(self, po_, pi_, pc_):
...
+    return bz2_filename
+
+
+class TestBz2Support:
+    """Test bz2 decompression support for METimageNCBaseFileHandler."""
+
+    @mock.patch("satpy.readers.core.metimage_nc.METimageNCBaseFileHandler._perform_geo_interpolation")
+    def test_bz2_reading(self, pgi_, metimage_base_nc_file, metimage_base_nc_file_bz2, metimage_filename_info):
+        """Test that a bz2-compressed file is transparently decompressed and read correctly."""
+        pgi_.return_value = _interpolated_lonlat()
```
djhoese: Why does the interpolation need to be mocked?

Manny7717 (approved):
Verified locally on head 95c65e9b (worktrees against origin/main f46113284; version.py stub + shared satpy venv).

**Fix is correct and matches existing conventions.** `unzip_file()` (already used by ahi_l1b_gridded_bin etc.) returns a mkstemp path for `.bz2` inputs, None otherwise; wiring it in `METimageNCBaseFileHandler.__init__` before `super().__init__()` gives every VII/metimage variant transparent bz2 support. The yaml gains `.nc.bz2` patterns alongside the existing `.nc` ones (plain patterns kept first — matching priority preserved).

**`__del__` contract is sound:** `super().__del__()` runs first to close the netCDF handle (correct for Windows lock release), then removes the temp file, swallowing AttributeError/OSError so cleanup never raises during GC. Verified the partial-construction path (corrupt bz2 → `unzip_file` raises before `super().__init__`): OSError propagates cleanly to the caller and GC produces zero "Exception ignored" noise (`file_handle` defaults to None on the base class, so the inherited `__del__` is safe).

**Regression-proven:** PR test file applied to base worktree → `test_bz2_reading` and `test_del_swallows_cleanup_errors` FAIL on base (`OSError: NetCDF: HDF error: ...nc.bz2` — the reader tried to open compressed bytes directly), PASS on head; `test_corrupt_bz2_raises` passes on both (unzip_file raises regardless — correct contract pin).

**No regressions:** full test_metimage_base_nc.py 9/9 pass on head. Test refactor (setUp helper extraction) is behavior-preserving — all pre-existing tests pass on both base and head. All CI green (pre-commit.ci, CodeFactor, CodeScene, coveralls, RTD, full 3.10-3.14/OS matrix incl. macos).

Non-blocking nits: (1) `__init__` sets `self._unzipped` before `super().__init__()` — if the latter raised for a non-bz2 reason, the temp file would leak until GC; harmless in practice since the temp is only created for bz2 inputs. (2) DeprecationWarning on `lat[:] = 10.` (numpy 2.5 shape-set) is pre-existing in the extracted helper, not introduced.

satpy/tests/reader_tests/test_metimage_base_nc.py
FabMin94: It's mocked because the real _perform_geo_interpolation expects tie-point-grid-shaped input matching actual instrument geometry, and this test's fixture data isn't shaped for that — same reason setUp mocks it. The test is not actually checking interpolation correctness (that's covered separately in test_functions, unmocked); it just needs `__init__` to succeed so there's a constructed reader to test bz2 decompression / `__del__` (in test_del_swallows_cleanup_errors()) against.

satpy/tests/reader_tests/test_metimage_base_nc.py
FabMin94: Good point — swapped the mock for actually deleting the temp file first (reader.filename), then calling `__del__()`.

satpy/readers/core/metimage_nc.py
FabMin94: The `self.__unzipped__` attribute is not always set. If `unzip_file()` itself raises (e.g. through `_unzip_with_pbzip` or `_unzip_with_bz2`, in the case of a corrupt bz2 input), `__init__` never reaches the assignment (`self._unzipped = unzip_file(filename)`), so `_unzipped` is never created on that instance — but `__del__` still runs on partially-constructed objects, hence the guard.

Switched to contextlib.suppress, thanks.

djhoese: Understood on the partially constructed objects. Thanks for explaining.

djhoese:
Regarding @Manny7717's comment on the numpy warning, Claude confirmed that is in the netcdf4-python library and has been fixed [upstream](https://github.com/Unidata/netcdf4-python/pull/1469). Side note: @Manny7717 I sent you an email.

I'll make more comments about a few last changes...

satpy/readers/core/metimage_nc.py
djhoese: What do you think about making `_unzipped` a class attribute. Then we don't need to suppress the AttributeError here. This is similar to what is done already in the netcdf.py helper code:

https://github.com/pytroll/satpy/blob/d32cb3088d9399c2ed8ee2237325bb95f1404d3b/satpy/readers/core/netcdf.py#L76

satpy/readers/core/metimage_nc.py
```diff
@@ -47,6 +50,9 @@ class METimageNCBaseFileHandler(NetCDF4FileHandler):
 
     def __init__(self, filename, filename_info, filetype_info, orthorect=False):
         """Prepare the class for dataset reading."""
+        self._unzipped = unzip_file(filename)
+        if self._unzipped:
+            filename = self._unzipped
```
djhoese, line 62: Because this gets passed to the base class it becomes `self.filename` and then gets added to the metadata:

https://github.com/FabMin94/satpy/blob/2dfe8083f9ab7759f3008b04b992b4dda8b4a55c/satpy/readers/core/metimage_nc.py#L299

Do we want the original bz2 filename or the temporary uncompressed one to end up in the metadata? @ameraner?

satpy/etc/readers/metimage_l1b_nc.yaml
```diff
@@ -13,7 +13,9 @@ file_types:
...
-      'W_XX-EUMETSAT-Darmstadt,SAT,{spacecraft_name:s}-VII-1B-RAD_C_EUMT_{creation_time:%Y%m%d%H%M%S}_{mission_type:s}_{environment:s}_{sensing_start_time:%Y%m%d%H%M%S}_{sensing_end_time:%Y%m%d%H%M%S}_{disposition_mode:s}_{processing_mode:s}____.nc']
+      'W_XX-EUMETSAT-Darmstadt,SAT,{spacecraft_name:s}-VII-1B-RAD_C_EUMT_{creation_time:%Y%m%d%H%M%S}_{mission_type:s}_{environment:s}_{sensing_start_time:%Y%m%d%H%M%S}_{sensing_end_time:%Y%m%d%H%M%S}_{disposition_mode:s}_{processing_mode:s}_T__.nc.bz2',
+      'W_XX-EUMETSAT-Darmstadt,SAT,{spacecraft_name:s}-VII-1B-RAD_C_EUMT_{creation_time:%Y%m%d%H%M%S}_{mission_type:s}_{environment:s}_{sensing_start_time:%Y%m%d%H%M%S}_{sensing_end_time:%Y%m%d%H%M%S}_{disposition_mode:s}_{processing_mode:s}____.nc',
+      'W_XX-EUMETSAT-Darmstadt,SAT,{spacecraft_name:s}-VII-1B-RAD_C_EUMT_{creation_time:%Y%m%d%H%M%S}_{mission_type:s}_{environment:s}_{sensing_start_time:%Y%m%d%H%M%S}_{sensing_end_time:%Y%m%d%H%M%S}_{disposition_mode:s}_{processing_mode:s}____.nc.bz2']
```
djhoese, line 18: These bz2 variants should be added to the L2 reader YAML too, right?

satpy/tests/reader_tests/test_metimage_base_nc.py
djhoese: Can't you pass `interpolate=False` in the filetype info?

satpy/readers/core/metimage_nc.py
djhoese, line 1: Can something be added to the module docstring or maybe the file handler class docstring about this new functionality and where the temp files exist?

satpy/tests/reader_tests/test_metimage_base_nc.py
```diff
@@ -441,3 +377,174 @@ def test_dataset(self, po_, pi_, pc_):
...
+        assert reader.spacecraft_name == expected_reader.spacecraft_name
+        assert reader.start_time == expected_reader.start_time
+
+        variable = reader.get_dataset(None, {"file_key": "data/measurement_data/tpw", "calibration": None})
+        expected = expected_reader.get_dataset(None, {"file_key": "data/measurement_data/tpw", "calibration": None})
+        assert np.allclose(variable.values, expected.values)
+        # Not asserting the decompressed temp file is deleted immediately here -- cleanup
+        # timing depends on __del__/gc timing and, on Windows, on the underlying file handle
+        # being released first. The actual contract we care about (cleanup never raises,
+        # even if it fails) is covered separately by test_del_swallows_cleanup_errors.
```
djhoese: What if you did:

```python
import gc
del reader
gc.collect()
assert not os.path.exists(tmp_file)
```

I know you had issues with this on Windows, but have you already tried forcing garbage collection?

satpy/tests/reader_tests/test_metimage_base_nc.py
```diff
@@ -441,3 +377,174 @@ def test_dataset(self, po_, pi_, pc_):
...
+@pytest.fixture
+def metimage_base_nc_file_bz2(metimage_base_nc_file):
+    """Compress the netCDF fixture into a .bz2 file alongside it."""
+    bz2_filename = metimage_base_nc_file.parent / (metimage_base_nc_file.name + ".bz2")
+    with open(metimage_base_nc_file, "rb") as f_in, bz2.open(bz2_filename, "wb") as f_out:
+        f_out.write(f_in.read())
+    return bz2_filename
+
+
+class TestBz2Support:
```
djhoese: Since none of the tests below use any shared logic I think I'd prefer these be standalone functions and remove this class.

djhoese:
Last comment for now: Any interest in adding your name to the `AUTHORS.md` list?

satpy/readers/core/metimage_nc.py
FabMin94: Absolutely. Done — added `_unzipped = None` as a class attribute. AttributeError is no longer possible in `__del__` so it's dropped from suppress(), leaving just suppress(OSError).

satpy/tests/reader_tests/test_metimage_base_nc.py
FabMin94: Yes, that works, and it's definitely much cleaner. Done.
Maybe it's not worth to keep the _interpolated_lonlat() helper anymore then, since that logic is only used in setUp() now.

satpy/etc/readers/metimage_l1b_nc.yaml
FabMin94: Yes. Done.

satpy/readers/core/metimage_nc.py
FabMin94: Sure. Done.

satpy/tests/reader_tests/test_metimage_base_nc.py
FabMin94: I tried it locally on HEAD [e4b5fb8](https://github.com/pytroll/satpy/pull/3433/changes/e4b5fb806339b7040a072a08a7347e4cd0ee69e5), but it still fails on Windows — please see the attached log. [pytest_test_metimage_base_nc_windows.log](https://github.com/user-attachments/files/31826894/pytest_test_metimage_base_nc_windows.log)

Claude's guess:
"The log confirms the file was created via unzip_file, and `__del__` runs (no exception), but the temp file is still present after del reader + gc.collect(). My guess is something beyond the reader itself (xarray's CachingFileManager or a dask-held file handle on the decompressed file) still has it open, independent of the reader object's own refcount — gc.collect() can't force that closed."

satpy/tests/reader_tests/test_metimage_base_nc.py
FabMin94: Sure. Done.

satpy/tests/reader_tests/test_metimage_base_nc.py
djhoese: Eh clean refactored is still clean code. Maybe leave it.

satpy/tests/reader_tests/test_metimage_base_nc.py
```diff
@@ -441,3 +377,167 @@ def test_dataset(self, po_, pi_, pc_):
                                                      "interpolate": True})
         # Checks that the function returns None
         assert longitude is None
+
+
+# --- bz2 decompression support ---------------------------------------------
+# Written as plain pytest functions (rather than TestMETimageNCBaseFileHandler
+# methods) specifically to use tmp_path fixtures for automatic, cross-platform
+# temp-file cleanup -- fixture injection doesn't work inside unittest.TestCase
+# methods, so these live at module scope instead.
```
djhoese, line 386: I don't think this comment is necessary. Please remove it.

satpy/tests/reader_tests/test_metimage_base_nc.py
```diff
@@ -441,3 +377,167 @@ def test_dataset(self, po_, pi_, pc_):
...
+    """
+    return (
+        xr.DataArray(np.ones((10, 100)) * 250, name="longitude",
+                     dims=("num_pixels", "num_lines")),
+        xr.DataArray(np.ones((10, 100)) * 2., name="latitude",
+                     dims=("num_pixels", "num_lines")),
+    )
+
+
+def _create_metimage_base_nc_file(filename):
```
djhoese, line 489: Can these 3 helper functions be moved above the test functions but stay below the fixtures that use them? I was confused by the tests being in the middle.

satpy/tests/reader_tests/test_metimage_base_nc.py
```diff
@@ -441,3 +377,167 @@ def test_dataset(self, po_, pi_, pc_):
...
+
+
+def test_del_swallows_cleanup_errors(metimage_base_nc_file_bz2, metimage_filename_info):
+    """Test that __del__ never raises, even if removing the decompressed temp file is already gone."""
+    filetype_info = {
+        "cached_longitude": "data/measurement_data/longitude",
+        "cached_latitude": "data/measurement_data/latitude",
+        "interpolate": False,
+    }
+    reader = METimageNCBaseFileHandler(str(metimage_base_nc_file_bz2), metimage_filename_info, filetype_info)
```
djhoese, line 456: I notice all these tests pass the string of the filename, not the `Path` object. Is this the same as what is described in https://github.com/pytroll/satpy/issues/3456. Is that why that was needed?

satpy/tests/reader_tests/test_metimage_base_nc.py
djhoese: Yeah that doesn't *not* make sense. We've had difficulties in the past with the xarray cache holding on to things.

satpy/tests/reader_tests/test_metimage_base_nc.py
FabMin94: Not deliberately, no — the str() calls weren't a workaround for this. They're just how the fixtures happened to get used from early on: tmp_path returns Path objects, and str() was the easy way to pass a filename in, following the str type hint in unzip_file's signature. I wasn't aware of the Path-dispatch gap in unzip_file() until this issue. So my tests have been accidentally avoiding the bug this whole time, not exercising a fix for it.

satpy/tests/reader_tests/test_metimage_base_nc.py
FabMin94: Done.

satpy/tests/reader_tests/test_metimage_base_nc.py
FabMin94: Done.

satpy/tests/reader_tests/test_vii_base_nc.py
pnuu: Yeah, I don't think this is very relevant to test.

satpy/readers/core/metimage_nc.py
pnuu: I would lean on the original compressed file. That's the file which were ingested and might be used in cleaning or database functionality in production systems.

satpy/tests/reader_tests/test_metimage_base_nc.py
djhoese: If you have the time, if you remove the `str` call locally and run the tests, do they fail?

FabMin94: I tried it now. Yes, the tests fail, here is the traceback [pytest_unzip_file_with_path_input.log](https://github.com/user-attachments/files/31963910/pytest_unzip_file_with_path_input.log)
As expected, unzip_file silently returns None for the Path input, and netCDF4.Dataset() ends up trying to open the compressed bz2 file directly, resulting in a generic HDF error: `OSError: [Errno -101] NetCDF: HDF error: '...test_file_metimage_base_nc.nc.bz2'`

satpy/tests/reader_tests/test_vii_base_nc.py
FabMin94: Ok, thanks. I'll remove it then.

satpy/readers/core/metimage_nc.py
FabMin94: Ok, then I could do something like:
```
def __init__(self, filename, filename_info, filetype_info, orthorect=False):
    """Prepare the class for dataset reading."""
    self._original_filename = filename
    self._unzipped = unzip_file(filename)
    if self._unzipped:
        filename = self._unzipped
```
and
```
def _get_global_attributes(self):
    """Create a dictionary of global attributes to be added to all datasets."""
    attributes = {
        "filename": self._original_filename,
        ...
    }
```

satpy/tests/reader_tests/test_metimage_base_nc.py
djhoese: Yay! Claude was right. Thanks for checking.

satpy/readers/core/metimage_nc.py
djhoese: I was thinking:

```python
orig_filename = filename
...
super().__init__(...)
self.filename = orig_filename
```

Or do something like `file_to_use = filename` then later `file_to_use = self._unzipped`. I also haven't thought about this that much so I'm sure I'm missing something.

Eh...maybe your way is better @FabMin94 because then `self.filename` is always the file being loaded from and worked on.

FabMin94: Yeah, I think resetting `self.filename` to the original compressed filename after `__init__` would break something: since `cache_handle` defaults to `False` in the `NetCDF4FileHandler`, every subsequent `get_dataset()` call reopens `self.filename` directly, so it would try to open the .bz2 file and hit the same HDF error I reproduced earlier for the Path-dispatch bug.

Here's how I tested it:
```
import bz2
import os
from satpy.readers.core.metimage_nc import METimageNCBaseFileHandler
from satpy.tests.reader_tests.test_metimage_base_nc import _create_metimage_base_nc_file, _metimage_filename_info

nc_filename = "test_file_metimage_base_nc.nc"
_create_metimage_base_nc_file(nc_filename)

bz2_filename = "test_file_metimage_base_nc.nc.bz2"
with open(nc_filename, "rb") as f_in, bz2.open(bz2_filename, "wb") as f_out:
    f_out.write(f_in.read())

filetype_info = {
    "cached_longitude": "data/measurement_data/longitude",
    "cached_latitude": "data/measurement_data/latitude",
    "interpolate": False,
}
filename_info = _metimage_filename_info()

# bz2_filename is a str here
reader = METimageNCBaseFileHandler(bz2_filename, filename_info, filetype_info)
print(reader.filename)  # shows the decompressed temp path (C:\Users\FABIO~1.MIN\AppData\Local\Temp\tmpu9ds8c02 in my case)

reader.filename = bz2_filename

variable = reader.get_dataset(None, {"file_key": "data/measurement_data/tpw", "calibration": None})
...
OSError: [Errno -101] NetCDF: HDF error: 'C:\\...\\satpy\\test_file_metimage_base_nc.nc.bz2'
```

Your second suggestion seems like a close variant of mine — it has to land in the same place, since `_get_global_attributes` needs to read the original name after `__init__` has long finished (so it can't be a local variable), and `self.filename` still needs to stay usable for reads either way.

I'll implement my version for now.

djhoese: Yeah sounds good. Thanks.

djhoese (approved):
I think I'm done complaining. Thanks for sticking with this after all the changes I requested. I think this is in a good state (besides the #3456 bug that Claude found). @ameraner or @pnuu care to do one last review?

pnuu (approved):
Looks good to me.

ameraner (approved):
Looks good to me too, thank you for all the efforts!

PR #3434, opened 2026-07-31
Add FCI LSA SAF FRP csv and nc reader support

### LINKED ISSUE #3350: Add a reader for LSA SAF FRP
## Feature Request

MTG LSA SAF FRP offers a wide range of application options for the Fire Temperature RGB. 

I would like to add a satpy writer for the new MTG FRP-Pixel product.

This provides an additional option in addition to the existing FIR reader.

**Additional context**
More details regarding documentation and data: https://lsa-saf.eumetsat.int/en/data/products/fire-products/

alexhalbig:
## Summary

This PR adds tests for the FCI LSA SAF FRP L2 CSV and NC reader.
Including yaml-, utest-, composite-, and enhancement-files

Work finished, based on PWC Spring 2026 Issue - Closes #3350 

## Testing
- pre-commit run --all-files
- ruff check . --fix
- ruff format .
- relevant local unit tests passed


satpy/etc/enhancements/fci.yaml
```diff
@@ -203,3 +203,25 @@ enhancements:
...
+      kwargs:
+        palettes:
+        - values:
+            - 0
+            - 50
+            - 100
+            - 150
+            - 200
+            - 250
+          colors:
```
ameraner, line 221: Just for curiosity, did you design this colormap yourself or is it taken from somewhere?

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with satpy.  If not, see <http://www.gnu.org/licenses/>.
+
+"""MTG FCI Fire Radiative Power (FRP) Level-2 (L2) CSV reader.
+
+This reader supports reading the frp product from the LSASAF FRPPIXEL based Product.
+It can be e.g. used with SingleBandCompositor or MaskingCompositor.
```
ameraner:
```suggestion
It can be used standalone to read the data from the files, or to generate composites e.g. with SingleBandCompositor or MaskingCompositor.
```

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+
+More detailed information about the related product and data see:
+https://lsa-saf.eumetsat.int/en/data/products/fire-products/
+
+
+Per default, the reader retds and loads an sparse 1-D array
```
ameraner:
```suggestion
Per default, the reader reads and loads a 1-D array
```

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+of fire pixels and maps them on an 2-D grid based on the 1 km Full disk
```
ameraner:
```suggestion
of fire pixels and maps them on a sparse 2-D grid based on the 1 km Full disk
```

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+FCI grid. The resulting 2-D array is Compatible with later resampling.
```
ameraner:
```suggestion
FCI grid. The resulting 2-D array is compatible with later resampling.
```

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+
+NOTE: Currently method end_time is designed for 10 minute scan times, not for RSS.
```
ameraner:
```suggestion
NOTE: The reader currently assumes the product to be full-disc, with a 10-minute repeat cycle, and coming from Meteosat-12.
```

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+        """Get area definition for fci native 1 km grid."""
+        return get_area_def("mtg_fci_fdss_1km")
+
+    def get_array_on_fci_grid(self, data_array):
+        """Place 1D fire detections on sparse 2D FCI grid."""
+        rows = self["abs_line"]
+        cols = self["abs_samp"]
+        attrs = data_array.attrs.copy()
+
+        rows_int = (rows.astype(int) - 1).compute()
```
ameraner:
```suggestion
        rows_int = (rows.astype(int)).compute()
```

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+        cols_int = (cols.astype(int) - 1).compute()
```
ameraner:
```suggestion
        cols_int = (cols.astype(int)).compute()
```

satpy/readers/fci_lsasaf_frp_l2_csv.py
ameraner: from direct communication with Emanuel Dutra from LSA-SAF, the indexing in the FCI product is 0-based (while it was 1-based in the SEVIRI product).

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+            if key in dsinfo:
+                data.attrs[key] = dsinfo[key]
+
+        if name == "frp":
+            lons = self["longitude"].to_dask_array(lengths=True)
+            lats = self["latitude"].to_dask_array(lengths=True)
+            data = data.assign_coords({
+                "longitude": ("y", lons),
+                "latitude": ("y", lats),
+            })
```
ameraner: maybe I'm missing something, but what is the point of adding the lon-lat as coordinates on y here, if we then immediately after remap the data on the 2-d array?

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,154 @@
...
+
+        rows_int = (rows.astype(int) - 1).compute()
+        cols_int = (cols.astype(int) - 1).compute()
+
+        values = data_array.data.compute() if hasattr(data_array.data, "compute") else data_array.values
+
+        valid = (
+            (rows_int >= 0) & (rows_int < FRP_GRID_SHAPE[0]) &
+            (cols_int >= 0) & (cols_int < FRP_GRID_SHAPE[1])
+        )
```
ameraner: Have you ever observed values outside of the range? I don't know if we have to make the effort of checking the validity of the rows/cols here, I would delegate/trust it to the LSASAF, as it would be a major anomaly otherwise..

ameraner (changes requested):
Nice job, thank you for this! Just some minor comments/questions/edits, and a requried change for the pixel coordinates.

satpy/etc/enhancements/fci.yaml
alexhalbig: I designed the colormap myself. But I tried to stick closely to the EUMETView one, mainly for comparability reasons in the first place. I also designed one using ADAGUC-like colors, but found that the background contrast on this one seems better and colors are more intuitive for fires.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Adjusted as suggested.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Adjusted as suggested.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Adjusted as suggested.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Adjusted as suggested.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Adjusted as suggested.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: You are right. Dimensions are not needed here. 
Removed assign_cords()

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Adjusted as suggested. 
I double checked this on the data, you are right. FCI LSA SAF FRP Products (both CSV and NetCDF) are 0-based. Thus -1 would cause a offset error.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Adjusted as suggested.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Removed valid-check. 
My offset error forced me to implement this valid-check, causing an error at the very uppermost and upperleft row/col. Now this valid-check is not needed anymore.

alexhalbig:
The CI run still fails in the two TestParallaxCorrectionModifier.test_modifier_interface_cloud_moves_to_observer cases for foroyar. The failure is caused by an unevaluated Dask boolean array:
`assert dask.array<all-aggregate, shape=(), dtype=bool, ...>`

This behavior surprises me, as the CI tests passed 2 days ago. The changes made, suggested by Andreas Review, should not impact the testing success in my opinion. 

My branch is not behind origin/main (HEAD...origin/main = 8 0). Could you please confirm whether this is a known Dask/Pyresample compatibility issue and whether these two tests should be adjusted or marked as expected failures?

satpy/etc/readers/lsasaf_mtgfrp_l2.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+file_types:
+  csv_fci_frp:
+    file_reader: !!python/name:satpy.readers.fci_lsasaf_frp_l2_csv.FRPFileHandler
+    file_patterns:
+      - '{facility_or_tool}-509_{platform_name}_MTFRPPIXEL-ListProduct_MTG-{coverage}_{start_time:%Y%m%d%H%M}.csv'
+  nc_fci_frp:
+    file_reader: !!python/name:satpy.readers.fci_lsasaf_frp_l2_nc.FRPFileHandler
+    file_patterns:
+     - 'W_PT-{facility_or_tool},SATELLITE,LSA-509_{platform_name}_MTFRPPIXEL_MTG-{coverage}_{disposition_mode}_LPMG_{start_time:%Y%m%d%H%M%S}.nc'
```
ameraner, line 19: until we really support nc, this should be removed

satpy/readers/fci_lsasaf_frp_l2_csv.py
ameraner: could it be that the change was not pushed? The assign_coords is still there

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,149 @@
...
+
+        values = data_array.data.compute() if hasattr(data_array.data, "compute") else data_array.values
+
+        grid = np.full(FRP_GRID_SHAPE, np.nan, dtype=np.float32)
+        grid[rows_int, cols_int] = values
+        xarr = xr.DataArray(
+            grid,
+            dims=("y", "x"),
+            attrs=attrs,
+        )
```
ameraner: The `grid` array here should be daskified to adhere to the normal format of reader outputs, and to optimise memory usage. You can take inspiration from an old version of the LI L2 reader to do that: https://github.com/pytroll/satpy/blame/618fc0e498c7f1e1bfce011f1ed1b041895ed8ce/satpy/readers/li_l2_nc.py#L141-L148
(in that reader, later on we needed an `np.add.at` functionality - so if you look at the current reader it is much more complex. However you don't need that here, since we have only one value per pixel, hence my link to an older blame version).

ameraner (changes requested):
Thank you for the updates! I noticed a couple of things, see below. 

The failing tests are not due to your changes, they're probably due to some library update that broke the parallax tests.. they will need to be fixed separately (feel free to give it a go in another PR if you have time).

gerritholl:
I confirm that **tests are failing in satpy main** after `mamba update --all` in my local environment.  This is unrelated to the current PR, but must be due to a recent dependency update.

satpy/etc/readers/lsasaf_mtgfrp_l2.yaml
alexhalbig: I thought one step in advance, as I just need to finish the utests for the nc-reader. But I agree with you. Will be removed here, and added later.

gerritholl:
The test failures are unrelated to this PR.  Satpy main is also failing.  See https://github.com/pytroll/satpy/issues/3441 .

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: Thanks this was helpful. The new version is daskified now.

alexhalbig:
@ameraner @mraspaud @djhoese From my perspective, this PR is ready to merge.  I have completed the CSV and NC reader support.  Could you please take a look?

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,153 @@
...
+        attrs = data_array.attrs.copy()
+
+        rows_int = (rows.astype(int)).compute()
+        cols_int = (cols.astype(int)).compute()
+
+        # values is a 1-D List of np.ndarray
+        values = data_array.data.compute() if hasattr(data_array.data, "compute") else data_array.values
+
+        # Create an empty 1-D nan array for the results
+        flattened_result = np.nan * da.zeros((FRP_GRID_SHAPE[0] * FRP_GRID_SHAPE[1]), dtype=data_array.dtype)
```
alexhalbig, line 142: I’ve been thinking about whether this should be implemented with or without np.nan. What do you think?
- with np.nan: it becomes a masked array, which can still optionally be filled using fill_value=0
- without np.nan:  it remains a regular raster, automatically filled with fill_value=0 
As it's a fire detection variable, I prefer the first one.

satpy/etc/readers/lsasaf_mtgfrp_l2.yaml
djhoese: If NC support is finished then this thread can be resolved right? My next question though if that's true, should the NC files be listed first in the `file_type:`s below? I assume the netcdf files are better quality and better performance to read.

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,153 @@
...
+# (at your option) any later version.
+#
+# satpy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with satpy.  If not, see <http://www.gnu.org/licenses/>.
+
```
djhoese, line 15:
```suggestion
```

We just recently relicensed to Apache v2 which does not require a license header on each file. Correct @mraspaud?

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,153 @@
+# Copyright (c) 2022 Satpy developers
```
djhoese, line 1:
```suggestion
# Copyright (c) 2026 Satpy developers
```

I'm not sure where we landed on copyright comments with the license change, but either way the year needs to be correct.

satpy/readers/fci_lsasaf_frp_l2_csv.py
```diff
@@ -0,0 +1,153 @@
...
+        if name == "frp":
+            data = self.get_array_on_fci_grid(data)
+
+        return data
+
+    def get_area_def(self, dsid):
+        """Get area definition for fci native 1 km grid."""
+        return get_area_def("mtg_fci_fdss_1km")
+
+    def get_array_on_fci_grid(self, data_array):
```
djhoese, line 129: I'm not familiar with this data. Could you tell me what `abs_line` and `abs_samp` contain? Are they a single value or a 1D array the same size as the variable being gridded?

Looking at the fancy work here I'm wondering/hoping if there is a way to put this into a `map_blocks` call or something. So the function that dask would call per chunk would generate the NaN filled chunk numpy array, fill in the data related to that chunk, and then return it. If you're already computing the 1D array (or pass it to `map_blocks` as a single chunk) then it should "just work". If necessary there is also dask's `blockwise`. My point being that I'm wondering if this could be a bit more lazy.

satpy/readers/fci_lsasaf_frp_l2_nc.py
djhoese, line 1: If these aren't two separate readers then I think it would make sense and follow common practice to put them in the same python module. Same for the tests.

djhoese (changes requested):
I thought I had only a few questions so I made individual comments and then I just kept having more and more to say. Nice work on this. I just have a few questions/suggestions so far.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: abs_line represents the index of the active fire pixel, related to the same row in the native FCI projection grid. 
Accordingly, abs_samp is the column number (zero based) value in the FCI projection.

satpy/readers/fci_lsasaf_frp_l2_nc.py
alexhalbig: I agree that the tests and readers could be merged into one unified reader. That was my initial intention. However, I quickly noticed this would make the structure more complex to read and maintain.
The CSV files contain fewer variables than the NetCDF files and use a different filename structure. NetCDF provides detailed metadata; CSV provides none, except "encoded" in the filename. 
I therefore prefer a simple KISS-style solution with minimal merging and no extensive if/else logic at numerous places.  I therefore focused on making the user experience as convenient as possible. Users only need to specify "reader='lsasaf_mtgfrp_l2' " Satpy then determines from the filenames whether the NetCDF or CSV reader should be used.

satpy/readers/fci_lsasaf_frp_l2_csv.py
alexhalbig: map_blocks and/or blockwise sounds worth a test. Maybe even dask.delayed. Anyway, the whole overall processing is already very fast, taking only about 10 seconds. I would consider this as a future improvement step :-)

satpy/readers/fci_lsasaf_frp_l2_nc.py
djhoese: Unless I'm missing something there doesn't need to be any `if/else` logic. I'm not talking about merging them into one class. The two classes go in one module. One class gets called FRPCSVFileHandler and the other FRPNetCDFFileHandler and the YAML references those. The variable map global variables get a CSV or NC specific prefix to them. The chunk size and FRP shape variables don't even need to change. The CSV file handler is less than 100 lines of code and the NC reader is ~150. It is very common to have multiple file handlers in a single reader module and it is expected to be that way.

satpy/readers/fci_lsasaf_frp_l2_csv.py
djhoese: Sounds good. Semi-recently Dask has actually steered away from mixing Delayed objects and Dask arrays for performance reasons (it ends up duplicating some of the dask graph in some cases). Let's save it for later. that's fine by me.

satpy/readers/fci_lsasaf_frp_l2_csv.py
djhoese: I think a NaN filled raster makes the most sense if I'm understanding you correctly.

satpy/etc/readers/lsasaf_mtgfrp_l2.yaml
alexhalbig: Yes agree, they contain more reliant metadata information than parsing information from the filename. I admit, I didn't know the order of the file_types is important.

satpy/readers/fci_lsasaf_frp_l2_nc.py
alexhalbig: Gotcha, thanks for clarifying. That sounds reasonable. I will change and let you know.

ameraner (commented):
Thank you Alex for adding the NetCDF reader too! 
For better flexibility, I would propose to add the `with_area_definition` mechanism also to this reader, see the equivalent LI L2 implementation https://github.com/pytroll/satpy/blob/73ae7e4a90e632ba35fc5acf631aaa34bfd3946c/satpy/readers/li_l2_nc.py#L47-L51 . I would set it as default `True`, same as for LIL2 https://github.com/pytroll/satpy/blob/73ae7e4a90e632ba35fc5acf631aaa34bfd3946c/satpy/readers/li_l2_nc.py#L81 . 
That way, by default (the current behaviour) you get a 2-d array that can be used as a mask etc. . But if requested through the reader_kwargs, a user can get just the 1-d lists of lat, lon and FRP values. If you still have time to work on this, otherwise it can be added later!

PR #3435, opened 2026-07-31
Add pyresample, trollimage, trollsift and pykdtree to show_versions defaults

### LINKED ISSUE #3116: Add additional dependencies to `show_versions` and `check_satpy`

khlaifiabilel:
﻿Adds the four Pytroll packages requested in #3116 to the default package list of `satpy.utils.show_versions`, so `check_satpy()` output includes them.

Happy to also add `pycoast`, `pydecorate`, and `geotiepoints` here if desired - they were floated in the issue but not confirmed.

 - [x] Closes #3116
 - [ ] Tests added - existing `show_versions`/`check_satpy` tests already exercise the default list

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -55,6 +55,7 @@ The following people have made contributions to this project:
 - [Przemyslaw Juda (pjuda)](https://github.com/pjuda)
 - [Inderpreet Kaur](https://github.com/ikaur17)
 - [Pouria Khalaj](https://github.com/pkhalaj)
+- [Bilel Khlaifia (khlaifiabilel)](https://github.com/khlaifiabilel)
 - [Janne Kotro (jkotro)](https://github.com/jkotro)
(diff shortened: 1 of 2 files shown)


djhoese:
Duplicate of #3426

khlaifiabilel:
Apologies for the noise - I missed that #3426 already covered this. Thanks for the quick triage!

PR #3438, opened 2026-08-06
Silence AVHRR GAC/LAC test warnings

sfinkens:
Silence warnings in `test_avhrr_l1b_gaclac.py`

### DIFF
--- a/satpy/tests/reader_tests/test_avhrr_l1b_gaclac.py
+++ b/satpy/tests/reader_tests/test_avhrr_l1b_gaclac.py
@@ -18,6 +18,7 @@
 
 """Pygac interface."""
 import datetime as dt
+import warnings
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
@@ -80,6 +81,14 @@ LAC_KLM_FILENAMES = ["BRN.HRPT.M1.D14152.S0958.E1012.B0883232.UB",
 
 LAC_EOSIP_FILENAMES = ["N06_RPRO_AVH_L1B_1P_20061206T010808_20061206T012223_007961/image.l1b"]
 
+
+@pytest.fixture(autouse=True)
+def silence_pygac_warnings():
+    """Silence some warnings from Pygac."""
+    warnings.filterwarnings(action="ignore", message="Using CoeffStatus.PROVISIONAL", category=RuntimeWarning)
+    warnings.filterwarnings(action="ignore", message="Using the 'corr' argument", category=DeprecationWarning)
+    warnings.filterwarnings(action="ignore", message="divide by zero", category=RuntimeWarning)
+    warnings.filterwarnings(action="ignore", message="Unexpected record length", category=RuntimeWarning)
 
 class TestGACLACFile:
     """Test the GACLAC file handler."""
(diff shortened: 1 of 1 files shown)


satpy/tests/reader_tests/test_avhrr_l1b_gaclac.py
```diff
@@ -81,6 +82,14 @@
 LAC_EOSIP_FILENAMES = ["N06_RPRO_AVH_L1B_1P_20061206T010808_20061206T012223_007961/image.l1b"]
 
 
+@pytest.fixture(autouse=True)
+def silence_pygac_warnings():
+    """Silence some warnings from Pygac."""
+    warnings.filterwarnings(action="ignore", message="Using CoeffStatus.PROVISIONAL", category=RuntimeWarning)
+    warnings.filterwarnings(action="ignore", message="Using the 'corr' argument", category=DeprecationWarning)
+    warnings.filterwarnings(action="ignore", message="divide by zero", category=RuntimeWarning)
```
djhoese: Is this an expected warning or could this be removed by making more realistic test data?

satpy/tests/reader_tests/test_avhrr_l1b_gaclac.py
```diff
@@ -81,6 +82,14 @@
...
+    warnings.filterwarnings(action="ignore", message="Unexpected record length", category=RuntimeWarning)
```
djhoese: Is this expected in one specific "sad path" of the tests? Can we add this filter only for that test?

sfinkens:
Thanks for reminding me not to take easy shortcuts 😄  With modified test data both warnings are gone now. For KLM I was able to figure it out myself, for POD I needed some help from an LLM. Anyway I have a better understanding of Pygac internals now 💡

satpy/tests/reader_tests/test_avhrr_l1b_gaclac.py
sfinkens: I changed the number of scans to 56 so the warning disappears. Adapted end-timestamps accordingly.

satpy/tests/reader_tests/test_avhrr_l1b_gaclac.py
```diff
@@ -241,57 +254,86 @@ def _get_times(num_lines: int):
...
 
         hdr3 = np.zeros(1, dtype=pygac.pod_reader.header3)
 
         spare = np.zeros(pygac.gac_pod.GACPODReader().offset - (hdr0.itemsize + hdr3.itemsize), dtype="u1")
         return [hdr0, hdr3, spare, scans]
 
-    @staticmethod
-    def _get_telemetry(num_scans):
-        """Get encoded telemetry.
+    def _get_telemetry(self):
```
djhoese, line 285: Small nitpick, but I think in the long run this would be better as a function outside this class. Not worth changing now, but it only takes this single `num_scans`.

sfinkens: Yes I'm not sure about that class. It's just a container for a couple of helper methods. Before, all methods were static. Now I added `num_scans` as an attribute, so that less arguments need to be passed around.

PR #3439, opened 2026-08-08
Add AGENTS.md file

djhoese:
Includes CLAUDE.md to point to AGENTS.md and various fixes identified by Claude. The file was almost entirely generated by Claude with guidance from me. It still needs more information on development practices.

@mraspaud and @pnuu I figured you could add/update this with the development practices you've got locally or from other projects. Then we can review and try to come to an agreement.

Note I purposefully don't have anything in here about `git` because I, for the most part, don't let agents commit for me or push or make a pull request.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -3,24 +3,90 @@
 Satpy is a [Pytroll](https://pytroll.github.io/) library for reading, manipulating, and writing data
 from remote-sensing earth-observing satellite instruments.
 Satpy is an abstraction layer over `pyresample`, `pyspectral`, `trollimage`, `pycoast`, and
-`python-geotiepoints`.
+`python-geotiepoints` libraries.
+
+If something below is not clear, consult the Sphinx documentation in `doc/source/`; the
+"Where to read more" table at the bottom points at the most relevant pages.
 
 ## Core concepts
 
-- `Scene` (`satpy/scene.py`) is the user-facing container. Typical flow: `Scene(filenames, reader=)`
-  → `load([...])` → `resample(area)` → `save_datasets()`.
-- Data is always an `xarray.DataArray` wrapping a **dask array**. Dims are `y`, `x`, and `bands`.
-  Metadata lives in `.attrs`; `area`, `start_time`, `end_time`, and `sensor` are expected to be
-  present.
-- `DataID` / `DataQuery` (`satpy/dataset/dataid.py`) are the keys in a Scene's `DatasetDict`.
-  Default ID keys for imager bands: `name`, `wavelength`, `resolution`, `calibration`, `modifiers`.
-- Pipeline: **readers** → **composites** (combine bands) → **modifiers** (transform one band) →
-  **resampling** → **enhancements** (stretch to 0–1) → **writers**. `DependencyTree`
-  (`satpy/dependency_tree.py`) resolves composite/modifier prerequisites.
-- **Every component is a Python class *plus* a YAML config.** YAML under `satpy/etc/` points at
-  classes with `!!python/name:` and is loaded with PyYAML's `UnsafeLoader` (which is why `.bandit`
-  skips B506). Adding a class without a matching YAML entry does nothing.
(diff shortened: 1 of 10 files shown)


sfinkens:
Interesting, thanks for starting this! I've read that LLMs generate more reliable results given some domain abstractions. Should we explain Satpy's abstractions (Scene, Composites etc) in this file or maybe point to the documentation (`overview.rst` for example)?

djhoese:
@sfinkens Yes that is probably a good idea and is probably what I would have done if I had written this file from scratch. So maybe I should do that. This file was generated by Claude and it was told specifically to look at the docs as a starting point and then look at the docstrings in the python modules. Yeah...maybe I should rewrite some of this.

AGENTS.md
```diff
@@ -0,0 +1,244 @@
...
+- Autouse fixtures in `satpy/tests/conftest.py`: `_reset_satpy_config`, `_clear_function_caches`,
+  `_forbid_pyspectral_downloads`. Opt into `include_test_etc` to point config at `satpy/tests/etc`.
+- Helpers in `satpy/tests/utils.py`: `make_dataid`, `make_dsq`, `FakeCompositor`, `CustomScheduler`,
+  `assert_maximum_dask_computes`, `make_fake_scene`. Enhancement tests assert **zero** dask computes
+  via `satpy/tests/enhancement_tests/utils.py`.
+- `--strict-markers`, `--strict-config`, and `xfail_strict` are on; no custom markers are registered.
+- Tests must not emit new warnings. Escalate in this order: fix the underlying cause →
+  `pytest.warns(match=...)` → `@pytest.mark.filterwarnings` → global `filterwarnings` in
+  `pyproject.toml`.
+- Prefer semi-realistic fake data (`da.arange(...).reshape(...).rechunk(...)`) over `zeros`/`ones`.
```
pnuu, line 218: Unless realistic values are needed, in which case ... ?

pnuu (commented):
Looks great! Just one comment inline.

I'm not sure the `AGENTS.md` I added in some other package has anything needed here. The only addition there was the TDD, which might need reformulation based on the discussion on Slack.

AGENTS.md
djhoese: I think most AIs would figure that out where necessary. Like if it is checking calibration then having values that produce valid calibrated results would make sense. I'm not sure how it picked up on this note in line 218, but it is a real world gotcha where I've had bugs missed in tests because the tests used numpy arrays or zeros/ones rather than data that varies and is in a dask array.

djhoese:
Ok I think this is good where it is and could be merged. Future changes can be separate PRs. I think agents using TDD is questionable and doesn't seem to consistently produce better results depending on the agent and how the rest of the task is defined.

AGENTS.md
pnuu: And apparently I read it incorrectly, your version is absolutely what I wanted :joy:

pnuu (approved):
LGTM!

djhoese:
One thing that just came up in an AI video I was watching is to have Claude (or other agent) move things outside of AGENTS.md and into skills files. That way the AGENTS.md file stays very slim and you only have code instructions (like how to run tests or create a dev environment) in the AI context when they're needed.

sfinkens:
Nice work! For me it would be fine to extract skills later.

sfinkens (approved):

pnuu:
The skills sound good, but I'm not sure how we should distribute them. As far as I've understood, they need to be insttalled separatedly for each harness and are not part of the repository.

djhoese:
Claude automatically detects `.claude/skills` if it is in your home directory *or* the project root. The AGENTS.md file could point other agents to that same location.

djhoese:
Ok I got approval from 2 of the 3 people that talk about AI agents the most. If Martin has a problem then I can address it later. I'm going to merge this now and we can always update it with more rules as we go.

PR #3442, opened 2026-08-16
Fix required NetCDF root variable collection

### LINKED ISSUE #2704: UnboundLocalError when required_netcdf_variables contains only root variables without /
**Describe the bug**

When a reader derives from `NetCDF4FileHandler` and uses `required_netcdf_variables` using only root variables, thus no groups and no required variables containing a `/`, instantiation will fail with `UnboundLocalError`.

**To Reproduce**

```python
# Your code here
from satpy.readers.netcdf_utils import NetCDF4FileHandler
import numpy as np
import xarray as xr

ft_info = {"file_reader": NetCDF4FileHandler, "file_patterns": ["{stem}.nc"], "required_netcdf_variables": ["panarea", "strómboli"]}

nm = "/tmp/test.nc"
ds = xr.Dataset()
ds["panarea"] = xr.DataArray(np.array([[0, 1, 2]]), dims=["y", "x"])
ds["strómboli"] = xr.DataArray(np.array([[1, 1, 2]]), dims=["y", "x"])
ds.to_netcdf(nm)

h = NetCDF4FileHandler("/tmp/test.nc", {"stem": "test"}, ft_info)
```

**Expected behavior**

I expect that I can create the file handler, or that it perhaps errors later due to not having implemented some mandatory pseudo-abstract methods.

**Actual results**

```python-traceback
  File "/data/gholl/checkouts/protocode/mwe/satpy-req-nc-vars-bug.py", line 16, in <module>
    h = NetCDF4FileHandler("/tmp/test.nc", {"stem": "test"}, ft_info)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/gholl/checkouts/satpy/satpy/readers/netcdf_utils.py", line 114, in __init__
    self._collect_listed_variables(file_handle, listed_variables)
  File "/data/gholl/checkouts/satpy/satpy/readers/netcdf_utils.py", line 182, in _collect_listed_variables
    if p != "attr":
       ^
UnboundLocalError: cannot access local variable 'p' where it is not associated with a value
```

**Environment Info:**
 - Satpy Version: main (v0.46.0-46-g989d9f0bd)

**Additional context**

Although I haven't tested it, it would seem to not collect any root variables following an attribute.

I ran into this when writing unit tests for #2686.

sylvesterkaczmarek:
Fixes #2704.

`_collect_listed_variables` currently relies on the loop variable `p` after iterating over path components. For a root variable that loop does not execute, which raises `UnboundLocalError`. A root variable following an attribute can also inherit the previous `p == "attr"` state and be skipped.

This replaces that implicit loop-variable state with an explicit per-item attribute flag, preserving existing grouped-variable and attribute handling while allowing root variables to be collected correctly.

Regression coverage includes:
- a required root variable with no `/` in its name
- a root variable immediately following a required attribute

- [x] Closes #2704
- [x] Tests added

Documentation is not changed because this is a reader-internal correctness fix with no public API change.


satpy/tests/reader_tests/test_netcdf_required_variables.py
djhoese, line 1: Could the tests in this file be added to the existing test module (`satpy/tests/reader_tests/test_netcdf_utils.py`)? I also think they could maybe be combined into a single test and use pytest's parametrize? Or maybe I'm missing the point of the tests too much?

djhoese (changes requested):
Thanks for this. This makes sense and is a clean fix. I'm surprised after all the refactoring that this bug took this long to find. I had one small request and one possibly invalid request that I made in an inline comment.

satpy/tests/reader_tests/test_netcdf_required_variables.py
djhoese: Oh even better, would it make sense to have these tests absorbed into existing tests?

sylvesterkaczmarek: Thanks, agreed. I've moved the regression coverage into the existing test_netcdf_utils.py module, combined the two root-variable cases using pytest parametrization, and removed the standalone test file.

djhoese (approved):
Looks good to me. @pnuu @ameraner reviews would be appreciated.

pnuu (approved):
LGTM, thanks!

ameraner (approved):
LGTM too, thanks!

djhoese:
Thanks @sylvesterkaczmarek!

PR #3295, opened 2025-11-13
Bump supported Python version to 3.12, 3.13 and 3.14

pnuu:
Drop Python 3.11 support and add Python 3.14 to CI.

 - [x] Closes #3267

PR #3443, opened 2026-08-18
Fix parallax tests after PROJ 9.8.0/1 updates

### LINKED ISSUE #3441: Tests failing after upgrading proj 9.7.1 → 9.8.1
**Describe the bug**

Tests in satpy main are failing with proj 9.8.1.  They work with proj 9.7.1.

**To Reproduce**

```bash
pytest -k test_modifier_interface_cloud_moves_to_observer modifier_tests/test_parallax.py
```

**Expected behaviour**

Tests pass.

**Actual results**

Tests fail.  For example, see the CI run for #3434 at https://github.com/pytroll/satpy/actions/runs/31711751053/job/94486215575 .

**Environment Info:**
 - OS: both mac and linux, it seems
 - Satpy Version: satpy main
 
**Additional context**

In the short term, we can pin proj to 9.7.1.  But we need to investigate the problem and either report a bug with the proj crew or adapt the code in satpy.

djhoese:
The EPSG 4087 was affected by this PROJ PR https://github.com/OSGeo/PROJ/pull/4656 which adds spherical `eqc` formulas. The end result is that one of the grids in the parallax tests shifted north by a bit and made the hardcoded pixel offsets incorrect. Claude offered to make the tests smarter and more flexible, but this fix was simpler. It does require PROJ 9.8+ to pass though.

 - [x] Closes #3441 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/tests/modifier_tests/test_parallax.py
+++ b/satpy/tests/modifier_tests/test_parallax.py
@@ -657,7 +657,7 @@ class TestParallaxCorrectionModifier:
         cloud_location = {
                 "foroyar": {
                     7500: (197, 202, 152, 172),
-                    15000: (238, 243, 164, 184)},
+                    15000: (239, 244, 165, 185)},
                 "ouagadougou": {
                     7500: (159, 164, 140, 160),
                     15000: (163, 168, 141, 161)}}


djhoese:
Ugh the Python 3.11 environment installs older PROJ :man_facepalming:

gerritholl:
> Ugh the Python 3.11 environment installs older PROJ 🤦‍♂️

Shouldn't we [have dropped Python 3.11 already](https://scientific-python.org/specs/spec-0000/)?

djhoese:
@gerritholl Yes, the PR has existed for a while https://github.com/pytroll/satpy/pull/3295. Would you prefer the simpler solution (the first commit in this PR) and we merge that drop-3.11 PR?

gerritholl:
Yes, I would prefer if we can merge that PR, so the solution here can be simpler.  I find the shrink and grow names a bit confusing.  But #3295 has failing tests and merge conflicts.

djhoese:
Ok I've reverted the "flexible" solution and gone back to the simpler one. I've also rebased on main which now has #3295 merged. @gerritholl if/when tests pass are you OK with me merging this?

gerritholl (approved):
Thanks for fixing this!

ameraner:
Very nice, thanks for investigating and fixing!

PR #3445, opened 2026-08-18
Fix AWIPS Tiled writer failing if valid_range missing

djhoese:
While trying to work with MetImage data in the AWIPS tiled writer I got an error because the writer assumed all datasets have a `valid_range` attribute. In MetImage's case it has a valid_min/valid_max. I will make another PR where I'll remove the attributes.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese:
I need this for some Polar2Grid work. Small change on "my" writer, going to merge. If someone wants to give feedback later I'll fix things then.

PR #3446, opened 2026-08-18
Remove valid_min/max attributes from MetImage DataArrays

djhoese:
The metimage reader(s) don't remove valid_min and valid_max attributes which causes issues for me later in the writers (ex. AWIPS tiled writer). These attributes, once the data is scaled, are no longer valid so there should be no harm removing them...at least I think. An alternative solution would be to scale them just like the data values are scaled so they are in the same unit space.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 


ameraner (approved):
LGTM, thanks for the fix! I also think it's ok to just remove them.

PR #3447, opened 2026-08-20
Fix negligible coordinate dimension matching

### LINKED ISSUE #3353: drop_coordinates behaviour contradicts documentation
**Describe the bug**

According to the documentation, `satpy.composites.core.CompositeBase.drop_coordinates` drops coordinates that do not correspond to any dimension:

https://github.com/pytroll/satpy/blob/58ce5367ea81a8083794b1e53c643b4bb85a5005/satpy/composites/core.py#L207-L214

In reality, it also drops coordinates that do correspond to a dimension.

**To Reproduce**

```python
import numpy as np
import xarray as xr
import satpy.composites.core

ds = xr.DataArray(
        np.arange(3*4).reshape(3, 4),
        dims=("y", "x"),
        coords={"time": xr.DataArray(
            np.arange(3).reshape(3), dims=("y",))})

print(ds.coords)

# this dataset has a coordinate (time) corresponding to a dimension (y)

ds_new = satpy.composites.core.CompositeBase.drop_coordinates([ds])[0]

# but the coordinate is still dropped

print(ds_new.coords)
```

**Expected behaviour**

I'm not sure what the intention is of the code.

I expect that either

- the documentation accurate describes that coordinates are dropped even if they correspond to a dimension, unless that dimension has the same name as the coordinate, or
- this coordinate is retained, as is documented.

**Actual results**

The coordinate is dropped:

```
Coordinates:
    time     (y) int64 24B 0 1 2
Coordinates:
    *empty*
```

**Environment Info:**
 - OS: openSUSE Leap 15.6
 - Satpy Version: v0.60.0

**Additional context**

This behaviour was originally introduced in #796.  It breaks #3168, which intends to retain time information.

sylvesterkaczmarek:
Fixes #3353.

`CompositeBase.drop_coordinates` currently checks only whether a coordinate name also appears in `DataArray.dims`. This can preserve `time(y)` whenever the array separately has a `time` dimension, even though that coordinate does not describe the `time` dimension.

This changes the check to preserve a negligible coordinate only when its own dimensions are exactly its same-named dimension, for example `time(time)`.

Tests cover:
- dropping `time(y)` without a `time` dimension
- dropping `time(y)` when a separate `time` dimension exists
- preserving a real `time(time)` dimension coordinate

Testing:
- Python syntax checks for the modified module and new test
- focused xarray reproduction for all three coordinate cases


djhoese:
>  This can preserve time(y) whenever the array separately has a time dimension, even though that coordinate does not describe the time dimension.

Sorry, I'm confused by this. So we have DataArray A that has dimensions `(time, y, x)` and one of its coordinates is "time" which is a DataArray with dimensions `(y,)`?

sylvesterkaczmarek:
> > This can preserve time(y) whenever the array separately has a time dimension, even though that coordinate does not describe the time dimension.
> 
> Sorry, I'm confused by this. So we have DataArray A that has dimensions `(time, y, x)` and one of its coordinates is "time" which is a DataArray with dimensions `(y,)`?

Yes, exactly. xarray allows a DataArray with dimensions (time, y, x) to have a coordinate named time whose own dimensions are (y,). That is the second reproducer in #3353. 

The current logic only checks whether the coordinate name time exists in ds.dims, so it incorrectly treats that time(y) coordinate as the coordinate variable for the time dimension. This PR instead checks the coordinate's own dimensions and only preserves the true dimension-coordinate case time(time).

djhoese:
That is a strange case and it seems I said the same thing in my comment in #3353. Unfortunately @gerritholl is going on holiday for a while so we can't get his perspective on the issue as he saw it.

Two requests:

1. Could you move the tests into an existing test module? We typically don't make test modules for specific cases, but more for specific units/modules of code.
2. What do you think about making a helper function or multiple helper functions to implement this filtering. For example a `_is_coord_var` or something?

sylvesterkaczmarek:
> That is a strange case and it seems I said the same thing in my comment in #3353. Unfortunately @gerritholl is going on holiday for a while so we can't get his perspective on the issue as he saw it.
> 
> Two requests:
> 
> 1. Could you move the tests into an existing test module? We typically don't make test modules for specific cases, but more for specific units/modules of code.
> 2. What do you think about making a helper function or multiple helper functions to implement this filtering. For example a `_is_coord_var` or something?

Thanks, both suggestions make sense. I’ve moved the regression cases into the existing test_core.py module alongside the other CompositeBase coordinate tests, and extracted the coordinate-variable check into _is_coord_var. 

The filtering now reads explicitly in terms of whether the coordinate is the variable for its same-named dimension. CI is rerunning now.

Manny7717 (approved):
Verified locally on CPU (head f091316e vs the pre-fix core.py blob from the PR diff base 1b124eb88a).

**Bug is real:** old logic kept any negligible coord whose NAME is a dim (`coord not in ds.dims` was False), so `time(y)` riding on an array that separately has a `time` dimension survived `drop_coordinates` even though it does not describe the `time` dimension. The new `_is_coord_var` check (`ds.coords[name].dims == (name,)`) is the correct criterion for "this coordinate is the variable for its own dimension".

**Regression proof (executed):** the 3 new tests pass at head; with the pre-fix `drop_coordinates` logic, `test_drop_coordinates_drops_mismatched_same_named_coordinate` FAILS (the genuine #3353 case — `time(y)` retained alongside a real `time` dim) while the other two pass pre-fix as expected (they lock pre-existing behavior: true dimension coord kept, non-dim `time(y)` without a `time` dim dropped). Honest, targeted coverage.

**No collateral:** full satpy/tests/compositor_tests/test_core.py → 36/36 at head. `NEGLIGIBLE_COORDS = ["time"]` substring matching is unchanged; the helper only tightens the retention rule for same-named coords on foreign dims.

**Not stale:** the PR diff-base core.py is byte-identical to current main's core.py except the removed GPL header (relicense #3451), so the change applies cleanly post-relicense; mergeable=MERGEABLE.

Repo CI already fully green on this head (CodeFactor, CodeScene, codecov, coveralls, RTD, pre-commit.ci, full test matrix). No human reviews yet — this is the first independent human-equivalent verification.

djhoese:
@gerritholl I think you're back at work right? Any chance you have time to review this since it was your original issue?

satpy/tests/compositor_tests/test_core.py
djhoese, line 1: I'm sorry it has been a while, but I have one more request. Could you combine these 3 new tests into a single test and use pytest's `parametrize`? The data creation is very similar and the checks are very similar so I'm hoping it only needs to be 2-4 arguments and maybe even improved checks (checking dims and coords in the result).

djhoese: Have you had any time to work on this @sylvesterkaczmarek ?

sylvesterkaczmarek: Done in `f407d99e`. I combined the three coordinate-filtering cases into one parametrized test and strengthened it to assert the result dimensions/shape plus the retained coordinate dimensions and values. The full `test_core.py` module passes 36/36, the repository pre-commit hooks are clean, and the commit is GitHub Verified.

djhoese (approved):
Thanks!

sylvesterkaczmarek:
Thanks for the review and merge, and for the suggestion to consolidate the regression cases. The parametrized version is cleaner and keeps the #3353 edge case explicit.

PR #3448, opened 2026-08-21
Fix Metop-SG-A1 to Metop-SGA1 naming scheme

djhoese:
Also add entries for D/E/F in the hopes that the spacecraft global attribute will be changed when the instrument is operational.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/core/metimage.py
+++ b/satpy/readers/core/metimage.py
@@ -19,9 +19,12 @@
 """Utilities for the management of METimage (VII) products."""
 
 PLATFORM_NAME_TRANSLATE = {
-    "SGA1": "Metop-SG-A1",
-    "SGA2": "Metop-SG-A2",
-    "SGA3": "Metop-SG-A3"
+    "SGA1": "Metop-SGA1",
+    "SGA2": "Metop-SGA2",
+    "SGA3": "Metop-SGA3",
+    "D": "Metop-D",
+    "E": "Metop-E",
+    "F": "Metop-F",
 }
 
 # PLANCK COEFFICIENTS FOR CALIBRATION AS DEFINED BY EUMETSAT
--- a/satpy/tests/reader_tests/test_metimage_base_nc.py
+++ b/satpy/tests/reader_tests/test_metimage_base_nc.py
@@ -168,7 +168,7 @@ class TestMETimageNCBaseFileHandler(unittest.TestCase):
                                               hour=17, minute=41, second=17, microsecond=555000)
         assert self.reader.end_time == expected_end_time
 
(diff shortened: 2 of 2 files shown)


adybbroe:
https://space.oscar.wmo.int/satellites/view/metop_sg_a1
Should we not follow Oscar here?
I think we have the Oscar style for Pyspectral, so wouldn't this break it for Pyspectral?

djhoese:
I've submitted a request to Oscar to rename them. EUMETSAT officials have complained to @ameraner and some of my SSEC colleagues to use the naming in this PR. Oscar is wrong as far as following the spacecraft organization's standards (EUMETSAT).

strandgren:
I wouldn't bee too optimistic about anything changing in the format specification and file attributes once the satellite becomes operational, I suspect it will stay as now to be honest. 

Also note that we have the `Metop-SG-B` satellites as well, carrying mainly instruments operating the microwave range. Since the two sets of satellites are launched in turns, I think the final operational names will become something like this:

```
Metop-SG-A1 -> Metop-D
Metop-SG-B1 -> Metop-E
Metop-SG-A2 -> Metop-F
Metop-SG-B2 -> Metop-G
Metop-SG-A3 -> Metop-H
Metop-SG-B3 -> Metop-I
```

but depending on the exact launch sequence and in which order the satellites are declared operational.

ameraner:
I don't think that the filename/content will change once the spacecraft is operational either, so we'll probably never hit the new dict entries, but instead we'll have to update the dict value (I know, painful). 

Honestly, I would propose to leave things as-is and not merge this, until the spacecraft goes operational. Otherwise we probably have to adapt pyspectral twice again. And Oscar, as of now, still agrees with satpy.

djhoese:
Thanks everyone for the feedback. This is exactly why I created this PR; to better my understanding.

> Honestly, I would propose to leave things as-is and not merge this, until the spacecraft goes operational. Otherwise we probably have to adapt pyspectral twice again. And Oscar, as of now, still agrees with satpy.

@ameraner @strandgren Let's say Oscar gets back to me with "yes, you're right, we'll change that right away". After it is changed, do you think that it'd still be better to wait for the operational change to SGA1/D to make changes here in satpy and in pyspectral?

strandgren:
I think staying aligned with Oscar is a good idea. So for now keep as is, but if OSCAR do change the name, we should change satpy and pyspectral as well.

And as long as WMO OSACR is "wrong" I don't think it's fair by EUMETSAT to complain about the naming.

PR #3449, opened 2026-08-24
Fix mws_l1b_nc reader not scaling and masking properly

djhoese:
Also lowercases sensor name to 'mws'.

Claude did almost all the coding on this one, but all code reviewed by me.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/etc/readers/mws_l1b_nc.yaml
+++ b/satpy/etc/readers/mws_l1b_nc.yaml
@@ -453,7 +453,7 @@ datasets:
         name: mws_lon
         resolution: 17000
         file_type: mws_l1b_nc
-        file_key: data/navigation/mws_lat
+        file_key: data/navigation/mws_lon
         standard_name: longitude
         units: degrees_east
 
--- a/satpy/readers/mws_l1b.py
+++ b/satpy/readers/mws_l1b.py
@@ -21,8 +21,8 @@ Documentation: https://www.eumetsat.int/media/44139
 import datetime as dt
 import logging
 
-import dask.array as da
 import numpy as np
+import xarray as xr
 from netCDF4 import default_fillvals
 
 from satpy.readers.core.netcdf import NetCDF4FileHandler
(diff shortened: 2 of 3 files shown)


Manny7717 (approved):
## Verified locally (head cc06db56, base ce376141)

**Regression-proven:** the PR's rewritten test file (with `PACKING_ATTRIBUTES` import removed for base compatibility) yields **4 failures on base** — `test_sensor` (MWS vs mws), `test_get_dataset_get_channeldata_counts`, `test_get_dataset_get_channeldata_bts`, `test_get_global_attributes` — and **24/24 pass on head**. All 4 base failures trace to the intended behavior changes, not unrelated breakage.

**Core bug confirmed in the old code:**
- `_get_dataset_channel` (old) applied only valid-range masking and NEVER applied `scale_factor`/`add_offset` — brightness temperatures came back as raw packed counts. New `mask_and_scale()` masks fill + valid range BEFORE scaling, casts to float32, applies scale/offset, and strips the packing attrs (they no longer describe the scaled array). Correct netCDF packing semantics.
- `_get_dataset_aux_data` (old) only scaled when BOTH `scale_factor` AND `add_offset` existed, and did `variable.attrs["missing_value"]` — a hard `KeyError` when the attr is absent. New path is robust to missing attrs (falls back to netCDF `default_fillvals` and ±inf range).
- YAML: `mws_lon` file_key pointed at `data/navigation/mws_lat` (copy-paste) — fixed to `mws_lon`. Real data bug.
- `sensor` now lowercases `/attr/instrument`, matching the reader config `sensors: [mws]` and satpy's case-sensitive `supports_sensor` set intersection.

**Edge-checked:** counts path keeps integer dtype + `_FillValue` attr (dtype-typed fill); float path NaN-fills; fill-vs-range precedence (FillValue → _FillValue → missing_value) sensible; `xr.set_options(keep_attrs=True)` preserves attrs through where/astype.

No blocking issues. Clean, well-tested fix.

djhoese:
@Manny7717 Who are you? I'm sorry if I missed some conversation about this. Did someone ask you for these AI reviews?

mraspaud (approved):
LGTM, thanks an lot for the simplifications!

PR #3450, opened 2026-08-25
Fix nucaps reader failing when H2O variable has missing valid_range

djhoese:
This was reported by @asierra to the Polar2Grid project. Two issues:

1. Valid range is missing from the H2O variable but the reader expected it to be there and I think in older file revisions it was always there. That's no longer true so the bracket `[]` get of the valid range had to be changed to default of `(None, None)`. The None handling was already implemented so it was a simple fix.
2. If a variable failed to load but pressure-based filtering was enabled then the reader still tried to filter variables that had failed causing a KeyError which propagated to the top causing all processing to stop.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese:
Merging. Simple fixes. Specific to a reader that I'm in charge of and I don't think many people use.

PR #3452, opened 2026-08-27
Update MetImage readers to produce scan-based dask chunks

djhoese:
I've done this type of change in other readers. The idea is to chunk the data aligned with instrument scans for a couple reasons:

1. Align dask chunks with on-disk file chunks (better loading performance)
2. Align chunks between resolutions/bands/data types. Makes comparisons and compositing have better performance (in theory) as they don't require rechunking.
3. Making chunks scan-based helps some other operations like the EWA resampling where it rechunks to scan boundaries.

This PR includes rewriting the l1b tests to use pytest.

Code was written by Claude AI but reviewed by me. I'm not a huge fan of the refactored tests, but it is also the right move I think since previously a single test was doing a lot of work. I'm tempted to rewrite the calibration logic to be a separate class with coefficients and things passed in so it is easier to test, but that seems like too much for this PR.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/satpy/readers/core/metimage.py
+++ b/satpy/readers/core/metimage.py
@@ -15,5 +15,9 @@ C2 = 1.4387863e+4  # [K·µm]
 TIE_POINTS_FACTOR = 8    # Sub-sampling factor of tie points wrt pixel points
 SCAN_ALT_TIE_POINTS = 4  # Number of tie points along the satellite track for each scan
 
+# Number of pixel rows per instrument scan. The tie points of a scan bound
+# ``SCAN_ALT_TIE_POINTS - 1`` intervals of ``TIE_POINTS_FACTOR`` pixels each.
+ROWS_PER_SCAN = (SCAN_ALT_TIE_POINTS - 1) * TIE_POINTS_FACTOR
+
 # MEAN EARTH RADIUS AS DEFINED BY IUGG
 MEAN_EARTH_RADIUS = 6371008.7714  # [m]
--- a/satpy/readers/core/metimage_nc.py
+++ b/satpy/readers/core/metimage_nc.py
@@ -5,13 +5,25 @@
 import datetime as dt
 import logging
 
+import numpy as np
 import xarray as xr
 from geotiepoints.viiinterpolator import tie_points_geo_interpolation, tie_points_interpolation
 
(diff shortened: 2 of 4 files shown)


djhoese:
CodeScene is mistaken. It is complaining about the `setUp` method but that doesn't exist anymore.

satpy/readers/core/metimage_nc.py
```diff
@@ -5,14 +5,26 @@
...
+    TIE_POINTS_FACTOR,
+)
 from satpy.readers.core.netcdf import NetCDF4FileHandler
+from satpy.utils import normalize_low_res_chunks
 
 logger = logging.getLogger(__name__)
 
+# Row/column dimension name pairs used by the various METimage products. The
+# equivalent renaming is done by ``METimageNCBaseFileHandler._standardize_dims``.
+PIXEL_DIMS = (("num_lines", "num_pixels"), ("num_points_alt", "num_points_act"))
```
strandgren: As far as I can see `num_points_act` is not a dimension actually used in the data (neither L1 nor L2)?

satpy/readers/core/metimage.py
```diff
@@ -15,5 +15,9 @@
 TIE_POINTS_FACTOR = 8    # Sub-sampling factor of tie points wrt pixel points
 SCAN_ALT_TIE_POINTS = 4  # Number of tie points along the satellite track for each scan
 
+# Number of pixel rows per instrument scan. The tie points of a scan bound
+# ``SCAN_ALT_TIE_POINTS - 1`` intervals of ``TIE_POINTS_FACTOR`` pixels each.
+ROWS_PER_SCAN = (SCAN_ALT_TIE_POINTS - 1) * TIE_POINTS_FACTOR
```
strandgren: Hm is this really what you want? There is the dimension `num_pixels_alt` which is the number of along-track pixels per scan (24).

you then have the number of scans per granule, `num_scans`, which is 35. The product of the two result in 840 which is equivalent to `num_lines`.

So to me it seems that the number you want here is simply the `num_pixels_alt`?

ameraner (approved):
LGTM, thanks for this!

PR #3454, opened 2026-08-28
fix(composites): mask the discarded side in day/night-only output without alpha

### LINKED ISSUE #3003: DayNightCompositor with `day_only` does not mask out night part if no alpha layer is present, defeating its purpose

Manny7717:
## What

Fixes #3003: with `day_night: day_only` (or `night_only`) and `include_alpha: False`, the discarded portion of the image was not masked — night pixels came out as 0 (the raw fill value) instead of NaN, so the result looked identical to not using the compositor at all.

## Root cause

`_mask_weights` correctly converts the discarded side of the weight map to NaN, but `_fill_weights` then refilled those NaNs with 0 (day_only) or 1 (night_only) before blending — undoing the mask. This also contradicted the documented contract: *"include_alpha=False means the output is a single-band image with undesired pixels being masked out (replaced with NaNs)"*.

## Change

Remove the refill (`_fill_weights`); the NaN mask survives into the blend, so the discarded side is NaN in the output.

## Tests

- New `test_day_only_sza_without_alpha_masks_night_pixels`: bottom row (SZA 94/100 → night) is all-NaN, top row keeps data. Fails on the previous implementation (verified) and passes with the fix.
- Updated `test_day_only_sza_without_alpha` and the two `night_only` no-alpha tests to the documented NaN behavior (their expected arrays encoded the buggy refill; comments note the geometry).
- `test_fill.py`: 24 passed; compositor suite 198 passed (1 pre-existing unrelated failure in test_viirs, fails on baseline too). ruff clean.

### DIFF
--- a/satpy/composites/fill.py
+++ b/satpy/composites/fill.py
@@ -213,15 +213,8 @@ class DayNightCompositor(GenericCompositor):
             weights: xr.DataArray,
             attrs: dict,
     ) -> list[xr.DataArray]:
-        weights = self._fill_weights(weights)
         data = self._merge_bands_with_weights(day_data, night_data, weights, attrs)
         return data
-
-    def _fill_weights(self, weights):
-        if not self.include_alpha:
-            fill = 1 if self.day_night == "night_only" else 0
-            weights = weights.where(~np.isnan(weights), fill)
-        return weights
 
     def _merge_bands_with_weights(self, day_data, night_data, weights, attrs):
         data = []
--- a/satpy/tests/compositor_tests/test_fill.py
+++ b/satpy/tests/compositor_tests/test_fill.py
@@ -104,7 +104,8 @@ class TestDayNightCompositor(unittest.TestCase):
             comp = DayNightCompositor(name="dn_test", day_night="night_only", include_alpha=False)
             res = comp((self.data_a, self.sza))
             res = res.compute()
-        expected = np.array([[0., 0.11042609], [0.6683502, 1.]], dtype=np.float32)
+        # The day pixel (SZA 80) is masked with NaN; night pixels keep their data
+        expected = np.array([[np.nan, 0.11042609], [0.6683502, 1.]], dtype=np.float32)
         assert res.dtype == np.float32
         np.testing.assert_allclose(res.values[0], expected)
         assert "A" not in res.bands
@@ -131,7 +132,8 @@ class TestDayNightCompositor(unittest.TestCase):
             comp = DayNightCompositor(name="dn_test", day_night="night_only", include_alpha=False)
             res = comp((self.data_b,))
             res = res.compute()
-        expected = np.array([[np.nan, 0.], [0., 0.]], dtype=np.float32)
+        # The test area is entirely in daylight at this time, so all pixels are masked
+        expected = np.array([[np.nan, np.nan], [np.nan, np.nan]], dtype=np.float32)
         assert res.dtype == np.float32
         np.testing.assert_allclose(res.values[0], expected)
         assert "A" not in res.bands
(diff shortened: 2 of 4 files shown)


pnuu:
@gerritholl this is fixing your issue, do you want to have a look?

djhoese:
@Manny7717 It looks like your CF writer PR (fmt) got mixed into this PR. Could you remove it please?

satpy/composites/fill.py
djhoese, line 85:
```suggestion
            include_alpha (bool): Whether the discarded portion of a "day_only"/"night_only"
                                  result is expressed through an alpha band (True, default) or by
                                  replacing the data with NaN (False). See the class docstring for
                                  the exact behaviour per input mode. For "day_night" this only
                                  controls whether pixels that are missing on one side but valid on
                                  the other become NaN (True) or 0 (False).
```

Claude suggested this edit to make this more clear along with my other comments.

djhoese, line 69:
```suggestion
    """A compositor that blends day data with night data.

    Using the ``day_night`` flag it is also possible to provide only a day product
    or only a night product and mask out the opposite portion of the image
    (night or day), either through an alpha band or by setting it to NaN.
    See the documentation below for more details.

    Single-side products ("day_only" / "night_only")
    ------------------------------------------------

    Only the first input is used. Its pixels fall into three zones based on
    solar zenith angle: kept (SZA below ``lim_low`` for day, above
    ``lim_high`` for night), blending (between the limits), and discarded.

    With ``include_alpha=True``:

    * L or RGB input: an alpha band is appended (LA / RGBA output). The
      data bands are left untouched in every zone; alpha is 1 in the kept
      zone, the SZA weight in the blending zone and 0 in the discarded zone.
    * LA or RGBA input: no band is added. The existing alpha band is
      multiplied by the SZA weight, so it is unchanged in the kept zone,
      faded in the blending zone and 0 in the discarded zone.

    With ``include_alpha=False``:

    * No alpha band is added and the output mode equals the input mode
      (L -> L, RGB -> RGB, RGBA -> RGBA).
    * Every band, including an existing alpha band, is multiplied by the
      SZA weight: unchanged in the kept zone, fading towards 0 through the
      blending zone, and NaN in the discarded zone.

    In both cases pixels that are NaN in the input (for example space
    pixels, or pixels where the solar zenith angle cannot be computed)
    stay NaN.

    The NaN output is what writers see: without a ``fill_value`` most
    writers will turn it into a transparent pixel; pass ``fill_value=0``
    to most writers to get an opaque black background instead.
    """
```

djhoese: Claude also suggested editing `composites.rst` in the docs line 167-176 with:

```rst
By default, the image under ``day_only`` or ``night_only`` flag will come out
with an alpha band to display its transparency. It could be changed by
setting ``include_alpha`` to False if there's no need for that alpha band.
In that case the opposite portion is set to NaN in every band instead;
most writers turn NaN into a transparent pixel unless a ``fill_value`` is
given, so pass ``fill_value=0`` when saving to get a single-band image with
a black background. See :class:`~satpy.composites.fill.DayNightCompositor`
for the exact behaviour per input mode (L, RGB, LA, RGBA). In the case
below, the image shows its day portion and day/night transition with the
night portion set to NaN::

    >>> from satpy.composites.fill import DayNightCompositor
    >>> compositor = DayNightCompositor("dnc", lim_low=85., lim_high=88., day_night="day_only", include_alpha=False)
    >>> composite = compositor([local_scene['true_color'])
```

But I can't edit that file without making a commit on this PR and there is still the CF commit here.

PR #3458, opened 2026-09-04
Change Scene iteration to return DataIDs instead of DataArrays

djhoese:
This PR prepares the `Scene.__iter__` method to act more like a dictionary/xarray Dataset/xarray DataTree. Currently iterating over a `Scene` (ex. `for x in scn:`) yield the underlying `DataArray`s rather than the `DataID` keys. This does not match the dict-like container nature of the `Scene`. This was mentioned back in #1419 but is not completely closed by this.

This PR adds a config option to control what behavior Scene iteration should have and includes a deprecation warning. If `None`, the default, then the existing behavior is used but a warning is logged. If the user uses `False` then you get existing behavior but no warning. If `True` then you get the new Satpy 1.0 behavior of DataIDs being yielded.

Claude was used for most of this work.

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/doc/source/config.rst
+++ b/doc/source/config.rst
@@ -246,6 +246,45 @@ will download and cache any necessary data files to :ref:`data_dir_setting`
 when needed. If ``False`` then pre-downloaded files will be used, but any
 other files will not be downloaded or checked for validity.
 
+.. _scene_iter_keys_setting:
+
+Scene Iteration Keys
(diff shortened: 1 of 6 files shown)


mraspaud (approved):
LGTM. Should we have a marker to know that we need to change this in 1.0?

djhoese:
Yes! I'll try to add that Monday maybe.

PR #3461, opened 2026-09-09
Revise NDVI green band correction based on quantitative tuning

strandgren:
This PR implements a slight revision of the NDVI hybrid green correction compositor, used for the FCI `true_color` composite and also available for AHI and AMI recipes.

The correction was previously tuned qualitatively during FCI commissioning. We have now performed a quantitative tuning using collocated FCI and OLCI observations, fitting the hue of the corrected FCI green against the OLCI 0.56 µm green band. Using hue reduces dependencies on illumination and BRDF effects due to the different viewing geometries. The resulting fit is a third-order polynomial mapping NDVI to the NIR blend factor.

The differences compared to `main` are generally small, but desert surfaces have a more natural colour (the slight pinkish tone stemming from the FCI native green band signal is reduced) and some vegetation becomes slightly brighter. More importantly, the correction is now quantitatively derived rather than based on qualitative tuning and eyeballing.

A comparison between the two can be seen below:

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

We are currently preparing a paper describing this work. We would like to refer to the open-source availability through the Satpy implementation in the paper. It would therefore be preferable to have this implementation merged into Satpy, and ideally included in a release, before the paper is submitted. In the spirit of promoting Satpy in the paper, I hope this approach is fine. Typically, the method would first be published and only then implemented in Satpy, so we are somewhat doing this in the opposite order here.

This work has been done together with @TristanCantatGaudin.


strandgren:
Perhaps some more context regarding the code changes: the previous method was using a linear mapping between NDVI and NIR blend fraction, but by setting the `strength` parameter the mapping could also be made non-linear. In the end this worked fine, but there were many tunable parameters (NDVI limits, blend fraction limits and the strength). 

With the new implementation proposed in this PR, we only have a single third-order polynomial with hard NDVI limits, hence the NDVI limits and the regression coefficients are defined in the `__init__` and cannot be modified through yaml configuration. Hence, the previous tunable parameters have been deprecated and for now a warning is issued if a user uses one of these deprecated arguments.

satpy/composites/spectral.py
```diff
@@ -118,34 +117,37 @@ class NDVIHybridGreen(SpectralBlender):
...
+        deprecated_args = {"ndvi_min", "ndvi_max", "limits", "strength"}
+        for name in deprecated_args & kwargs.keys():
+            warnings.warn(
+                f"'{name}' has been deprecated for the NDVIHybridGreen Compositor and will be ignored.",
+                UserWarning,
+                stacklevel=2,
+            )
+
+        for name in deprecated_args:
+            kwargs.pop(name, None)
```
djhoese: Why isn't this pop part of the above loop?

strandgren: good catch, but no good answer what it was like that.. fixed now

satpy/composites/spectral.py
```diff
@@ -118,34 +117,35 @@ class NDVIHybridGreen(SpectralBlender):
...
+        for name in deprecated_args & kwargs.keys():
+            warnings.warn(
+                f"'{name}' has been deprecated for the NDVIHybridGreen Compositor and will be ignored.",
+                UserWarning,
+                stacklevel=2,
+            )
+            kwargs.pop(name)
+
+        self.ndvi_min = 0.1
+        self.ndvi_max = 0.9
```
pnuu, line 139: If they are deprecated and ignored, do they need to be attributes here?

satpy/composites/spectral.py
```diff
@@ -118,34 +117,35 @@ class NDVIHybridGreen(SpectralBlender):
...
+        self.a = 0.3940
+        self.b = -1.3701
+        self.c = 2.0206
+        self.d = -1.0349
```
pnuu: Any more descriptive names for these?

pnuu: Ok, possibly not looking at the equation they are used in. Or maybe `factor_0` ... `factor_3` showing the order of the weight (from 0 being constant to 3 being cubic).

pnuu (commented):
Some comments in-line.

satpy/composites/spectral.py
TristanCantatGaudin: We could also have them as a single list, where the variable name can be as explicit as we like:

`polynomial_coefs_descending = [
    -1.0349,  
     2.0206,  
    -1.3701,  
     0.3940
]`

satpy/composites/spectral.py
TristanCantatGaudin: They are still used, but whatever the user provides will be ignored and overwritten by these limits.

satpy/composites/spectral.py
strandgren: I added them as a list following Tristan's suggestions and then later unpack them as `coef_3`, `coef_2`, `coef_1` and `coef_0` making it more clear which order of the polynomial they refer to.

satpy/composites/spectral.py
strandgren: indeed, before they were set to the corresponding input arguments of `init`, but now we no longer allow that and hard-code them instead (since these are the limits of the polynomial regression fit)

PR #3462, opened 2026-09-10
Add reader for GMS-4

### LINKED ISSUE #3201: GMS-4 VISSR reader (revisited).
## Feature Request

Depends on how you look at it, there's only so much documentation on this subject, I'm revisiting older satellite data, and I had noticed that you had a GMS-5 VISSR reader, wonderful. I have since realized why there is not a GMS-4 VISSR and earlier reader, given how different the formats are. I noticed there was an [old post](https://github.com/pytroll/satpy/issues/486) made by CyanideCN a few years back, and it did also mention the older formats aside from GMS-5, I don't know if you guys had continued on it or forgot about it. Being limited to only GMS-5 VISSR data is fine but like I said, limiting. I would like to be able to utilize older data to produce updated images of say tropical storms or events and anomalies back then.

I would like to have the option to look back at older VISSR data from GMS-4 and earlier. 

None that I really know of.

**Additional context**
Data source nowadays; ftp://gms.cr.chiba-u.ac.jp

Updated links from CyanideCN's post; 

GMS-4 VISSR Format: https://www.data.jma.go.jp/mscweb/en/operation/fig/VISSR_FORMAT_GMS-4.pdf

GMS User Guide: https://www.data.jma.go.jp/mscweb/en/operation/fig/GMS_Users_Guide_3rd_Edition_Rev1.pdf

KNLargadas:
This PR aims to add support for GMS-1/2/3/4 satellite data. Code was created by Claude with guidance from me.

 - [x] Closes #3201 
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/satpy/readers/gms/gms1_4_vissr_l1b.py
+++ b/satpy/readers/gms/gms1_4_vissr_l1b.py
@@ -1,0 +1,576 @@
+"""Reader for GMS-1-4 VISSR Level 1B data.
+
+Introduction
+------------
+The ``gms1_4_vissr_l1b`` reader can decode, navigate and calibrate Level 1B data
+from the Visible and Infrared Spin Scan Radiometer (VISSR) in `VISSR
+archive format`. Corresponding platforms are GMS-1 to GMS-4
+(Japanese Geostationary Meteorological Satellite).
+
+Unlike GMS-5, GMS-1-4 VISSR only has two channels, each stored in a separate file:
+
+.. code-block:: none
+
+    VS901110.Z23
+    IR901110.Z23
+
+This is how to read them with Satpy:
+
+.. code-block:: python
+
+    from satpy import Scene
+    import glob
+
+    filenames = glob.glob(""/data/VS*")
+    scene = Scene(filenames, reader="gms1-4-vissr_l1b")
+    scene.load(["VIS"])
+
+
+References:
+~~~~~~~~~~~
+
+Details about platform, instrument and data format can be found in the
+following references:
+
+    - `VISSR Format Description`_
+    - `GMS User Guide`_
+
+.. _VISSR Format Description:
+    https://www.data.jma.go.jp/mscweb/en/operation/fig/VISSR_FORMAT_GMS-4.pdf
+.. _GMS User Guide:
+    https://www.data.jma.go.jp/mscweb/en/operation/fig/GMS_Users_Guide_3rd_Edition_Rev1.pdf
+
+
+Calibration
+-----------
+
+Sensor counts are calibrated by looking up reflectance/temperature values in the
+calibration tables included in each file. See section 2.2 in the VISSR user
+guide.
+
+
+Navigation
+----------
+
+VISSR images are oversampled and not rectified.
+
+
+Oversampling
+~~~~~~~~~~~~
+VISSR oversamples the viewed scene in E-W direction by a factor of ~2:
+IR/VIS pixels are 1.46/2.92 urad on a side, but the instrument samples every
+4.8/2.4 urad in E-W direction. That means pixels are actually overlapping on
+the ground.
+
+This cannot be represented by a pyresample area definition, so each dataset
+is accompanied by 2-dimensional longitude and latitude coordinates. For
+resampling purpose a full disc area definition with uniform sampling is provided
+via
+
+.. code-block:: python
+
+    scene[dataset].attrs["area_def_uniform_sampling"]
+
+
+Rectification
+~~~~~~~~~~~~~
+
+VISSR images are not rectified. That means lon/lat coordinates are different
+
+1) for all channels of the same repeat cycle, even if their spatial resolution
+   is identical
+2) for different repeat cycles, even if the channel is identical
+
+However, the above area definition is using the nominal subsatellite point as
+projection center. As this rarely changes, the area definition is pretty
+constant.
+
+
+Performance
+~~~~~~~~~~~
+
+Navigation of VISSR images is computationally expensive, because for each pixel
+the view vector of the (rotating) instrument needs to be intersected with the
+earth, including interpolation of attitude and orbit prediction. For IR channels
+this takes about 10 seconds, for VIS channels about 160 seconds.
+
+
+Space Pixels
+------------
+
+VISSR produces data for pixels outside the Earth disk (i.e. atmospheric limb or
+deep space pixels). By default, these pixels are masked out as they contain
+data of limited or no value, but some applications do require these pixels.
+To turn off masking, set ``mask_space=False`` upon scene creation:
+
+.. code-block:: python
+
+    import satpy
+    import glob
+
+    filenames = glob.glob("VS*")
+    scene = satpy.Scene(filenames,
+                        reader="gms5-vissr_l1b",
+                        reader_kwargs={"mask_space": False})
+    scene.load(["VIS"])
+
+
+Metadata
+--------
+
+Dataset attributes include metadata such as time and orbital parameters,
+see :ref:`dataset_metadata`.
+
+Partial Scans
+-------------
+
+On demand a special Typhoon schedule would be activated between
+03:00 and 05:00 UTC.
+"""
+
+import numpy as np
+import dask.array as da
+import xarray as xr
+
+from satpy.readers.core.file_handlers import BaseFileHandler
+from satpy.readers.hrit_jma import mjd2datetime64
+from satpy.utils import datetime64_to_pydatetime
+import satpy.readers.core._geos_area as geos_area
+
+import satpy.readers.gms.gms5_vissr_navigation as nav_shared
+import satpy.readers.gms.gms1_4_vissr_format as fmt
+
+
+def _mjd_to_datetime(mjd):
+    return datetime64_to_pydatetime(mjd2datetime64(np.array(mjd)))
+
+
+class GmsVissrFileHandler(BaseFileHandler):
+    """File handler for GMS-1..4 native VISSR archive files."""
+
+    def __init__(self, filename, filename_info, filetype_info):
+        super().__init__(filename, filename_info, filetype_info)
+        self._l1b = GmsVissrL1bFile(filename)
+
+    @property
+    def start_time(self):
+        return _mjd_to_datetime(float(self._l1b.coord["scheduled_observation_time"]))
+
+    @property
+    def end_time(self):
+        if self._l1b.scan_times.size:
+            last_valid = self._l1b.scan_times[np.isfinite(self._l1b.scan_times)]
+            if last_valid.size:
+                return _mjd_to_datetime(float(last_valid.max()))
+        return self.start_time
+
+    @property
+    def sensor_names(self):
+        return {"VISSR"}
+
+    def combine_info(self, all_infos):
+        if len(all_infos) == 1:
+            combined_info = dict(all_infos[0])
+            new_dict = self._combine(all_infos, min, "start_orbit")
+            new_dict.update(self._combine(all_infos, max, "end_orbit"))
+            new_dict.update(self._combine_orbital_parameters(all_infos))
+            combined_info.update(new_dict)
+            return combined_info
+        return super().combine_info(all_infos)
+
+    def get_dataset(self, dataset_id, ds_info):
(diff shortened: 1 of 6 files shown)


satpy/readers/gms/gms1_4_vissr_l1b.py
```diff
@@ -0,0 +1,614 @@
...
+        return res
+
+    def _calibrate(self, counts):
+        lut = self._calib_table
+        mask = self._mask
+
+        def _lookup(block, lut=lut, mask=mask):
+            return lut[block.astype(np.int64) & mask]
+
+        return counts.map_blocks(_lookup, dtype=np.float32)
```
djhoese: A couple things:

1. Is `counts` an xarray `DataArray`? If so then it may be useful to see if `_lookup` is being called by dask or xarray with dummy inputs to determine the type of array that will be returned. In a pure dask `dask.array.map_blocks` call you normally should pass `meta=np.array((), dtype=dtype)` to get around this. I'm not sure what xarray's version of `map_blocks` does here.
2. The `_lookup` function chould be moved outside the class. Dask has trouble serializing local functions that it can't access from the global namespace. This doesn't normally come up with Satpy's typical dask threaded scheduler usage, but we try to avoid it in case people use dask's distributed scheduler. If your Claude thinks this should be added to AGENTS.md then I think that's a good idea.
3. How do you know the dtype is 32-bit float?

satpy/readers/gms/gms1_4_vissr_l1b.py
```diff
@@ -0,0 +1,614 @@
...
+        estimator = AreaDefEstimator(self._l1b, self._platform_name())
+        return estimator.get_area_def_uniform_sampling(dataset_id)
+
+    def _platform_name(self):
+        try:
+            name = bytes(self._l1b.mode["satellite_name"]).split(b"\x00")[0].decode("ascii", "replace").strip()
+            if name:
+                return name
+        except Exception:
+            pass
```
djhoese: This use of `except Exception` is frowned upon. If there are specific exceptions that are being guarded against then I'd prefer if they were listed explicitly. As for the `if name:`, is that something that can actually happen with the data files you've tried reading?

satpy/readers/gms/gms1_4_vissr_l1b.py
```diff
@@ -0,0 +1,614 @@
...
+        raise NotImplementedError(
+            "No fixed-grid area definition for GMS 1-4 archive data -- "
+            "confirmed E-W oversampling (~1.46x on VIS, ~2.92x on IR) "
+            "means a clean AreaDefinition can't represent this data "
+            "accurately. Use the per-pixel longitude/latitude "
+            "coordinates instead (SwathDefinition, built automatically "
+            "by Satpy from get_dataset()'s attached coords), or the "
+            "dataset's 'area_def_uniform_sampling' attribute for "
+            "resampling purposes."
+        )
```
djhoese: I think this is swallowed by the base reader class so there isn't much use in a long error message here (a user will never see it) and the base class already provides a NotImplementedError.

satpy/readers/gms/gms1_4_vissr_l1b.py
```diff
@@ -0,0 +1,614 @@
...
+        data_array.coords["longitude"].attrs["resolution"] = nadir_resolution
+        data_array.coords["latitude"].attrs["resolution"] = nadir_resolution
+
+        from pyresample.geometry import SwathDefinition
+        swath_def = SwathDefinition(
+            lons=data_array.coords["longitude"],
+            lats=data_array.coords["latitude"],
+        )
+        swath_def.name = f"{self._l1b.channel}_swath"
+        data_array.attrs["area"] = swath_def
```
djhoese: All of this is meant to be handled by the reader class by the datasets in the YAML defining the longitude and latitude coordinate datasets that correspond to their geolocation. This should not be done here or by a file handler.

satpy/readers/gms/gms1_4_vissr_l1b.py
```diff
@@ -0,0 +1,614 @@
...
+        return {"VISSR"}
+
+    def combine_info(self, all_infos):
+        if len(all_infos) == 1:
+            combined_info = dict(all_infos[0])
+            new_dict = self._combine(all_infos, min, "start_orbit")
+            new_dict.update(self._combine(all_infos, max, "end_orbit"))
+            new_dict.update(self._combine_orbital_parameters(all_infos))
+            combined_info.update(new_dict)
+            return combined_info
```
djhoese: If the length is 1, why not just return `all_infos[0]`?

djhoese (changes requested):
I made some initial comments. I'm a little confused by some of the approaches and that Claude made them. I would have thought our AGENTS.md file would have stopped it, but also it looks like you didn't have `pre-commit` installed which heavily limited the quality of the code Claude produced. The `pre-commit.ci` failures in CI will need to be fixed.

On a higher-level scale though, I'm wondering if this reader should maybe just be called `gms4_vissr_l1b` rather than `1-4`. I find the `1-4` confusing as first glance and I think anyone familiar with the data would know that if something supports GMS 4 that it also supports GMS 1 through 3 as well. What do you think?

satpy/readers/gms/gms1_4_vissr_l1b.py
KNLargadas: For 1 and 2, ``counts`` is a ``dask.array.Array``, not a xarray ``DataArray``. This has been fixed.
For 3, the format spec for VIS_CALIBRATION_TABLE and IR_CALIBRATION is defined as R4 (>f4), so the LUT lookup produces float32 without any casting.

satpy/readers/gms/gms1_4_vissr_l1b.py
KNLargadas: The ``except Exception`` has been removed. For the ``if name``, the empty-name is possible since we found spin_rate unpopulated on pre-1987 GMS files, but I haven't seen it actually happening.

satpy/readers/gms/gms1_4_vissr_l1b.py
KNLargadas: This has been removed

satpy/readers/gms/gms1_4_vissr_l1b.py
KNLargadas: This has been resolved

satpy/readers/gms/gms1_4_vissr_l1b.py
KNLargadas: Also resolved.

satpy/readers/gms/gms4_vissr_l1b.py
```diff
@@ -0,0 +1,604 @@
...
+                                    params["orbit_prediction_2"]["dtype"])
+
+        cal_key = "ir_calibration" if self.channel == fmt.IR_CHANNEL else "vis_calibration"
+        self.calibration = _read_struct(self._raw, params[cal_key]["offset"],
+                                         params[cal_key]["dtype"])
+
+        self._line_chunks = line_chunks
+        self._parse_image_data(spec)
+
+    # -----------------------------------------------------------------
```
djhoese, line 345: Is there a reason for these horizontal bars? Can these be removed?

satpy/readers/gms/gms4_vissr_l1b.py
```diff
@@ -0,0 +1,604 @@
...
+        """
+        if calibration == "counts":
+            return counts
+        res = self._calibrate(counts)
+        res = self._postproc(res, calibration)
+        return res
+
+    def _calibrate(self, counts):
+        lookup = functools.partial(_lookup_calibration_value, lut=self._calib_table, mask=self._mask)
+        return counts.map_blocks(lookup, dtype=np.float32, meta=np.array((), dtype=np.float32))
```
djhoese: You may want to check with some print statements in `_lookup_calibration_value`, but if `self._calib_table` is a dask array then it may be getting passed as a dask array instead of a numpy array which is what is desired. I also am not sure you need a partial here. I think you can pass they kwargs directly to `map_blocks` and any kwargs it doesn't know about will be passed on to the caller. Or maybe they need to before the dask kwargs... 🤔

satpy/readers/gms/gms4_vissr_l1b.py
```diff
@@ -0,0 +1,604 @@
...
+
+    def _postproc(self, res, calibration):
+        if calibration == "reflectance":
+            res = self._convert_to_percent(res)
+        return res
+
+    def _convert_to_percent(self, res):
+        return res * 100
```
djhoese:
```suggestion
    def _postproc(self, res, calibration):
        if calibration == "reflectance":
            # convert to percent
            return res * 100
        return res
```

satpy/readers/gms/gms4_vissr_l1b.py
KNLargadas: I forgot to remove those, that was where the original navigation module for this reader (before connecting it to the existing gms5 navigation) was placed. Fixing it

satpy/readers/gms/gms4_vissr_l1b.py
KNLargadas: ``self._calib_table`` was getting passed as a plain numpy array. Simplified it and removing functools

djhoese (commented):
I'm sorry, I thought I mentioned this earlier, could you please add a test module for this? In modern Satpy testing we've been trying to have tests create fake files on disk and then read them rather than monkeypatching the file handler code.

djhoese:
Ah and the tests are why the CI is failing too. The old gsm5 tests are looking for the renamed geolocation module.

satpy/tests/reader_tests/gms/test_gms4_vissr_l1b.py
```diff
@@ -0,0 +1,398 @@
...
+    Regression coverage for the "Bumpy Road" refactor: the original
+    implementation was a per-line Python loop with two nested ifs; the
+    vectorized replacement was verified against it over 200 randomized
+    trials during development, including the asymmetric-clamping edge
+    cases below (w is only clamped from below, e only from above).
+    """
+
+    def test_basic_mask(self):
+        """Test a plain, well-formed earth mask."""
+        obj = vissr.GmsVissrL1bFile.__new__(vissr.GmsVissrL1bFile)
```
djhoese: I'm not sure I like using `__new__` directly. It looks like you have fake files in the other tests, can those be used here? Additionally, the tests in this class look very similar, could they be combined using pytest's `parametrize`?

satpy/etc/readers/gms4-vissr_l1b.yaml
```diff
@@ -0,0 +1,56 @@
...
+    sensor: gms4-vissr
+    wavelength: [0.55, 0.73, 0.9]
+    resolution: 1250
+    calibration:
+      counts:
+        standard_name: counts
+        units: 1
+      reflectance:
+        standard_name: toa_bidirectional_reflectance
+        units: "%"
```
djhoese: @KNLargadas Do you know if the reflectance data coming out of this reader has `/cos(SZA)` already applied to it? Satpy is trying to move towards distinguishing between visible data calibrated to be a reflectance before `/ cos(solar zenith angle)` is applied and after. You can see this work here:

https://github.com/pytroll/satpy/pull/3292

If the data for this reader does not have `/ cos(SZA)` already applied then could you change the name of this calibration level to `unnormalized_reflectance` and change the standard_name to `product_of_cosine_solar_zenith_angle_and_toa_bidirectional_reflectance`? We hope to merge the above pull request soon so anything that doesn't work in this reader should work soon after you make this change.

djhoese (changes requested):
A couple more changes.

satpy/etc/readers/gms4-vissr_l1b.yaml
KNLargadas: It does not have ``/ cos(SZA)`` already applied. This is resolved

satpy/tests/reader_tests/gms/test_gms4_vissr_l1b.py
KNLargadas: Yes they can be combined. Added that into the test reader

KNLargadas:
Hi, would like to know if it's good or if it needs to be improved still. Thanks!

PR #3464, opened 2026-09-11
Recipes for PPS-CBH and ADCOT

ninahakansson:
Add recipes for NWC SAF PPS cloud base height and all day optical thickness .

 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/satpy/etc/composites/visir.yaml
+++ b/satpy/etc/composites/visir.yaml
@@ -321,6 +321,20 @@ composites:
     - cma
     standard_name: pps_cma
 
+  pps_adcot:
+    compositor: !!python/name:satpy.composites.cloud_products.CloudCompositorCommonMask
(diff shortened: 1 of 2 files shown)


satpy/etc/composites/visir.yaml
```diff
@@ -321,6 +321,20 @@ composites:
     - cma
     standard_name: pps_cma
 
+  pps_adcot:
+    compositor: !!python/name:satpy.composites.cloud_products.CloudCompositorCommonMask
+    prerequisites:
+    - adcot
+    - adcot
```
pnuu, line 328: should this be `cma` or does `adcot` work similarly?

pnuu (commented):
One question inline. Also example images would be nice for reference.

PR #3466, opened 2026-09-11
Fix METimage L1 orthorectification

ameraner:
This PR fixes a bug in the applied orthorectification formula, where in the case of the longitude correction, a factor of the cosine of the latitude was missing. 

See according snippet from the format spec:
image

On a side, it also updates the filepatterns to allow different processing facility entries.

 - [x] Tests added 
 - [x] Fully documented


satpy/etc/readers/metimage_l1b_nc.yaml
```diff
@@ -12,10 +12,8 @@ reader:
...
-      'W_XX-EUMETSAT-Darmstadt,SAT,{spacecraft_name:s}-VII-1B-RAD_C_EUMT_{creation_time:%Y%m%d%H%M%S}_{mission_type:s}_{environment:s}_{sensing_start_time:%Y%m%d%H%M%S}_{sensing_end_time:%Y%m%d%H%M%S}_{disposition_mode:s}_{processing_mode:s}____.nc',
-      'W_XX-EUMETSAT-Darmstadt,SAT,{spacecraft_name:s}-VII-1B-RAD_C_EUMT_{creation_time:%Y%m%d%H%M%S}_{mission_type:s}_{environment:s}_{sensing_start_time:%Y%m%d%H%M%S}_{sensing_end_time:%Y%m%d%H%M%S}_{disposition_mode:s}_{processing_mode:s}____.nc.bz2']
+    file_patterns: ['W_XX-EUMETSAT-Darmstadt,SAT,{spacecraft_name:s}-VII-1B-RAD_C_EUMT_{creation_time:%Y%m%d%H%M%S}_{mission_type:s}_{environment:s}_{sensing_start_time:%Y%m%d%H%M%S}_{sensing_end_time:%Y%m%d%H%M%S}_{disposition_mode:s}_{processing_mode:s}_{processing_facility:s}__.nc',
+                    'W_XX-EUMETSAT-Darmstadt,SAT,{spacecraft_name:s}-VII-1B-RAD_C_EUMT_{creation_time:%Y%m%d%H%M%S}_{mission_type:s}_{environment:s}_{sensing_start_time:%Y%m%d%H%M%S}_{sensing_end_time:%Y%m%d%H%M%S}_{disposition_mode:s}_{processing_mode:s}_{processing_facility:s}__.nc.bz2',]
```
djhoese, line 16: I was thinking this on the other metimage PR, can the compression be added to the end? Something like `.nc{compression_ext}`?

djhoese: Then it is just one pattern.

djhoese:
Do you have a resampled image of with and without orthorectification? Any idea on the added time/memory when this is used?

ameraner:
The gif below shows:
- Google Maps high-res orthorectified image
- METimage without orthorectification
- METimage with the old orthorectification
- METimage with the current PR
orthorectification_example2

satpy/etc/readers/metimage_l1b_nc.yaml
ameraner: Hmm yes we could do that, but are we sure we support any type of compression? Looking at unzip_file we only support bz2 no?

djhoese: Yes, true, I was mostly thinking about simplifying the file patterns, not supporting additional compression types. I'm fine leaving it as is.

ameraner:
On the execution time, I've done some timing tests and I can't really see a difference.

satpy/etc/readers/metimage_l1b_nc.yaml
ameraner: Ok, then I'd rather leave it explicit like this I'd say, if we start having futher compression types we can reconsider.

djhoese (approved):

djhoese:
On slack we got approval by @simon-sat and myself for making orthorectification the default. Would you mind making that change here? @strandgren @mraspaud @pnuu or anyone else have an opinion?

pnuu:
Ok by me. I think that's the default also on VIIRS (compact- and the normal SDRs if all geolocation are given).

strandgren:
Fine with me as well, as long as it's well documented how to disable it if desired

PR #3469, opened 2026-09-14
Check optional datasets in NIREmissivePartFromReflectance

### LINKED ISSUE #2460: using modifier nir_reflectance gives error
Loading SLSTR band S7 with modifier nir_reflectance fails with error:
ValueError: operands could not be broadcast together with shapes (2400, 3000) (1200, 1500)
**To Reproduce**

```python
import satpy
satpy.config.set(config_path=['/home/hp/Documents/mySatpyConfigs/'])

from satpy import Scene
from satpy.composites import GenericCompositor
import pyresample
from IPython.display import Image
from pathlib import Path
from glob import glob

dataPath=str(Path().home())+'/DATA/s3/x'
files=glob(dataPath+'/*/*')

scn=Scene(reader='slstr_l1b',filenames=files)
area='pyrenees'
cmp='s7refl'
code=' SLSTR S7refl'
scn.load([cmp])
```

**Expected behavior**

code to load/display the reflective part of the band

**Actual results**
```
Cell In[2], line 20
     18 cmp='s7refl'
     19 code=' SLSTR S7refl'
---> 20 scn.load([cmp])
     21 scn=scn.resample(area)
     23 prodPath=str(Path().home())+'/y/'

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/satpy/scene.py:1360, in Scene.load(self, wishlist, calibration, resolution, polarization, level, modifiers, generate, unload, **kwargs)
   1358 self._read_datasets_from_storage(**kwargs)
   1359 if generate:
-> 1360     self.generate_possible_composites(unload)

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/satpy/scene.py:1423, in Scene.generate_possible_composites(self, unload)
   1416 def generate_possible_composites(self, unload):
   1417     """See which composites can be generated and generate them.
   1418 
   1419     Args:
   1420         unload (bool): if the dependencies of the composites
   1421                        should be unloaded after successful generation.
   1422     """
-> 1423     keepables = self._generate_composites_from_loaded_datasets()
   1425     if self.missing_datasets:
   1426         self._remove_failed_datasets(keepables)

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/satpy/scene.py:1442, in Scene._generate_composites_from_loaded_datasets(self)
   1439 trunk_nodes = self._dependency_tree.trunk(limit_nodes_to=self.missing_datasets,
   1440                                           limit_children_to=self._datasets.keys())
   1441 needed_comp_nodes = set(self._filter_loaded_datasets_from_trunk_nodes(trunk_nodes))
-> 1442 return self._generate_composites_nodes_from_loaded_datasets(needed_comp_nodes)

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/satpy/scene.py:1448, in Scene._generate_composites_nodes_from_loaded_datasets(self, compositor_nodes)
   1446 keepables = set()
   1447 for node in compositor_nodes:
-> 1448     self._generate_composite(node, keepables)
   1449 return keepables

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/satpy/scene.py:1506, in Scene._generate_composite(self, comp_node, keepables)
   1503     return
   1505 try:
-> 1506     composite = compositor(prereq_datasets,
   1507                            optional_datasets=optional_datasets,
   1508                            **comp_node.name.to_dict())
   1509     cid = DataID.new_id_from_dataarray(composite)
   1510     self._datasets[cid] = composite

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/satpy/modifiers/spectral.py:70, in NIRReflectance.__call__(self, projectables, optional_datasets, **info)
     65 """Get the reflectance part of an NIR channel.
     66 
     67 Not supposed to be used for wavelength outside [3, 4] µm.
     68 """
     69 projectables = self.match_data_arrays(projectables)
---> 70 return self._get_reflectance_as_dataarray(projectables, optional_datasets)

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/satpy/modifiers/spectral.py:81, in NIRReflectance._get_reflectance_as_dataarray(self, projectables, optional_datasets)
     78 da_sun_zenith = self._get_sun_zenith_from_provided_data(projectables, optional_datasets)
     80 logger.info('Getting reflective part of %s', _nir.attrs['name'])
---> 81 reflectance = self._get_reflectance_as_dask(da_nir, da_tb11, da_tb13_4, da_sun_zenith, _nir.attrs)
     83 proj = self._create_modified_dataarray(reflectance, base_dataarray=_nir)
     84 proj.attrs['units'] = '%'

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/satpy/modifiers/spectral.py:125, in NIRReflectance._get_reflectance_as_dask(self, da_nir, da_tb11, da_tb13_4, da_sun_zenith, metadata)
    123 """Calculate 3.x reflectance in % with pyspectral from dask arrays."""
    124 reflectance_3x_calculator = self._init_reflectance_calculator(metadata)
--> 125 return reflectance_3x_calculator.reflectance_from_tbs(da_sun_zenith, da_nir, da_tb11, tb_ir_co2=da_tb13_4) * 100

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/pyspectral/near_infrared_reflectance.py:275, in Calculator.reflectance_from_tbs(self, sun_zenith, tb_near_ir, tb_thermal, **kwargs)
    273 corrected_thermal_emiss_one = thermal_emiss_one * self._rad3x_correction
    274 nomin = l_nir - corrected_thermal_emiss_one
--> 275 denom = self._solar_radiance - corrected_thermal_emiss_one
    276 data = nomin / denom
    277 mask = denom < EPSILON

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/dask/array/core.py:223, in check_if_handled_given_other.<locals>.wrapper(self, other)
    221     return NotImplemented
    222 else:
--> 223     return f(self, other)

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/dask/array/core.py:2387, in Array.__sub__(self, other)
   2385 @check_if_handled_given_other
   2386 def __sub__(self, other):
-> 2387     return elemwise(operator.sub, self, other)

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/dask/array/core.py:4762, in elemwise(op, out, where, dtype, name, *args, **kwargs)
   4758     shapes.append(out.shape)
   4760 shapes = [s if isinstance(s, Iterable) else () for s in shapes]
   4761 out_ndim = len(
-> 4762     broadcast_shapes(*shapes)
   4763 )  # Raises ValueError if dimensions mismatch
   4764 expr_inds = tuple(range(out_ndim))[::-1]
   4766 if dtype is not None:

File ~/anaconda3/envs/satpy/lib/python3.10/site-packages/dask/array/core.py:4690, in broadcast_shapes(*shapes)
   4688         dim = 0 if 0 in sizes else np.max(sizes)
   4689     if any(i not in [-1, 0, 1, dim] and not np.isnan(i) for i in sizes):
-> 4690         raise ValueError(
   4691             "operands could not be broadcast together with "
   4692             "shapes {}".format(" ".join(map(str, shapes)))
   4693         )
   4694     out.append(dim)
   4695 return tuple(reversed(out))

ValueError: operands could not be broadcast together with shapes (2400, 3000) (1200, 1500)
```

**Screenshots**
not applicable

**Environment Info:**
 - Linux
 - Satpy Version: 0.42.0]
 - PyResample Version:
 

**Additional context**

composite.yaml:

```yaml
  s7refl:
    prerequisites:
      - name: S7
        modifiers: [nir_reflectance]
    standard_name: s7
```

DKilkenny:
- [x] Closes #2460
 - [x] Tests added
 - [x] Fully documented

`NIRReflectance.__call__` passes `projectables + optional_datasets` to `match_data_arrays`, but `NIREmissivePartFromReflectance.__call__` passes only `projectables`. Both classes read the optional datasets in the shared `_get_nir_inputs`, for the sun zenith angle and for the 13.4µm brightness temperature, so in the subclass a mismatched optional skips the geometry check entirely and reaches pyspectral unvalidated.

That is the bug behind #2460. @djhoese identified the cause in the thread in December 2023:

> Ooohhh `optional_datasets` isn't included in the projectables but it should be. That's the "proper" fix for this I think. [...] In that case `match_data_arrays` would have raised `IncompatibleAreas` and the "derive" function never would have been called.

The parent class was fixed. The subclass was not, and still reproduces the original report on current `main`. With a full-resolution NIR channel and a half-resolution sun zenith angle:

```
NIRReflectance                   -> IncompatibleAreas (correct, caught before compute)
NIREmissivePartFromReflectance   -> ValueError: operands could not be broadcast together
                                    with shapes (1200, 1500) (2400, 3000)
```

**The quieter failure matters more.** A shape mismatch at least fails loudly. An optional dataset with a *matching* shape but a different area broadcasts fine, so nothing raises at all and the emissive part is derived from a sun zenith angle belonging to a different place:

| optional dataset | `NIRReflectance` | `NIREmissivePartFromReflectance` |
| --- | --- | --- |
| same shape, different area | `IncompatibleAreas` | **no exception** |
| no `area` attribute | `ValueError: Missing 'area' attribute` | **no exception** |

Both rows raise after this change.

### The change

Rather than copying the matching line into the subclass, this moves it into a shared `_match_projectables_and_optionals` helper on `NIRReflectance` that both `__call__`s use. Copying was how the two diverged in the first place; a shared helper makes divergence structurally impossible for this pair. Happy to collapse it back to a mirrored line if you would rather keep the diff minimal.

Nothing public changes. `match_data_arrays` remains the overridable seam, so a third-party subclass overriding it still gets its override invoked through the helper.

**A second, smaller bug came along with it.** Both classes advertise `optional_datasets=None` in their signature, and both crash on `main` if you actually pass it:

```
NIRReflectance                   -> TypeError: can only concatenate list (not "NoneType") to list
NIREmissivePartFromReflectance   -> TypeError: 'NoneType' object is not iterable
```

The `optional_datasets or []` in the helper handles both. Nothing in `Scene._generate_composite` passes `None` today, since `_get_prereq_datasets` always returns a list, so this only affects direct callers. Say the word if you would rather see it separately.

### Testing

Five tests, each parametrized over both classes, covering: differing resolutions on the sun zenith optional, differing resolutions on the 13.4µm optional, a same-shape/different-area optional, an optional with no `area`, and no optionals at all.

I mutation tested these rather than trusting a green run. Reverting the subclass to the original line fails five of them. Checking only the sun-zenith optional and skipping the CO2 one fails exactly the two CO2 parameters, which is why that test exists rather than being redundant. Dropping the `None` guard fails the no-optionals parameters. Crossing the two returned slices fails six pre-existing tests.

**One mutation my tests do not kill**, stated rather than papered over: running the check and then passing the *unmatched* arrays on to `_get_nir_inputs` passes everything. I looked for an honest assertion that would catch it and could not find one. Only `.data` is ever read from an optional, so `drop_coordinates` and `align_geo_coordinates` cannot reach the output, and dask unifies chunks during blockwise regardless, which I confirmed by chunking the sun zenith angle differently from the NIR channel and getting identical result chunks either way. For the optionals it looks like an equivalent mutant. A chunk assertion would pass for the wrong reason, so I left it out.

Two intentional behaviour changes in the subclass worth flagging. Chunk unification now spans the optionals as well as the projectables, so output chunking can be a finer common refinement than before; values are unchanged and this matches what the parent already did. And an optional lacking an `area` now raises `ValueError` rather than being ignored, which `Scene._generate_composite` does not catch the way it catches `IncompatibleAreas`. Both are existing parent-class semantics now extended to the subclass, and there is a test pinning the second one.

I checked that no configuration relies on a mismatched optional. In `visir.yaml`, `nir_reflectance` and `nir_emissive` have identical prerequisites and optional prerequisites and differ only in the class, and `viirs.yaml` deliberately pins the sun zenith resolution to the band (`M15` with 742, `I05` with 371). So there is no config in which the subclass receives an optional set the parent does not also receive under the same geometry, which means anything that would newly raise here would already be raising in the parent today.

`376 passed, 2 xfailed` across `satpy/tests/test_modifiers.py`, `satpy/tests/modifier_tests/` and `satpy/tests/compositor_tests/`, clean under `-W error`, ruff clean.


satpy/tests/test_modifiers.py
```diff
@@ -324,10 +324,17 @@ def test_provide_sunz_threshold_and_masking_limit(self, tmp_path, comp_kwargs):
...
-    def test_nir_multiple_resolutions(self):
-        """Check that multiple resolutions in the optional datasets produce an IncompatibleArea."""
+    @pytest.mark.parametrize("modifier_name", ["NIRReflectance", "NIREmissivePartFromReflectance"])
+    def test_nir_multiple_resolutions(self, modifier_name):
+        """Check that multiple resolutions in the optional datasets produce an IncompatibleArea.
+
+        Both modifiers consume the optional datasets in ``_get_nir_inputs``, so both must
+        reject a mismatched one. Only ``NIRReflectance`` used to, which is the bug behind
+        GH#2460: the emissive variant let the mismatch through and failed later with a
+        broadcasting error from pyspectral.
```
djhoese: This comment is great for the AI that did this work, but is unnecessary to the long term understanding of this test. Please remove it.

satpy/tests/test_modifiers.py
```diff
@@ -338,11 +345,93 @@ def test_nir_multiple_resolutions(self):
...
+        with pytest.raises(IncompatibleAreas):
+            comp([self.nir, self.ir_], optional_datasets=[co2], **info)
+
+    @pytest.mark.parametrize("modifier_name", ["NIRReflectance", "NIREmissivePartFromReflectance"])
+    def test_nir_same_shape_different_area(self, modifier_name):
+        """Check that an optional dataset with a matching shape but a different area is rejected.
+
+        This is the quiet version of the bug: the shapes broadcast happily, so nothing
+        raises and the modifier silently derives a product from a sun zenith angle that
+        belongs somewhere else.
```
djhoese: Same here. I don't think this contributes much to this test that can't be determined from the test name and the first sentence in the docstring. If you're modifying this with AI I'd even be curious if these two tests (this one and the one above) could be combined in a single parametrized test.

djhoese (changes requested):
Thanks for this contribution. In the future it'd be nice if you could prove a human was in the loop on this, but I also realize we don't have an official policy on this. I've made some small requests. I also wish so many tests didn't need to be added for such a small edge case, but they also make sense. I requested in one of my comments that the tests be combined, but that may be easier said than done.

DKilkenny:
My apologies, I wasn't trying to hide any AI usage. I did check your contribution guidelines beforehand and the `CLAUDE.md` and `AGENTS.md` files as well. I wrote the main function but used Claude to help write the test cases, since I lean toward more tests and wanted to be thorough as an outside contributor. That said, I removed two other tests it came up with because they weren't realistic.  

It was definitely a miss to have those extra docstrings so I removed that and then actually parameterized four tests into one for a total of two.

djhoese:
No problem. Thanks for the changes. Yeah we need to come up with something. Claude is extra noisy on its PR descriptions where it basically feels like reading its entire context in english in the PR description then reading the context in code in the code changes. I personally am leaning towards shorter (human written if possible) PR descriptions and trying to get Claude to be smarter about what tests are needed. It's all understandable and we're all figuring this out. I appreciate the contributions regardless.

satpy/tests/test_modifiers.py
```diff
@@ -258,6 +258,25 @@ def setup_method(self):
                 "units": "K",
             })
 
+    def _bad_optional(self, kind):
```
djhoese, line 261: 🤔 How much do the tests depend on the actual data of these arrays? It is all checking for an exception right? So maybe the `DataArray` creation could be part of the test with the parameters being `dims` and `attrs`? Oh but then there's if it has an area or a wrong area... 🤔 

It'd probably be OK to import IncompatibleAreas in the overall module to simplify this too (then there's no need for the `if` on that case to get the exception).

djhoese: Maybe `kind`, `shape`, and `area_type`? Sorry, my brain isn't working, but this block of `if` statements seems like it could be simplified and based on the actual inputs to `DataArray` rather than the higher-level case being tested.

DKilkenny: The tests don't really rely on the data since every case raises an error before anything computes. It's just an `arange` of the data right now. 

Good call on the parameters, I added the three you mentioned and removed the helper. I left the one `if` clause to set area in the `attrs` when there is one like you called out.

DKilkenny: One note is that self.area_hr isn't used any more since the 4x4 area comes from the existing `_sunz_bigger_area_def` so I removed it but let me know if you want me to bring it back. 

Other thing is that I set two constants at the module level to keep the code a little cleaner and since I saw a couple in the other test files. I can change that too if you'd rather not.

```
SUNZ_ATTRS = {"standard_name": "solar_zenith_angle"}
CO2_ATTRS = {"wavelength": (12.0, 13.0, 14.0), "units": "K",
             "start_time": dt.datetime(2020, 1, 1, 12, 0, 0)}
 ```

DKilkenny:
Since the `kind` to real inputs change added two more parameters, I moved the `DataArray` construction into the parameter itself so each row is just the dataset being rejected. It did add a module level helper but I figure it's ok since it takes in the `shape`, `attrs`, and (optionally) `area` rather than a bunch of case names.

djhoese:
After trying to review the tests, I'm remembering that we didn't move many of the tests for modifiers to their own modules. Could you move these tests for these NIR modifiers to a new `satpy/tests/modifier_tests/test_spectral.py` module?

DKilkenny:
Moved those tests and brought it up to date with the latest in `main`

djhoese (approved):
Thanks!

djhoese:
One CI job was failing so I've restarted it.

PR #3471, opened 2026-09-16
Cache METimage global attributes for small performance improvement

djhoese:
Claude noticed that the global attributes were parsed from the input files for every variable load. This PR makes a small change to cache this and reuse it. When loading all regular bands this has a small performance improvement but I don't remember the exact numbers at the time. It seems like a pretty easy win to do this. It is also wrapped in a mapping proxy object so it is read-only. Claude says this is a "free" guard against someone modifying the cached dictionary and has little to no performance penalty.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


ameraner (approved):
LGTM, thanks!

PR #3473, opened 2026-09-16
Add METimage support to CREFL

djhoese:
Idea and implementation started by @dougaschu at SSEC and then finished by me. The current opinion by the higher-ups is that the CREFL version of the METimage true color looks better than the pyspectral correction (see https://github.com/ssec/polar2grid/issues/837 for details and some images). It is also very close though. Regardless, for Polar2Grid I've been told to use the CREFL solution which requires adding the support into Satpy.

Note: This PR does not add any composites that actually use the crefl for metimage. I'll do that in Polar2Grid and it will be up to any users that want to use it to add it to their own configs.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


pnuu (approved):
LGTM. Nice refactor on the tests.

PR #3474, opened 2026-09-17
Add support for VIIRS EDR LST v2r2 files

djhoese:
See https://github.com/ssec/polar2grid/issues/842

Claude did the coding here. It also found that LST data was never masked properly. Valid range is in the file in the packed (in-file) range, so bytes, rather than the real data range. For LST data that made it all NaNs. This fixes that. FYI to @kathys. If they fix this to be CF standard (`valid_range` in the real data values, not the scaled range) then this will likely filter to the wrong range.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented 


kathys:
@djhoese   I read your comment about the `valid_range` not being compliant.  I can inform NOAA.

djhoese:
Hm apparently I'm wrong. Claude says so at least:


**Packed values.** `valid_range` (and `valid_min`, `valid_max`, `_FillValue`) should be in the same units and data type as the **stored, packed data**, not the unpacked physical values.

The CF Conventions state this in two places:

- **Section 8.1 (Packed Data):** when packed data contains missing values, the attributes that indicate missing values (`_FillValue`, `valid_min`, `valid_max`, `valid_range`) must have the same data type as the packed data.
- **Section 2.5.1 (Missing Data):** applications should *first* check whether a value is valid, and *then* apply `scale_factor`/`add_offset`. So the validity test runs on the raw stored numbers.

This follows the older NetCDF User Guide (NUG) convention, and the common libraries implement it that way, including netCDF-Java and netCDF4-python's auto mask-and-scale.

For example, packing temperature as `short` with `scale_factor = 0.01` and `add_offset = 273.15`:

```
short temp(time) ;
    temp:scale_factor = 0.01f ;
    temp:add_offset = 273.15f ;
    temp:valid_range = -5000s, 5000s ;   // packed: 223.15 K to 323.15 K unpacked
    temp:_FillValue = -32768s ;
```

Two practical cautions:

- **Legacy files:** some older datasets (notably some satellite products) wrongly put unpacked values in `valid_range`. The telltale sign is an attribute type that doesn't match the variable's type, such as a `float` range on a `short` variable. Some readers try to detect this, but you shouldn't rely on it.
- **Type matching:** the attribute type should exactly match the variable type (`short` → `short`). A mismatch is both a CF compliance issue and a common source of ambiguity. The CF compliance checker will flag it.

PR #3475, opened 2026-09-17
fix(sgli_l1b): Unwrap one-item array attributes on numpy >=2.4 and up…

### LINKED ISSUE #3470: Reading any real GCOM-C/SGLI L1B granule fails with numpy >= 2.4
**Describe the bug**
```
TypeError: only 0-dimensional arrays can be converted to Python scalars
```

JAXA stores `Resampling_interval`, `Number_of_lines` and `Number_of_pixels` as **1-element
arrays**, and `HDF5SGLI.interpolate_spherical` passes them to numpy unchanged. Converting an
array with `ndim > 0` to a scalar was deprecated in numpy 1.25 and [raises `TypeError` as of
numpy 2.4.0](https://github.com/numpy/numpy/blob/main/doc/source/release/2.4.0-notes.rst) so the reader works silently (with a `DeprecationWarning`) up to numpy 2.3
and errors from 2.4 on.

**To Reproduce**

```python
import h5py
import numpy as np
from satpy.readers.sgli_l1b import HDF5SGLI

with h5py.File("/tmp/sgli_repro.h5", "w") as f:
    geo = f.create_group("Geometry_data")
    lon = geo.create_dataset("Longitude", data=np.linspace(10, 20, 30).reshape(5, 6))
    lat = geo.create_dataset("Latitude", data=np.linspace(40, 50, 30).reshape(5, 6))
    for dset in (lon, lat):
        dset.attrs["Resampling_interval"] = np.array([10])   # a one-item list, as JAXA writes it
    img = f.create_group("Image_data")
    img.attrs["Number_of_lines"] = np.array([41])
    img.attrs["Number_of_pixels"] = np.array([51])

fh = HDF5SGLI("/tmp/sgli_repro.h5", {"resolution": "Q"}, {})
fh.get_lon_lats({"name": "longitude_v"})
```

The normal way in is just a Scene load:

```python
scn = Scene(reader="sgli_l1b", filenames=[vnrdq_file, irsdq_file])
scn.load(["latitude_v", "longitude_v", "solar_zenith_angle"], resolution=250)
```

## Expected behavior

The lat/lon and angle datasets load, as they still do on numpy 2.3 and earlier.

## Actual results

```python-traceback
  File "repro.py", line 17, in <module>
    fh.get_lon_lats({"name": "longitude_v"})
  File ".../satpy/readers/sgli_l1b.py", line 169, in get_lon_lats
    lons, lats = self.interpolate_spherical(lons, lats, resampling_interval)
  File ".../satpy/readers/sgli_l1b.py", line 183, in interpolate_spherical
    tie_lines = np.arange(0, polar_angle.shape[0] * resampling_interval, resampling_interval)
TypeError: only 0-dimensional arrays can be converted to Python scalars
```

If you hand it the spacing as a plain number, it gets one step further and then fails on the
image size instead:

```python-traceback
  File ".../satpy/readers/sgli_l1b.py", line 187, in interpolate_spherical
    new_azi, new_pol = interpolator.interpolate_to_shape(full_shape, chunks="auto")
  File ".../geotiepoints/interpolator.py", line 360, in interpolate_to_shape
    fine_points = [np.arange(size) for size in shape]
TypeError: only 0-dimensional arrays can be converted to Python scalars
```
The tests don't catch this because they build the test file with plain numbers
(`longitude.attrs["Resampling_interval"] = 10`, `image_data.attrs["Number_of_lines"] = 1955`),
while real files use one-item lists. Writing them as one-item lists in the fixtures would make
the tests fail the same way.

The fix could be to take the number out of the list where the attributes are read:

```python
resampling_interval = attrs["Resampling_interval"].item()   # in get_lon_lats / get_full_angles

full_shape = (self.h5file["Image_data"].attrs["Number_of_lines"].item(),
              self.h5file["Image_data"].attrs["Number_of_pixels"].item())
```

## Environment Info

- OS: Linux (Ubuntu 24.04 container)
- Python: 3.12.3
- Satpy: 0.59.0 (0.60.0 and `main` have the same reader code)
- numpy: 2.5.3 — also fails on 2.4.0; works on 2.3.5 and earlier
- h5py: 3.14.0
- python-geotiepoints: 1.9.0

cshatto:
JAXA stores `Resampling_interval`, `Number_of_lines` and `Number_of_pixels` as one-element arrays, which the SGLI reader passed to numpy unchanged — relying on the implicit array-to-scalar conversion removed in numpy 2.4, so every real GCOM-C/SGLI granule now fails with `TypeError: only 0-dimensional arrays can be converted to Python scalars`. This PR calls `.item()` at the three read sites and rebuilds the test fixtures with one-item arrays so they match real files.

Closes #3470


cshatto:
Hello, I just wanted to check-in on the status here. Thanks in advance!

mraspaud (approved):
Thanks a lot for the fix! Looks good to me.

mraspaud:
@cshatto sorry for the late reaction, I'm the author of this reader and am currently travelling (EUM conference).
The changes look good to me, however I have a question: I'm pretty sure that when I developped the reader, the scalars were not arrays. Do you know if that is a recent addition, and if old data from sgli have the same format? Otherwise, we might need to support both formats...

cshatto:
Hi @mraspaud thank you and no problem at all! This small change was blocking our numpy upgrade to > 2 so that is why I poked the issue again. [Numpy 2.4 changed the behavior of a 1D array being converted to a scalar](https://github.com/numpy/numpy/blob/15a1636a6d68bbee2056494978f9867ee40b9d25/doc/source/release/2.4.0-notes.rst#raise-typeerror-on-attempt-to-convert-array-with-ndim--0-to-scalar), so it could be that the data always behaved as a scalar because numpy was converting it silently from 1D array to scalar.

mraspaud:
Ok, thanks for the explanation. Let's just merge this, and if it breaks with some old data in the future, we'll fix it then. Thanks for your contribution!

PR #3477, opened 2026-09-18
Refactor Scene resampling to new helper class

djhoese:
This is an evolution of @pnuu's #3178. I asked Claude to do what I mentioned in my comment from that PR and move @pnuu's refactored code into its own object. Claude then went a step further and reorganized how ancillary datasets are handled by using recursion and caching. I think the code looks really clean, but the biggest downside is the tests which contain repeats for the tests that exist for the Scene's resampling. The changes were big enough that I thought it made sense to create a separate pull request.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


gerritholl:
On a quick glance, this looks interesting.  The whole refactoring was triggered by my work on #3168, which now has a merge conflict with this branch, which I don't have time to look at before 28 September.  After 28 September I can test how it interacts with #3168.

pnuu (approved):
Yes. Definitely yes.

I'm closing my earlier version.

satpy/resample/base.py
```diff
@@ -190,3 +191,132 @@ def resample_dataset(dataset, destination_area, **kwargs):
...
+
+        Results are memoized by DataID, so a dataset (typically an ancillary
+        variable) that is encountered multiple times is only resampled once and
+        the same resampled object is returned each time.
+        """
+        ds_id = DataID.from_dataarray(dataset)
+        try:
+            return self._resampled[ds_id]
+        except KeyError:
+            pass
```
mraspaud, line 241: This should be an lru_cache, or at least some memory-limited version of the dictionary

satpy/resample/base.py
```diff
@@ -190,3 +191,132 @@ def resample_dataset(dataset, destination_area, **kwargs):
...
+        return self.resample(anc)
+
+    def _get_resampler(self, source_area):
+        """Get the resampler for *source_area*, creating it on first use."""
+        try:
+            return self._resamplers[source_area][1]
+        except KeyError:
+            key, resampler = prepare_resampler(source_area, self.destination_area, **self.resample_kwargs)
+            self._resamplers[source_area] = (key, resampler)
+            return resampler
```
mraspaud, line 274: Same here, this is a potential memory leak

satpy/resample/base.py
```diff
@@ -190,3 +191,132 @@ def resample_dataset(dataset, destination_area, **kwargs):
...
+            LOG.info("Not reducing data before resampling.")
+            return dataset, source_area
+        return self._slice_data(dataset, slices, reduced_area), reduced_area
+
+    def _get_reduction(self, source_area):
+        """Get the slices and reduced area for *source_area*, reusing previous results."""
+        try:
+            return self._reductions[source_area]
+        except KeyError:
+            pass
```
mraspaud, line 295: Third case of the potential memory issue.
I think lru_cache should be used

mraspaud (changes requested):
Great cleanup, that's really nice!
I think I spotted three potential memory leaks though where the dicts defined in the DatasetResampler can grow indefinitely. I can see that the way that class is used in Scene would not be a problem, but this is a public class, and it would be very tempting to use it for a custom script where one would just want to resample a full archive of something, leading to very large amounts of datasets being cached in the DatasetResampler object.

Another thing: the name makes me think of a potential accessor in the xarray.Dataset object: I know this is not the objective here, but did you have that in mind when writing this? 
And that leads me to another question: is Satpy the right place for this? I can see some satpy-specific things here but they don't look like they can't be abstracted?

Anyway, sorry for all the questions, but the caching needs to be fixed imo.

satpy/resample/base.py
djhoese: What do you think the limit should be for all of these? 100?

djhoese:
@mraspaud 

> Another thing: the name makes me think of a potential accessor in the xarray.Dataset object: I know this is not the objective here, but did you have that in mind when writing this?
And that leads me to another question: is Satpy the right place for this? I can see some satpy-specific things here but they don't look like they can't be abstracted?

I did not have an accessor in mind. I basically had Claude take my comments from Panu's PR and guided it into this form. Most of Panu's code did the structure and Claude just put it into a separate class.

Regarding this going in Pyresample or somewhere else: It is on my long term TODO list to move the "get resampler" logic into pyresample. It just hasn't been the top priority.

satpy/resample/base.py
djhoese: Claude points out that the `._resamplers` in the `Scene` has this problem too and has since it was added. Maybe that should be bounded too?

satpy/resample/base.py
mraspaud: I guess it really depends on what data we load, but yeah, maybe 100 works?

satpy/resample/base.py
mraspaud: yes, good catch!

djhoese:
Ok I had Claude take a look at this and it went way further than I thought, but I also don't have a strong argument against the way it went. But I want to put Claude's changes here to look at while finishing other more important PRs. I'll come back to this and see how I feel later. For those reading this in the mean time, my hang ups are:

1. `resample_all` and `resample` on the helper class. I don't like the name `resample_all` but don't have a better name. I'm also not sure we can't just treat this helper class like a private helper...but as @mraspaud said maybe we should design this to go in pyresample in the future.
2. The way keys and values are split in `scene.py` to resample all the `DataArray`s and then assign those resampled DataArrays to the new Scene.

Issue #2947, opened 2024-10-22
In night-time geo_color, mid-level clouds give false impression of clear-sky conditions


gerritholl:
**Describe the bug**

For the night part, the `geo_color` composite uses a blending of `geo_color_high_clouds` and `geo_color_low_clouds`.  For the high clouds, brightness temperature is shown as transparency using `HighCloudCompositor` / `CloudCompositor`, where 300 K are fully transparent (not a high cloud) and a latitude-dependent threshold (either 210 K or 230 K) are fully opaque / white.

When mid-level clouds cover low-level clouds, those mid-level clouds are shown mostly transparent.  However, they are actually opaque, and it is unknowable whether low-level clouds exist underneath the mid-level clouds.  By displaying opaque clouds as semi- or even nearly fully transparent, the user may be misled to believe there are clear-sky conditions.

An example can be seen on 2024-10-15 over Central Europe.  At 04:20, low clouds cover valleys in German mountain valleys:

![wmsMap_2024-10-15T0420](https://github.com/user-attachments/assets/76485138-a670-468d-b064-26c37aedfbc5)

By 06:00, advection of mid-level clouds looks just like the fog is dissipating.  Note that the mid-level clouds themselves are nearly invisible in this visualisation, but the city lights underneath are clearly shown.

![wmsMap_2024-10-15T0600](https://github.com/user-attachments/assets/7045a8bb-b75f-4986-9486-ac9cf95fe146)

After sunrise, at 07:20, we can see that actually, the low-level clouds didn't go anywhere, and are partially covered by higher clouds, now clearly visible in the true colour:

![wmsMap_2024-10-15T0720](https://github.com/user-attachments/assets/1dcf3f26-9b98-4185-ae0c-1308a668959d)

For additional views or videos, see [eumetview](https://view.eumetsat.int/productviewer?v=103644).

**To Reproduce**

```python
import hdf5plugin
import os
from satpy import Scene
from glob import glob
from sattools.io import plotdir
fci_files = {
    "good": glob("/media/nas/x23352/MTG/FCI/L1c-cases/20241015-hidden-fog/04/W_XX*BODY*C_0027*nc"),
    "bad": glob("/media/nas/x23352/MTG/FCI/L1c-cases/20241015-hidden-fog/06/W_XX*BODY*C_0037*nc"),
    "day": glob("/media/nas/x23352/MTG/FCI/L1c-cases/20241015-hidden-fog/08/W_XX*BODY*C_0052*nc")}
[...]

PR #3478, opened 2026-09-18
Improve the geo_color composite to address existing limitations

strandgren:
This PR introduces some changes and improvements for the `geo_color` compsite based on some recurring user feedback that I have received, including some issues reported in https://github.com/pytroll/satpy/issues/2947, as well as personal preference🙂.

The current implementation introduced in https://github.com/pytroll/satpy/pull/2557 closely follows the GeoColor concept described in [Miller et al. (2020)](https://doi.org/10.1175/JTECH-D-19-0134.1) with some minor adaptions to better fit the Satpy architecture. But the implementation has some limitations that I try to address here (and plan to publish as well).

> [!NOTE]
> The PR is still **draft** since I need to work on the tests. I also need to get green light for these changes in-house at EUMETSAT where the Satpy `geo_color` is used quite a bit. But I still publish the PR here to allow for early feedback also from this community.

### User feedback and issues
The city lights are too bright and distracting. The use of the black marble also introduced a lot of clear-sky surface signal that comes from static VIIRS data and thus doesn't represent the data for which the composite is created, e.g. FCI.

### Changes in this PR
To overcome this, this PR creates an alternative composite `geo_color_city_lights` which uses the same black marble as input, but applies stretching and gamma correction to reduce the brightness of both the city lights and the surface (essentially becoming black).

### User feedback and issues

There are obvious false alarms over some arid surfaces.

Clouds have different transparency depending on if the surface below is land or water, leading to possible discontinuities at the coastline. For the same reason, some low clouds over land are also hard to see because of high level of transparency.

### Changes in this PR

Another test is added to remove bare soil false alarms. The test uses the IR10.5-IR8.7 split window difference, the satellite zenith angle and two configurable thresholds to identify and remove bare soil false alarms. This is a low cloud detection test taken from the NWC-SAF GEO cloud detection algorithm. The data needed for this is optional and can be skipped if the false alarms is not a problem in a certain area being processed.

Decouple the limits for the low cloud detection from the limits used for the alpha transition such that the same alpha limits are used independent of surface type.

### User feedback and issues
Thin cirrus are too transparent. Currently the IR10.5 brightness temperature is both stretched using a greyscale colormap and assigned an alpha channel with increasing transparency with increasing temperature. This leads to thin clouds becoming both black and transparent, and thus hardly visible against a dark background. This is a flaw in the current implementation.

### Changes in this PR
To overcome this the high clouds are always set to white colour and only the alpha channel is used to modify the appearance of the clouds. Hence, thin clouds are now as transparent, but white, thus increasing the contrast against the dark background.

This also allows to remove the latitude dependency of the alpha limits (at least for full disk processing) and we can simply use the `CloudCompositor` instead and avoid computing the latitude. 

## Combined effects for final composite
The fact that the background layer is now essentially black except for the city lights and the layer focusing on high clouds also contains visible signal for clear-sky, the surface signal that we see in `geo_color` is the temperature signal from the actual instrument, e.g. FCI, rather than VIIRS. This also allow us to better see low clouds that are not necessarily detection by the low-level cloud detection test.

Furthermore, the day-night blending limits are modified to include more day-time imagery.

## Results

The figures below shows an overview on how the different layer change with this PR as well as some examples of the final composite. Full resolution imagery are temporarily available here: https://sftp.eumetsat.int/public/folder/do6cxnkrfko9limqhf_ukg/geo_color

Slide1
Slide2
Slide3
Slide4
Slide5

------------------

> [!IMPORTANT]
> The `geo_color` composite still uses **simple** spectral thresholding and static thresholds to characterize the clouds and low clouds (high or low) may not be optimally displayed/visualized or even not visible at all. For that reason `geo_color` should be used with care for any type of weather forecasting and is primarily meant for communication/outreach purposes.

 - [ ] Tests added 
 - [x] Add composites without city lights and without low-clouds
 - [x] Create global composites to verify consistent behavior for geo-ring
 - [x] Fully documented 

_AI was used to support this PR, mainly for refactoring and docstrings. All such changes have been reviewed by me._


strandgren:
I also added two alternative recipes to generate geo_color without the city lights (e.g. if you do not want to display any static auxiliary data) or without the blue colorization of low clouds. Below are some examples, also demonstrating that the new recipes work for both FCI, ABI and AHI.

**geo_color**
geo_color_world_mosaic_20260915T160000

**geo_color_no_lights**
geo_color_no_lights_world_mosaic_20260915T160000

**geo_color_no_blue**
geo_color_no_blue_world_mosaic_20260915T160000

